PR #4, opened 2016-05-09
Fixing ambiguous column SQL error

dominiclord:
Lack of a table name was throwing a SQL error.

> Column 'ID' in on clause is ambiguous

This should technically make things clearer and fixes the error.


Chouby:
Thanks for the PR.
May I ask you in which circumstances you encounter this SQL error?

Chouby:
I noticed that the PR creates a database error Unknown column 'wp_posts.ID' in 'on clause' in get_adjacent_post()

PR #12, opened 2016-06-03
add order to output of PLL_Switcher::get_elements()

diggy:
Adds order (`term_group` value) to output of `PLL_Switcher::get_elements()`. Useful, if not essential, for building a custom language switcher using `pll_the_languages()` with `raw` param set to `true`.


Chouby:
`PLL_Switcher::get_elements()` already returns the elements in the correct order. Could you explain your use case where the order information is essential?

diggy:
Yes, I'm sorry, you're right, I had something running that was messing up the order. Regardless, it might be useful to send along the order values, right now the order of items needs to be "extracted" from the array keys of the output of `pll_the_languages()`, no exact numeric values available. Your call :)

PR #17, opened 2016-06-10
Truly pluggable default flags

southan:
My biggest frustration working with this nonetheless-superb plugin is the use of hardcoded URL's within the flag "API", and the very limited `pll_get_flag` filter (it only passes the final `` HTML and the language slug).

If I can essentially add/remove flags with `pll_predefined_flags` I should be able to go the whole 9 yards - with the attached suggestion, any plugin/theme author can replace the entire flag set with their own, hosted anywhere, and with any file extension.


Chouby:
Default flags are displayed on the backend. That's the main reason why I did not add any filter to change them. Custom flags can have any size and and most of them will not fit well on the backend.

On the frontend however, you can use whatever custom flag you want. See: https://polylang.pro/doc/can-i-use-my-own-flags-for-the-language-switcher/

southan:
But the fact that the filter _does_ affect the backend implies a dev _can_ add/edit flags globally. I would really like to be able to at least add flags that aren't out-of-the-box and have them visible for an end-user in the admin. Let me know your thoughts.

JoryHogeveen:
This would be a big +1 for me.

> Custom flags can have any size and and most of them will not fit well on the backend.

I'd say this should be the responsibility of the developer or website owner. Though this should definately be a separate filter...

PR #25, opened 2016-06-17
added capability of adding a class on first item in switcher

acki:
Added capability of adding a class on first item in switcher. This could be useful for CSS styling - when you have to style a space between other menu items and the switcher.


Chouby:
Merged with a minor change in https://github.com/polylang/polylang/commit/73783f082270e478e37e445cac5de5935a18a018

PR #37, opened 2016-08-01
UTF-8 Turkish Language error solved

Natgho:
"[error] 471#0: *67 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Class 'PLL_Install' not found"
If using turkish (or UTF-8) and using Wp ALL import together, packages confict, this change solve the problem.

### DIFF
--- a/include/class-polylang.php
+++ b/include/class-polylang.php
@@ -67,7 +67,7 @@ class Polylang {
 			return;
 		}
 
-		$class = str_replace( '_', '-', strtolower( substr( $class, 4 ) ) );
+		$class = str_replace( '_', '-', mb_strtolower( substr( $class, 4 ), "UTF-8" ) );
 		$to_find = array( 'media', 'share', 'slug', 'slugs', 'sync', 'translate', 'wpml', 'xdata' );
 		$dir = implode( '-', array_intersect( explode( '-', $class ), $to_find ) );
 


Chouby:
Hi!

I can't use the function 'mb_strtolower' as is as it's not available on all PHP installs. More over, I don't understand why 'Install' would need need the multibyte function to be lower-cased. The string is just plain ASCII. Could you please give more details to reproduce the issue?

Natgho:
Hi Chouby;

I am talking about this function;
http://php.net/manual/en/function.mb-strtolower.php
You used default "strtolower" function. Polylang,  "install" has the file named. If use strtolower convert file name, name equal "Install", so polylang cant find this folder. If use mb_strtolower, file name correctly convert name. The reason is to use the Turkish language and use UTF-8. I am searched internet, found people who have similar problem . I hope I can explain?

Chouby:
As explained above, I can't use this function because it is not enabled by default in PHP. See http://php.net/manual/en/mbstring.installation.php

I have successfully made tests with all sorts of scripts such as Turkish, Chinese, Arabic which all need a correct UTF-8 setup. WordPress itself is using `strtolower` more than 100 times.

I may expect that a character such as `Ï` is not correctly lowercased by strtolower because it's not plain ASCII but a character such as `I` which is plain ASCII should be correctly lowercased by strtolower on any system. Moreover I am using `substr`, not `mb_substr`, which does not seem to cause issues on your system.

Chouby:
Could you please tell me what is the output of
`echo setlocale(LC_ALL, 0);`
on your system?

Natgho:
Hi, response just "C".

Chouby:
According to the documentation of [strtolower](http://php.net/manual/en/function.strtolower.php), the function takes the current locale to determine which characters are lowercased. The 'C' locale is the most generic one and works perfectly on my test site. It does not convert accented characters but here we don't need it. Plain ASCII characters are all converted correctly.

PR #55, opened 2016-11-07
Avoid cross domain issues when linking files via `home_url`.

hubeRsen:
Tackling issues in combination with other plugins (for example Jetpack, which uses `home_url` for managing infiniteScroll request).


Chouby:
We can't filter directly 'home_url' because it applies to all links. Among possible other issues, this breaks links to the translation (in the language switcher).
If you have a compatibility issue with a plugin, the link must be filtered at a higher level.
If this is for Jetpack, please don't hesitate to open an issue explaining the problem.

grappler:
Yes, we have just realized that it affects the language switcher too.

Issue #53, opened 2016-11-04
Polylang lang_choice dropdown widget displays without Label - non-accessible form control


Twintails:
ref: [(http://webaim.org/techniques/forms/controls#select)]

**Line 44-46 on include/widget-languages.php**
```
	if ( $title ) {
				echo $args['before_title'] . $title . $args['after_title'];
			}
```

**Suggestion without theme fallback:**

```
if ( $title ) {
  echo $instance['dropdown'] ? '<label for="lang_choice_'.$args['widget_id'].'">' . $title : $args['before_title'] . $title;
  echo $instance['dropdown'] ? '</label>' : $args['after_title'];
}
```

**Suggestion with Theme Fallback:**

      if ( $title ) {
        echo $instance['dropdown'] ? '<label  for="lang_choice_'.$args['widget_id'].'">' . $args['before_title'] . $title : $args['before_title'] . $title;
        echo $instance['dropdown'] ? $args['after_title'] . '</label>' : $args['after_title'];
      }

This way them devs could still wrap the title from the $before_title and $after_title hooks, with 
```
<sapn class="visuallyhidden">Widget Title</span> or <h2>Widget Title</h2>
```
I couldn't determine where the 'lang_choice_' portion of the select(dropdown) id comes from. perhaps someone more familiar with the code could look into integrating this?

Chouby:
Thanks for reporting this issue.

The code was initially inspired by the core widgets archives and categories. But I missed the accessibility improvement done in WP 4.2 (https://core.trac.wordpress.org/ticket/18650).

I'd like however to keep the consistency with the core widgets and thus use the same code. See https://github.com/WordPress/WordPress/blob/4.6.1/wp-includes/widgets/class-wp-widget-archives.php#L59

Thus I propose:
`
if ( $instance['dropdown'] ) {
	if ( $title ) {
		echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . $title . '</label>';
	}
	echo $list;
} else {
	echo "<ul>\n" . $list . "</ul>\n";
}
`

Twintails:
Excellent! 
It may need to be able to wrap the title still from the theme register_sidebar's before_title and after_title. 

```
echo $args['before_title'] . '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . $title . '</label>' . $args['after_title'];

```
[...]

PR #56, opened 2016-11-07
Issue #53 change

Twintails:
If widget set to dropdown, wraps the widget title in Label element; with accessibility improvement done in WP 4.2. Prepends and appends before_title and after_tilte around the Label in case theme supplies a block level parent.

### DIFF
--- a/include/widget-languages.php
+++ b/include/widget-languages.php
@@ -48,7 +48,7 @@ class PLL_Widget_Languages extends WP_Widget {
 					if ( $title ) {
 						echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . $title . '</label>';
 					} else {
-						echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . __( 'Language Switcher', 'polylang' ) . '</label>';
+						echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . __( 'Choose a language', 'polylang' ) . '</label>';
 					}
 					echo $list;
 				} else {
--- a/include/widget-languages.php
+++ b/include/widget-languages.php
@@ -32,30 +32,28 @@ class PLL_Widget_Languages extends WP_Widget {
 	 * @param array $instance The settings for the particular instance of the widget
 	 */
 	 function widget( $args, $instance ) {
- 		// Sets a unique id for dropdown
- 		$instance['dropdown'] = empty( $instance['dropdown'] ) ? 0 : $args['widget_id'];
+		// Sets a unique id for dropdown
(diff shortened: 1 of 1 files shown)


Chouby:
As is, it won't display the title if the widget options are set to display a list.

If you look at the core widgets, the title is displayed in all cases. And the accessibility label is added only if the options are set to display a dropdown.

Chouby:
I am not a specialist. But I guess that the screen reader will read the label first, then the widget title, and finally the select. So may be it would confuse the reader to have the widget title between the label and the select. 

In core widgets, the label is after the title. See:
https://github.com/WordPress/WordPress/blob/master/wp-includes/widgets/class-wp-widget-archives.php#L51-L60

That would give this for the Languages widget:
```
if ( $title ) {
	echo $args['before_title'] . $title . $args['after_title'];
}
if ( $instance['dropdown'] ) {
	if ( $title ) {
		echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . $title . '</label>';
	}
	echo $list;
} else {
	echo "<ul>\n" . $list . "</ul>\n";
}
```
What do you think?

Twintails:
I see what you are saying about WP Core, and I can make the change if it's your preference.
I see the Wordpress core themes visually hide .screen-reader-text with styling, but it sure seems redundant to have it read as you move through the page with arrow keys.
"Widget Title", "Widget Title", "Selected Option" "Widget Title"
 with the screen reader. 

My preference would be to use the Widget title as the Select's Label, as the widget is displayed as a list or a dropdown. However, I can understand if you want to follow Wordpress Core. 

Please Confirm and I'll make the change to this PR Branch.

Twintails:
I think it also needs a fallback to retain accessibility with a label on the select if the end user chooses not to have a Widget Title?
```
if ( $title ) {
    echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . $title . '</label>';
} else {
    echo '<label class="screen-reader-text" for="' . esc_attr( 'lang_choice_' . $instance['dropdown'] ) . '">' . __( 'Language Switcher', 'polylang' ) . '</label>';
}
```
Is that the correct way to re-use that from line 18 ?

Chouby:
Yes. You are right. I thought to that too. Up to now, I avoided to output any string on frontend (currently, the Polylang translation files contain only strings used on admin side), to minimize the performance impact of Polylang on frontend. But since I want to keep the possibility not to have widget title (at least not to break existing designs), there is now a good reason to output a translated string on frontend.

Since we have now to use the translated string, it would make sense to use it even when there is a title to avoid the redundancy you mentioned. 

"Language swicher" is a good string. But there is also "[Choose a language](https://github.com/polylang/polylang/blob/2.0.9/settings/view-tab-lang.php#L42)" among existing translated strings. Maybe it would be even better for this purpose.

Twintails:
agreed, please verify I used the correct method '__()'

PR #57, opened 2016-11-09
New filter for the HREFLANG code

junaidbhura:
Theme developers and SEO professionals need more flexibility with the HREFLANG code. This might depend on how their theme is built, or their SEO strategy.

This filter gives the ability to update language code right before it is displayed, giving more flexibility.


Chouby:
See filter ''pll_rel_hreflang_attributes' added in https://github.com/polylang/polylang/commit/0871c6feb9fe167dc8165b553aaa0e86472e658a

PR #62, opened 2016-11-15
Get custom flags from theme directory

m2de:
The current use of the `wp-content/polylang` directory seems like bad practice. This allows custom flags placed inside the current theme (or child theme) directory to be loaded. Simply place flags in a `polylang/flags` directory inside your theme folder.


marcodewsign:
Can this be merged please or an explanation given why it shouldn't be merged in?

grappler:
It might be interesting to use `get_theme_file_uri()` instead. https://make.wordpress.org/core/2016/09/09/new-functions-hooks-and-behaviour-for-theme-developers-in-wordpress-4-7/

Issue #20, opened 2016-06-13
String translations for non-admins


JoryHogeveen:
Is it possible to allow other roles to access String translations only? Maybe a custom permission?

Chouby:
I am using the 'manage_options' capability because the strings translations are intended to translate options. Thus I believe that only users having this capability should have access to it.

JoryHogeveen:
Ah ok, I am also using them for front-end stuff like strings in a theme and widgets. My clients all have the editor role so they can't get to settings they could break but I'd like to allow them to translate these strings.
If you have any idea on how to accomplish this without the need of extra plugins or .mo files, please let me know.

Chouby:
Theme options and widgets should not be accessible to editors. Am I wrong?

JoryHogeveen:
You are correct but I change these and added the "edit_theme_options" cap to the editor role.
Thus I know this is not a default configuration :)

I'll close this issue because I guess it is out of the scope for your support but if you have an idea for a workaround, please let me know (maybe a filter for the capability??)

diggy:
>strings translations are intended to translate options

I think a lot of people (you know, those who don't read the docs) "abuse" the strings translations for translating strings in theme template files. Maybe the strings translations tab could benefit from a subtle warning with regard to this?

Chouby:
@JoryHogeveen Changing the capability level is already possible:
`
add_action( 'init', function() {
	if ( PLL_ADMIN ) {
		remove_action( 'admin_menu', array( PLL(), 'add_menus' ) );
		add_action( 'admin_menu', function() {
			add_submenu_page( 'options-general.php', $title = __( 'Languages', 'polylang' ), $title, 'edit_theme_options', 'mlang', array( PLL(), 'languages_page' ) );
		} );
	}
} );`
But it impacts all the tabs, not only the 'strings translations', so that may not be what you want. Having a separate capability would probably require a structure change. Maybe, I would need to create a specific menu instead of using a submenu of settings.

@diggy If you are referring to https://wordpress.org/plugins/polylang-theme-strings/, it does not disturb me as we can consider that translating the theme requires the 'manage_options' capability. What's more disturbing is using this to translate content as NextGen Gallery does.

JoryHogeveen:
@Chouby Yes I thought of that solution but this indeed allows users to see all settings, not just the translations.
If I can find some time I will look into this. It's not a high priority issue, more a thought to enhance this functionality :)

Chouby:
[...]

PR #66, opened 2016-11-26
Allow the admin page capability to be filtered

JoryHogeveen:
Related: https://github.com/polylang/polylang/issues/20

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -87,13 +87,14 @@ class PLL_Admin_Base extends PLL_Base {
 		$tabs = apply_filters( 'pll_settings_tabs', $tabs );
 
 		foreach ( $tabs as $tab => $title ) {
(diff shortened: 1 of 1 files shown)


Chouby:
I understand why you propose a filter. I believe that the [currently proposed way](https://github.com/polylang/polylang/issues/20#issuecomment-269211385) to control the capabilities is not complex. 

But I see 2 issues at the moment:
Here the filter controls all submenus at once when it would be preferable to control them separately. 
As explained [here](https://github.com/polylang/polylang/issues/20#issuecomment-269483382), my goal in the future would be to create capabilities attached to each language to allow "translator roles". Since I am not sure how I will implement that, there is a risk of conflict with the filter as proposed and then I would have difficulties to maintain backward compatibility. So for the moment I prefer to keep this unmerged.

Issue #82, opened 2017-01-11
Add function to escape string


jegbagus:
WordPress escape both __() & _e() with esc_html__() & esc_html_e()
please also add function to escape html from translation.

Thank you

PR #83, opened 2017-01-11
 Add function to escape string for translation in the string translation panel

jegbagus:
this commit related to ticket #82

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -137,7 +137,7 @@ function pll__( $string ) {
  * @param string $string the string to translate
  * @return string translation in the current language
  */
-function pll_esc_htlm__ ( $string ) {
+function pll_esc_html__( $string ) {
 	return esc_html( pll__( $string ) );
 }
 
@@ -172,9 +172,8 @@ function pll_e( $string ) {
  * @param string $string the string to translate
  */
 function pll_esc_html_e( $string ) {
-	echo pll_esc_htlm__( $string );
+	echo pll_esc_html__( $string );
 }
-
 
 /**
  * Translates a string ( previously registered with pll_register_string )
--- a/include/api.php
+++ b/include/api.php
@@ -138,7 +138,7 @@ function pll__( $string ) {
  * @return string translation in the current language
  */
 function plll_esc_htlm__ ( $string ) {
-	return esc_html__( pll__( $string ) );
+	return esc_html( pll__( $string ) );
 }
 
 /**
--- a/include/api.php
+++ b/include/api.php
@@ -137,8 +137,20 @@ function pll__( $string ) {
  * @param string $string the string to translate
  * @return string translation in the current language
  */
-function plll_esc_htlm__ ( $string ) {
+function pll_esc_htlm__ ( $string ) {
 	return esc_html( pll__( $string ) );
+}
+
+/**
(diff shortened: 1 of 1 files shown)


include/api.php
```diff
@@ -130,6 +130,18 @@ function pll__( $string ) {
...
 /**
+ * Translates a string ( previously registered with pll_register_string ) and escapes it for safe use in HTML output.
+ * 
+ * @since 2.1
+ * 
+ * @param string $string the string to translate
+ * @return string translation in the current language
+ */
+function plll_esc_htlm__ ( $string ) {
+	return esc_html__( pll__( $string ) );
```
grappler: You are missing the text domain here. You want to do `return esc_html( pll__( $string ) );`.

include/api.php
```diff
@@ -130,6 +130,18 @@ function pll__( $string ) {
...
 
 /**
+ * Translates a string ( previously registered with pll_register_string ) and escapes it for safe use in HTML output.
+ * 
+ * @since 2.1
+ * 
+ * @param string $string the string to translate
+ * @return string translation in the current language
+ */
+function plll_esc_htlm__ ( $string ) {
```
grappler: There is a "l" too many in `plll` and html is spelled incorectly

Chouby: Could you please rename the function to pll_esc_html__ including in the function where it is used below?

grappler:
`pll_esc_attr_e()` & `pll_esc_attr__()` should be included too if `pll_esc_html__()` is going to added.

Chouby:
Hi!

I ma ok to introduce these new functions. We could also add the corresponding functions for `esc_attr() `

jegbagus:
hi @grappler thank you for correction. Also add pll_esc_attr (without _e and __). WordPress it self not having esc_attr__ or esc_attr_e function. so i think we don't need to have it also. 

Thank you

include/api.php
```diff
@@ -130,6 +130,30 @@ function pll__( $string ) {
...
 
 /**
+ * Translates a string ( previously registered with pll_register_string ) and escapes it for safe use in HTML output.
+ * 
+ * @since 2.1
+ * 
+ * @param string $string the string to translate
+ * @return string translation in the current language
+ */
+function pll_esc_htlm__ ( $string ) {
```
grappler: There is a space between the function name and the opening bracket. HTML has still a typo.

include/api.php
```diff
@@ -141,6 +165,18 @@ function pll_e( $string ) {
...
+ *
+ * @since 2.1
+ *
+ * @param string $string the string to translate
+ */
+function pll_esc_html_e( $string ) {
+	echo pll_esc_htlm__( $string );
+}
+
+
```
grappler: There are two line breaks instead of one.

grappler:
> WordPress it self not having esc_attr or esc_attr_e function

I am not sure what you mean. You can see these two function below.
https://developer.wordpress.org/reference/functions/esc_attr_e/
https://developer.wordpress.org/reference/functions/esc_attr__/

Chouby:
I renamed `pll_esc_attr()` to `pll_esc_attr__()` to follow the same pattern as similar functions. I also added `pll_esc_attr_e()`

PR #84, opened 2017-01-13
Menue item shows base64 encoded flag image as title tag

jhit:
The menue Item shows the base64 encoded flag image code as title tag for the language link.

### DIFF
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -87,7 +87,7 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 				foreach ( $the_languages as $lang ) {
 					$lang_item = clone $item;
 					$lang_item->ID = $lang_item->ID . '-' . $lang['slug']; // A unique ID
-					$lang_item->title = $options['show_flags'] && $options['show_names'] ? $lang['flag'] . '<span style="margin-left:0.3em;">' . esc_html( $lang['name'] ) . '</span>' : ( $options['show_flags'] ? $lang['flag'] : esc_html( $lang['name'] ) );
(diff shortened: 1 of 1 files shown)


Chouby:
Hi,

I can't merge as the PR introduces a bug. Selecting only "Show flags" in the menu item would display names instead of the expected flags.

The issue you are encountering is a frequent issue found in themes. Such themes are duplicating the html content of the menu item in the title attribute. Whether this practice is good or not is another debate, but it can work only if the menu item contains text only. It cannot work if the item contains html tags such as an image. Basically the theme is not compatible with images in items. Of course, such issue does not occur with Twenty themes.

jhit:
Sorry, but I think I cannot follow you in all points.

First we did run your plugin on a very basic custom made template that was converted from a plain html Template. No fancy stuff there, no manipulation of standard Wordpress menu functionality.
Second we did configure the menue to only show the lanuage flag icon. In that case we saw the title tag to be filled with the flag image code in base64 encoding.
The Image code directly comes from your plugin code at `: ( $options['show_flags'] ? $lang['flag'] : esc_html( $lang['name'] )` and not as you say from some template code.
We checked the change I made on our site and it worked. Not sure what else should go wrong with this small change.

Chouby:
I am quite sure that there is an issue with your theme (or possibly a plugin if you menu is displayed by a plugin).

The line that you proposed to modify does not fill the title attribute but the menu item label. The title attribute is in $item->attr_title. Polylang keep this property empty. WordPress does not use it if it is empty. See https://github.com/WordPress/WordPress/blob/4.7.1/wp-includes/class-walker-nav-menu.php#L156

So if the title attribute is not empty in your case, it means that it is filled by some 3rd party. As I wrote earlier, it is a frequent issue, that theme developers duplicate the item label in the title attribute. This could be ok if the label has only text, but is wrong if the label includes html tags (such as an image).

I tested your code and it displays names instead of flags when I choose to display only flags in the menu item.

jhit:
Ok. Not sure why this is borked in our theme. I'll just keep the coude that works for us and will not bother you any longer ;)
Cheers.

Chouby:
In that case, I suggest that you escape it: `esc_html( $lang['name'] )`

Issue #91, opened 2017-02-02
language selection don't work for menu


hechi:
Hey guys,

Polylang: 2.1
Wordpress: 4.7.2

Behavior: 
On the Top of the admin-page is the language selection, but the links are without the domain (see screenshot) 

![screenshot_20170202_150153](https://cloud.githubusercontent.com/assets/3808354/22553258/59185df8-e95c-11e6-9ed0-5e0fb56c51c4.png)

But the main problem is, that if we select a language in **http://domain.de/de** to **http://domain.de/en** the main menu does not change.

Left side is the Germanpage and right side is the chinesepage both have the english menu. 
![screenshot_20170202_153523](https://cloud.githubusercontent.com/assets/3808354/22553626/bf166f54-e95d-11e6-9bff-8fa3b7537c45.png)

It worked with the previous versions. So i didn't change anything. 
What i currently tried:
* empty cache
* check static pages for the languages 
* downgrade plugin

some suggestions? Which infos did you need :)

Chouby:
Please describe one problem per issue otherwise it's impossible to follow. Did you make tests with a default theme (Twenty Seventen) and no other plugins activated?

hechi:
Hey @Chouby i disabled all plugins (except AuthLDAP) and downloaded the theme and there is the same Problem. I change the language but the Menu is not translated and links to the language before.

hechi:
@Chouby i created a test VM and copied everything so its a clone of the original. So you can have a look 
https://test.sopat.de/en/
* Menu should look like (Home, Products, Contact Us)
https://test.sopat.de/de/
* Menu should look like (Home, Produkte, Kontakt)

I disabled all plugins expect AuthLDAP and Polylang, i removed unnecessary languages and pages.

![screenshot_20170206_111604](https://cloud.githubusercontent.com/assets/3808354/22643162/ad3871c6-ec5d-11e6-89ca-f7181166fe17.png)

hechi:
Little update if i change *URL modifications* from **The language is set from the directory name in pretty permalinks** to **The language is set from content** it works. But this is not what we wanted ^^

![screenshot_20170206_122549](https://cloud.githubusercontent.com/assets/3808354/22645468/6c862a06-ec67-11e6-9d9e-88e99cb34f12.png)

hechi:
oh oh oh... it was my fault :( my nginx proxy did not set the `$_SERVER['HTTP_HOST']` right. Now it works. I am so sry that i bothered you with this.

PR #92, opened 2017-02-06
throw error if $_SERVER['HTTP_HOST'] is not set correctly

hechi:
Accordingly to my own fault in https://github.com/polylang/polylang/issues/91 the polylang throws an error into the logs if `$_SERVER['HTTP_HOST']` is not set (if its empty)

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -314,7 +314,8 @@ class PLL_Admin_Base extends PLL_Base {
 	 * @param object $wp_admin_bar
 	 */
 	public function admin_bar_menu( $wp_admin_bar ) {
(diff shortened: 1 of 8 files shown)


Chouby:
Thanks for your PR.

However I am not sure that it's up to Polylang to fire errors in this case. WordPress itself is using `$_SERVER['HTTP_HOST']` in several places:
See https://github.com/WordPress/WordPress/search?utf8=%E2%9C%93&q=%24_SERVER%5B%27HTTP_HOST%27%5D

Moreover, the WordPress practice that I am following is to log errors only when WP_DEBUG is true. And it that case I expect that you would see PHP notices for an undefined variable.

hechi:
Hey Chouby,

sure it is not really Polylang's problem it self but my wordpress worked fine expect the Polylang plugin. So it is only a suggestion to write this to the logs because it was not obvious that i screwed up the WP config ^^ because the rest worked fine.

Thx for looking into this
Servus

Issue #93, opened 2017-02-07
Option to discard item spacing when calling pll_the_languages


chesio:
Hi Chouby,

A small feature request: It would be nice to have an option to request output of _pll_the_languages_ function without any whitespace. This is useful for styling list items with `display: inline;`.

For the sake of consistence (and backwards compatibility), the option could be named `item_spacing` and accept values `preserve` (default) and `discard` - this is exactly what WordPress introduced for [wp_nav_menu](https://developer.wordpress.org/reference/functions/wp_nav_menu/) in 4.7.

Actually, you might want to adhere to this new option also when rendering language switcher from within _wp_nav_menu_.

Cheers,
Česlav

chesio:
> Actually, you might want to adhere to this new option also when rendering language switcher from within wp_nav_menu.

This remark obviously can be ignored, I should have looked into Polylang code first.

The  _item_spacing_ option is implemented in #95. One note: values for _item_spacing_ are not validated (ie. it can be any string, not only 'discard' and 'preserve'), but any other value than 'discard' results in fall back to default (current) behavior.

Chouby:
My CSS competences are quite poor but I get what I expect if I write
```
.lang-item {
  display: inline;
}
```
I have read https://core.trac.wordpress.org/ticket/35206 and did not understand the benefit. When I added this indent and line breaks, my intention was only to produce more readable html.

> This remark obviously can be ignored, I should have looked into Polylang code first.
Yes. Hopefully the menu language switcher should honor every wp_nav_menu options. So your suggestion would be needed only for the widget when displayed as a list.

chesio:
The problem with `display: inline` is that whitespace around inline elements affects layout (unlike whitespace around block elements). This is often undesirable, for example when you don't want extra space in between language menu items. There are several ways how to [fight it](https://css-tricks.com/fighting-the-space-between-inline-block-elements/), but the most reliable one is to simply not have any whitespace around such elements.

I would provide some examples, but the linked article explains this problem better than I ever could :)

> When I added this indent and line breaks, my intention was only to produce more readable html.

I totally understand and that's why I propose an option to change this behavior instead of complete removal. Also, backwards compatibility is important here (as mentioned in the Trac ticket).

Chouby:
Thanks. The link is indeed very clear and instructive.

PR #95, opened 2017-02-08
Add option to discard item spacing in output of PLL_Switcher.

chesio:
Hi, #93 turns out quite easy to implement, so I'm submitting a PR.


Chouby:
I want to merge this. Could you just modify the comments to avoid a reference to the menu as it does not impact the language switcher outputted in the menu?

chesio:
Sure, I changed the comment to _"whether to preserve or discard whitespace between list items"_ - is it ok?

Chouby:
That's perfect. Thank you.

PR #100, opened 2017-02-20
Space after HTML id

szepeviktor:
https://wordpress.org/support/topic/space-after-html-id/

@Chouby Do you have time to close simple WP.org support topics?


Chouby:
I am sorry. I must admit that I don't read all support requests on the forum. Even if I sometimes pick one or two. Thanks for the PR :)

PR #101, opened 2017-02-21
add body class filter

diggy:
Note: I prefer lowercase, dashed body classes (e.g. `pll-nl-be`), so did not use `sanitize_html_class()` (which would result in e.g. `pll-nl_BE`)

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -31,6 +31,9 @@ class PLL_Filters {
 		// Converts the locale to a valid W3C locale
 		add_filter( 'language_attributes', array( $this, 'language_attributes' ) );
 
+		// Adds prefixed and sanitized locale to body classes
+		add_filter( 'body_class', array( $this, 'body_class' ) );
+
 		// Prevents deleting all the translations of the default category
 		add_filter( 'map_meta_cap', array( $this, 'fix_delete_default_category' ), 10, 4 );
 	}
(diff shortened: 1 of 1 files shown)


chesio:
Hi,

If you want to style your content based on the language, why not use lang attribute of html element? Instead of `body.pll-nl-be` you can use `html[lang="nl-BE"] body` ? I don't think extra class on body element is necessary, it brings nothing new.

diggy:
- WP users are familiar with body classes and their power
- WP itself uses body classes extensively
- The body classes are filterable, thus providing flexibility

chesio:
All true, but not really arguments for adding duplicate information to page body. Someone skilled enough to discover that there is a body class with page language can certainly add a simple filter to functions.php, if he finds out there is no such class:

```php
add_filter('body_class', function (array $classes) {
    $classes[] = str_replace('_', '-', get_locale());
    return $classes;
});
```

Also, having this filter in Polylang means that as soon as you stop using it, you lose the extra class and any rules in your CSS tied to it stop working. Whereas the snippet above or the `html[lang="nl-BE"] body` selector both work even with other multilanguage plugins (or even without any multilanguage plugin at all).

diggy:
Thank you for your opinion, but aren't you slightly overthinking this? It's just some body classes. There won't be any noticeable performance impact because of this change. Also, something does not have to be "new" to be useful. And fwiw, if you switch from Polylang to another multilingual plugin, I'm pretty sure you'll have some other, more severe headaches to deal with than changing a few CSS selectors in your stylesheet. I'm definitely "skilled enough", but still I'd like to see this functionality out of the box. Laziness, it's a good thing.

chesio:
> ...aren't you slightly overthinking this?

No, I prefer (and encourage) generic solutions to problems, so there's less for me to remember (and for others too). Laziness is a good thing, as you say.

Anyway, I said what I had on my mind and ultimately it's up to @Chouby to decide.

Chouby:
I tend to agree with @chesio.

There are least 2 other ways to use language specific styles without Polylang dependency which, in my opinion are better to use than an extra class.

1. The `:lang` pseudo selector is a standard CSS rule. See https://www.w3schools.com/cssref/sel_lang.asp
2. If you have more than a few specific rules, you can use a stylesheet dedicated to one locale, for example en_US.css or fr_FR.css at the root of your theme directory. The stylesheet is automatically loaded by WP.

diggy:
No sweat, wrapped it in a plugin: https://github.com/diggy/polylang-body-class

PR #103, opened 2017-03-03
Fix incorrect usage of `add_action()` in wpml-compat.php

diggy:
cf. https://github.com/polylang/polylang/search?utf8=%E2%9C%93&q=pll_get_strings


Chouby:
Yes. It should work as is, but that's much better for readability.

PR #106, opened 2017-03-13
require is a statement

szepeviktor:
@Chouby Please review all of require/include/_once statements. They may be misused.
Thanks!


Chouby:
Thanks. I have always seen both ways of writing require and include and thought it was indifferent. Reading examples form the PHP doc make it clear that's not.

PR #135, opened 2017-05-19
fixed related posts filter to use slug instead of name

stevekaeser:
Related posts aren't showing up because the value being search against `taxonomy.language.slug` is the `name` instead of the `slug`. This PR fixes that issue by requesting the `slug` from `pll_get_post_language()`.


Chouby:
I guess that `sanitize_title()` is not needed anymore. Do you agree?

Chouby:
Merging this even if I did not succeeded to test it myself (unable to connect to Jetpack server).

PR #141, opened 2017-06-10
use `s` type specifier, escape `$term_id`

diggy:
`esc_attr()` returns a string


Chouby:
You are right. There is probably a useless double conversion string -> int -> string.
However, `esc_attr` is rather overkill for an integer. Using `intval` (and thus keeping %d) should be sufficient.

PR #157, opened 2017-08-04
Fix dropdown rendering

olegstepura:
* Remove url property of `$item`:
   if empty string here - the link leads to 'undefined'
* Add `attr_title` to all items: 
   else `title` property is used and you see some HTML code with `<img src=...` on hover instead of pure text title.

### DIFF
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -80,7 +80,7 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 				// parent item for dropdown
 				if ( ! empty( $options['dropdown'] ) ) {
 					$item->title = $options['show_flags'] && $options['show_names'] ? $this->curlang->flag . '&nbsp;' . esc_html( $this->curlang->name ) : ( $options['show_flags'] ? $this->curlang->flag : esc_html( $this->curlang->name ) );
-					$item->url = '';
+					$item->attr_title = esc_html( $this->curlang->name );
 					$item->classes = array( 'pll-parent-menu-item' );
 					$new_items[] = $item;
 					$offset++;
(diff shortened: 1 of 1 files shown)


Chouby:
Thanks for the PR.

1. It seems that HTML5 allows not to add an `href` attribute for the `a` tag. Does it cause an issue for you? Otherwise not setting `$item->url` creates a link to `mysite/#pll_switcher` which is not very elegant. I would prefer  `$item->url = '#'` which would link to `mysite/#`. What do you think?

2. I am reluctant to add any title attribute because the included information is at best useless, at worst painful for users using a screen reader (See the ticket https://core.trac.wordpress.org/ticket/24766 which decided the removal of title attributes in WP Core). Moreover other menu items would get an attribute title only if it has been filled for each item in Appearence > Menus (empty by default). So there is a high risk not to be consistent with other menus items on most websites.

I guess the title attribute is added by your theme as a copy of the item html (which causes the  `<img src=...`). As explained, I believe that's wrong. I hope some day theme authors will stop doing that. I suggest that you use a filter to just remove this title attribute if it is a copy of the html (hopefully you'll find one after the theme has added it).

olegstepura:
1. Actually `mysite/#pll_switcher` is ok since it can even scroll to an element if it is not visible. I think leaving it so it better.
2. maybe it would be good to set `attr_title` to an empty string. This sounds like best solution to both worlds.

Please let me know if I need to modify my PR

Chouby:
I am not sure that an empty string for `attr_title` will fix all themes copying the item html. But if it does for some of them, then the change will be positive. Let's do like that.

olegstepura:
I updated the PR making `attr_title` an empty string

Chouby:
I did not merged your PR directly because of the remain non empty `attr_title` at line 93

olegstepura:
Oh, sorry. How do we proceed?

Chouby:
No worry. I fixed it myself in https://github.com/polylang/polylang/commit/5876975282c0b16d5a8c3e307209baa717908b36
I just wanted to explain why I did not merge the PR as is.

PR #166, opened 2017-08-24
Catch WP_Error in PLL_Translated_Object::get_object_term()

diggy:


Chouby:
Could you tell me under what circumstances the funtion was passed a WP_Error object?

diggy:
Got the `Empty term` error in CLI context, no logs unfortunately.

PR #185, opened 2017-10-24
Enable admin function in WP_CLI

chrisschrijver:
I wanted to do a copy_post as batch through WP_CLI (polylang pro). I wasn't able to do so since these functions only gets loaded in admin and on ajax requests. This simple change makes it possible to use the function in WP_CLI as well.

### DIFF
--- a/include/class-polylang.php
+++ b/include/class-polylang.php
@@ -123,7 +123,7 @@ class Polylang {
 
 		// Admin
 		if ( ! defined( 'PLL_ADMIN' ) ) {
-			define( 'PLL_ADMIN', defined( 'DOING_CRON' ) || ( is_admin() && ! PLL_AJAX_ON_FRONT ) );
+			define( 'PLL_ADMIN', defined( 'DOING_CRON' ) || ( defined( 'WP_CLI' ) && WP_CLI ) || ( is_admin() && ! PLL_AJAX_ON_FRONT ) );
 		}
 
 		// Settings page whatever the tab


szepeviktor:
@chrisschrijver Do you know https://github.com/dereckson/wp-cli-polylang ?

chrisschrijver:
@szepeviktor I did not, but this plugin does not handle the functionality I needed. I think It would be nice to merge my (pro) features in that plugin. But probably I would face the same issue there. 

The command I created are able to:
- Assign a language to posts that have no language assigned yet (handy if polylang is activated in an existing website)
- (Only for PRO user) Bulk copy content from one language to another.

szepeviktor:
> Assign a language to posts that have no language assigned yet

It may be achieved my piping wp-cli to itself: `list ID-s | set lang meta`

chrisschrijver:
It may. But it's great to have a "shortcut" through a specific command which do al the work for you

diggy:
There's a post duplication feature in https://github.com/diggy/polylang-cli#wp-pll-post-duplicate

Chouby:
Is there any risk that we would need to load the frontend side of Polylang in a WP CLI context?

chrisschrijver:
@diggy Not an option to bulk create missing languages or assign a language to a post that doesn't have one yet

chrisschrijver:
@Chouby Not to my knowlegde. It's already loaded into ajax request, so why not on WP CLI.

szepeviktor:
Thanks.

PR #189, opened 2017-11-07
Added a span wrapper for the current language name

AmitT:
span wrapper added to current language name so its syntax will fit to other language items and will be easier to style.

### DIFF
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -79,7 +79,7 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 
 				// parent item for dropdown
 				if ( ! empty( $options['dropdown'] ) ) {
-					$item->title = $options['show_flags'] && $options['show_names'] ? $this->curlang->flag . '<span>&nbsp;' . esc_html( $this->curlang->name ) . '</span>' : ( $options['show_flags'] ? $this->curlang->flag : esc_html( $this->curlang->name ) );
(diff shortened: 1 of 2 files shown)


Chouby:
I woul rather do the same as what's already done for other items.
See https://github.com/polylang/polylang/blob/2.2.5/frontend/frontend-nav-menu.php#L92
It seems that I forgot this one.

AmitT:
Fixed. 
But, I think we should come with a better solution because margin-left:0.3em will work only for LTR languages.

Chouby:
You are right for RTL languages. What's currently done doesn't work.
The original code was:
```php
$lang['flag'] . '&nbsp;' . esc_html( $lang['name'] )
```
In v1.9, I modified it to the current version:
```php
$lang['flag'] . '<span style="margin-left:0.3em;">' . esc_html( $lang['name'] ) . '</span>'
```
because someone requested the span wrapper as you did. I thought that removing the non breaking space was good but I needed to add the inline style not to break the style of existing sites. And I did not notice that I broke RTL languages :/

One solution could be to keep using the inline style but make it dependant on the text direction.

Another solution could be to add the non breaking space back. But  I would rather keep it outside the span wrapper like this:
```php
$lang['flag'] . '&nbsp;<span>' . esc_html( $lang['name'] ) . '</span>'
```
None of these solutions are perfect. What do you think?

AmitT:
I think the first solution will be easier to override and  that's why it is better.
Actually, (if we leaving the backward compatibility issue) I prefer to add a class to every span and let the developers customize the CSS as they wish.

PR #193, opened 2017-11-15
Fix typos in word whether in phpDoc

pahan35:
Fix other typos which was skipped on previous whether typos fix


Chouby:
I often make the mistake. Thank you!

PR #196, opened 2017-11-15
Fix wrong get lang from url

pahan35:
IMHO there is no better way to avoid mismatch except compare with hosts


Chouby:
Thanks. I noticed that your proposal would break `get_language_from_url()` when used with a parameter. SO I added corresponding tests. But your proposed fix made me think it was even better to reuse the same method already developped for multiple domains.

PR #197, opened 2017-11-16
Fix typos in comments (mostly in frontend and include files)

pahan35:


changelog.txt
```diff
@@ -980,8 +980,8 @@ This version does include important changes in database. More than ever, make a
 = 0.9.4 (2012-10-23) =
 
 * Add Afrikaans translation contributed by [Kobus Joubert](http://translate3d.com/)
-* Add Belarusian translation contributed by [Alexander Markevitch](http://fourfeathers.by/)
-* Add Afrikaans (af) and Belarusian (be_BY) to predefined languages list (automatic download and update of language files won't work)
+* Add Belorussian translation contributed by [Alexander Markevitch](http://fourfeathers.by/)
+* Add Afrikaans (af) and Belorussian (be_BY) to predefined languages list (automatic download and update of language files won't work)
```
Chouby: I believe that Belarusian is the correct word: https://en.wikipedia.org/wiki/Belarusian_language

admin/admin-filters-columns.php
```diff
@@ -323,7 +323,7 @@ public function ajax_update_term_rows() {
 
 		$translations = empty( $_POST['translations'] ) ? array() : explode( ',', $_POST['translations'] ); // collect old translations
 		$translations = array_merge( $translations, $this->model->term->get_translations( (int) $_POST['term_id'] ) ); // add current translations
-		$translations = array_unique( $translations ); // remove doublons
+		$translations = array_unique( $translations ); // remove doubloons
```
Chouby: doublon is a French word meaning "duplicate". So I would rather use "duplicates" rather than "doubloons" (which according to the dictionnary is an old Spanish coin, out of context here).

pahan35:
I'm pleased to help and. That's some new experience for me to work with pull requests, Travis and PHPUnit. And it's exciting at the same time :)

Chouby:
Thanks :)

PR #220, opened 2018-02-16
Remove duplicated entry in list of available PLL_Choose_Lang classes

ocean90:


ocean90:
I don't think that failing test is related to this PR. Maybe a restart helps, @Chouby?

Chouby:
This is intentional. There are 4 options fo the option `force_lang`:
0 => Language set form the content => `PLL_Choose_Lang_Content`
1 => Language set from the directory name in url => `PLL_Choose_Lang_Url`
2 => Language set from the subdomain => `PLL_Choose_Lang_Url`
3 => Language set from the domain name => `PLL_Choose_Lang_Domain`

Is the PR related to some error you experienced?

PR #252, opened 2018-04-23
fix incompatability with php 7.1 (PHP Fatal error: Uncaught Error: []…

JanneAalto:
… operator not supported for strings)


Chouby:
Coud you clarify in which circumstances you encountered this error? It is expected that the parameter 'tax_query' is an array (See https://developer.wordpress.org/reference/classes/wp_query/#taxonomy-parameters). Thus such error should never occur.

PR #257, opened 2018-05-10
UI/UX: Prepend flags to widget titles in admin & customizer

JoryHogeveen:
Enhances the widget admin UI for different widgets per language.

**Todo:** @Chouby
- [x] Update doc version tags
- [ ] Minify JS file
  I don't know which tool you use to minify so I'll leave that up to you!

### Examples:

**Widget admin screen**
image

**Customizer**
image


JoryHogeveen:
@Chouby 
**Todo:** Minify JS file
I don't know which tool you use to minify so I'll leave that up to you!

admin/admin-base.php
```diff
@@ -130,6 +133,18 @@ public function admin_enqueue_scripts() {
 		wp_enqueue_style( 'polylang_admin', plugins_url( '/css/admin' . $suffix . '.css', POLYLANG_FILE ), array(), POLYLANG_VERSION );
 	}
 
+	/**
+	 * Enqueue scripts to the WP Customizer.
+	 *
+	 * @since 1.3.x
```
JoryHogeveen: @Chouby 
**Todo:** Change version to the actual version this would be merged.

JoryHogeveen:
Tests work fine, just a load error in travis on 1 of the jobs.

js/widgets.js
```diff
@@ -0,0 +1,75 @@
...
+			 * Make sure the widget's contents are embedded; normally this is done
+			 * when the control is expanded, for DOM performance reasons.
+			 */
+			control.embedWidgetContent();
+
+			// Now we know for sure the widget is fully embedded.
+			add_flag( control.container.find( '.widget' ) );
+		}
+		wp.customize.control.each( customize_add_flag );
+		wp.customize.control.bind( 'add', customize_add_flag );
```
JoryHogeveen: Could be better to change this to `.on()`?
http://api.jquery.com/bind/

chesio: It could be also better to change `jQuery( document ).ready(function( $ ) {` to just `jQuery( function( $ ) {`
https://jquery.com/upgrade-guide/3.0/#deprecated-document-ready-handlers-other-than-jquery-function

chesio:
Just came to GitHub today to ask for this feature... Would be great to have it in Polylang!

js/widgets.js
JoryHogeveen: @chesio Good one, new handler is set:
```js
( function( $ ) {
    // Code
} ( jQuery ) );
```

The use of `.on()` is not possible since `wp.customize.control` is not a jQuery object.

include/walker-dropdown.php
```diff
@@ -22,9 +22,10 @@ class PLL_Walker_Dropdown extends Walker {
 	function start_el( &$output, $element, $depth = 0, $args = array(), $current_object_id = 0 ) {
 		$value = $args['value'];
 		$output .= sprintf(
-			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
+			"\t" . '<option value="%1$s"%2$s%3$s%4$s>%5$s</option>' . "\n",
 			esc_attr( $element->$value ),
 			method_exists( $element, 'get_locale' ) ? sprintf( ' lang="%s"', esc_attr( $element->get_locale( 'display' ) ) ) : '',
+			( ! empty( $element->flag_url ) ) ? sprintf( ' flag="%s"', esc_attr( $element->flag_url ) ) : '',
```
JoryHogeveen: @Chouby I could add an `is_admin()` check in there so the flag url's only get added in the WP admin. I understand that this walker is used on the frontend as well.

Chouby:
I just tested the feature. It looks nice. I am planning this for the next major release.

JoryHogeveen:
Great, thanks!

JoryHogeveen:
@Chouby Any update on the schedule for PLL 2.4?
Thanks!

include/walker-dropdown.php
JoryHogeveen: @Chouby please let me know if you want the `is_admin()` check in there.

Chouby:
@JoryHogeveen Sorry. I am looking seriously at this only now.

First, the plan for the 2.4 is a beta realease in the first half of October and a final release beginning of November. 

Regarding your PR, I would prefer not access directly to the flag files but rather re-use the encoded flags. The property `$language->flag` already contains a complete `` tag which should avoid us to build a new one it in js. We could export an array with language code as key and `$language->flag` as value with `wp_localize_script`, which in the end should avoid modifying the dropdown walker specifically for the admin. What d you think?

include/walker-dropdown.php
Chouby: See my global comment https://github.com/polylang/polylang/pull/257#issuecomment-420930371

JoryHogeveen:
Hi @Chouby 

> First, the plan for the 2.4 is a beta realease in the first half of October and a final release beginning of November.

Okidoki!

> The property $language->flag already contains a complete  tag which should avoid us to build a new one it in js.

This property won't contain any custom flags the theme might set?

> We could export an array with language code as key and $language->flag as value with wp_localize_script, which in the end should avoid modifying the dropdown walker specifically for the admin. What d you think?

Might work, but your since you enqueue scripts based on the current screen do you want to make this available on all pages where any polylang script is loaded?
Something like:
```php
wp_localize_script( 'polylang_admin', 'pll_admin', array(
    'languages' => pll_languages_list(),
) );
```

JoryHogeveen:
@Chouby 
There is one issue with this because of a limitation in the current language select box.
The `lang` attribute content is the `display` (w3c) style of a locale, not the actual locale: [`$lang->get_locale( 'display' )`](https://github.com/polylang/polylang/blob/master/include/walker-dropdown.php#L27).
This works fine in most cases, but with special locale types it doesn't work. Example: `nl_NL_formal`.

The `display` part renders the w3c compatible locale. So I could simply use that as the flag key in the localized object, but.. if (for whatever reason) a user has both formal and informal (`nl_NL` and `nl_NL_formal`) languages enabled and also uses different flags for them this won't work anymore as the w3c part of these two languages is the same (`nl-NL`).

My suggestion is to add a `data-locale` property to the dropdown on admin pages and use that for flag selection. If you have an better idea, let me know!

Chouby:
Thank you. It looks really good :)

PR #262, opened 2018-06-11
Fix "s.trim is not a function" error

woutervanvliet:
When settings.data is an instance of FormData, wpAjax.unserialize fails. This PR should fix that.

I used `uglifyjs post.js --compress --mangle` to minify post.js, which seems to produce slightly different results from what the minified file looked like originally. Let me know if you want me to minify it in some other way.


woutervanvliet:
@Chouby The fails don't seem to have anything with my PR. Please advice.

manooweb:
@woutervanvliet don't worry about your code. In fact it is our continuous integration system that detected some line indented error in our php code.

Chouby:
@woutervanvliet Thank for your PR. Could you provide a use case to reproduce the error and test your code?

woutervanvliet:
@Chouby To be honest, I'm a little but unsure about the exact trigger for the error - but I'll describe what I know.

In my WordPress installation (in development) I have a few custom types, meta boxes, and the Gutenberg editor. When autosave happens on my custom post type, or when I manually save, the Polylang global ajax handler is invoked and assumes it can unserialize the request data with `wpAjax.unserialize`. 

However, from Gutenberg the request data is not JSON, but a FormData instance - triggering `s.split is not a function`.

Not sure if it also happens on standard Page or Post types, haven't experienced it there - when I submitted this PR I hadn't created any Pages or Posts.

woutervanvliet:
Perfect, thanks

PR #270, opened 2018-06-26
Added PLL_SET_COOKIE_IN_JS helping varnish based caches

gerdneuman:
Some hosters (mine included) offer a [varnish](https://varnish-cache.org/) based caching solution. For instance, [Pantheon](https://pantheon.io/docs/caching-advanced-topics/#using-styxkey) and [gandi.net Simple Hosting](https://doc.gandi.net/en/simple/cache#cookies) offer such a cache. It is great! 

Unfortunately, like most varnish setups these do not cache if there's a `Set-Cookie` in the response.

This patch allows to define a `PLL_SET_COOKIE_IN_JS` to be set in `wp-config.php` and if this is set together with `WP_CACHE` then JS is used for cookie setting instead of the HTTP response. Hence the cookie is not set a-priori in the response but when the page got loaded using JavaScript. So there should be not side effects (as far as I can see, or course)

This means that also the *first-time* visitor response can be cached and delivered be delivered from varnish cache, not just every response *after* the first one. 

This helps:
* in case of a [hug of death](https://en.wikipedia.org/wiki/Slashdot_effect) if you have some media coverage where all the many visitors get to your page the first time
* deliver fast to crawlers like Bingbot or GoogleBot (usually do not keep cookies in between sessions)
* deliver fast to anyone else getting to your website first time

With "fast" I mean 50ms instead of 1500ms as in our case for First Byte Time (backend processing), so approx. 30 times faster and also much better scaled and load balanced since it's a full-fledged cache and not php-fpm processes.

I think the bug surface / side effects of this PR are minimal, but let me know what you think?

### DIFF
--- a/frontend/choose-lang.php
+++ b/frontend/choose-lang.php
@@ -74,15 +74,23 @@ abstract class PLL_Choose_Lang {
 	}
 
 	/**
-	 * set a cookie to remember the language.
-	 * possibility to set PLL_COOKIE to false will disable cookie although it will break some functionalities
+	 * Set a cookie to remember the language.
+	 * Possibility to set PLL_COOKIE to false will disable cookie although it will break some functionalities
+	 *
+	 * Defining PLL_SET_COOKIE_IN_JS will disable cookie setting via Set-Cookie response,
+	 * but instead language cookie can be set with JS in `add_cookie_script` of `modules/plugins/cache-compat.php`.
+	 * This helps `varnish` based caches (cannot cache if `Set-Cookie` present)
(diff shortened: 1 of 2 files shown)


Chouby:
I'd rather not want to introduce a new constant as it makes things more difficult if not impossible to test with phpunit. I propose you a different approach with a filter in #274. Let me know your thoughts. I close this PR in favor of the other.

gerdneuman:
yes, #274 looks fine to me. Thanks and have a nice weekend!

PR #271, opened 2018-06-26
date.setTime() expects ms not s, also make code more robust by closure around vars

gerdneuman:
Fixes two issues with the current code:
1. The setTime() in JS takes *milli*seconds not seconds. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setTime with `represented by a number of milliseconds` So currently instead of 1 year the expiration time was set to something like 4 hours ;)
2. I think it's not a good idea to have a global variable name `date`, which is a very general name. This can lead to all kind of (really hard to debug) problems if another plugins decides to create the a var by the same name. It is better to scope the variable by using a [self-executing anonymous function](http://markdalgleish.com/2011/03/self-executing-anonymous-functions/).


Chouby:
Thanks for the fix. However, the size of the time in miliseconds will lead to an integer overflow on 32 bits installs. I will make the multiplication in javascript instead of PHP to avoid this.

PR #272, opened 2018-06-28
Ensure correct scheme for flag url on load

JoryHogeveen:
The language dropdown in #257 is rendering flag url's without a check for the correct scheme because the handlers are not triggered.
This PR sets the correct scheme for flag url's on `__construct()`.

**Enhancement idea:**
Let me know if you want me to change this.

Currently you are using `is_ssl()` and `str_replace()` to change the scheme.
Example: https://github.com/polylang/polylang/blob/master/include/language.php#L169-L175

Why not use WP core `set_url_scheme()` without a second parameter. Less code and it let's users use the filter: `set_url_scheme` from WP core.
https://developer.wordpress.org/reference/functions/set_url_scheme/

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -87,6 +87,9 @@ class PLL_Language {
 				$this->facebook = $languages[ $this->locale ]['facebook'];
 			}
 		}
+
+		// Ensure correct scheme.
+		$this->flag_url = set_url_scheme( $this->flag_url );
 	}
 
 	/**


Chouby:
@JoryHogeveen I believe that, if there is an issue, this is not the right place for the fix. Indeed, `$this->flag_url` is not always defined at this line (when the language has not been stored in the transient yet).

Today I believe that there should be no issue as Polylang always uses the encoded flag on admin. But since you encountered an issue while developing #257, I'd like to fix it anyway.

However, I did not succeed to reproduce. If I look at the flag url property on an https site, the scheme is correct for me. I expect it to be as [the url is evaluated with `plugins_url()`](https://github.com/polylang/polylang/blob/2.3.10/include/language.php#L102) which applies the correct scheme.

Could you provide me the steps to reproduce?

JoryHogeveen:
Hi @Chouby,

I can't recall exactly how I encountered this. I _think_ this happened because I accessed the `flag_url` property before `set_flag()` was triggered.
This was related to dev of #257 which now uses the base64 images.

Aside from that. `set_url_scheme()` might be good to use on multiple locations.
Like here: https://github.com/polylang/polylang/blob/ac7537c6bb41fdde6df0edb484e257de39a4da56/include/language.php#L211-L219
And here: https://github.com/polylang/polylang/blob/ac7537c6bb41fdde6df0edb484e257de39a4da56/include/language.php#L172-L178
You could simplify this method to 2 lines by using `set_url_scheme()` and with that let WP do the scheme work.

Chouby:
Yes that's true for urls but not for `$this->flag` as this is an html snippet.

Chouby:
I'll close this PR. If you experience the same issue again in the future, feel free to open a new one or an issue. Meanwhile, I will follow your suggestion with `set_url_scheme()` and use relative urls for `$this->flag` as there is no need to have absolute url there.

PR #274, opened 2018-06-29
Allow to load the cache compatibility with a filter

Chouby:
Introduces the filter `pll_is_cache_active` to allow third party developers to load the cache compatibility even when no known plugin is active.


gerdneuman:
@Chouby LGTM, thank you!

PR #299, opened 2018-10-03
Integrate with Events Manager plugin

jasonhildebrand:
This PR implements basic support for the Events Manager plugin, so that Events Manager knows which languages are configured in polylang.  If more than one language is configured in polylang, Events Manager will enable its multilingual support.


jomuji:
Hello Jason,

i have applied your changes ... no luck, i can't get it to work.

PR #319, opened 2018-12-12
Use PHPDoc format for plugin file header

szepeviktor:
See https://codex.wordpress.org/File_Header#Plugin_File_Header_Example

### DIFF
--- a/polylang.php
+++ b/polylang.php
@@ -1,17 +1,24 @@
 <?php
 
 /**
-Plugin Name: Polylang
-Plugin URI: https://polylang.pro
-Version: 2.5
-Author: Frédéric Demarle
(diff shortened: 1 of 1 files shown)


Chouby:
What's the purpose of duplicating the information in a different format? I am using the official doc https://developer.wordpress.org/plugins/the-basics/header-requirements/ and there is no field starting by `@`.

chesio:
@szepeviktor I guess you should name your PR "Use PHPDoc format for plugin file header", then your intention would be more clear.

szepeviktor:
@Chouby Could you decide on this?

Have this combined PHPDoc+WP meta headers is a modern thing.

PR #331, opened 2019-01-28
Added new filter for wpml strings in backend which need to be translated

pascalknecht:
Currently compatibility only spans the frontend for strings. However this poses some problems in dealing with e-mails that can be triggered with certain plugins via backend.
Hereby I provide a filter `pll_always_translate_string` which can be used to translate a string even if in backend. This might need some further work in the admin table.

Would love to get feedback whether you consider this for merging.


Chouby:
It seems to me that it would be rather dangerous to filter and translate options on admin side. There is a high risk to mess the original option.

Chouby:
This should now be solved by #613

PR #332, opened 2019-02-02
Check if language is a valid object before returning

MarcusDalgren:
So we get this error pointing to wp-content/plugins/polylang/include/links.php on line 34 with varying degrees of regularity in some of our projects. The function get_language which is called specifies object|bool as return type so it needs to be checked again before trying to use it as an object.


Chouby:
Could you tell me in which context you receive this error? Are you calling the method directly? There is a API function to get the home url per language `pll_home_url()`. See https://github.com/polylang/polylang/blob/master/include/api.php#L85-L99

PR #340, opened 2019-03-11
Add pll_class filter

epeli:
Hi!

We're working on a WPGraphQL extension for Polylang here:

https://github.com/valu-digital/wp-graphql-polylang

It happens that the `PLL_Frontend` class is instantiated for the GraphQL requests which not really useful for us because it automatically filters posts etc. by the current language which is a concept that does not make any sense in the GraphQL API because one can query multiple things in multiple languages if wanted.

We would like to customize this behaviour (PLL_REST_Request seems better) and maybe pass in a custom subclass of PLL_Base.

Filter like this would help a lot. Thanks.

### DIFF
--- a/include/class-polylang.php
+++ b/include/class-polylang.php
@@ -197,15 +197,22 @@ class Polylang {
 		$class = apply_filters( 'pll_model', PLL_SETTINGS ? 'PLL_Admin_Model' : 'PLL_Model' );
 		$model = new $class( $options );
 		$links_model = $model->get_links_model();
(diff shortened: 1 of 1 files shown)


Chouby:
Hi!

I haven't merged the PR as is due to conflicts. But I took the idea except that I renamed the filter to `pll_context`. It went in 2203b9e16532797fa530f9b73024b53885d728ef and will be included in Polylang 2.6.

epeli:
Awesome, thanks!

PR #343, opened 2019-03-19
Enhancement: Add filters to the language switcher

JoryHogeveen:
This PR will add filters to the language switcher widget and the switcher options to allow developers to extend this widget.

**Filter: `pll_get_switcher_options`**
Filters the available switcher options.

**Filter: `pll_widget_language_switcher`**
Allows to modify the output of the languages switcher widget.

**Action: `pll_widget_language_switcher_form`**
Allows adding fields to the language switcher widget

**Filter: `pll_widget_language_switcher_update`**
Allow to append custom settings to the widget update method


JoryHogeveen:
@Chouby let me know what you think! I need to modify the behavior/style of the switcher a bit for a project and it would be great if that is possible while still using the regular PLL widget.

Chouby:
Hello,

Could you provide details on your plan? Seeing all the filters and the action, I wonder if it wouldn't be easier to extend `PLL_Widget_Languages` . And then keep only one widget by unregistering the Polylang widget as I do for the widget calendar here: https://github.com/polylang/polylang/blob/2.5.2/include/base.php#L45-L46

JoryHogeveen:
Hi @Chouby,

I thought of that at first as well but this solution would make all widgets act the same in every case. In my case I would like to have the option to add some settings/changes.
IMO this makes it easier for users to extend your widget instead of doing a full rewrite.
With this PR I believe I've made extending the widget as flexible as possible without losing backwards compatibility.

PR #359, opened 2019-04-26
Get current/default language object using API

JoryHogeveen:
Currently the API only supports getting a language field.
This PR adds the option to return the current or default language object.

Example:
```php
pll_current_language( OBJECT );
pll_default_language( OBJECT );
```


Chouby:
We can add this. However what do you think to use the constant `OBJECT` instead of the string `'object'`? It seems to me that it would closer to the "WordPress way".

JoryHogeveen:
@Chouby Both is the "WordPress way" since there isn't really a default.

`wpdb::get_results` uses the constants but (for example) `get_taxonomies` uses strings (`names` || `objects`).
Both methods seem fine to me :) if you prefer the constant then lets go for that!

JoryHogeveen:
On second thought I'd prefer the constant as well to make a clear difference with the string type parameter (which is for a specific language field).

PR #368, opened 2019-06-04
Adding alias support for terms join clause.

mch0lic:
Similar to posts join clause; would benefit 3rd party integrations.

### DIFF
--- a/include/translated-term.php
+++ b/include/translated-term.php
@@ -139,11 +139,17 @@ class PLL_Translated_Term extends PLL_Translated_Object {
 	 *
 	 * @since 1.2
 	 *
+	 * @param string $alias Alias for $wpdb->terms table
 	 * @return string join clause
 	 */
-	public function join_clause() {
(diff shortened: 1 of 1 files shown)


include/translated-term.php
```diff
@@ -139,11 +139,17 @@ public function delete_translation( $id ) {
 	 *
 	 * @since 1.2
 	 *
+	 * @param string $alias Alias for $wpdb->terms table
 	 * @return string join clause
 	 */
-	public function join_clause() {
+	public function join_clause( $alias = '' ) {
```
chesio: It's just a nitpicking, but `$alias = 't'` would spare the `if (empty ( $alias )` check below.

mch0lic: Kept the similar coding style as seen in the posts join method. Thanks for merging!

Chouby: @chesio You are right. I will change that.

Issue #361, opened 2019-04-30
Polylang breaks TinyMCE of PODS custom fields


phrench:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] This bug happens with a default WordPress theme active
- [x] I can reproduce this bug consistently

## Steps to reproduce the issue

1. Install the PODS plugin
2. Create a custom post type
3. Add a custom field of the type WYSIWYG (Visual Editor) to the custom post type
4. Add a new entry (of the custom post type)
5. TinyMCE field cannot be access with the mouse cursor. (= actual behavior)
6. Disable Polylang plugin
7. TinyMCE field can now be accessed with the mouse cursor. (= expected behavior)

phrench:
Also noticed that collapsing some metaboxes from plugins such as SEOPress doesn't work anymore. But this is only when editing a custom post type. When editing a normal blog post, collapsing the very same metabox works.

Thus looks like some conflict of Polylang with custom post types or the PODS plugin.

Chouby:
Do you confirm that this is the same issue as https://github.com/pods-framework/pods/issues/5365 and fixd by the PR #369?

phrench:
Confirmed, it's the same issue and fixed by the PR.

When applying the change the issue is gone and the UI works as expected.

Chouby:
Closed by #369

PR #369, opened 2019-06-05
Only append lang to preload paths if the current post has a language.

JoryHogeveen:
Fixes bug with Pods WYSIWYG meta field (and most likely other plugins that use `wp_editor()`: https://github.com/pods-framework/pods/issues/5365

### DIFF
--- a/admin/admin-block-editor.php
+++ b/admin/admin-block-editor.php
@@ -31,6 +31,9 @@ class PLL_Admin_Block_Editor {
 	 */
 	public function preload_paths( $preload_paths, $post ) {
 		$lang = pll_get_post_language( $post->ID );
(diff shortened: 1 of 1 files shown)


Chouby:
I am proposing a new patch. The idea here is to truly verify that the post type is translatable, in order to correctly handle the case where the post is translatable but has no language yet (if it was written before Polylang has been activated).

Is it ok for you like that?

Of course this a bug, but I am surprised by the consequences. I would have expected that he only consequence would be that REST API requested are not corretly preloaded and thus executed again the block editor.

Fix #361

JoryHogeveen:
Hello @Chouby,

I was surprised as well, took me a while before I even thought this could be related at all.
Your addition looks good to me!

PR #391, opened 2019-07-19
fix(static-pages) : Page on front and page for posts are now filtered…

raaaahman:
… to match the current language when querying the corresponding options

May need further testing

Fixes https://github.com/polylang/polylang-pro/issues/257


raaaahman:
Issue is related to the UI, the solution should be related to the UI also.

PR #392, opened 2019-07-22
Reload wpml-config.xml before save post

FloLech:
This changes the behavior of loading the wpml-config.xml to get refreshed before the post is saved.
Originally the config file gets loaded on init() and stays in this state until the class gets reinitialized. This leads to the problem of outdated translations options, when changing wpml-config.xml before save. To overcome this problem it is necessary to reload the wpml-config.xml in the beginning of the following methods:

- copy_post_metas
- copy_term_metas
- translate_types
- translate_taxonomies

The function to load the config file is put into the function load_wpml_config_file and added to the end of the file.

### DIFF
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -55,20 +55,7 @@ class PLL_WPML_Config {
 			}
 		}
 
-		// Theme
-		if ( file_exists( $file = ( $template = get_template_directory() ) . '/wpml-config.xml' ) && false !== $xml = simplexml_load_file( $file ) ) {
-			$this->xmls[ get_template() ] = $xml;
-		}
-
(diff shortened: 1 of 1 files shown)


Chouby:
Thanks for the PR.

However, wpml-config.xml files are aimed to be static files which should not be modified during a page load. Moreover, reading the files at each call to `copy_post_metas` would have a negative performance impact as the filter is called many times. Thus, it is intented that each file is read only once. 

If you need to dynamically change the behavior controlled by wpml-config.xml files, I suggest that you use the filters `pll_copy_post_metas`, `pll_copy_term_metas`, `pll_get_post_types`, `pll_get_taxonomies`, possibly storing options in the datatabase.

PR #393, opened 2019-07-22
feat(core) : Show translated page in static pages dropwdown

raaaahman:
Also, I would verify if update_show_on_front() is still needed 

 Fixes polylang-pro#257, use this one instead of #391 as it has less risks of side effects

### DIFF
--- a/admin/admin-static-pages.php
+++ b/admin/admin-static-pages.php
@@ -7,6 +7,8 @@
  */
 class PLL_Admin_Static_Pages extends PLL_Static_Pages {
 	/**
+	 * Used to build links in admin notices
+	 *
 	 * @var PLL_Links
 	 */
 	protected $links;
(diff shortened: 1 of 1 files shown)


raaaahman:
Need to verify other calls to the filter :

## Dropdown usage

Because the dropdown calls wordpress `get_ppages()` function, it triggers our filter [PLL_Filters::get_pages()](https://github.com/polylang/polylang/blob/249ea8f82d998689e11136250dca689eb86d9baf/include/filters.php#L131-L182)

### Admin pages

[wordpress/wp-admin/options-reading.php](https://github.com/WordPress/WordPress/blob/b77cf42c0beb326a2fe0336272405245f87527dc/wp-admin/options-reading.php#L92-L119)

The 'selected' option comes from a straight `get_option()`.

[wordpress/wp-admin/privacy.php](https://github.com/WordPress/WordPress/blob/96b4570781bbba0a87c0cfa277c6afc65a7efef4/wp-admin/options-privacy.php#L215-L223)

This is handled by  polylang here: [PLL_Filters::translate_privacy_policy_page()](https://github.com/polylang/polylang/blob/249ea8f82d998689e11136250dca689eb86d9baf/include/filters.php#L287-L289)

### Classic editor 

[polylang/admin-classic-editor.php](https://github.com/polylang/polylang/blob/145333d029636ce3bb6ae1d702627c065cfabaa1/admin/admin-classic-editor.php#L194-L212)

Here the parent page SHOULD already be in the language of the child page, no need for finding translation.

### Woocommerce 

Woocommerce ets users define custom Shop, Checkout, Commercial Conditions, etc. These are not translated when the admin dashboard is filtered.

Though, the ones for which there are missing translations are reported in Woocommerce > State / Info page.

The checkout has a special link towards the privacy policy that has lead to the aforementioned PLL\_Filters::translate\_privacy\_policy\_page().

PR #400, opened 2019-08-20
Set cookie with "httponly" set to true

mweimerskirch:
Safari has started to automatically expire cookies after one week unless they are only accessible over the HTTP protocol.
See https://webkit.org/blog/8613/intelligent-tracking-prevention-2-1/
This means Safari users will have to choose their preferred language each week.

The change proposed here will prevent this from happening. The downside is however, that the cookie is no longer accessible via JavaScript.


Chouby:
Unfortunately, we can't support this change. The cookie may be written in Javascript. See https://github.com/polylang/polylang/blob/2.6.8/modules/plugins/cache-compat.php#L33. Also extensions may need to access to the cookie in Javascript too.

PR #408, opened 2019-09-18
Fix: Uncaught Error - Cannot unset string offsets

sawwd:
Hi,

I do not have time to check and open an issue for this but it happens in some cases as expected because maybe_unserialize returns mixed. 

Thanks,


Marianne380:
@sawwd Thank for your PR. Could you provide a use case to reproduce the error ?

Chouby:
Closed in favor of #641

PR #422, opened 2019-11-14
Escape output for vip

rfair404:
A little bit of context. I'm trying to run this plugin in a VIP Go environment. The phpcs scanner is complaining about a lot of things, but the escaping is one that can't get past code review. I added escaping functions everywhere that the scanner was complaining about. While this PR may not be the right approach, it would be great if you could bring the code quality in this plugin up to speed with what VIP requires.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -240,23 +240,23 @@ class PLL_Admin_Base extends PLL_Base {
 					$.ajaxPrefilter(function (options, originalOptions, jqXHR) {
 						if ( -1 != options.url.indexOf( ajaxurl ) || -1 != ajaxurl.indexOf( options.url ) ) {
 							if ( 'undefined' === typeof options.data ) {
-								options.data = ( 'get' === options.type.toLowerCase() ) ? '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>' : <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>;
+								options.data = ( 'get' === options.type.toLowerCase() ) ? '<?php esc_html_e( $str ); ?>' : <?php esc_html_e( $arr ); ?>;
 							} else {
 								if ( 'string' === typeof options.data ) {
 									if ( '' === options.data && 'get' === options.type.toLowerCase() ) {
-										options.url = options.url+'&<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>';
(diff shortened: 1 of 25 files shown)


Chouby:
It looks like there is a misunderstanding of what `esc_html` or similar functions do.
For example, when you replace:
`echo '<div class="calendar_wrap">';`
by
`esc_html_e( '<div class="calendar_wrap">' );`
The output will be:
`&lt;div class=&quot;calendar_wrap&quot;&gt;`
That means that the browser will display `<div class="calendar_wrap">` to the visitor instead of opening a div.

I took this example because, it happens that this line in Polylang is an exact copy of WordPress: https://github.com/WordPress/WordPress/blob/5.3/wp-includes/widgets/class-wp-widget-calendar.php#L62 and, as the string is hardcoded, it's easier to understand the effect. 

In other places, you escaped variables that were already previously escaped, resulting in the same kind of effect. For example the html code of links are displayed instead of the links int the posts list table or in the metabox. 

Our plugin includes phpunit tests. Running the tests would have shown you many errors. Even if you dd not run the phpunit tests, visiting the admin should have shown you many broken things.

Late escaping often goes against DRY. That's why, after careful check, we whitelist PHPCS escaping errors. So PHPCS shouldn't have warned you about these non escaped strings. As you can see in our travis config, we are running PHPCS tests, even if some rules are excluded.

Moreover, you often use `esc_html_e()`. This function not only escapes a string but also translates it. It means that the function will look for the string in the WordPress translations files and returns the translation if it finds it. As strings like `<div class="calendar_wrap">` are not in the translations, this of course does not work. PHPCS should have indirectly warned you about this misuse, because in our plugin, you must always pass the text domain as second parameter of `esc_html_e()` or other translations functions, and you did not pass a second parameter to this function. 

That said, escaping is related to security. Unless people want to put hundreds of thousands of websites at risk, we expect that, when a developer discovers a security issue, he contacts us privately rather than opening a public PR or discussion.

rfair404:
Hey, thank you for the reply. My intention isn't to imply that this plugin isn't secure, or that anyone is at risk. The bigger issue, that I'm asking for the development team of this plugin to consider, is making the plugin compliant with WP VIP standards, which I realize can often be difficult to do. 

As it stands currently, it isn't possible to use this plugin in a VIP environment, without having their automated code reviews complain about a lot of the code in the plugin. It seems that it was at one point approved for VIP, however recent changes may mean that you can't install the current version. 

Can you perhaps suggest a route towards how we could get this into a VIP environment?

PR #425, opened 2019-11-22
Add AMP compatibility for accessing front page

westonruter:
The AMP plugin uses an `amp` query var to indicate that the AMP version of a page should be served when the plugin is configured with the Transitional or Reader template modes. This query var causes Polylang to think that a request for `/?amp` is not actually the homepage due to the presence of that `amp` query var. So this PR adds compatibility for this scenario: it includes the `amp` query var among the list of query vars that are subtracted from the query vars in the current query which should be ignored for front page queries.

Originally reported at https://wordpress.org/support/topic/static-home-page-with-polylang-2/

### DIFF
--- a/modules/plugins/plugins-compat.php
+++ b/modules/plugins/plugins-compat.php
@@ -124,6 +124,11 @@ class PLL_Plugins_Compat {
 		if ( function_exists( 'custom_post_widget_plugin_init' ) && class_exists( 'PLL_Content_Blocks' ) ) {
 			add_action( 'pll_init', array( $this->content_blocks = new PLL_Content_Blocks(), 'init' ) );
 		}
+
+		// AMP
+		if ( function_exists( 'amp_get_slug' ) ) {
+			add_filter( 'pll_front_page_ignored_query_vars', array( $this, 'add_amp_to_front_page_ignored_query_vars' ) );
+		}
 	}
 
 	/**
@@ -333,5 +338,16 @@ class PLL_Plugins_Compat {
 			// Otherwise rely on MU Domain Mapping
 			redirect_to_mapped_domain();
 		}
+	}
+
+	/**
+	 * Append the AMP query var to the supplied query vars.
+	 *
+	 * @param array $query_vars Query vars.
+	 * @return array Query vars with AMP added.
+	 */
+	public function add_amp_to_front_page_ignored_query_vars( $query_vars ) {
+		$query_vars[] = amp_get_slug();
+		return $query_vars;
 	}
 }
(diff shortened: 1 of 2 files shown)


swissspidy:
@Chouby Hey Frédéric, what are your thoughts on adding a filter to `PLL_Frontend_Static_Pages::is_front_page()` as done in this PR?

modules/plugins/plugins-compat.php
```diff
@@ -124,6 +124,11 @@ public function plugins_loaded() {
 		if ( function_exists( 'custom_post_widget_plugin_init' ) && class_exists( 'PLL_Content_Blocks' ) ) {
 			add_action( 'pll_init', array( $this->content_blocks = new PLL_Content_Blocks(), 'init' ) );
 		}
+
+		// AMP
+		if ( function_exists( 'amp_get_slug' ) ) {
+			add_filter( 'pll_front_page_ignored_query_vars', array( $this, 'add_amp_to_front_page_ignored_query_vars' ) );
```
raaaahman: Since you propose to add a filter hook, wouldn't it make sense to register this filter callback that triggers one of AMP functions, inside the AMP plugin?

swissspidy: This is a great question!

The problem for the AMP plugin is that adding all the plugin compatibility layers is not really scalable and difficult to maintain. That's why the AMP plugin has always been trying to advocate to other plugins and themes in the ecosystem to handle that in their projects; with the AMP plugin serving as the "framework" to support that.

So I'd argue to keep this in Polylang. If you ever change something in this logic, people wouldn't have to wait for an accompanying change in the AMP plugin but could instead just update Polylang alone and it would _just work_. Does that  makes sense for you?

cc @westonruter in case you want to add something

westonruter: Yeah, that sounds right. Also, Polylang is already adding compatibility for other plugins, so it seems to fit nicely here.

Nikita-Sp:
Hi!
As I understand there is no this fix in the current public update, right?
Thanks.

Nikita-Sp:
@swissspidy Hi!

I'm sorry, you've got two updates in a row this days, and I'm tired a bit to add this code manually.
Maybe you could review and add it to release already? Maybe I don't understand something, but it seems everyone just forgot about this tread.

Thanks.

swissspidy:
@Nikita-Sp I don't work on Polylang. You might want to ping someone else :-)

Nikita-Sp:
@manooweb please review this update. We're adding this code from version to version manually.
Thanks.

jamesozzie:
Hi @manooweb,

Any update on this? It would be great to know Polylang is fully AMP compatible. I use it on a couple of my own sites and it's great, only in the case of a homepage as above. This came up in a recent [support topic on the AMP plugins support forums](https://wordpress.org/support/topic/polylang-and-amp-homepage-problem/).

PR #428, opened 2019-11-29
Filter to modify widget properties for string translation.

JoryHogeveen:
Currently there is no way (without unnessesary custom code) to add widget properties for string translation.
The "heavy" word is already done within Polylang, with this filter plugin developers like myself can easily integrate their widgets with Polylang through the filter `pll_translate_widget_properties`.

The only thing they will have to do besides this is apply `pll__` within their code where applicable.


JoryHogeveen:
Hi @Chouby 
I see this PR has not been merged yet, any feedback on this? Let me know!

PR #430, opened 2019-12-05
Use preg_quote() to escape regular expression characters

ocean90:
Also, replace the `/` delimiter with `#` for consistency and to avoid the need for escaping slashes which makes reading the patterns not so easy. Using `preg_quote()` ensures that dots in domains are also escaped.

### DIFF
--- a/include/links-directory.php
+++ b/include/links-directory.php
@@ -59,10 +59,10 @@ class PLL_Links_Directory extends PLL_Links_Permalinks {
 		if ( ! empty( $lang ) ) {
 			$base = $this->options['rewrite'] ? '' : 'language/';
 			$slug = $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : $base . $lang->slug . '/';
-			$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '/^https?:\/\//', '://', $this->home . '/' . $this->root );
+			$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '#^https?://#', '://', $this->home . '/' . $this->root );
 
 			if ( false === strpos( $url, $new = $root . $slug ) ) {
-				$pattern = str_replace( '/', '\/', $root );
+				$pattern = preg_quote( $root, '#' );
 				$pattern = '#' . $pattern . '#';
 				return preg_replace( $pattern, $new, $url, 1 ); // Only once
 			}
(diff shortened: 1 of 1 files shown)


Chouby:
@ocean90 Thank you for your PR. It's true that I uselessly escaped `/` characters by ignorance. Not only here. Did you encounter a practical case where the unescaped `.`  character caused a bug?

include/links-directory.php
```diff
@@ -89,10 +89,10 @@ public function remove_language_from_link( $url ) {
 		}
 
 		if ( ! empty( $languages ) ) {
-			$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '/^https?:\/\//', '://', $this->home . '/' . $this->root );
+			$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '#^https?://#', '://', $this->home . '/' . $this->root );
 
-			$pattern = str_replace( '/', '\/', $root );
-			$pattern = '#' . $pattern . ( $this->options['rewrite'] ? '' : 'language\/' ) . '(' . implode( '|', $languages ) . ')(\/|$)#';
+			$pattern = preg_quote( $root, '#' );
+			$pattern = '#' . $pattern . ( $this->options['rewrite'] ? '' : 'language/' ) . '(' . implode( '|', $languages ) . ')(/|$)#';
```
raaaahman: Shouldn't we also escape `$this->option['rewrite']` and '$languages`, or the whole `$pattern` after line 95?

include/links-directory.php
```diff
@@ -117,8 +117,8 @@ public function get_language_from_url( $url = '' ) {
 		$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : $this->home . '/' . $this->root;
 
 		$pattern = wp_parse_url( $root . ( $this->options['rewrite'] ? '' : 'language/' ), PHP_URL_PATH );
-		$pattern = str_replace( '/', '\/', $pattern );
-		$pattern = '#' . $pattern . '(' . implode( '|', $this->model->get_languages_list( array( 'fields' => 'slug' ) ) ) . ')(\/|$)#';
+		$pattern = preg_quote( $pattern, '#' );
+		$pattern = '#' . $pattern . '(' . implode( '|', $this->model->get_languages_list( array( 'fields' => 'slug' ) ) ) . ')(/|$)#';
```
raaaahman: Same remark: shouldn't we escape the whole `$pattern`?

include/links-directory.php
ocean90: `$this->option['rewrite']` isn't used, `$languages` can be escaped but it's only a list of slugs (`[a-z-]`) so won't cause issues.

ocean90:
@Chouby I only noticed this when writing my own link model. `.` basically means one occurrence of any character. Can't think of a domain which may cause  issues here. One line is about path escaping (`get_language_from_url()`) which is more likely to cause issues , through, I don't have any practical examples right now. I guess it's more of a "better safe than sorry" thing.

include/links-directory.php
Chouby: I don't think so. This would escape the pipes and parentheses.

PR #456, opened 2020-02-14
Replace option to remove all data at uninstall by a PHP constant

Chouby:
This PR removes the Tools module in Polylang settings. This module was used only for the option to remove all data at uninstall (when clicking on the red delete link in the plugins list table).

The goal is to avoid that a user accidentally removes Polylang data by checking an option that was too easy to check and forget.

The option is  replaced by a PHP constant that can be added to the wp-config.php when removing all data is absolutely needed.

This is inspired by what's done by WooCommerce with the constant `WC_REMOVE_ALL_DATA`. See: https://github.com/woocommerce/woocommerce/blob/3.9.2/uninstall.php#L28

NB: This also means that on a multisite, the option is not anymore controlled for each site individually.


tests/phpunit/tests/test-install.php
```diff
@@ -63,20 +67,22 @@ function test_uninstall() {
 
 		update_post_meta( $item_id, '_pll_menu_item', array() );
 
+		$this->assertFalse( defined( 'PLL_REMOVE_ALL_DATA' ), 'The constant PLL_REMOVE_ALL_DATA must not be defined for this test'  );
```
raaaahman: I don't believe we should perform assertions on the test environment: we need to be absolutely sure about how we are running our test. Anyway, if the constant was defined, the test would fail at the next assertion...

tests/phpunit/tests/test-install.php
```diff
@@ -63,20 +67,22 @@ function test_uninstall() {
...
+
 		if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
 			define( 'WP_UNINSTALL_PLUGIN', true );
 		}
 
 		include_once dirname( dirname( dirname( dirname( __FILE__ ) ) ) ) . '/uninstall.php';
 		new PLL_Uninstall();
```
raaaahman: Actually, the instantiation of this class as already been done in https://github.com/polylang/polylang/blob/0a50edc71c80ce29582a0991ab22013ebe6ba706/uninstall.php#L159 , that you trigger the line above.

Also, this class only contains one method, that is executed upon instantiation... That makes me believe that it is actually a script that you wrapped into a class, but you don't get OOP benefits here, and this doesn't really make the code clearer IMO.

uninstall.php
```diff
@@ -20,6 +20,11 @@ class PLL_Uninstall {
 	public function __construct() {
 		global $wpdb;
 
+		// Don't do anything except if the constant PLL_REMOVE_ALL_DATA is explicitely defined and true.
+		if ( ! defined( 'PLL_REMOVE_ALL_DATA' ) || ! PLL_REMOVE_ALL_DATA ) {
+			return;
```
raaaahman: See my previous comment.

Returning from a constructor outlines the fact that this class isn't used in an OOP way. I believe we should never perform any logic in a class constructor, as this makes them quite hard to instantiate in our automated tests.

tests/phpunit/tests/test-install.php
```diff
@@ -63,20 +67,22 @@ function test_uninstall() {
...
 
-		// Option Uninstall = true
-		$options['uninstall'] = 1;
-		update_option( 'polylang', $options );
+		// Constant PLL_REMOVE_ALL_DATA = true
+		define( 'PLL_REMOVE_ALL_DATA', true );
+
 		new PLL_Uninstall(); // Fire uninstall process again
 
 		// No options
```
raaaahman: The following assertions would be better as part of another test, as you have two opposite cases here:

- The case were we **don't** want to remove the data
- The case were we want to remove our data

Each of these cases might be working independently from the other, so two tests would provide better information here.

tests/phpunit/tests/test-install.php
```diff
@@ -16,6 +16,10 @@ function test_activate() {
 		$this->assertEquals( POLYLANG_VERSION, $options['version'] );
 	}
 
+	function test_no_uninstall() {
+		$this->assertFalse( defined( 'PLL_REMOVE_ALL_DATA' ), 'The constant PLL_REMOVE_ALL_DATA mut not be defined for this test'  );
```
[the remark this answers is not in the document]
Chouby: That was indeed the intention. Not as a useful test by itself but as a helper for future tests developments if the developer doesn't initially pay attention to the constant. If you believe that it's useless, then we can remove it.

tests/phpunit/tests/test-install.php
Chouby: Same as above.

tests/phpunit/tests/test-install.php
Chouby: I did not change anything here. Classes used as namespaces. All of this has been written at at time where we had to support PHP 5.2.

uninstall.php
Chouby: I did not want to rewrite everything. Just add this constant to control the uninstall. Although it has downsides for testing.

tests/phpunit/tests/test-install.php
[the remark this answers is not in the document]
raaaahman: Okay, I think I get it: because of constants being globals, some tests couldn't be performed if the constant is already set.

Maybe it would be clearer to use `PHPUnit::markTestSkipped()` with a corresponding message in the `setUp()` function of this class, like the[ example from PHPUnit documentation.](incomplete-and-skipped-tests-skipping-tests-examples-databasetest-php)

tests/phpunit/tests/test-install.php
```diff
@@ -16,10 +16,29 @@ function test_activate() {
 		$this->assertEquals( POLYLANG_VERSION, $options['version'] );
 	}
 
+	/**
+	 * This test requires the definition of the constants WP_UNINSTALL_PLUGIN
+	 * The constant PLL_REMOVE_ALL_DATA must not be defined
+	 */
 	function test_no_uninstall() {
```
raaaahman: The test looks good to me. Maybe we could now rename it, because it tests the uninstall, but without removing the data.

So maybe: 'test_uninstall_without_removing_data'?

tests/phpunit/tests/test-install.php
```diff
@@ -16,10 +16,29 @@ function test_activate() {
...
+		// Constant PLL_REMOVE_ALL_DATA undefined => nothing deleted
+		$options = get_option( 'polylang' );
+		$this->assertNotEmpty( $options );
 	}
 
+	/**
+	 * This test requires the definition of the constants WP_UNINSTALL_PLUGIN and PLL_REMOVE_ALL_DATA
+	 * It is expected that only this test defines PLL_REMOVE_ALL_DATA
+	 */
 	function test_uninstall() {
```
raaaahman: Same as previous comment: 'test_uninstall_do_remove_data' ?

tests/phpunit/tests/test-install.php
```diff
@@ -67,19 +86,10 @@ function test_uninstall() {
 
 		update_post_meta( $item_id, '_pll_menu_item', array() );
 
-		$this->assertFalse( defined( 'PLL_REMOVE_ALL_DATA' ), 'The constant PLL_REMOVE_ALL_DATA must not be defined for this test'  );
-
 		if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
```
raaaahman: Since you defined it in `PLL_Install_Test::test_uninstall()`, no need to do it here. Besides, you already need `PLL_Install_Test::test_uninstall()` to be executed before this one, because the `PLL_REMOVE_ALL_DATA` constant set in this code would break the test if they were performed in the other order.

It seems that test are performed in sequential order in the class they're written on. Otherwise, you might use the [@depends](https://phpunit.de/manual/5.7/en/appendixes.annotations.html#appendixes.annotations.depends) annotation if tests needs to be played in a specific order (beware, if the dependency test fails, the dependent test would be skipped).

tests/phpunit/tests/test-install.php
```diff
@@ -16,10 +16,29 @@ function test_activate() {
 		$this->assertEquals( POLYLANG_VERSION, $options['version'] );
 	}
 
+	/**
+	 * This test requires the definition of the constants WP_UNINSTALL_PLUGIN
+	 * The constant PLL_REMOVE_ALL_DATA must not be defined
+	 */
 	function test_no_uninstall() {
-		$this->assertFalse( defined( 'PLL_REMOVE_ALL_DATA' ), 'The constant PLL_REMOVE_ALL_DATA mut not be defined for this test'  );
+		if ( ! defined( 'WP_UNINSTALL_PLUGIN' ) ) {
```
raaaahman: This still bothers me that we don't know whether this constant is defined or not in our environment. In fact, it might even be better to define it anyway, so if the test outputs an error, it will give us an insight on our test environment.

Chouby: Do you mean taht you would prefer to define the constant `WP_UNINSTALL_PLUGIN` in the boostrap once and for all?

tests/phpunit/tests/test-install.php
Chouby: The 2 tests are not really dependent on each other. So I prefer to keep theme completely separated.

tests/phpunit/tests/test-install.php
raaaahman: On second thought, no, that would be less explicit. Just let it as it is.

Issue #452, opened 2020-02-02
Homepage redirect should include "Vary: Accept-Language" header


chesio:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] This bug happens with a default WordPress theme active
- [x] I can reproduce this bug consistently

## The issue

When Polylang performs [302 redirect](https://github.com/polylang/polylang/blob/master/frontend/choose-lang.php#L288) for home page request, the response may be cached at various places (fastcgi cache at webserver, proxy cache between webserver and browser, browser's cache). However, the URL being redirected to may depend on the value of `Accept-Language` header sent by browser, therefore the response should include `Vary: Accept-Language` header. This response header signals to all the different caches that the response may differ when `Accept-Language` header in request differs (see this [thread](https://stackoverflow.com/questions/36925218/which-http-headers-should-i-set-for-multi-language-site) on SO).

## Expected behavior and actual behavior

I encountered this issue when I have set up fastcgi cache at my Nginx server - by default, all 301 and 302 redirects are cached too. Without the `Vary: Accept-Language` header, the first redirect response is cached and then served to all requests to homepage (ie. without language), so all users are redirected to the same language version. With the header, the cache is only hit when the value of `Accept-Language` header matches.

manooweb:
Hello @chesio,

Thank you for reporting this issue.
 
We did the modification to add `Vary: Accept-Language` header in the response when a 302 redirection occurs.

It's difficult for us to test because we don't have any environment quickly and easily ready.

Could you please test it? You can find code changes in our PR #459 or in the specific branch here https://github.com/polylang/polylang/blob/fix%23452/frontend/choose-lang.php#L288

Otherwise do you know a cache plugin which also caches 302 redirection?

Thank you very much.

chesio:
Hello @manooweb,

I can confirm that the PR #459 fixes the issue when deployed to my server (I've tested it with Polylang Pro 2.6.10 with the patch from the PR applied).

Thanks a lot for addressing the issue!

manooweb:
Hello @chesio,

[...]

PR #459, opened 2020-02-19
Add Vary: Accept-Language http header

manooweb:
Add `Vary: Accept-Language` http header when a 302 redirection occurs on home page.

See #452


manooweb:
@chouby We had the feedback from the customer see #452 latest comments. Maybe we can put this modification in the 2.6.x branch ?

Chouby:
I prefer to keep this in 2.7.

PR #460, opened 2020-02-20
Twenty Seventeen: Translate header video Youtube url

### LINKED ISSUE #420: Header video youtube url not translatable in Twenty Seventeen
Twenty Seventeen has an option to set a header video from an external url in Appearance > Customize > Header Media. It is not possible to have a different url per language.

See: https://wordpress.org/support/topic/polylang-translation-of-header-video/

Chouby:
Twenty Seventeen has a specific setting to add a Youtube video as header. This PR allows to use a different url per language by adding the url to the strings translations.
Fix #420

### DIFF
--- a/modules/plugins/plugins-compat.php
+++ b/modules/plugins/plugins-compat.php
@@ -266,10 +266,18 @@ class PLL_Plugins_Compat {
 	 * @since 2.0.10
 	 */
 	public function twenty_seventeen_init() {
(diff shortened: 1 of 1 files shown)


manooweb:
👍
It's strange to go back to strings translations to change add URL for the other language but I guess that it is more complicated to do it directly in the customizer. It will need to be documented somewhere in our support documentation.
Otherwise I have no further remarks.

PR #471, opened 2020-03-25
Forbid access to wizard steps when no language is defined

manooweb:
After removing the condition for the WooCommerce step and fix the https://github.com/polylang/polylang-wc/issues/263 issue, @chouby suggest to forbid access to steps which needs that at least one language is defined.

This is the goal og this PR to forbid access to all the wizard steps except for "Licenses" and "Languages" steps of course.


manooweb:
There is no other places where we test if no language is defined. We will also always can cheat by bypassing the `Content ` step and have a homepage with no language assigned. So I'd prefer to keep the test just above the one you notice in your comment.

manooweb:
> I would also remove useless code such as https://github.com/polylang/polylang/blob/2.7-beta2/modules/wizard/view-wizard-step-home-page.php#L109-L115

Done in the last commit.

PR #472, opened 2020-03-25
Fix deactivate license not working in settings

Chouby:
This PR solves a conflict created by an undesirable ajax response sent from the wizard in settings.
Fix https://github.com/polylang/polylang-pro/issues/465


manooweb:
The button is now really reactivated in settings and triggers the ajax request but on my side I have always an http 403 error on a WP install. On an another install, it works but only after refreshing the settings page. Indeed, just after I validate the licence key the `deactivate` button appears but doesn't work as in the https://github.com/polylang/polylang-pro/issues/465 .

So this PR partially corrected the problem.

PR #474, opened 2020-03-26
Apply _wp_specialchars to the name when checking if a term exists

Chouby:
When inserting a term, the filter `_wp_specialchars()` is applied to the term name (through the `sanitize_term()` method.
For example, `Cook & eat` is transformed to `Cook &amp; eat`. Thus we also need to apply this filter when comparing term names in our `term_exists()` method.

NB: This is not done in the original `term_exists()` function from WordPress (this probably should). However, the comparison is made first on the term slug, so in most cases, the slug comparison will match, thus bypassing the name comparison. So it's less probable to encounter the issue with the WP function, compared to our function which checks only the term name. To reproduce the bug with the WP function, you would need to change the slug to something not default.

Fix https://github.com/polylang/polylang-wc/issues/218

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -428,6 +428,7 @@ class PLL_Model {
 		global $wpdb;
 
 		$term_name = trim( wp_unslash( $term_name ) );
+		$term_name = _wp_specialchars( $term_name );
 
 		$select = "SELECT t.term_id FROM $wpdb->terms AS t";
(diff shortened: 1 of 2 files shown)


Chouby:
A ticket is already open for the same issue in WP: https://core.trac.wordpress.org/ticket/47351

PR #475, opened 2020-03-26
Add $languages initialization wrongly removed

manooweb:
wrongly removed by this commit https://github.com/polylang/polylang/pull/471/commits/927609795cb8e1792a13a2549d2d39a7fc9cabf8

Closes https://github.com/polylang/polylang-pro/issues/468

and potentially avoid an issue when main homepage has no language yet.

We need to improve this case by enforcing the assignement of a language to all the Content.
See https://github.com/polylang/polylang-pro/issues/469


manooweb:
Strange, I rebased before pushing. I will make another PR

PR #477, opened 2020-03-27
Enforce content step as next steps depend on it

manooweb:
In fact unlike I said in the https://github.com/polylang/polylang-pro/issues/469 the message display when there is one language and already assigned to the homepage is necessary to warn the user that I will have to translate the future homepage in each language he will add.

To be sure the homepage has a language we now enforce to pass through the `Content` step before.

Because the homepage will now always have a language assigned, some code is now useless in the `create_home_page_translations` method. Then I removed this useless code.

This modification is also needed in Polylang Pro (wizard-pro.php). Another PR is then made in its repository.

Closes https://github.com/polylang/polylang-pro/issues/469

### DIFF
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -186,7 +186,7 @@ class PLL_Wizard {
 		$home_page_id = get_option( 'page_on_front' );
 		$home_page_language = $this->model->post->get_language( $home_page_id );
 
-		if ( ! $home_page_language && 'home-page' === $this->step ) {
+		if ( ! $home_page_language && in_array( $this->step, array( 'home-page', 'wc-pages' ) ) ) {
 			wp_safe_redirect( esc_url_raw( $this->get_step_link( 'untranslated-contents' ) ) );
 			exit;
 		}
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -186,7 +186,7 @@ class PLL_Wizard {
 		$home_page_id = get_option( 'page_on_front' );
 		$home_page_language = $this->model->post->get_language( $home_page_id );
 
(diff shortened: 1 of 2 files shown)


modules/wizard/wizard.php
```diff
@@ -183,6 +183,13 @@ public function setup_wizard_page() {
 			exit;
 		}
 
+		$home_page_id = get_option( 'page_on_front' );
+		$home_page_language = $this->model->post->get_language( $home_page_id );
+
+		if ( ! $home_page_language && 'home-page' === $this->step ) {
+			wp_safe_redirect( esc_url_raw( $this->get_step_link( 'untranslated-contents' ) ) );
+			exit;
+		}
```
Chouby: Shouldn't we protect all steps after the content step as we did for the languages? This would protect the WooCommerce step too.

manooweb: Done.

manooweb:
> I suggest to simplify this line: https://github.com/polylang/polylang/blob/2.7-beta2/modules/wizard/view-wizard-step-home-page.php#L46

Done too 👍

modules/wizard/wizard.php
```diff
@@ -183,6 +183,13 @@ public function setup_wizard_page() {
 			exit;
 		}
 
+		$home_page_id = get_option( 'page_on_front' );
+		$home_page_language = $this->model->post->get_language( $home_page_id );
+
+		if ( ! $home_page_language && ! in_array( $this->step, array( 'licenses', 'languages', 'media', 'untranslated-contents' ) ) ) {
```
Chouby: I propose to check `$this->model->get_objects_with_no_lang( 1 )` rather than only the language of the home page. Because the issue is the home page language only for the homepage step, but for other steps, the issue is fr other pages (WC pages).

manooweb: I did like that at the beginning before pushing the first time 😁

PR #481, opened 2020-03-31
Fix database error when the setup wizard is launched the first time

manooweb:
Fix by adding the same test as in PLL_Settings::notice_objects_with_no_lang method to ensure at least on language is defined to test if untranslated contents exist.

Closes https://github.com/polylang/polylang-pro/issues/474

### DIFF
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -184,7 +184,7 @@ class PLL_Wizard {
 			exit;
 		}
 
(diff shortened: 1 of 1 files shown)


Chouby:
Testing `! empty( $this->options['default_lang'] )` is an indirect way to check that at least one language has been defined to avoid the database error in `PLL_Admin_Model::get_objects_with_no_lang()`. What do you think of testing `count( $languages )` instead?

manooweb:
I wondered and I chose the consistency with PLL_Settings::notice_objects_with_no_lang. However you're right is more consistent with the other condition in the wizard.

PR #495, opened 2020-04-12
Prevent undefined key error with default_lang

davisshaver:
Interacting with a site via WP CLI, if Polylang is activated but not yet configured with the install wizard, a warning can be thrown due to accessing an unset `default_lang` key. E.g. https://github.com/polylang/polylang/blob/master/admin/admin-base.php#L335

### DIFF
--- a/install/install.php
+++ b/install/install.php
@@ -92,6 +92,7 @@ class PLL_Install extends PLL_Install_Base {
 			'domains'          => array(),
 			'version'          => POLYLANG_VERSION,
 			'first_activation' => time(),
+			'default_lang'     => '',
 		);
 	}
 


Chouby:
Hello,

Could you provide the steps to reproduce?

I expect that this line is not executed if there is no languages yet due to this test: https://github.com/polylang/polylang/blob/2.7.1/admin/admin-base.php#L61-L63 and the default language should be set as soon as the first language is created. https://github.com/polylang/polylang/blob/2.7.1/admin/admin-model.php#L51-L54

davisshaver:
Sure! It happens under very specific circumstances.

- Plugin is enabled via `mu-plugins`
- Polylang option does not exist yet
- Activation hook was not run because of being enabled with `mu-plugins`

I want to note that this can be worked around by modifying the `mu-plugins` loader to only load the plugin if the `polylang` option is defined, but I still think it might be worth addressing with this patch.

I believe the reason this happens is that although Polylang will [backfill with default options](https://github.com/polylang/polylang/blob/master/include/class-polylang.php#L220-L223) if the `polylang` option isn't set on a request, other areas of the codebase will expect the `default_lang` key to exist.

The use case here is interacting with a site programmatically for CI. We have a CLI command to add the Polylang languages but it is noisy with errors if we don't activate the plugin via the traditional method. I believe adding this key would enable the use case of enabling via `mu-plugins` without needing to do the safety check against `polylang` option.

Chouby:
I have setup a WP fresh install and put Polylang in the mu-plugins directory. The polylang option is not present in the database. What should I do then to reproduce the error? For the test, I have run a random CLI command `wp core update` and I got no error.

davisshaver:
@Chouby In `mu-plugins` is there a loader file that is including polylang?

Chouby:
I am not 100% why you ask the question, but I confirm that the plugin is correctly loaded and the behavior is what I expect when no languages are defined.

davisshaver:
No problem! I think there needs to be a loader file at top-level in `mu-plugins` that is including Polylang as that would be the case where the activation hook isn't hit. Not clear if you are saying that's what you have and it's not an issue or maybe it was installed/activated some other way?

PR #497, opened 2020-04-15
Fix a fatal error with InfiniteWP

Chouby:
I don't understand how it's possible, but it looks like under some circumstances, InfiniteWP is messing with the actions `wp_loaded` and `admin_init` causing objects used by `PLL_Sync_Post` (mainly `PLL_Sync_Post_Model` and `PLL_Bulk_Translate` not to be instantiated). The symptom seen by customers is a fatal error in the `PLL_Sync_Post` constructor.

Adding a check for `$polylang->bulk_translate` in the constructor would solve the fatal error woudl fix it but would leave other notices. I found out it was better to prevent the instantiation of `PLL_Sync_Post` if `PLL_Sync_Post_Model` hasn't been instantaited first (knowing that PLL_Bulk_translate is instatiated in the same `add_filters` method.

To reproduce the issue, I am only guessing as the customers did not provide the steps. Thus I used this:
```php
add_action( 'wp_loaded', function() {
	do_action('admin_init');
}, 1);
```
Although it is *not* the code of InfiniteWP, I hope it reproduces the same behavior. 

Fix https://github.com/polylang/polylang-pro/issues/488

### DIFF
--- a/admin/admin.php
+++ b/admin/admin.php
@@ -190,7 +190,8 @@ class PLL_Admin extends PLL_Admin_Base {
 	/**
 	 * Load the post synchronization object, depending on the editor in use.
 	 *
-	 * We must make sure to instantiate the class only once, as the function may be called from a filter and that the synchronization model has been instantiated (due to InfiniteWP messing the actions wp_loaded and admin_init).
+	 * We must make sure to instantiate the class only once, as the function may be called from a filter,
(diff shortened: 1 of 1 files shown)


raaaahman:
Since this method exists only to instantiate the correct class for our synchronization module, I'd suggest we write the different cases directly in its documentation block, rather than the comment.

About the fix, this seems to be an issue that a dependency injection container would solve.

PR #504, opened 2020-05-07
Prevent cookie hijacking

tbolender:
Our site gets regularly scanned by a vulnerability scanner. Since we employ Polylang, our CVSS rating has dramatically increased. We found out that Polylang employs cookies without the restriction to HTTP usage (`httpOnly`). This allows [session hijacking](https://www.owasp.org/index.php/HttpOnly).

While it is possible to hack relabeling in configuration for PHP or nginx, this is not a solution for everybody. I am totally aware that the language setting is not super sensible. It is just not a good practice in my option.

I propose to simply set the parameter for `httpOnly` to true when setting the cookie. I added it to the cookie function and the JavaScript cache fallback. This allows us to use Polylang without any manual hacking to receive a good security assessment.

P.S. Thanks a lot for your work, it made translating our page very comfortable!

### DIFF
--- a/frontend/choose-lang.php
+++ b/frontend/choose-lang.php
@@ -102,7 +102,8 @@ abstract class PLL_Choose_Lang {
 				time() + $expiration,
 				COOKIEPATH,
 				2 == $this->options['force_lang'] ? wp_parse_url( $this->links_model->home, PHP_URL_HOST ) : COOKIE_DOMAIN,
-				is_ssl()
+				is_ssl(),
+				true
 			);
 		}
 	}


modules/plugins/cache-compat.php
```diff
@@ -36,7 +36,7 @@ public function add_cookie_script() {
 			'(function() {
 				var expirationDate = new Date();
 				expirationDate.setTime( expirationDate.getTime() + %d * 1000 );
-				document.cookie = "%s=%s; expires=" + expirationDate.toUTCString() + "; path=%s%s";
+				document.cookie = "%s=%s; expires=" + expirationDate.toUTCString() + "; path=%s%s; HttpOnly";
```
chesio, line 39: I don't think this will work - according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#JavaScript_access_using_Document.cookie):

> Cookies created via JavaScript cannot include the `HttpOnly` flag.

Would be little weird if it had worked when the point of `HttpOnly` flag is that you cannot access cookies via `document.cookie`.

tbolender: To be honest, I was not sure about it and must have skipped the sentence during research. I thought it might work as a write-only operation. I will remove it.

chesio: You still have the problem remaining that this snippet will stop working at all with cookies that are HTTP only :-) I don't know if Polylang authors are willing to sacrifice this piece of functionality.

tbolender: You mean that a cookie set this way is not protected? True, but at least the default cookie is protected. I do not know about a way to fix it and I would leave it this way for now.

chesio: Sorry, disregard my last comment. I somehow assumed that cookie value needs to be readable in JS too, but this is not the case.

Chouby:
This PR is a duplicate of #400 that I closed because, even if Polylang itself doesn't read the cookie with Javascript, there are extensions doing this.

PR #513, opened 2020-06-02
Add tests to detect the modification of WordPress functions

Chouby:
This PR adds phpunit tests to check if WordPress functions that we have copied with or without modifications need to be updated.

This is done by reading the WordPress function or method, evaluating its md5 and comparing it to the last known md5. The test is done only for WordPress versions older than a defined versions to avoid errors when testing the backward compatibility with older versions of WordPress.

The PR also updates the methods get_calendar() and admin_bar_search_menu() with the
code base of WP 5.4.1.

### DIFF
--- a/frontend/frontend-filters-search.php
+++ b/frontend/frontend-filters-search.php
@@ -73,8 +73,8 @@ class PLL_Frontend_Filters_Search {
 	}
 
 	/**
-	 * Rewrites the admin bar search form to pass our get_search form filter. See #21342
-	 * Code base last checked with WP 4.9.7
+	 * Rewrites the admin bar search form to pass our get_search_form filter. See #21342.
+	 * Code last checked: WP 5.4.1.
 	 *
 	 * @since 0.9
 	 *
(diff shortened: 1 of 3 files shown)


Chouby:
I agree with the separation of the tests. But I would prefer keeping the version check (including now the call to `markTestSkipped()` in `check_method()` rather than writing it multiple times. I would also separate the tests for `get_calendar()` and `WP_Widget_Calendar::widget()`.
>Running them each time we modify Polylang might be superfluous, we should instead run at each WordPress realease
We need to know before the release. And since these tests run fast, I don't believe it's worth to seperate them.

raaaahman:
> We need to know before the release. And since these tests run fast, I don't believe it's worth to separate them.

On local, we will need to manually update our WordPress environment whenever we want to test if changes have been made to these functions. They will otherwise pass and we will just forget they exists.

On Travis, they could potentially fail for reasons external to the PR we are trying to push. We will then need to double check if the tests that are failing are some tests that are allowed to fail or not...

There are few of these tests for now, but we might also add more in the future (like the tests we used to have for YOAST).

raaaahman:
I've requested reviews from @manooweb @Marianne380 as well, because we all need to be aware that these tests exists, whatever the final decision is here.

Chouby:
I added a commit with proposal I made above.

> On Travis, they could potentially fail for reasons external to the PR we are trying to push. We will then need to double check if the tests that are failing are some tests that are allowed to fail or not...

They will not only "potentially" fail. They *will* fail out of context and that's the goal. These tests are not standard tests. Their goal is to alert us when a potentially impacting change occurs. And I have no better tool than Travis for that. 

Note however that the PR will not be marked as errored, because we allow the tests to fail with the WordPress development version. And I expect that we will fix all such issues before the final version is released; This however requires that we manually check Travis results for the development version during the beta period to make sure that every tests will pass when the final version is released. I of course already do that as there are many other reasons why WP deveopments could break our tests.

raaaahman:
> Note however that the PR will not be marked as errored, because we allow the tests to fail with the WordPress development version.

Okay, that was my concern.

PR #516, opened 2020-06-08
Displays flags instead of dashicons-yes icon in list tables

### LINKED ISSUE #515: Display the flags on post list
This idea come from @webaxones on [Helpscout](https://secure.helpscout.net/conversation/1112039236/9759) (private)
The main idea is to replace the tick with the language flag:
polylang
For history, this issue was previously opened on Polylang Pro. https://github.com/polylang/polylang-pro/issues/528

sebastienserre:
Add flags on posts & term lists
Closes #515


admin/admin-filters-columns.php
```diff
@@ -1,7 +1,4 @@
 <?php
```
manooweb, line 1: Strange because it makes a WPCS error but I haven't this error on local on my branches where I haven't the `@package Polylang` too

manooweb:
It seems that unit tests is broken by these modifications.

sebastienserre:
@manooweb / @Chouby I don't know what to do on this PR. Could you please check it ?

Chouby:
Rebased and force-pushed to fix conflicts.

PR #517, opened 2020-06-09
Add abbr wrapper

manooweb:
After the correction of https://github.com/polylang/polylang-pro/issues/526 in this PR https://github.com/polylang/polylang-pro/pull/531 where I decided to wrap the language slug (when a language has no flag) in a `abbr` tag, this PR propose to do the same on legacy metaboxes (post and term) and on colums on WordPress lists (posts and taxonomies) for consistency.

### Post
image

image

### Term and Taxonomy
image

image

image

**Edit**: we decided no to use title attribute for accessibilty reasons https://trello.com/c/PFmIHrC7/115-what-will-we-do-for-accessibility 
It's the goal of the last commit to remove the title attribute https://github.com/polylang/polylang/pull/517/commits/d955521396bbd60b115c937e454236cac11d3213
I decided to keep the `abbr` tag if we will decide to add a tooltip in the future.

### DIFF
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -66,7 +66,7 @@ class PLL_Admin_Filters_Columns {
 		}
 
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[ 'language_' . $language->slug ] = $language->flag ? $language->flag . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>' : esc_html( $language->slug );
+			$columns[ 'language_' . $language->slug ] = ( $language->flag ? $language->flag : sprintf('<abbr title="%1$s">%2$s</abbr>', esc_attr( $language->name ), esc_html( $language->slug ) ) ) . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
 		}
 
 		return isset( $end ) ? array_merge( $columns, $end ) : $columns;
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -66,7 +66,7 @@ class PLL_Admin_Filters_Columns {
 		}
 
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[ 'language_' . $language->slug ] = $language->flag ? $language->flag : '<abbr>' . esc_html( $language->slug ) . '</abbr>' . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
+			$columns[ 'language_' . $language->slug ] = $language->flag ? $language->flag : '<abbr>' . esc_html( $language->slug ) . '</abbr><span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
 		}
 
 		return isset( $end ) ? array_merge( $columns, $end ) : $columns;
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -66,7 +66,7 @@ class PLL_Admin_Filters_Columns {
 		}
 
 		foreach ( $this->model->get_languages_list() as $language ) {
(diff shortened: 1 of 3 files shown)


admin/admin-filters-columns.php
```diff
@@ -66,7 +66,7 @@ protected function add_column( $columns, $before ) {
 		}
 
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[ 'language_' . $language->slug ] = $language->flag ? $language->flag . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>' : esc_html( $language->slug );
+			$columns[ 'language_' . $language->slug ] = ( $language->flag ? $language->flag : sprintf( '<abbr title="%1$s">%2$s</abbr>', esc_attr( $language->name ), esc_html( $language->slug ) ) ) . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
```
raaaahman, line 69: I find this hard to read. I would prefer that we compute the logic and generate the output separately. Consider the following example:

```php
if ( false === $language->flag ) {
  $tag           = 'abbr';
  $attributes = array( 'title' => $language->name );
  $content    = $language->name;
}

$element = $document->createElement( $tag, $content );
foreach( $attributes as $name => $value ) {
  $element->setAttribute( $name, $value );
}

return $document->saveHTML();
```

This code use PHP's built-in `DOMDocument` as an example, but it may be switched for some less verbose HTML parsing / generating library now that we use composer to manage our dependencies.

admin/view-translations-post.php
```diff
@@ -34,7 +34,27 @@
 		}
 		?>
 		<tr>
-			<th class = "pll-language-column"><?php echo $language->flag ? $language->flag : esc_html( $language->slug ); // phpcs:ignore WordPress.Security.EscapeOutput ?></th>
+			<th class = "pll-language-column">
+			<?php
+			if ( $language->flag ) {
+					printf(
```
raaaahman, line 40: I don't find that mixing HTML and PHP's output writing functions makes the clearest code to read. Could we just use one or the other?

admin/admin-filters-columns.php
```diff
@@ -66,7 +66,7 @@ protected function add_column( $columns, $before ) {
 		}
 
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[ 'language_' . $language->slug ] = $language->flag ? $language->flag . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>' : esc_html( $language->slug );
+			$columns[ 'language_' . $language->slug ] = $language->flag ? $language->flag : '<abbr>' . esc_html( $language->slug ) . '</abbr><span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
```
Chouby, line 69: This removes the accessibilty text when displaying a flag. Could you fix?

admin/view-translations-post.php
```diff
@@ -34,7 +34,7 @@
 		}
 		?>
 		<tr>
-			<th class = "pll-language-column"><?php echo $language->flag ? $language->flag : esc_html( $language->slug ); // phpcs:ignore WordPress.Security.EscapeOutput ?></th>
+			<th class = "pll-language-column"><?php echo $language->flag ? $language->flag : '<abbr>' . esc_html( $language->slug ) . '</abbr><span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';  // phpcs:ignore WordPress.Security.EscapeOutput ?></th>
```
Chouby, line 37: There is no accessibility text for the flag because there is one mentioning the language later in the row. I propose not to add the accessibility text for this abbreviation.

Chouby:
Not a bug. Punting to 2.8.

admin/admin-filters-columns.php
manooweb: Hum 🤔 `alt` attribute included in the `$language->flag` is enough for accessibility and thus there is no need to duplicate this information.

manooweb:
> Not a bug. Punting to 2.8.

I propose to postponed later than 2.8 to discuss and process all accessibilty improvment in one time everywhere

manooweb:
We've not discussed about this subject anymore, So I propose to close this PR.

PR #518, opened 2020-06-09
Add attribute title on pencil icon on legacy metabox

manooweb:
During the correction of polylang/polylang-pro#526 in this PR polylang/polylang-pro#531 where I decided to wrap the language slug (when a language has no flag) in a abbr tag, I noticed that no title attribute is set on pencil icon on legacy metabox unlike on plus icon.
This PR propose to add the title attribute as on the plus icon.

### post
image

### term
image


manooweb:
I close this PR because as we decided we need to avoid use of title attribute for accessibility reason.
We need to think to another solution to display a tooltip but this an enhancement.

PR #520, opened 2020-06-10
Move instantiation of CRUD classes to PLL_Base

Chouby:
Replaces https://github.com/polylang/polylang-pro/pull/516

Compared to the original PR:
* I wrapped the instantiation in a test to check if there are some languages.
* I added some documentation for the other PLL_Base properties.

### DIFF
--- a/include/base.php
+++ b/include/base.php
@@ -9,7 +9,40 @@
  * @since 1.2
  */
 abstract class PLL_Base {
-	public $links_model, $model, $options;
+	/**
+	 * Stores the plugin options.
+	 *
+	 * @var array
+	 */
+	public $options;
+
+	/**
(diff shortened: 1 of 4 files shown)


include/base.php
```diff
@@ -41,6 +74,11 @@ public function __construct( &$links_model ) {
 	 * @since 2.6
 	 */
 	public function init() {
+		if ( $this->model->get_languages_list() ) {
```
raaaahman, line 77: I have a problem with writing a condition in a method that is documented to 'instantiate the classes always needed'. I don't know what to change though (the method doc comment maybe).

Chouby:
I just force pushed a rebase on the master to solve conflicts with a change to the `PLL_Base::init()` description to better match what the method does.

PR #522, opened 2020-06-10
site-health

sebastienserre:
Add 1 test (if the front page are correctly translated) and information like Options and languages settings in the Site Health information tab

### DIFF
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -85,18 +85,12 @@ class PLL_Admin_Site_Health {
 					$value = __( 'N/A', 'polylang' );
 				}
 
-				$to_be_removed = apply_filters(
-					'pll_site_heath_ignore_list',
-					array(
-						'flag' => true, // remove the flag as filter only display plain text
-						'host' => true, // Key not used by Polylang yet
-					)
-				);
-				if ( $to_be_removed[ $key ] ) {
+				if ( in_array( $key, array( 'flag', 'host' ) ) ) {
 					continue;
 				}
 					$fields[ $key ]['label']   = $key;
 					$fields[ $key ]['value']   = $value;
+
 			}
 			if ( empty( $language->flag ) ) {
 				$language->flag = __( 'Undefined', 'polylang' );
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -21,13 +21,13 @@ class PLL_Admin_Site_Health {
 	protected $model;
 
 	/**
-	 * A reference to the PLL_Admin_Links instance.
+	 * A reference to the PLL_Admin_Static_Pages instance.
 	 *
 	 * @since 2.8
 	 *
-	 * @var PLL_Admin_Links
+	 * @var PLL_Admin_Static_Pages
 	 */
-	protected $links;
+	protected $static_pages;
 
 	/**
 	 * PLL_Admin_Site_Health constructor.
@@ -36,7 +36,7 @@ class PLL_Admin_Site_Health {
 	 */
 	public function __construct( &$polylang ) {
 		$this->model = &$polylang->model;
-		$this->links = &$polylang->links;
+		$this->static_pages = &$polylang->static_pages;
 
 		// Information tab.
 		add_filter( 'debug_information', array( $this, 'info_options' ), 15 );
@@ -242,41 +242,26 @@ class PLL_Admin_Site_Health {
 	 */
 	public function homepage_test() {
 		$result = array(
-			'label'       => __( 'All languages have a translated home page', 'polylang' ),
+			'label'       => __( 'All languages have a translated homepage', 'polylang' ),
 			'status'      => 'good',
 			'badge'       => array(
 				'label' => __( 'Polylang', 'polylang' ),
+				'color' => 'blue',
 			),
 			'description' => sprintf(
 				'<p>%s</p>',
-				__( 'A website can\'t be displayed without homepage.', 'polylang' )
+				__( 'It is mandatory to translate the static front page in all languages.', 'polylang' )
 			),
 			'actions'     => '',
 			'test'        => 'pll_homepage',
 		);
-		$untranslated = array();
-		foreach ( $this->model->get_languages_list() as $language ) {
-			if ( ! $language->page_on_front ) {
-				$untranslated[] = sprintf(
-					'<a href="%s">%s</a>',
-					esc_url( $this->links->get_new_post_translation_link( get_option( 'page_on_front' ), $language ) ),
-					esc_html( $language->name )
-				);
-			}
-		}
-		if ( ! empty( $untranslated ) ) {
-			$result['status'] = 'critical';
-			$result['label'] = __( 'Translation of Home page missing in one or more languages', 'polylang' );
-			$result['description'] = sprintf(
-			/* translators: %s is a comma separated list of native language names */
-				esc_html__( 'You must translate your static front page in %s.', 'polylang' ),
-				implode( ', ', $untranslated ) // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-			);
-			$result['actions'] .= '';
-			$result['badge']       = array(
-				'label' => __( 'Polylang', 'polylang' ),
-				'color' => 'blue',
-			);
+
+		$message = $this->static_pages->get_must_translate_message();
+
+		if ( ! empty( $message ) ) {
+			$result['status']      = 'critical';
+			$result['label']       = __( 'The homepage is not translated in all languages', 'polylang' );
+			$result['description'] = sprintf( '<p>%s</p>', $message );
 		}
 		return $result;
 	}
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -8,13 +8,15 @@ if ( ! defined( 'ABSPATH' ) ) {
 } // Exit if accessed directly.
 
 /**
- * Class PLL_Admin_Site_Health
+ * Class PLL_Admin_Site_Health to add debug info in WP Site Health
  *
- * @since
+ * @since 2.8
  */
 class PLL_Admin_Site_Health {
 	/**
 	 * PLL_Admin_Site_Health constructor.
+	 *
+	 * @since 2.8
 	 */
 	public function __construct() {
 		add_filter( 'debug_information', array( $this, 'pll_info_options' ), 15 );
@@ -22,10 +24,11 @@ class PLL_Admin_Site_Health {
 	}
 
 	/**
-	 * @param $debug_info array
+	 * Add Polylang Options to Site Health Informations tab.
 	 *
-	 * @return mixed
-	 * @since
+	 * @param array $debug_info array options to display.
+	 * @return array
+	 * @since   2.8
 	 */
 	public function pll_info_options( $debug_info ) {
 		$options = get_option( 'polylang' );
@@ -34,12 +37,12 @@ class PLL_Admin_Site_Health {
 			if ( ! is_array( $value ) ) {
 				switch ( $key ) {
 					case 'first_activation':
-						$fields[ $key ]['label']   = __( $key, 'polylang' );
+						$fields[ $key ]['label']   = $key;
 						$fields[ $key ]['value']   = date_i18n( 'd/m/Y', $value );
 						$fields[ $key ]['private'] = false;
 						break;
 					default:
-						$fields[ $key ]['label']   = __( $key, 'polylang' );
+						$fields[ $key ]['label']   = $key;
 						$fields[ $key ]['value']   = $value;
 						$fields[ $key ]['private'] = false;
 						break;
@@ -47,11 +50,11 @@ class PLL_Admin_Site_Health {
 			}
 			if ( is_array( $value ) ) {
 				if ( empty( $value ) ) {
-					$fields[ $key ]['label']   = __( $key, 'polylang' );
+					$fields[ $key ]['label']   = $key;
 					$fields[ $key ]['value']   = __( 'N/A', 'polylang' );
 					$fields[ $key ]['private'] = false;
 				} else {
-					$fields[ $key ]['label']   = __( $key, 'polylang' );
+					$fields[ $key ]['label']   = $key;
 					$fields[ $key ]['value']   = implode( ', ', $value );
 					$fields[ $key ]['private'] = false;
 				}
@@ -67,6 +70,13 @@ class PLL_Admin_Site_Health {
 		return $debug_info;
 	}
 
+	/**
+	 * Add Polylang Languages settings to Site Health Informations tab.
+	 *
+	 * @param array $debug_info array options to display.
+	 * @return array
+	 * @since   2.8
+	 */
 	public function pll_info_languages( $debug_info ) {
 		$languages = PLL()->model->get_languages_list();
 		$fields = array();
@@ -75,12 +85,17 @@ class PLL_Admin_Site_Health {
 				if ( empty( $value ) ) {
 					$value = __( 'N/A', 'polylang' );
 				}
-				// remove the flag as filter only display plain text
-				if ( 'flag' !== $key ) {
-					$fields[ $key ]['label']   = __( $key, 'polylang' );
+
+				$to_be_removed = array(
+					'flag' => true, // remove the flag as filter only display plain text
+					'host' => true, // Key not used by Polylang yet
+				);
+				if ( $to_be_removed[ $key ] ) {
+					continue;
+				}
+					$fields[ $key ]['label']   = $key;
 					$fields[ $key ]['value']   = $value;
 					$fields[ $key ]['private'] = false;
-				}
 			}
 			if ( empty( $language->flag ) ) {
 				$language->flag = __( 'Undefined', 'polylang' );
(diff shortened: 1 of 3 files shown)


modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	/**
+	 * Add Polylang Options to Site Health Informations tab.
+	 *
+	 * @param array $debug_info array options to display.
+	 * @return array
+	 * @since   2.8
+	 */
+	public function pll_info_options( $debug_info ) {
```
Chouby, line 38: You don't need to prefix the method nam with `pll_`.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+		$options = get_option( 'polylang' );
+		$fields = array();
+		foreach ( $options as $key => $value ) {
+			if ( ! is_array( $value ) ) {
+				switch ( $key ) {
+					case 'first_activation':
```
Chouby, line 44: I am not sure that it's useful to display this option. It useful only to know when displaying the review notice.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+						$fields[ $key ]['label']   = $key;
+						$fields[ $key ]['value']   = date_i18n( 'd/m/Y', $value );
+						$fields[ $key ]['private'] = false;
+						break;
+					default:
+						$fields[ $key ]['label']   = $key;
+						$fields[ $key ]['value']   = $value;
+						$fields[ $key ]['private'] = false;
+						break;
```
Chouby, line 53:
```php
$fields[ $key ] = array(
	'label'   => $key,
	'value'   => $value,
	'private' => false,
);
```
would be more readable to me. Let say that's in our coding standards although wie have no WPCS sniff to check this ;-).

Reading the post you linked to, it looks like the `private` key is not necessary when to false.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+				}
+			}
+			if ( is_array( $value ) ) {
```
Chouby, line 56: Using `else` rather than making a second test would look more logical to me.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+					$fields[ $key ]['value']   = implode( ', ', $value );
+					$fields[ $key ]['private'] = false;
+				}
+			}
+		}
+
+		$debug_info['polylang'] = array(
+			'label'    => __( 'Polylang Options', 'polylang' ),
+			'fields' => $fields,
+			'show_count' => true,
```
Chouby, line 72: I believe that we should remove the count.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+					$value = __( 'N/A', 'polylang' );
+				}
+
+				$to_be_removed = array(
+					'flag' => true, // remove the flag as filter only display plain text
+					'host' => true, // Key not used by Polylang yet
+				);
+				if ( $to_be_removed[ $key ] ) {
+					continue;
+				}
```
Chouby, line 100: This code raises lots of `PHP Notice:  Undefined index` at line 98.
I suggest to simplify this by checking directly `in_array( $key, array( 'flag', 'host' ) )`
I suggest removing other keys: 'taxonomy', 'description', 'parent', 'filter'.

We should discuss if we need to remove 'flag' and 'custom_flag'.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+					$fields[ $key ]['value']   = $value;
+					$fields[ $key ]['private'] = false;
+			}
+			if ( empty( $language->flag ) ) {
+				$language->flag = __( 'Undefined', 'polylang' );
+			}
+			$debug_info[ $language->name ] = array(
+				'label'  => $language->name,
+				// translators: placeholder is the flag image
+				'description' => sprintf( __( 'Flag used in the language switcher: %s', 'polylang' ), $language->flag ),
```
Chouby, line 111: This may be not true. If the user has setup a custom flag, then the language switcher will display the custom flag, not this one.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+				'fields' => $fields,
+				'show_count' => true,
```
Chouby, line 113: I propose removing the count.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+				}
+					$fields[ $key ]['label']   = $key;
+					$fields[ $key ]['value']   = $value;
+					$fields[ $key ]['private'] = false;
+			}
+			if ( empty( $language->flag ) ) {
+				$language->flag = __( 'Undefined', 'polylang' );
+			}
+			$debug_info[ $language->name ] = array(
+				'label'  => $language->name,
```
Chouby, line 109: I propose to extend the label to something like 'Language: %s'.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+	 * Add a Site Health test on Home Page translation
+	 *
+	 * @param array $tests array with tests declaration data
+	 * @return array
+	 * @since   2.8
+	 */
+	public function pll_is_homepage( $tests ) {
+		// add test only if static page on front page.
+		if ( '0' !== get_option( 'page_on_front' ) ) {
+			$tests['direct']['pll_hp'] = array(
```
Chouby, line 129: I prefer full names rather than abbreviations. Here `pll_home_pages` would be clearer.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+				'label' => __( 'Home Page Translated', 'polylang' ),
```
Chouby, line 130: Please don't capitalize the strings.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+	 * @return array $result array with test results
+	 * @since 2.8
+	 */
+	public function pll_homepage_test() {
+		$result = array(
+			'label'       => __( 'All languages have a translated home page', 'polylang' ),
+			'status'      => 'good',
+			'badge'       => array(
+				'label' => __( 'i18n', 'polylang' ),
+				'color' => 'green',
```
Chouby, line 149: It seems that WordPress doesn't use specific colors. Maube we should not use the too.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+	 *
+	 * @return array $result array with test results
+	 * @since 2.8
+	 */
+	public function pll_homepage_test() {
+		$result = array(
+			'label'       => __( 'All languages have a translated home page', 'polylang' ),
+			'status'      => 'good',
+			'badge'       => array(
+				'label' => __( 'i18n', 'polylang' ),
```
Chouby, line 148: As you already mentionned yourself, 'Polylang' would be a better label than 'i18n'.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+			$result['description'] = sprintf(
+			/* translators: %s is a comma separated list of native language names */
+				esc_html__( 'You must translate your static front page in %s.', 'polylang' ),
+				implode( ', ', $untranslated ) // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			);
+			$result['actions'] .= sprintf(
+				'<p><a href="%s">%s</a></p>',
+				esc_url( admin_url( 'edit.php?post_type=page' ) ),
+				__( 'Translate Missing Home Page', 'polylang' )
+			);
```
Chouby, line 180: The description already displays link to translate the pages (which open directly the editor). This on looks uselless to me.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
...
+			$result['badge']       = array(
+				'label' => __( 'i18n', 'polylang' ),
+				'color' => 'red',
```
Chouby, line 183: Same a absove for the label and color.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,188 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+} // Exit if accessed directly.
```
Chouby, line 8: Useless for a class definition where there is no directly executable code.

sebastienserre:
As I've already created the class in the module directory, I don't understand what I've to do to match your general comment.

Chouby:
[Site health](https://trello.com/c/u4Lycx1o/95-site-health)

Chrystll:
I noticed that the "Polylang" badge is not surrounded as the others, see:

screenshot-localhost_10016-2020 06 25-16_12_33

Chouby:
> As I've already created the class in the module directory, I don't understand what I've to do to match your general comment.

Forget. I misread the code :(

Chrystll:
[media_support]
Currently in the Poylang Diag plugin the value of media_support is empty when the Media module is deactivated. We should get 0. If the Media module is acivated we get 1 as expeted. Could you please see how we can get 0 when the Media module is deactivated?

[uninstall]
The Uninstall module doesn't exist anymore. I think that we can remove this line.

[nav_menus]
In the Polylang Options of the Site healh info the array is empty whereas I get the menus informations in the Polylang diag, see:
screenshot-localhost_10016-2020 06 25-16_43_04
screenshot-localhost_10016-2020 06 25-16_43_36

Could you please add these same informations in Polylang Options?

modules/site-health/admin-site-health.php
sebastienserre: I don't know if having the flag and flag url is relevant in support request.

modules/site-health/admin-site-health.php
sebastienserre: @Chouby is flag_url modified when a custom flag used ?

modules/site-health/admin-site-health.php
sebastienserre: done

Chrystll:
Languages block
If I change the name of a language I can't edit it anymore, see:
screenshot-localhost_10016-2020 06 25-16_53_03 (1)

screenshot-localhost_10016-2020 06 25-16_52_29

sebastienserre:
> I noticed that the "Polylang" badge is not surrounded as the others, see:
> 
> screenshot-localhost_10016-2020 06 25-16_12_33

Corrected

modules/site-health/admin-site-health.php
Chouby: Yes on frontend, no on admin. You can see the switch at: https://github.com/polylang/polylang/blob/2.7.4/include/model.php#L105-L110

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,248 @@
...
+							$fields[ $key ]['label'] = $key;
+							$fields[ $key ]['value'] = implode( ', ', $this->model->model->get_translated_post_types() );
+							break;
+						case 'taxonomies':
+							$fields[ $key ]['label'] = $key;
+							$fields[ $key ]['value'] = implode( ', ',
+								$this->model->model->get_translated_taxonomies() );
+							break;
+						case 'nav_menus':
+							$current_theme = wp_get_theme();
```
Chouby, line 92: I would evaluate this outside the loop. Also you could use directly `get_stylesheet()`.

modules/site-health/admin-site-health.php
sebastienserre: I don't remember if made that
```
/**
	 * Return the flag used in the language switcher
	 *
	 * @param object $language language object
	 * @return mixed
	 * @since   2.8
	 */
	public function get_flag( $language ) {
		if ( empty( $language->flag ) ) {
			$language->flag = __( 'Undefined', 'polylang' );
		}
		if ( ! empty( $language->custom_flag_url ) ) {
			$language->flag = '<img src="' . $language->custom_flag_url . '" height="11">';
		}
		return $language;
	}
``` to correct this comment on another but my custom flag is well displayed

File: modules/site-health/admin-site-health.php [hunk 18]
```diff
@@ -0,0 +1,248 @@
...
+
+	/**
+	 * Add Polylang Options to Site Health Informations tab.
+	 *
+	 * @param array $debug_info array options to display.
+	 * @return array
+	 * @since   2.8
+	 */
+	public function info_options( $debug_info ) {
+		$options = get_option( 'polylang' );
```
raaaahman, line 71: This information is stored in the `$polylang->options` property when it is passed to the constructor: https://github.com/polylang/polylang/pull/522/files#diff-cb22506afd9748f751e395fe4f93c2e7R21

I believe it is better to store it in a property and avoid using WordPress functions throughout our code: we'll have more control over the code, we may avoid unnecessary calls to the database, and testing will be made easier.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,248 @@
...
+							$fields[ $key ]['label'] = $key;
+							$fields[ $key ]['value'] = implode( ', ',
+								$this->model->model->get_translated_taxonomies() );
+							break;
+						case 'nav_menus':
+							$current_theme = wp_get_theme();
+							foreach ( $value as $menus ) {
+								if ( key( $value ) !== $current_theme->stylesheet ) {
+									continue;
+								}
```
raaaahman: I believe there's only one entry in the `$value` array that will satisfy this condition, and we already know its key. If so, it would be easier to read if you simply access the entry with `$value[ $current_theme->stylesheet ]`, instead of looping to skip all the cases.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,272 @@
...
+
+	/**
+	 * Add Polylang Languages settings to Site Health Informations tab.
+	 *
+	 * @param array $debug_info array options to display.
+	 * @return array
+	 * @since   2.8
+	 */
+	public function info_languages( $debug_info ) {
+		$languages = PLL()->model->get_languages_list();
```
raaaahman, line 153: You already have access to this function from the `$model` property: https://github.com/polylang/polylang/pull/522/files#diff-cb22506afd9748f751e395fe4f93c2e7R14

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,272 @@
...
+	public function get_flag( $language ) {
+		if ( empty( $language->flag ) ) {
+			return $flag = '<span class="no-flag">' . __( 'Undefined', 'polylang' ) . '</span>';
+		}
+		if ( ! empty( $language->custom_flag_url ) ) {
+			return $flag = '<img src="' . $language->custom_flag_url . '" height="11">';
+		} else {
+			return $flag = $language->flag;
+		}
+	}
```
raaaahman, line 206: I don't believe you are giving a faithful information about the flag used in the switcher if you rewrite the logic inside the site health module, plus it will be two places to modify the code each time we change the flag display logic. Also, didn't we made the language code as a placeholder for unavailable flags?

Instead, [the code that displays the flag in the switcher](https://github.com/polylang/polylang/blob/0c8e04725c5810bdeed206f02c52206ba111d914/include/switcher.php#L96) should be made as a function to reuse here. This may induce additional changes because I believe that the flags will be differently set in the context of the admin dashboard than on the frontend. 

@manooweb needed to display the flags for the switcher in the admin dashboard too, so the two features may require the same change.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,272 @@
...
+			),
+			'description' => sprintf(
+				'<p>%s</p>',
+				__( 'A website can\'t be displayed without homepage.', 'polylang' )
+			),
+			'actions'     => '',
+			'test'        => 'pll_homepage',
+		);
+		$untranslated = array();
+		foreach ( PLL()->model->get_languages_list() as $language ) {
```
raaaahman, line 247: Same remark than https://github.com/polylang/polylang/pull/522/files#diff-cb22506afd9748f751e395fe4f93c2e7R71

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,272 @@
...
+			if ( ! $language->page_on_front ) {
+				$untranslated[] = sprintf(
+					'<a href="%s">%s</a>',
+					esc_url( PLL()->links->get_new_post_translation_link( get_option( 'page_on_front' ), $language ) ),
```
raaaahman, line 251: You can have the links as a dependency of this class in the same way than you did for the `PLL_Model`.

modules/site-health/admin-site-health.php
```diff
@@ -0,0 +1,272 @@
...
+				$fields[ $key ]['value'] = $value;
+
+				if ( 'term_group' === $key ) {
+					$fields[ $key ]['label'] = _x( 'order', 'Order of the language in the language switcher', 'polylang' );
+				}
+			}
+			$flag = $this->get_flag( $language );
+
+			$lang_name = sanitize_title( $language->name );
+			$debug_info[ $lang_name ] = array(
```
raaaahman, line 179: From [WordPress' Site Health Check documentation](https://make.wordpress.org/core/2019/04/25/site-health-check-in-5-2/):

> Each new section that’s added should use your plugin or theme slug to avoid name collisions. All core entries are prefixed with wp-.

We should have a way to distinguish our languages from those of another multilingual plugin, otherwise, they may be overridden.

Chouby:
Rebased locally and force pushed to fix the conflicts

PR #524, opened 2020-06-19
Reorganize modules and integrations

Chouby:
* All modules are now responsible to instantiate their classes based on conditions defined in the file `load.php`.

* Modules with settings add their section with the filter `pll_settings_module`. To ease the ordering of the settings sections, the order is now explicitely defined by array keys.

* Integrations are now moved to their own directory instead of a module directory. The class `PLL_Plugins_Compat` is renamed to `PLL_Integrations` to better reflect its usage. `PLL_Plugins_Compat` is kept as a class alias for backward compatibility. Each integration has its own directory and its file `load.php` to decide loading conditions. Integrations for Twenty Seventeen and MU Domain Mapping are moved to their own class. Only the most simple integrations are kept in `PLL_Integrations`


admin/admin.php
```diff
@@ -24,18 +24,14 @@
...
  * classic_editor  => reference to PLL_Admin_Classic_Editor object
  * filters_media   => optional, reference to PLL_Admin_Filters_Media object
- * wizard          => reference, a PLL_Wizard object
  *
  * @since 1.2
- * @since 2.7 Added a reference to a PLL_Bulk_Translate instance.
- * @since 2.7 Added a reference to a PLL_Wizard object.
  */
 class PLL_Admin extends PLL_Admin_Base {
-	public $filters, $filters_columns, $filters_post, $filters_term, $nav_menu, $sync, $filters_media;
```
raaaahman, line 38: This will need to be reformatted to be parsed by PHPDocumentor.

integrations/cache/load.php
```diff
@@ -0,0 +1,21 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Don't access directly.
+};
+
+add_action(
+	'plugins_loaded',
+	function() {
+		if ( pll_is_cache_active() ) {
+			add_action( 'pll_init', array( PLL_Integrations::instance()->cache_compat = new PLL_Cache_Compat(), 'init' ) );
```
raaaahman, line 16: I like that the hooks are registered from a method separated from the constructor. But this is not consistent with how our modules register their hooks.

integrations/domain-mapping/domain-mapping.php
```diff
@@ -0,0 +1,75 @@
...
+			// Don't redirect theme customizer
+			if ( isset( $_POST['customize'] ) && isset( $_POST['theme'] ) && 'on' === $_POST['customize'] ) { // phpcs:ignore WordPress.Security.NonceVerification
+				return;
+			}
+
+			// If we can't associate the requested domain to a language, redirect to the default domain
+			$requested_url  = pll_get_requested_url();
+			$requested_host = wp_parse_url( $requested_url, PHP_URL_HOST );
+
+			$hosts = PLL()->links_model->get_hosts();
```
raaaahman, line 61: Could the `$links_model` instance be passed in the constructor?

raaaahman, line 61: Could we pass the `$links_model` instance as a dependency here?

integrations/domain-mapping/domain-mapping.php
```diff
@@ -0,0 +1,75 @@
...
+			$lang  = array_search( $requested_host, $hosts );
+
+			if ( empty( $lang ) ) {
+				$status   = get_site_option( 'dm_301_redirect' ) ? '301' : '302'; // Honor status redirect option
+				$redirect = str_replace( '://' . $requested_host, '://' . $hosts[ $options['default_lang'] ], $requested_url );
+				wp_safe_redirect( $redirect, $status );
```
raaaahman, line 67: Could we use the `PLL_Settings::redirect()` here?

modules/wpml/load.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Don't access directly
+};
+
+if ( $polylang->model->get_languages_list() ) {
+	if ( ! defined( 'PLL_WPML_COMPAT' ) || PLL_WPML_COMPAT ) {
+		PLL_WPML_Compat::instance(); // WPML API
+		PLL_WPML_Config::instance(); // wpml-config.xml
```
raaaahman, line 15: Why does this two have their own singletons?

settings/settings.php
```diff
@@ -96,6 +88,7 @@ public function register_settings_modules() {
 		 * @param array $modules the list of module classes
 		 */
 		$modules = apply_filters( 'pll_settings_modules', $modules );
+		ksort( $modules );
```
raaaahman, line 91: This is a good thing that we may be able to register settings independently for each module.

But I would advocate to use an `$priority` property to sort them in the settings view, because setting manually the numeric indexes in an array could result in entries overriding others. I have coded such a sorting function for [PLL_Bulk_Translate::display_form()](https://github.com/polylang/polylang-pro/blob/c48d41da19babe9f07dd13db5bb984e0bc440be5/modules/bulk-translate/bulk-translate.php#L295-L300), that could be a good thing that we use the same logic and names (you can change them in the `bulk-translate` module if you find something more fitting).

admin/admin.php
Chouby: I remember that you opened a separate issue for this: https://github.com/polylang/polylang-pro/issues/544

integrations/integrations.php
```diff
@@ -0,0 +1,149 @@
...
+	 *
+	 * @return object
+	 */
+	public static function instance() {
+		if ( empty( self::$instance ) ) {
+			self::$instance = new self();
+		}
+
+		return self::$instance;
+	}
```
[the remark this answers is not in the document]
Chouby: I did not change this.

integrations/integrations.php
```diff
@@ -0,0 +1,149 @@
...
+	 * @param array $args
+	 * @return array
+	 */
+	public function no_category_base_get_terms_args( $args ) {
+		if ( doing_filter( 'category_rewrite_rules' ) ) {
+			$args['lang'] = '';
+		}
+		return $args;
+	}
+}
```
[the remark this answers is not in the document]
Chouby: You are right. One class, 2 files (as we have to add the `load.php` here), just for a single line function seems overkill for me.

For the history, this class was initially a god class including all plugins compatibilities. I then started separating some, indeed based on their size and complexity.

integrations/cache/load.php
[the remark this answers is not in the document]
Chouby: I did not change this. I rather think that this one is going the righr direction while older integrations still add hooks in the constructor.

integrations/domain-mapping/domain-mapping.php
Chouby: I did not change this. I consider that plugins integrations are completely external to the plugin so I don't mind using `PLL()` inside inegrations code. In some cases we could pass dependencies by getting the `$polylang` object from the `pll_init` hook. But this doesn't apply to all cases. `PLL_Integrations` is instatiated before Polylang is initialized (because we need that in a few cases, for example for Jetpack).

integrations/domain-mapping/domain-mapping.php
Chouby: I don't think so. We are in a completely different context.

modules/wpml/load.php
Chouby: I did not change this.

settings/settings.php
Chouby: You are right. That's a big issue of what I have written. Especially `pll_settings_modules` is a filter that can be used by 2 plugin developpers who don't know about the other.

We could use, for each module:
```php
array(
'class' => 'class_name',
'priority' => integer,
)
```
But we have to find some way to be backward compatible. That could be not too difficult if a developper just added a class. But if a 3rd party developper attempted to replace a class as I did here: https://github.com/polylang/polylang-pro/blob/0fb329d7fffe3b3b595710c8487379be3ccc0a4c/modules/media/load.php#L10-L11, I fear that there is no proper solution.

Chouby: On GitHub, I found only this repo using the filter: https://github.com/washingtonstateuniversity/WSUWP-Extended-Polylang/blob/4a5bcf23f395cb50b9cc2b8b1bea2c483a055352/includes/wsuwp-extended-polylang.php#L32-L46. No plugins or themes are using the filter in wp.org directory.

raaaahman: > But we have to find some way to be backward compatible. That could be not too difficult if a developper just added a class. But if a 3rd party developper attempted to replace a class as I did here: https://github.com/polylang/polylang-pro/blob/0fb329d7fffe3b3b595710c8487379be3ccc0a4c/modules/media/load.php#L10-L11, I fear that there is no proper solution.

If a 3rd party developer as replaced a Polylang Pro settings' module by accessing it through its numeric id in the `$modules` array, this PR already breaks i compatibility because the numeric ids have changed.

One way to keep compatibility could be to set the priority as a property of the `PLL_Settings_Module` subclasses, with a default set in the parent class. You can then keep the `$modules` array in the exact same order, and sort the modules in the [PLL_Settings_Table::prepare_items()](https://github.com/polylang/polylang/blob/9405a129d61c7ddb719ffc2b18562e51992e8e89/settings/table-settings.php#L169-L172) method, just before they are displayed.

raaaahman: > On GitHub, I found only one public repo using the filter: https://github.com/washingtonstateuniversity/WSUWP-Extended-Polylang/blob/4a5bcf23f395cb50b9cc2b8b1bea2c483a055352/includes/wsuwp-extended-polylang.php#L32-L46.

It seems that the goal of this code is only to not include the Pro version settings, which is what these modification will do. We could tell the plugin authors that this code is not needed anymore once we've merged.

integrations/domain-mapping/domain-mapping.php
raaaahman: I guess you are right. This feature don't seem to require most of the code that is inside `PLL_Settings::redirect()`.

raaaahman:
@Chouby I am able to tell, in most of the cases, what this PR changes and what it keeps. But as I stated in my more general comment, I believe that some parts of the code could be changed as well to be consistent with our structure, this will diminish the time we will need to search for the correct parts of the code.

Chouby:
> This is definitely going in a good way for modularity purposes. I don't really get why the modules and integrations needs to be different technically though. In the end, this is not much a problem as long as the modules are consistent with other modules and the integrations are consistent with other integrations (but that seems to become less evident with `PLL_WPML_Config`, which is a bit of both).

In my opinion modules and inegrations should be kept separated for clarity.

The loading concept is the same. The timing and dependencies are not the same. 
* We need to load the integrations as soon as possible in some cases (ex Jetpack). This requires that integrations classes must not depend on the `$polylang` object.
* On the other hand, modules gneerally depend on the `$polylang` object, so we need to wait for it to be available.

As for the "WPML compatibility mode", this is definitely a module, as this is not an integration with WPML but an implementation of the WPML API.

raaaahman:
> The loading concept is the same. The timing and dependencies are not the same.
> 
>     * We need to load the integrations as soon as possible in some cases (ex Jetpack). This requires that integrations classes must not depend on the `$polylang` object.
> 
>     * On the other hand, modules gneerally depend on the `$polylang` object, so we need to wait for it to be available.

The fact is that we still convey integrations dependencies through the `PLL_Base` subclasses, but in a different manner than for the modules. So when the `PLL_Jetpack` integration will actually [do its business logic](https://github.com/polylang/polylang/blob/77c5585700f931373484cf1602e90c62ce697c2d/integrations/jetpack/jetpack.php#L41-L43), the `PLL_Base` subclass still needs to be instantiated.

Both modules and integrations don't really depend on the `$polylang` instance itself, but on the modules, core components and integrations it holds as properties. These are the instances we ultimately need to share, the `$polylang` object is just a tool for doing so.

It looks like this has been an ongoing topic for some time now: https://trello.com/c/eW2SOV0K/79-about-pll-instance

raaaahman:
The changes in https://github.com/polylang/polylang/pull/524/commits/77c5585700f931373484cf1602e90c62ce697c2d are exactly what how I envisioned it.

Since we use twice [a similar anonymous function](https://github.com/polylang/polylang-pro/blob/a23ed7378e94398dd315b008196fc6d556134ff0/modules/bulk-translate/bulk-translate.php#L300-L302), maybe we can create it as a named function?

Even better, we could defined some `PLL_Sortable_View_Item` interface, that both `PLL_Settings_Modules` and `PLL_Bulk_Translate_Option` may implement, to clearly state which classes can be passed as argument to the commonalized function.

modules/wpml/load.php
manooweb: https://github.com/polylang/polylang/blob/2.7.x/include/class-polylang.php#L285-L288

integrations/integrations.php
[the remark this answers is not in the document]
manooweb: https://github.com/polylang/polylang/blob/2.7.x/modules/plugins/plugins-compat.php#L63-L76

integrations/integrations.php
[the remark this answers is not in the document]
manooweb: Even if it is "overkill" I agree about we decided to separate all plugin integration in their own folder. It's more readable and searchable to have a folder corresponding on the plugin. It's more consistent too.

manooweb:
> The changes in [77c5585](https://github.com/polylang/polylang/commit/77c5585700f931373484cf1602e90c62ce697c2d) are exactly what how I envisioned it.
> 
👍


> @Chouby I am able to tell, in most of the cases, what this PR changes and what it keeps. But as I stated in my more general comment, I believe that some parts of the code could be changed as well to be consistent with our structure, this will diminish the time we will need to search for the correct parts of the code.

See my comment here https://github.com/polylang/polylang/pull/524#discussion_r445995312

manooweb:
> @Chouby I am able to tell, in most of the cases, what this PR changes and what it keeps. But as I stated in my more general comment, I believe that some parts of the code could be changed as well to be consistent with our structure, this will diminish the time we will need to search for the correct parts of the code.

> @Chouby I am able to tell, in most of the cases, what this PR changes and what it keeps. But as I stated in my more general comment, I believe that some parts of the code could be changed as well to be consistent with our structure, this will diminish the time we will need to search for the correct parts of the code.

See my comment here https://github.com/polylang/polylang/pull/524#discussion_r445995312

settings/settings.php
manooweb: > But we have to find some way to be backward compatible. That could be not too difficult if a developper just added a class. But if a 3rd party developper attempted to replace a class as I did here: https://github.com/polylang/polylang-pro/blob/0fb329d7fffe3b3b595710c8487379be3ccc0a4c/modules/media/load.php#L10-L11, I fear that there is no proper solution.

I think we will have to define `$priority` property in Polylang Pro `PLL_Settings_Advanced_Media` class.
The `pll_settings_modules` allows to replace the Polylang  `PLL_Settings_Media` class by Polylang Pro `PLL_Settings_Advanced_Media` class but not with the same order to display it in the settings list table.

PR #526, opened 2020-06-25
Switcher block

manooweb:
Because of the reorganization of Polylang and Polylang Pro this PR https://github.com/polylang/polylang-pro/pull/493 is outdated.

This PR is the part which needs to be adapted in Polylang to make the language switcher block to work.


include/switcher.php
```diff
@@ -55,22 +57,31 @@ protected function get_elements( $links, $args ) {
...
+			$current_lang = ''; // Avoids potential notice
+
+			if ( 0 === $args['admin_render'] ) {
+				if ( $current_lang = $links->curlang->slug == $slug ) {
+					if ( $args['hide_current'] && ! ( $args['dropdown'] && ! $args['raw'] ) ) {
+						continue; // Hide current language except for dropdown
+					} else {
+						$classes[] = 'current-lang';
+					}
 				}
```
raaaahman, line 69: I don't understand, wasn't that code supposed to be executed outside of a language context? Why checking on the current language here?

include/switcher.php
```diff
@@ -55,22 +57,31 @@ protected function get_elements( $links, $args ) {
...
+					$url = get_permalink( $tr_id );
+				} elseif ( null === $args['post_id'] ) {
+					$url = $links->get_translation_url( $language );
+				}
+			} else {
+				if ( isset( $args['lang'] ) && $args['lang'] == $slug ) {
+					if ( $args['hide_current'] && ! ( $args['dropdown'] && ! $args['raw'] ) ) {
+						continue; // Hide current language except for dropdown
+					}
+				}
```
raaaahman, line 81: This is very similar to the condition above. I believe we should rather filter the languages list to exclude the current language before [iterating on it](https://github.com/polylang/polylang/blob/f6fbb96d14808fd5b54040c124aa237f61f0d3e5/include/switcher.php#L53).

Chouby: I agree that we should simplify this.
1. You are using two parameters, one documented ,`admin_render`, the other undocumented, `lang`. I would use only one. Let's call it `current_lang`, a it is what your `lang` is. You impose the current language in the language switcher.
2. The we could do something like:
```php
$current_lang  = empty( $args['current_lang'] ) ? $links->curlang->slug : $args['current_lang'];
```
This should allow you not to duplicate the test.
3. Then we should have to solve the issue with
```php
elseif ( null === $args['post_id'] ) {
				$url = $links->get_translation_url( $language );
				}
```
This could be done by checking `$args['current_lang']` in the test to avoid the line to be executed.

include/switcher.php
```diff
@@ -55,22 +57,31 @@ protected function get_elements( $links, $args ) {
...
 			}
 
-			if ( $no_translation = empty( $url ) ) {
+			if ( $no_translation = empty( $url ) && 0 === $args['admin_render'] ) {
```
Chouby, line 84: Does the class `no-translation` disturbs the block renderig? Otherwise I believe that we could remove the adidional test.

include/api.php
```diff
@@ -28,11 +29,8 @@
  * @return null|string|array null if displaying, array if raw is requested, string otherwise
  */
 function pll_the_languages( $args = '' ) {
-	if ( PLL() instanceof PLL_Frontend ) {
-		$switcher = new PLL_Switcher();
-		return $switcher->the_languages( PLL()->links, $args );
-	}
-	return '';
```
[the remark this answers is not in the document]
manooweb: Revert this code because of the direct use of PLL_Switcher::the_languages method to render the language switcher block.

I don't believe a API `pll_switcher block` function is needed because `pll_the_languages` function already does the job.

include/api.php
```diff
@@ -28,11 +29,8 @@
...
+	$switcher = new PLL_Switcher();
+	return $switcher->the_languages( PLL()->links, $args );
```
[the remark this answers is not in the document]
manooweb: Reverted to the old code

include/switcher.php
```diff
@@ -29,7 +29,9 @@ public static function get_switcher_options( $type = 'widget', $key = 'string' )
 			'hide_current'           => array( 'string' => __( 'Hides the current language', 'polylang' ), 'default' => 0 ),
 			'hide_if_no_translation' => array( 'string' => __( 'Hides languages with no translation', 'polylang' ), 'default' => 0 ),
 		);
-
+		if ( 'all' === $key ) {
+			return $options;
+		}
```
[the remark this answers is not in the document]
manooweb: Reverted by using twice `PLL_Switcher::get_switcher_options`: once with 'string' value, once with 'default' for the `$key` parameter in `PLL_Block_Editor_Switcher_Block::register_block_polylang_language_switcher` method

include/switcher.php
manooweb: I suppose that I had some problem when rendering the block in the admin context but I don't remember why and I can't reproduce the case anymore.
So I removed the test.

include/switcher.php
manooweb: In some REST API requests (post save for example) the language we set in admin_current_language could be `null`.
So we should test something like that `if ( ! is_null( $args['admin_current_language'] && empty( $args['admin_current_lang'] ) )`. I think it is difficult to understand because a `null` value is considered as an `empty` value too.
Thus I prefered to use two parameters `admin_render` as before and the new `admin_current_lang` (replacing lang) for testing what you need to do.

include/switcher.php
manooweb: Just because of the language switcher feature: option 'Hides the current language'

manooweb:
Notice I remove the unit test modifications because I reverted the pll_the_languages function which isn't usable in a admin context.
So this function always return an empty string in this context

PR #528, opened 2020-06-26
Always filter WP_Query by the current language

### LINKED ISSUE #469: Sticky filter is showing posts for all languages when a specific language is selected
## Prerequisites

1. Add at least two languages in Polylang.
2. Add a bunch of Sticky posts for both languages.
3. When you use the "Sticky" filter for one language sticky posts for all languages are shown:
Screenshot 2020-03-16 at 09 10 04

When using "Sticky" filter for one language posts for all languages are shown. I expect that only posts for selected language will be visible.

Tested on WordPress 5.3.2 and Polylang PRO 2.6.10.

Chouby:
Fixes #469 
Closes https://github.com/polylang/polylang-pro/issues/434

On admin siide, `WP_Query` wasn't filtered by the current language when the query included query vars which specify objects with their language, typically, `p', `page_id`, 'cat'...

This caused several conflicts that we used to fix by removing these exclusions with the filter  `pll_filter_query_excluded_query_vars`. This also caused troubles with the REST API. So it the end, it looks like the idea was bad from the start.

This PR removes completely these exclusions and fixes or adds phpunit tests accordingly.

We will be able to remove the usage of the filter `pll_filter_query_excluded_query_vars` in `PLL_Featured_Content` once this PR and #524 are merged.

### DIFF
--- a/include/query.php
+++ b/include/query.php
@@ -9,30 +9,6 @@
  * @since 2.2
  */
 class PLL_Query {
-
-	protected static $excludes = array(
-		'p',
-		'post_parent',
-		'attachment',
-		'attachment_id',
-		'name',
-		'pagename',
-		'page_id',
-		'category_name',
-		'tag',
-		'tag_id',
-		'cat',
-		'category__in',
-		'category__and',
-		'post__in',
-		'post_name__in',
-		'tag__in',
-		'tag__and',
-		'tag_slug__in',
-		'tag_slug__and',
-		'post_parent__in',
-	);
-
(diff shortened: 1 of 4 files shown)


Chrystll:
Your PR fixes the bug. I don't get it anymore when selecting a language filter first then select the sticky filter.

Chouby:
For Polylang fo WooCommerce, we need to drop the support of Polylang < 2.8 before we can remove the filter.

I will wait for #524 to merge.

PR #529, opened 2020-06-26
Move all integrations to their own directory.

Chouby:
This PR follows the comments of @raaaahman and @manooweb for the plugins integrations https://github.com/polylang/polylang/pull/524#discussion_r443506907

It of course depends on #524.

* All integrations are moved on their own directory. A `load.php` and a class for each.
* For some plugins I added a test to check if the 3rd party plugin is indeed active (these tests did not exist before for the smallest integrations).
* Adapt phpunit test for YARPP.

### DIFF
--- a/tests/phpunit/tests/plugins/test-yarpp.php
+++ b/tests/phpunit/tests/plugins/test-yarpp.php
@@ -4,12 +4,13 @@ class YARPP_Test extends PLL_UnitTestCase {
 	function setUp() {
 		parent::setUp();
 
-		$GLOBALS['polylang'] = &self::$polylang; // avoid conflicts when other tests are executed before
+		$GLOBALS['polylang'] = &self::$polylang; // Avoid conflicts when other tests are executed before.
 	}
 
 	// bug introduced in 1.8 and fixed in 1.8.2
 	function test_yarpp_support() {
-		do_action( 'setup_theme' );
+		define( 'YARPP_VERSION', '1.0' ); // Fake.
+		do_action( 'plugins_loaded' );
 		do_action( 'init' );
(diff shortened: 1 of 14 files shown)


tests/phpunit/tests/plugins/test-yarpp.php
```diff
@@ -4,12 +4,13 @@ class YARPP_Test extends PLL_UnitTestCase {
...
 	}
 
 	// bug introduced in 1.8 and fixed in 1.8.2
 	function test_yarpp_support() {
-		do_action( 'setup_theme' );
+		define( 'YARPP_VERSION', '1.0' ); // Fake.
+		do_action( 'plugins_loaded' );
 		do_action( 'init' );
 		$this->assertEquals( 1, $GLOBALS['wp_taxonomies']['language']->yarpp_support );
 	}
```
raaaahman, line 16: I don't believe this test will allow us to detect breaking changes from YARPP, because if they change their constant's name, it is still hardcoded in our test.

Besides, it seems that this test tests our solution (setting a property in a global object) rather than testing that the problem is solved (and I can't deduce what problem it is made to solve by reading the test).

Also, the `PLL_Integrations` instance is set once during the test bootstrap [that calls PLL_Integrations::instance()](https://github.com/polylang/polylang/blob/c12fc8abfe41d5d429a906c11cc2a869994ab9cf/include/class-polylang.php#L58) and then shared between all of our tests. Then previous tests can trigger the hooks that will trigger a functionality of an integration, which may alter the state of our test environment (like here, by setting a global variable); before we have set up our environment properly to test this functionality.

I believe it would be better to test the conditions inside each integration `init()` method. That way, we will be able to test them independently from each other, and you may hook directly these `init()` methods instead of anonymous functions.

PR #531, opened 2020-06-30
Add compatibility with sitemaps

### LINKED ISSUE #451: Add compatibility with XML Sitemaps
## Prerequisites

- [x] This bug happens with only Polylang and XML Sitemaps feature plugin active

1. Install and activate XML Sitemaps feature plugin (I used [master](https://github.com/GoogleChromeLabs/wp-sitemaps) from GitHub, but there is also [version](https://wordpress.org/plugins/core-sitemaps/) in WP Plugins Directory)
2. Open the sitemap.xml (sitemap index) - observe sitemap contents reflecting current front-end language only
3. Go to post or page sitemap (actual sitemap) - observe sitemap contents reflecting current front-end language only

I would expect the sitemap to reflect (include) content from all active languages.

## Footnote

Here's the [feature proposal post](https://make.wordpress.org/core/2019/06/12/xml-sitemaps-feature-project-proposal/) and here's a recent [follow-up post](https://make.wordpress.org/core/2020/01/27/feature-plugin-xml-sitemaps/) - I have no idea what current timing for inclusion in core is, whether it's expected to land in 5.4 (I doubt), but it might be good idea to get in touch with folks behind the initiative early in development cycle.

Chouby:
This PR adds compatibility with the WP core sitemaps which are planned to be introduced in WP 5.5. See https://make.wordpress.org/core/tag/sitemaps/

The sitemaps should work out of the box for subdomains and multiple domains. So the compatibility code is loaded only when the language is set from the content or from the directory name in the url. In this case, the index is unique for all languages. However the index lists one sitemap per language for each existing sitemap type.

Fixes #451

**Edit**: for testing after installing and activating `Core Sitemaps` feature plugin, just enter the URL for the default XML sitemap file name `http://mon-site.com/sitemap.xml` which redirects to `http://mon-site.com/wp-sitemap.xml`

Then we should see the sitemap index with URLs for each each existing sitemap type and one per language when Polylang is activated too

image

### DIFF
--- a/modules/sitemaps/sitemaps-provider-decorator.php
+++ b/modules/sitemaps/sitemaps-provider-decorator.php
@@ -1,4 +1,7 @@
 <?php
+/**
+ * @package Polylang
+ */
 
 /**
  * Sitemaps providers decorator
@@ -28,6 +31,9 @@ class PLL_Sitemaps_Provider_Decorator extends WP_Sitemaps_Provider {
 	 * Constructor.
 	 *
 	 * @since 2.8
+	 *
+	 * @param WP_Sitemaps_Provider $provider An instance of a WP_Sitemaps_Provider child class.
+	 * @param string               $lang     Language code.
 	 */
 	public function __construct( $provider, $lang ) {
 		$this->name = $provider->name . '-' . $lang;
--- a/modules/sitemaps/sitemaps-provider-decorator.php
+++ b/modules/sitemaps/sitemaps-provider-decorator.php
@@ -59,6 +59,9 @@ class PLL_Sitemaps_Provider_Decorator extends WP_Sitemaps_Provider {
 	/**
 	 * Gets the max number of pages available for the object type.
 	 *
+	 * Returns 0 for all languages but the default language for untranslated post types
+	 * and untranslated taxonomies to avoid to duplicate the corresponding sitemaps.
+	 *
 	 * @since 2.8
 	 *
 	 * @param string $object_subtype Optional. Object subtype. Default empty.
--- a/modules/sitemaps/sitemaps-provider-decorator.php
+++ b/modules/sitemaps/sitemaps-provider-decorator.php
@@ -1,0 +1,89 @@
+<?php
+
+/**
+ * Sitemaps providers decorator
+ *
+ * @since 2.8
+ */
+class PLL_Sitemaps_Provider_Decorator extends WP_Sitemaps_Provider {
+	/**
+	 * The decorated sitemaps provider.
+	 *
+	 * @since 2.8
+	 *
+	 * @var WP_Sitemaps_Provider
+	 */
+	protected $provider;
+
+	/**
+	 * Language code.
+	 *
+	 * @since 2.8
+	 *
+	 * @var string
+	 */
+	protected $lang;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 2.8
+	 */
+	public function __construct( $provider, $lang ) {
+		$this->name = $provider->name . '-' . $lang;
+		$this->object_type = $provider->object_type;
+
+		$this->provider = $provider;
+		$this->lang = $lang;
+	}
+
+	/**
+	 * Gets a URL list for a sitemap.
+	 *
+	 * @since 2.8
+	 *
+	 * @param int    $page_num       Page of results.
+	 * @param string $object_subtype Optional. Object subtype name. Default empty.
+	 * @return array Array of URLs for a sitemap.
+	 */
+	public function get_url_list( $page_num, $object_subtype = '' ) {
+		return $this->provider->get_url_list( $page_num, $object_subtype );
+	}
+
+	/**
+	 * Gets the max number of pages available for the object type.
+	 *
+	 * @since 2.8
+	 *
+	 * @param string $object_subtype Optional. Object subtype. Default empty.
+	 * @return int Total number of pages.
+	 */
+	public function get_max_num_pages( $object_subtype = '' ) {
+		return $this->provider->get_max_num_pages( $object_subtype );
+	}
+
+	/**
+	 * Gets the URL of a sitemap entry.
+	 *
+	 * @since 2.8
+	 *
+	 * @param string $name The name of the sitemap.
+	 * @param int    $page The page of the sitemap.
+	 * @return string The composed URL for a sitemap entry.
+	 */
+	public function get_sitemap_url( $name, $page ) {
+		$url = $this->provider->get_sitemap_url( $name, $page );
+		return str_replace( $this->provider->name, $this->name, $url );
+	}
+
+	/**
+	 * Returns the list of supported object subtypes exposed by the provider.
+	 *
+	 * @since 2.8
+	 *
+	 * @return array List of object subtypes objects keyed by their name.
+	 */
+	public function get_object_subtypes() {
+		return $this->provider->get_object_subtypes();
+	}
+}
--- a/modules/sitemaps/sitemaps-provider-decorator.php
+++ b/modules/sitemaps/sitemaps-provider-decorator.php
@@ -65,6 +65,19 @@ class PLL_Sitemaps_Provider_Decorator extends WP_Sitemaps_Provider {
 	 * @return int Total number of pages.
 	 */
 	public function get_max_num_pages( $object_subtype = '' ) {
+		switch ( $this->provider->name ) {
+			case 'posts':
+				if ( ! PLL()->model->is_translated_post_type( $object_subtype ) && PLL()->model->options['default_lang'] !== $this->lang ) {
+					return 0;
+				}
+				break;
+			case 'taxonomies':
+				if ( ! PLL()->model->is_translated_taxonomy( $object_subtype ) && PLL()->model->options['default_lang'] !== $this->lang ) {
+					return 0;
+				}
+				break;
+		}
+
 		return $this->provider->get_max_num_pages( $object_subtype );
 	}
 
--- a/modules/sitemaps/sitemaps-provider-decorator.php
+++ b/modules/sitemaps/sitemaps-provider-decorator.php
@@ -19,28 +19,48 @@ class PLL_Sitemaps_Provider_Decorator extends WP_Sitemaps_Provider {
 	protected $provider;
 
 	/**
-	 * Language code.
+	 * The PLL_Links_Model instance.
+	 *
+	 * @since 2.8
+	 *
+	 * @var PLL_Links_Model
+	 */
+	protected $links_model;
+
+	/**
+	 * The PLL_Model instance.
+	 *
+	 * @since 2.8
+	 *
+	 * @var PLL_Model
+	 */
+	protected $model;
+
+
+	/**
+	 * Language used to filter queries for the sitemap index.
 	 *
 	 * @since 2.8
 	 *
 	 * @var string
 	 */
-	protected $lang;
+	private $filter_lang;
 
 	/**
 	 * Constructor.
 	 *
 	 * @since 2.8
 	 *
-	 * @param WP_Sitemaps_Provider $provider An instance of a WP_Sitemaps_Provider child class.
-	 * @param string               $lang     Language code.
+	 * @param WP_Sitemaps_Provider $provider    An instance of a WP_Sitemaps_Provider child class.
+	 * @param PLL_Links_Model      $links_model The PLL_Links_Model instance.
 	 */
-	public function __construct( $provider, $lang ) {
-		$this->name = $provider->name . '-' . $lang;
+	public function __construct( $provider, $links_model ) {
+		$this->name = $provider->name;
 		$this->object_type = $provider->object_type;
 
 		$this->provider = $provider;
-		$this->lang = $lang;
+		$this->links_model = $links_model;
+		$this->model = $links_model->model;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+		$qv = get_query_var( 'sitemap' );
+		if ( $qv ) {
+			$arr = explode( '-', $qv );
+			$lang = end( $arr );
+			$this->curlang = $this->model->get_language( $lang );
+		}
+	}
+
+	/**
+	 * Whitelists the home url filter for the sitemaps
```
raaaahman, line 69: Do you mean that it adds the wp-sitemap calls to the `home_url()` WordPress function in our whitelist for results that should not be filtered by language?

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @since 2.8
+	 *
+	 * @param array $whitelist White list.
+	 * @return array;
+	 */
+	public function home_url_white_list( $whitelist ) {
+		$whitelist[] = array( 'file' => 'sitemaps' );
```
raaaahman, line 77: Ain't that too generic for the usage we are doing it here? Would we want to automatically whitelist any plugin that have 'sitemap' in its files names?

I believe you have made it that way to not have to change it every time the sitemap feature changes its code, but would it cover all the cases? What happens the call comes from a function in the `functions.php` file for example? What if a third party plugin triggers a call from a wp-sitemap class, but from the context of this 3rd party plugins files, thus not bearing the 'sitemap' mention in its file name?

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @since 2.8
+	 *
+	 * @param array $rules Rewrite rules.
+	 * @return array
+	 */
+	public function rewrite_rules( $rules ) {
+		$new_rules = array();
+		$languages = '(?:' . implode( '|', $this->get_active_languages() ) . ')';
+
+		foreach ( $rules as $key => $rule ) {
```
raaaahman, line 93: As there is great number of rewrite rules, it would be easier to debug if you use an `array_filter` first, and then only iterate over the ones we need to modify.

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+			} else {
+				$new_rules[ $key ] = $rule;
+			}
+		}
+		return $new_rules;
+	}
+
+	/**
+	 * Replaces the list of sitemaps providers by our decorators.
+	 *
```
raaaahman, line 106: It would be good to mention that it creates a provider by language for each registered provider.

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @param array $providers Array of WP_Sitemaps_Provider objects keyed by their name.
+	 * @return array
+	 */
+	public function providers( $providers ) {
+		$new_providers = array();
+
+		foreach ( $providers as $key => $provider ) {
+			foreach ( $this->get_active_languages() as $lang ) {
+				$new_providers[ "$key-$lang" ] = new PLL_Sitemaps_Provider_Decorator( $provider, $lang );
```
raaaahman, line 117: Since we don't know what other plugins will throw in this array, it would be better to only decorate the ones that implement `PLL_Sitemaps_Provider`.

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @since 2.8
+	 */
+	protected function get_active_languages() {
+		$languages = $this->model->get_languages_list();
+		if ( wp_list_filter( $languages, array( 'active' => false ) ) ) {
+			return wp_list_pluck( wp_list_filter( $languages, array( 'active' => false ), 'NOT' ), 'slug' );
+		}
+		return wp_list_pluck( $languages, 'slug' );
+	}
```
raaaahman, line 135: Maybe this method could belong to the `PLL_Model` class? I can imagine it being used elsewhere.

modules/sitemaps/sitemaps.php
manooweb: I don't understand the issue you tried to explain during the DM and what it is the goal to whitelist `sitemaps` filenames.

Chrystll:
I get a bug with custom post type and taxonomies for which I don't enable the translation as below:

screenshot-localhost-2020 07 02-16_34_24

screenshot-localhost-2020 07 02-16_34_49

They are displayed as here:
- http://localhost/wordpress/explore-porto/wp-sitemap-posts-en-book-1.xml
- http://localhost/wordpress/explore-porto/wp-sitemap-posts-fr-book-1.xml
- http://localhost/wordpress/explore-porto/wp-sitemap-taxonomies-en-genre-1.xml
- http://localhost/wordpress/explore-porto/wp-sitemap-taxonomies-fr-genre-1.xml

Chrystll:
When using subdomains I get all the time the sitemap in the default language, see: 
Capture du 2020-07-02 17-06-32


I don't get this issue with domains.

modules/sitemaps/sitemaps-provider-decorator.php
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * @package Polylang
+ */
+
+/**
+ * Sitemaps providers decorator
+ *
+ * @since 2.8
+ */
+class PLL_Sitemaps_Provider_Decorator extends WP_Sitemaps_Provider {
```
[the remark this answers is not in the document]
Chouby: What do you suggest?

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+	protected $model;
+
+	/**
+	 * A reference to the current language.
+	 *
+	 * @since 2.8
+	 *
+	 * @var PLL_Language
+	 */
+	protected $curlang;
```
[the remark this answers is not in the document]
Chouby: in `set_language()`.

modules/sitemaps/sitemaps.php
```diff
@@ -0,0 +1,136 @@
...
+		add_filter( 'rewrite_rules_array', array( $this, 'rewrite_rules' ) );
+		add_filter( 'wp_sitemaps_register_providers', array( $this, 'providers' ), 99 ); // 99 in an attempt to have all sitemaps providers in our filter.
+	}
+
+	/**
+	 * Sets the language for the current sitemap being rendered.
+	 *
+	 * @since 2.8
+	 */
+	public function set_language() {
```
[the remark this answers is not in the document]
Chouby: The method does really set the sitemap language at the last line.

modules/sitemaps/sitemaps.php
Chouby: By default, our `home_url` filter is active for themes and some whitelisted functions of WP core, but not for plugins. For sitemaps, we always need to filter the home url. So I am whitelisting all files with 'sitemaps' somewhere in their path. I don't believe it's risky. If some conflict arises at some time we could try to narrow the whitelist.

modules/sitemaps/sitemaps.php
Chouby: See above. For 3rd party plugins, we need to whitelist the filter on a case by case basis. As done for Yoast: https://github.com/polylang/polylang/blob/a73de2b21fbfde0587a7c3df71f309cebf3e6a73/integrations/wpseo/wpseo.php#L40

modules/sitemaps/sitemaps.php
Chouby: I understand your proposal. The issue is that we need to insert the new rules at the right postion. So in the end it would make the code probably more complex to insert the new rules in the old array.

modules/sitemaps/sitemaps.php
Chouby: That's something I thought to. The question is what would be the most common case? Translated content or untranslated content? I opted for the first case.

Chouby:
@Chrystll First issue fixed by the last commit. Both urls linked in fact to the same (duplicated) sitemap. I don't reproduce teh second issue.

modules/sitemaps/sitemaps.php
Chouby: That's somethign that we can indeed consider. I am not particularly satisfied of it as it is something useful only for Polylang Pro that I include in Polylang.

modules/sitemaps/sitemaps-provider-decorator.php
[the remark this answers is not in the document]
raaaahman: `PLL_Sitemaps_Multilingual` ?

modules/sitemaps/sitemaps.php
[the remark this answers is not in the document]
raaaahman: And once it is set, where is it accessed? Which method needs a `$curlang`?

modules/sitemaps/sitemaps.php
raaaahman: Okay, so the whitelisted files and functions will always be filtered by language, is that right?

modules/sitemaps/sitemaps.php
raaaahman: My concern are more about our code breaking because some 3rd party plugin inserted an object that doesn't implement the `PLL_Sitemaps_Provider`, so our delegated method calls will break because that object will not have the methods we expected.

modules/sitemaps/sitemaps.php
raaaahman: If it's the separation between Polylang and Polylang Pro that bothers you, the simplest solution I can think about would be to subclass this class in Polylang Pro, and move this method to this subclass. Here, the `PLL_Stiemaps::get_active_language()` (or maybe renamed as 'get_languages_list') would only delegate its call to `PLL_Model::get_languages_list()`.

That's similar to Strategy Pattern by the way.

modules/sitemaps/sitemaps.php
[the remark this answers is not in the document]
raaaahman: Okay then.

modules/sitemaps/sitemaps.php
raaaahman: What I'm trying to tell, is that the [PLL_Frontend_Filter::home_url()](https://github.com/polylang/polylang/blob/7bea3700a7293f8ef31b700fa0fe84647420c53b/frontend/frontend-filters-links.php#L308) method will whitelist calls from files where only a portion of the filename matches the string.

So whitelisting the `sitemaps` string seems to imply that every call to the home url coming from a file having `sitemap` in its filename will come from WordPress Core's sitemap feature, and every call to the home url from WordPress Core's sitemap feature will come from a file with `sitemap` in it's filename. I don't believe that this will always be the case, and I came with the examples listed above.

modules/sitemaps/sitemaps.php
Chouby: Any call to `home_url()` done from the whitelisted files and functions are indeed filtered.

modules/sitemaps/sitemaps-provider-decorator.php
```diff
@@ -65,6 +65,19 @@ public function get_url_list( $page_num, $object_subtype = '' ) {
 	 * @return int Total number of pages.
 	 */
 	public function get_max_num_pages( $object_subtype = '' ) {
+		switch ( $this->provider->name ) {
+			case 'posts':
+				if ( ! PLL()->model->is_translated_post_type( $object_subtype ) && PLL()->model->options['default_lang'] !== $this->lang ) {
```
raaaahman, line 73: Maybe we can pass the list of translated object subtypes as an argument of the `PLL_Sitemaps_Provider_Decorator` constructor [here](https://github.com/polylang/polylang/blob/6226a8f29b1be571cf890fc73d537c2cca025e24/modules/sitemaps/sitemaps.php#L118) ?

Chrystll:
The sitemap is wrong in Plain permalinks when visualize it at explore-porto/?sitemap=index. See:

Capture du 2020-07-03 14-32-45

modules/sitemaps/sitemaps-provider-decorator.php
```diff
@@ -59,29 +79,113 @@ public function get_url_list( $page_num, $object_subtype = '' ) {
...
+	 *
+	 * @since 2.8
+	 *
+	 * @return array[] Array of sitemap types including object subtype name and number of pages.
+	 */
+	public function get_sitemap_type_data() {
+		$sitemap_data = array();
+
+		add_filter( 'wp_sitemaps_posts_query_args', array( $this, 'query_args' ) );
+		add_filter( 'wp_sitemaps_taxonomies_query_args', array( $this, 'query_args' ) );
```
raaaahman: These hooks are registered once per sitemap provider. That means that the second provider will also activate the 1st provider's hooks, and the third will activate the 1st and 2nd providers' hooks as well. I've noticed that as the current state of the code you've taken care that these former providers' hooks will have no effect, but I'm worried this implicit behavior could be overlooked in the future.

Also, these hooks are both called from two functions, and I'm not sure you want to call them in both cases, or just on one for each. See: 
- https://github.com/GoogleChromeLabs/wp-sitemaps/blob/c8c36818811fd7e2d339b86ddded08f532a075e4/inc/providers/class-wp-sitemaps-posts.php#L90-L93
- https://github.com/GoogleChromeLabs/wp-sitemaps/blob/c8c36818811fd7e2d339b86ddded08f532a075e4/inc/providers/class-wp-sitemaps-posts.php#L176-L180

An alternative to using hooks here, would be to decorate the `WP_Sitemaps_Post::get_posts_query_args` and `WP_Sitemaps_Taxonomies::get_taxonomies_query_args` methods and filter the lists here. Next step could be to ask the WP Sitemaps team to make merge these two functions in one (`WP_Sitemaps_Provider::get_query_args` for example, so we could only manage this in one place (and the other plugins developers could use this method name to be compatible with our decorator).

modules/sitemaps/sitemaps-provider-decorator.php
```diff
@@ -59,29 +79,113 @@ public function get_url_list( $page_num, $object_subtype = '' ) {
...
+				case 'taxonomies':
+					if ( $this->model->is_translated_taxonomy( $object_subtype_name ) ) {
+						foreach ( $this->get_active_languages() as $language ) {
+							$sitemap_data[] = $this->get_sitemap_data( $object_subtype_name, $language );
+						}
+					} else {
+						$sitemap_data[] = $this->get_sitemap_data( $object_subtype_name );
+					}
+					break;
+			}
```
raaaahman: The only change between these two block is the method call in the condition. You can remove duplication by making it a variable, computed before the `foreach` loop.

modules/sitemaps/sitemaps-provider-decorator.php
```diff
@@ -94,8 +198,17 @@ public function get_max_num_pages( $object_subtype = '' ) {
 	 * @return string The composed URL for a sitemap entry.
 	 */
 	public function get_sitemap_url( $name, $page ) {
-		$url = $this->provider->get_sitemap_url( $name, $page );
-		return str_replace( $this->provider->name, $this->name, $url );
+		$parts = explode( '-', $name );
+		$lang = end( $parts );
+		$lang = $this->model->get_language( $lang ); // Validates that we got an existing language code.
+
+		if ( $lang ) {
```
raaaahman: This comment made me think that a `PLL_Model::language_exists()` method could be helpful.

modules/sitemaps/sitemaps-provider-decorator.php
```diff
@@ -59,29 +79,113 @@ public function get_url_list( $page_num, $object_subtype = '' ) {
...
+	 *
+	 * @since 2.8
+	 */
+	protected function get_active_languages() {
+		$languages = $this->model->get_languages_list();
+		if ( wp_list_filter( $languages, array( 'active' => false ) ) ) {
+			return wp_list_pluck( wp_list_filter( $languages, array( 'active' => false ), 'NOT' ), 'slug' );
 		}
+		return wp_list_pluck( $languages, 'slug' );
+	}
```
raaaahman: You can take advantage of the Decorator pattern here to move this into Polylang Pro. You just need to create a second decorator, that decorates this `get_active_languages()` method, and hook its instantiation **after** the instantiation of this one.

Here the function will only call `$this->model->get_languages_list()`, but Polylang Pro's decorator will call its decorated instance's (which is the Polylang Decorator) `get_active_languges()` method and then filter to only get active languages.

modules/sitemaps/sitemaps-provider-decorator.php
Chouby: That's something we could introduce it if some day we need it. In most cases however, I use PLL_Model::get_language() to validate that the language exists and retrieve the corresponding object in one call.

modules/sitemaps/sitemaps-provider-decorator.php
Chouby: I will leave it here for now to avoid adding complexity.

PR #535, opened 2020-07-06
Wpml config translate metas

raaaahman:
This PR aims to allow the `wpml-config.xml` file to filter the metas that will be translated.

Moved from `polylang-po/2.8-dev`


Chouby:
Closed in favor of #1133

PR #536, opened 2020-07-07
Convert README to markdown

koppor:
GitHub uses Markdown as premier text language and renders it nicely.

Thus, I propose to update the README.md

Old:

grafik

New:

grafik

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,18 @@
-=== Polylang ===
-Contributors: Chouby, manooweb, raaaahman, marianne38
-Donate link: https://polylang.pro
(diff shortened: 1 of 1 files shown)


manooweb:
Thanks for this contribution but sorry we close this PR because we can't remove readme.txt because it is uses by WordPress.org to display  informations on its [description page](https://wordpress.org/plugins/polylang/#description).

To get information about Polylang and for non-developers people we recommend to refer to this page.

However we decided to use the readme.md markedown file to give information to developers who want to start with Polylang from Gihub.

Chouby:
See #817

PR #537, opened 2020-07-08
Use custom flags on demand

Chouby:
Up to now, and only on frontend, we replaced the default flags by the custom flags (when they exist) into the `PLL_Language` properties. This means that `PLL_Language::$flag` may contain a different flag depending on the context (frontend, backend, rest). This could confuse developpers and create bugs.

I propose to change the logic and to pull custom flags only on demand ( for the language switcher and the site health).
The PR 
* removes `PLL_Language:set_custom_flag()`
* introduces `PLL_Language:maybe_get_custom_flag()` and  `PLL_Language:maybe_get_custom_flag_url()` and uses this functions when needed.

Additionnaly, `PLL_Language:set_custom_flag()` was setting the url scheme for flag urls. This is moved to `PLL_Language:set_url_scheme()` which replaces `PLL_Language:set_home_url_scheme()`.

Fix https://github.com/polylang/polylang-pro/issues/556

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -240,21 +240,21 @@ class PLL_Language {
 	}
 
 	/**
-	 * Replace flag by custom flag
-	 * Takes care of url scheme
+	 * Returns the html of the custom flag if any, or the default flag otherwise.
 	 *
-	 * @since 1.7
+	 * @since 2.8
 	 */
-	public function set_custom_flag() {
-		// Overwrite with custom flags on frontend only
-		if ( ! empty( $this->custom_flag ) ) {
-			$this->flag = $this->custom_flag;
-			$this->flag_url = $this->custom_flag_url;
-			unset( $this->custom_flag, $this->custom_flag_url ); // hide this
-		}
+	public function maybe_get_custom_flag() {
+		return empty( $this->custom_flag ) ? $this->flag : $this->custom_flag;
+	}
 
-		// Set url scheme, also for default flags
-		$this->flag_url = set_url_scheme( $this->flag_url );
+	/**
+	 * Returns the url of the custom flag if any, or the default flag otherwise.
+	 *
+	 * @since 2.8
+	 */
+	public function maybe_get_custom_flag_url() {
+		return empty( $this->custom_flag_url ) ? $this->flag : $this->custom_flag_url;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


include/links-model.php
```diff
@@ -119,8 +119,8 @@ public function pll_after_languages_cache( $languages ) {
 				$this->set_home_url( $language );
 			}
 
-			// Ensures that the ( possibly cached ) home url uses the right scheme http or https
-			$language->set_home_url_scheme();
+			// Ensures that the ( possibly cached ) home and flag urls use the right scheme http or https.
+			$language->set_url_scheme();
```
raaaahman, line 123: Actually, the `PLL_Model::get_languages_list()` instantiate the `PLL_Languages` objects, then calls to its `pll_after_languages_cache` filter. (For some reason, [the languages may have already been stored once in the cache](https://github.com/polylang/polylang/blob/7bea3700a7293f8ef31b700fa0fe84647420c53b/include/model.php#L78), but they are also stored **after** this filter has been triggered...)

This triggers the `PLL_Links_Model::pll_after_languages_list()` method which calls the `PLL_Languages::set_url_scheme()` on each language instance.

Given that all this takes place at the time the `PLL_Language` objects are instantiated, I would find it clearer to read if it happened in the `PLL_Language` constructor.

Also, the `set_flag()` method seems to be systematically called each time the `PLL_Language` instances are created (see this [search](https://github.com/polylang/polylang/search?q=set_flag&unscoped_q=set_flag) ), this could be merged in the constructor as well.

include/language.php
```diff
@@ -240,21 +240,21 @@ public static function get_flag_html( $flag, $title = '', $alt = '' ) {
...
+	 * @since 2.8
 	 */
-	public function set_custom_flag() {
-		// Overwrite with custom flags on frontend only
-		if ( ! empty( $this->custom_flag ) ) {
-			$this->flag = $this->custom_flag;
-			$this->flag_url = $this->custom_flag_url;
-			unset( $this->custom_flag, $this->custom_flag_url ); // hide this
-		}
+	public function maybe_get_custom_flag() {
```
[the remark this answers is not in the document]
Chouby: When there is no flag, nothing is displayed on frontend. We don't have to display anything unlike on backend. Your proposed name is better.

include/links-model.php
Chouby: For `PLL_Languages::set_url_scheme()`, yes we are obliged to do like this, because urls are stored in the transient for both frontedn and amdin, and I have seen some sites with `http` on front and `https` on admin. That's probably less frequent now but I don't want to take the risk.
For `PLL_Languages::set_flag()`, probaby several modifications leaded to this situation.

raaaahman: > For `PLL_Languages::set_url_scheme()`, yes we are obliged to do like this, because urls are stored in the transient for both frontedn and amdin, and I have seen some sites with `http` on front and `https` on admin. That's probably less frequent now but I don't want to take the risk.

I still don't get it, but I'll pass.

PR #544, opened 2020-07-22
Re-minify all files

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/606

Some old code was remaining in minified js files. Thus the code could be present in 2 files and could be executed 2 times. I re-minified all js and and css files to remove all old traces.


Chouby:
Yes that's a practice that we have to give up.

Chrystll:
You fixed the issue. thank you.

PR #548, opened 2020-07-23
Prevent creating 2 terms with the same name and same language

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/557

This PR prevents creating a second term with the same name (and the same language) without specifying the slug. This was an inconsistent behaviour compared to WordPress which doesn't allow to this.

The PR introduces a new method `term_exists_by_slug()` which allows to check if a term slug already exists in a given language. This function is then used to check that the term with the default slug doesn't already exist before appending the language code to the evaluated slug.

A phpunit test tests the original bug.


Chrystll:
You fixed the issue +1

PR #552, opened 2020-08-07
Move sanitization filters for German and Danish to their own class

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/183

In fact, the PLL_Admin_Filters wasn't instantiated in a REST API context which is used by the block editor. Then sanitize_title hook wasn't trigger correctly in this context.

In addition the language code isn't always passed to the request when we save the post because the language is a block editor attribute of the post.

So this PR propose:

- to extract the code of PLL_Admin_Filters to a new class PLL_Admin_Filters_Sanitize to be able to instantiate it in both of the two contexts classic and block editor
- Instantiate the new PLL_Admin_Filters_Sanitize class at the right place for both of the two contexts
- Retrieve the language code just before dispatching the REST API request to set it as the current language to make sanitization work fine for German and Danish

Unit test has been added too but in Polylang Pro repository because we need the REST API compatibility which isn't implemented in Polylang but in Polylang Pro


admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,132 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * Setup specific admin filters usefull for sanitization.
```
raaaahman, line 7: Typo at 'usefull'

admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * Extract from PLL_Admin_Filters to be able to use in a REST API context.
```
raaaahman, line 9: This should rather be in the PR's comment. Actually, this class is also used in a non REST API context.

admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,132 @@
...
+			// When it's a post request on a new post the language is sent into the request.
+			if ( ! empty( $request->get_param( 'lang' ) ) ) {
+				$this->curlang = $this->model->get_language( sanitize_key( $request->get_param( 'lang' ) ) );
+			} else {
+				// Otherwise we need to get the language from the post itself.
+				$this->curlang = $this->model->post->get_language( sanitize_key( $request->get_param( 'id' ) ) );
+			}
+		}
+		return $result;
+	}
```
raaaahman, line 59: I don't believe this behavior is needed in the `PLL_Admin` context that also uses this class. This should be moved in a REST specific class.

admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,132 @@
...
+	/**
+	 * Maybe fix the result of sanitize_user() in case the languages include German or Danish
+	 *
+	 * @since 2.0
+	 *
+	 * @param string $username     Sanitized username.
+	 * @param string $raw_username The username prior to sanitization.
+	 * @param bool   $strict       Whether to limit the sanitization to specific characters. Default false.
+	 * @return string
+	 */
```
raaaahman, line 119: This doc block should mention that all the logic is performed in WordPress' [remove_accents()](https://github.com/WordPress/wordpress-develop/blob/c7783dc49a016b98480d2966b14a59674bf41588/src/wp-includes/formatting.php#L1584) function.

admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,132 @@
...
+	 * used for the admin interface and if one this language is German or Danish, some specific
+	 * characters such as ä, ö, ü, ß are incorrectly sanitized.
+	 *
+	 * @since 2.0
+	 *
+	 * @param string $title     Sanitized title.
+	 * @param string $raw_title The title prior to sanitization.
+	 * @param string $context   The context for which the title is being sanitized.
+	 * @return string
+	 */
```
raaaahman, line 96: This doc block should mention that all the logic is performed in WordPress' [remove_accents()](https://github.com/WordPress/wordpress-develop/blob/c7783dc49a016b98480d2966b14a59674bf41588/src/wp-includes/formatting.php#L1584) function.

admin/admin-filters-sanitize.php
manooweb: Maybe it wasn't clear but the code already existed before in PLL_Admin_Filters which is used in classic editor context.
One part of PLL_Admin_Filters_Sanitize class extract the code to sanitize title and username and be able to instantiate it in the two contexts.

admin/admin-filters-sanitize.php
manooweb: This part has been moved to Polylang Pro REST API module in PLL_REST_Post which already add the `rest_pre_dispatch` filter.
See the PR https://github.com/polylang/polylang-pro/pull/628

admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,126 @@
...
+	/**
+	 * Constructor: setups filters and actions
+	 *
+	 * @since 2.9
+	 *
+	 * @param object $polylang
+	 */
+	public function __construct( &$polylang ) {
+		$this->model = &$polylang->model;
+		$this->curlang = &$polylang->curlang;
```
raaaahman, line 37: `$curlang` property doesn't exists in `PLL_REST_Request`. By assigning a shared reference here, you are dynamically creating it in the `$polylang` object.

admin/admin-filters-sanitize.php
```diff
@@ -0,0 +1,126 @@
...
+	 * @param string $locale
+	 * @return string
+	 */
+	public function get_locale( $locale ) {
+		if ( isset( $_POST['post_lang_choice'] ) && $lang = $this->model->get_language( sanitize_key( $_POST['post_lang_choice'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$locale = $lang->locale;
+		} elseif ( isset( $_POST['term_lang_choice'] ) && $lang = $this->model->get_language( sanitize_key( $_POST['term_lang_choice'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$locale = $lang->locale;
+		} elseif ( isset( $_POST['inline_lang_choice'] ) && $lang = $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$locale = $lang->locale;
```
raaaahman, line 62: I don't believe these cases will be relevant in a REST request context.

raaaahman (commented):
I recall discussing about subclassing the sanitization filters in:

- One class for the admin context
- One class for the REST context

admin/admin-filters-sanitize.php
manooweb: PLL_Admin_Filters_Sanitize also concerns classical requests - ajax call for example when you modify the slug of the permalink in the classic editor context - not only REST ones

admin/admin-filters-sanitize.php
manooweb: $curlang has been replaced by $locale property which is used only internally to PLL_Filters_Sanitization class

admin/admin-filters-sanitize.php
manooweb: Finally this code has been moved in PLL_Admin class because PLL_Filters_Sanitization instantiation is done in PLL_Admin::add_filters() method and need the locale and thus use this code to retrieve the right locale in a classic editor context

raaaahman (approved):
The code looks cleaner now that you have separated the responsibilities. Good job.

admin/admin.php
```diff
@@ -8,30 +8,80 @@
...
+	public $filters_media;
+
+	/**
+	 * Instance of PLL_Filters_Sanitization
+	 *
+	 * @since 2.9
+	 * 
+	 * @var PLL_Filters_Sanitization
+	 */
+	public $filters_sanitization;
```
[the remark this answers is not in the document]
Chouby: I agree with @raaaahman but I much prefer that each PR does only one thing. Let's keep this for this time, but please open a new PR to complete the job on the properties documentation.

admin/admin.php
```diff
@@ -121,4 +172,27 @@ public function add_filters() {
 			$this->$obj = new $class( $this );
 		}
 	}
+	/**
+	 * Retrieve the locale according to the current language instead of the language
+	 * of the admin interface.
+	 *
+	 * @since 2.9
```
Chouby, line 179: I prefer to keep the original version `@since 2.0` as the funtion was not created for 2.9.

admin/admin.php
```diff
@@ -121,4 +172,27 @@ public function add_filters() {
...
+	 *
+	 * @return string
+	 */
+	public function get_locale() {
```
Chouby, line 183: I would rename this method now to something like `get_locale_for sanitization` or something like that to be more precise on its usage.

include/filters-sanitization.php
```diff
@@ -0,0 +1,116 @@
...
+	 * @param object $polylang
+	 * @param string $locale Locale code of the language.
+	 */
+	public function __construct( &$polylang, $locale ) {
+		$this->model  = &$polylang->model;
+		$this->locale = $locale;
+
+		// We need specific filters for some languages like German and Danish
+		$specific_locales = array( 'da_DK', 'de_DE', 'de_DE_formal', 'de_CH', 'de_CH_informal', 'ca', 'sr_RS', 'bs_BA' );
+		if ( array_intersect( $this->model->get_languages_list( array( 'fields' => 'locale' ) ), $specific_locales ) ) {
```
Chouby, line 42: Now that we pass `$locale` to this class, we should not need all languages anymore. So I suppose that we can just test that `$locale` is in the array `$specific_locales`. This should also simplify the class (no more need to pass `$polylang` and use the `PLL_Model` instance.

include/filters-sanitization.php
```diff
@@ -0,0 +1,116 @@
...
+			add_filter( 'sanitize_title', array( $this, 'sanitize_title' ), 10, 3 );
+			add_filter( 'sanitize_user', array( $this, 'sanitize_user' ), 10, 3 );
+		}
+	}
+
+	/**
+	 * Retrieve the locale code of the language.
+	 *
+	 * @since 2.9
```
Chouby, line 51: I would keep `@since 2.0`

include/rest-request.php
```diff
@@ -8,20 +8,61 @@
...
+	 * @var PLL_Frontend_Filters
+	 */
+	public $filters;
+
+	/**
+	 * Instance of PLL_Filters_Links
+	 *
+	 * @var PLL_Filters_Links
+	 */
+	public $filters_links;
```
Chouby, line 65: I don't believe that any modification to this file is necessary.

Chouby (changes requested):

Chouby:
Since the new class is meant to filter **sanitize**_tiltle and **sanitize**_user, what about to rename it `PLL_Filters_Sanitize` rather than `PLL_Filters_Sanitization`? Not a big deal. Up to you.

admin/admin.php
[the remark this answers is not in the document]
manooweb: 🤔 however you asked to add declaration and documentation in your comment here https://github.com/polylang/polylang-pro/pull/628#discussion_r492645535 😁

include/filters-sanitization.php
manooweb: Done

include/rest-request.php
manooweb: Due to the fact ay the beginning of the PR we instantiated PLL_Filters_Sanitization class in this class.
I'm going to revert these modifications

manooweb:
> Since the new class is meant to filter **sanitize**_tiltle and **sanitize**_user, what about to rename it `PLL_Filters_Sanitize` rather than `PLL_Filters_Sanitization`? Not a big deal. Up to you.

I named this class like that at the beginning but I'd prefer to not use verb but a noun for class name which modelizes an object. I'd prefer verb for methods which describe actions with the object or with something.

admin/admin.php
manooweb: Done

include/filters-sanitization.php
manooweb: Yes! Good Idea, I'm going to implement it.

include/filters-sanitization.php
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * Constructor: setups filters and actions.
+	 *
+	 * @since 2.9
+	 *
+	 * @param object $polylang
+	 * @param string $locale Locale code of the language.
+	 */
+	public function __construct( &$polylang, $locale ) {
+		$this->model  = &$polylang->model;
```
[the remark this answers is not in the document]
manooweb: See the @Chouby's suggestion https://github.com/polylang/polylang/pull/552#discussion_r492638303

include/filters-sanitization.php
manooweb: Done !

tests/phpunit/tests/test-admin-filters.php
```diff
@@ -17,27 +17,39 @@ function setUp() {
...
-		self::$polylang->filters->curlang = self::$polylang->model->get_language( 'en' );
+		$pll_admin = new PLL_Admin( self::$polylang->links_model );
+		$pll_admin->curlang = self::$polylang->model->get_language( 'en' );
+		$pll_admin->add_filters();
 		$this->assertEquals( 'angstrom', sanitize_user( 'ångström' ) );
 
-		self::$polylang->filters->curlang = self::$polylang->model->get_language( 'de' );
+		$pll_admin = new PLL_Admin( self::$polylang->links_model );
+		$pll_admin->curlang = self::$polylang->model->get_language( 'de' );
+		$pll_admin->add_filters();
```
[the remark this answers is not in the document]
manooweb: After applying the @Chouby's suggestion https://github.com/polylang/polylang/pull/552#discussion_r492638303 you're right: test_sanitize_title() breaks at the third assertion because the sanitization filter is still available.

I think it's interesting to keep all the case - and so assertions - in the same method. So I put a remove_filter() call before each case.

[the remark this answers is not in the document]
manooweb: After talking about during the last DM. I splitted these unit tests to separate each case.

tests/phpunit/tests/test-admin-filters.php
```diff
@@ -15,38 +15,42 @@ function setUp() {
...
 		$this->assertEquals( 'fullmenge', sanitize_title( 'Füllmenge' ) );
-
+	}
+	function test_sanitize_title_for_language_from_form_without_character_conversion() {
+			// Bug fixed in 2.4.1
+		$_POST['post_lang_choice'] = 'en';
+		$pll_admin = new PLL_Admin( self::$polylang->links_model );
+		$pll_admin->add_filters();
+		$this->assertEquals( 'fullmenge', sanitize_title( 'Füllmenge' ) );
+	}
```
raaaahman: These tests always pass, whether we set the filters or not.

tests/phpunit/tests/test-admin-filters.php
```diff
@@ -15,38 +15,42 @@ function setUp() {
...
-		$_POST['post_lang_choice'] = 'en';
-		$pll_admin = new PLL_Admin( self::$polylang->links_model );
-		$pll_admin->add_filters();
-		$this->assertEquals( 'fullmenge', sanitize_title( 'Füllmenge' ) );
-
 		$_POST['post_lang_choice'] = 'de';
 		$pll_admin = new PLL_Admin( self::$polylang->links_model );
 		$pll_admin->add_filters();
 		$this->assertEquals( 'fuellmenge', sanitize_title( 'Füllmenge' ) );
 	}
```
raaaahman: These are explicit names for tests, good choice!

tests/phpunit/tests/test-admin-filters.php
```diff
@@ -15,38 +15,42 @@ function setUp() {
...
 
-	function test_sanitize_user() {
+	function test_sanitize_user_without_character_conversion() {
 		$pll_admin = new PLL_Admin( self::$polylang->links_model );
 		$pll_admin->curlang = self::$polylang->model->get_language( 'en' );
 		$pll_admin->add_filters();
 		$this->assertEquals( 'angstrom', sanitize_user( 'ångström' ) );
+	}
```
raaaahman: This test never fails, whether we set the filters or not.

raaaahman (commented):
This is even simpler than before, well done.

However, I fear I didn't really understood the scope of the `PLL_Admin_Filters::get_locale()` method in the previous iterations...

tests/phpunit/tests/test-admin-filters.php
manooweb: 🤔 These tests exist before. I think that this the goal: verify that our filter doesn't break anything for standard languages - just removing accents.
Notice that WordPress apply a default sanitization by applying sanitize_title filter and then executing sanitize_title_with_dashes() formating function 
https://developer.wordpress.org/reference/functions/sanitize_title_with_dashes/

tests/phpunit/tests/test-admin-filters.php
manooweb: Same kind of answer than above https://github.com/polylang/polylang/pull/552#discussion_r494866974
Notice for sanitize_user filter the default sanitization function applied is strtolower

admin/admin.php
```diff
@@ -176,11 +176,11 @@ public function add_filters() {
 	 * Retrieve the locale according to the current language instead of the language
 	 * of the admin interface.
 	 *
-	 * @since 2.9
+	 * @since 2.0
 	 *
 	 * @return string
 	 */
-	public function get_locale() {
+	public function get_locale_for_sanitization() {
```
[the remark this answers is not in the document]
manooweb: It's locale to correctly instantiate PLL_Filters_Sanitization which is base on this local to activate or not our sanitization filters.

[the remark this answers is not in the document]
raaaahman: OK

raaaahman (approved):

PR #553, opened 2020-08-07
Whitelist jQuery methods for VIP WPCS

manooweb:
This PR propose after verification to whitelist all the jQuery methods: prepend, replaceWith, before and after.
In addition I noticed that some i18n strings weren't correctly escaped in the wizard which are used in the javascript code. So this PR also fixes this issue.

This is the part of javascript code to close this issue https://github.com/polylang/polylang-pro/issues/596

### DIFF
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -26,7 +26,14 @@ class PLL_Wizard {
 	/**
 	 * List of steps
 	 *
-	 * @var array $steps
+	 * @var array $steps {
+	 *     @type string $name      i18n string which names the step.
+	 *     @type callable $view    The callback function use to display the step content.
+	 *     @type callable $handler The callback function use to process the step after it is submitted.
+	 *     @type array $scripts    List of scripts handle needed by the step.
+	 *     @type array $styles     The list of styles handle needed by the step.
+	 *
+	 * }
 	 */
 	protected $steps = array();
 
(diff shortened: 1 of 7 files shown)


modules/wizard/wizard.php
```diff
@@ -427,27 +427,27 @@ public function add_step_languages( $steps ) {
...
+				'i18n_pll_invalid_flag'       => esc_html__( 'The flag does not exist', 'polylang' ),
+				'i18n_dialog_title'           => esc_html__( "A language wasn't added.", 'polylang' ),
+				'i18n_dialog_yes_button'      => esc_html__( 'Yes', 'polylang' ),
+				'i18n_dialog_no_button'       => esc_html__( 'No', 'polylang' ),
+				'i18n_dialog_ignore_button'   => esc_html__( 'Ignore', 'polylang' ),
+				'i18n_remove_language_icon'   => esc_html__( 'Remove this language', 'polylang' ),
 			)
 		);
 		wp_enqueue_script( 'pll-wizard-languages' );
 		wp_enqueue_style( 'pll-wizard-selectmenu', plugins_url( '/css/selectmenu' . $this->get_suffix() . '.css', POLYLANG_FILE ), array( 'dashicons', 'install', 'common', 'wp-jquery-ui-dialog' ), POLYLANG_VERSION );
```
raaaahman, line 448: The whole register / localize / enqueue scripts process could be encapsulated into a custom function. So we could automatically escape every script we 'localize'.

PR #556, opened 2020-08-14
2.6.x

seattlekoro:
My web site doesn't work with the newest version because php version is old. I want to down grade the plugin.


Chouby:
I don't understand the purpose of this PR.

PR #557, opened 2020-08-17
Fix duplicate post compatibility

Chouby:
Since version 2.8, we receive a warning when the plugin Duplicate Post is active. 
```
Warning: call_user_func_array() expects parameter 1 to be a valid callback, class 'PLL_Duplicate_Post' does not have a method 'taxonomies_blacklist' in /wp-includes/class-wp-hook.php on line 289
```
It's due to a mistake made during the integration refactor.
This PR fixes the issue.

### DIFF
--- a/integrations/duplicate-post/duplicate-post.php
+++ b/integrations/duplicate-post/duplicate-post.php
@@ -27,7 +27,7 @@ class PLL_Duplicate_Post {
 	 * @param array|string $taxonomies
 	 * @return array
 	 */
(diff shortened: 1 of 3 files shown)


Marianne380 (approved):

Chrystll:
Validated. Thank you, you fixed the issue.

tremblaydanny:
Thank you for fixing this issue! Duplicate post works perfectly now.

PR #558, opened 2020-08-19
Fix get_display_flag_url() returning html instead of url

Chouby:
In version 2.8, we introduced the method `PLL_Language::get_display_flag_url()` with a typo. Thus, it returns the html flag instead of the flag url. As a consequence, custom language switcher using `pll_the_languages()` with the `raw` option don't get what they expect for the flag.

This PR fixes the issue and adds a few more assertions in the test of `pll_the_languages()`.


raaaahman (changes requested):
Could you add a specific test for this case instead of writing more assertions in a test that is already used for another responsibility?

I'm not confident that I will remember where the custom flags are tested if I have to modify it in the future.

Also, I can't say if you are testing with a custom flag or a default one.

Chouby:
The issue is not with custom flags which are tested elsewhere: https://github.com/polylang/polylang/blob/2.8/tests/phpunit/tests/test-custom-data.php

The fact is that I completed a test that was not complete enough as a lot of elements of the returned array of `pll_the_languages()` were tested, but not the flag. Also `pll_the_languages()` is part of the API so it seemed good to me to test the output of this function rather than the internal code that we changed and may change in the future. This also through this function that our customers experienced the bug.

What else do you propose?

tests/phpunit/tests/test-flags.php
```diff
@@ -23,6 +23,12 @@ function setUp() {
 		self::$polylang = new PLL_Frontend( self::$polylang->links_model );
 	}
 
+	function test_default_flag() {
+		$lang = self::$polylang->model->get_language( 'en' );
+		$this->assertEquals( plugins_url( '/flags/us.png', POLYLANG_FILE ), $lang->get_display_flag_url() ); // Bug fixed in 2.8.1.
+		$this->assertEquals( 1, preg_match( '#<img src="data:image\/png;base64,(.+)" title="English" alt="English" width="16" height="11" style="(.+)" \/>#', $lang->get_display_flag() ) );
```
raaaahman, line 29: You have the [assertStringMatchesFormat()](https://phpunit.de/manual/5.7/en/appendixes.assertions.html#appendixes.assertions.assertStringMatchesFormat) assertions if you want to test with regexps.

raaaahman (approved):
I feel you have gone a little far to avoid duplicating a `sprintf()` call in https://github.com/polylang/polylang/pull/559/files ...

tests/phpunit/tests/test-flags.php
Chouby: If you are thinking to https://github.com/polylang/polylang/pull/559/files#diff-34d98b479a021607a28a4070fb0bea86R241-R252, my goal was not to avoid duplicating `sprintf()` calls but to find a way to display the inline style when we have a height and/or width and noting wen we have nothing. I chose this way rather than making several nested tests.

I didn't know  `assertStringMatchesFormat`. Useful.

Chrystll:
Validated. Thank you for the fixed.

PR #566, opened 2020-08-27
Make flag tests independent from other tests

raaaahman:
## The issue

By running the `Flags_Test` before the `Switcher_Test`, with no other tests, a fatal error would arise at the `PLL_Frontend_Filters::get_locale()` method, triggered by the [get_locale() call](https://github.com/polylang/polylang/blob/bb5e7f55ab90277363d4077586bb9eaabe12f752/include/olt-manager.php#L75) in the global instance of `PLL_OLT_Manager` (which is shared between tests because instantiated in the bootstrap and being a Singleton).

For some reasons, the filters set in the `Flags_Test` **where not removed** when running the `Switcher_Test`, and where triggered by the `PLL_Frontend::init()` calls in the `Switcher_Test::setUp()`, while some of the values needed for this filter have been cleaned after the `Flag_Test`...

## How has it been fixed?

By calling the `PLL_Frontend::init()` method in the `Flag_Test::setUp()`, it seems that the filters instantiated are resolved and then not shared with other test classes.

## Other issues

Setting a new `PLL_Model` instance to replace the shared `PLL_Admin_Model` instance in the `PLL_Frontend` instance, two more issues have been outlined:

- It appeared that `Flags_Test::test_default_flag_ssl()` and `Flags_Test::test_custom_flag_ssl()` used cached values (coming from the shared `PLL_Admin_Model::cache` shared instance). Setting a new `PLL_Model` has reset these cached values, which would then not be set for SSL protocol. Resetting the cache **before** each test would solve the issue.
- Because our `PLL_UnitTestCaseTrait::wpTearDownAfterTest()` relies on the `self::$polylang->model->delete_language()` method to clean the environment, we cannot substitute the static `self::$polylang->model` instance for another class than `PLL_Admin_Model`. This was solved by using a different instance in the test methods.

### DIFF
--- a/tests/phpunit/tests/test-flags.php
+++ b/tests/phpunit/tests/test-flags.php
@@ -2,6 +2,8 @@
 
 class Flags_Test extends PLL_UnitTestCase {
 
+	private $pll_env;
+
 	/**
 	 * @param WP_UnitTest_Factory $factory
 	 */
@@ -20,6 +22,15 @@ class Flags_Test extends PLL_UnitTestCase {
 
 		unlink( WP_CONTENT_DIR . '/polylang/fr_FR.png' );
 		rmdir( WP_CONTENT_DIR . '/polylang' );
+	}
+
+	function setUp() {
+		$options       = array_merge( PLL_Install::get_default_options(), array( 'default_lang' => 'en_US' ) );
+		$model         = new PLL_Model( $options );
+		$links_model   = new PLL_Links_Default( $model );
+		$this->pll_env = new PLL_Frontend( $links_model );
+		$this->pll_env->init();
+		$this->pll_env->model->cache->clean();
 	}
 
 	function test_default_flag() {
--- a/tests/phpunit/tests/test-flags.php
+++ b/tests/phpunit/tests/test-flags.php
@@ -35,12 +35,6 @@ class Flags_Test extends PLL_UnitTestCase {
 		$this->pll_env->model->cache->clean();
 	}
 
-	function tearDown() {
-		remove_filter( 'pll_custom_flag', array( $this, 'override_custom_flag' ) );
-
-		parent::tearDown();
-	}
-
 	function test_default_flag() {
 		$lang = self::$model->get_language( 'en' );
 		$this->assertEquals( plugins_url( '/flags/us.png', POLYLANG_FILE ), $lang->get_display_flag_url() ); // Bug fixed in 2.8.1.
--- a/tests/phpunit/tests/test-flags.php
+++ b/tests/phpunit/tests/test-flags.php
@@ -25,12 +25,20 @@ class Flags_Test extends PLL_UnitTestCase {
 	}
 
 	function setUp() {
+		parent::setUp();
+
 		$options       = array_merge( PLL_Install::get_default_options(), array( 'default_lang' => 'en_US' ) );
 		$model         = new PLL_Model( $options );
 		$links_model   = new PLL_Links_Default( $model );
 		$this->pll_env = new PLL_Frontend( $links_model );
 		$this->pll_env->init();
 		$this->pll_env->model->cache->clean();
+	}
+
+	function tearDown() {
+		remove_filter( 'pll_custom_flag', array( $this, 'override_custom_flag' ) );
+
+		parent::tearDown();
 	}
 
 	function test_default_flag() {
--- a/tests/phpunit/tests/test-flags.php
+++ b/tests/phpunit/tests/test-flags.php
@@ -22,7 +22,7 @@ class Flags_Test extends PLL_UnitTestCase {
 
 		$options       = array_merge( PLL_Install::get_default_options(), array( 'default_lang' => 'en_US' ) );
 		$this->model         = new PLL_Model( $options );
-		$links_model = new PLL_Links_Default( $model ); // Registers the 'pll_languages_list' and 'pll_after_languages_cache' filters.
+		$links_model = new PLL_Links_Default( $this->model ); // Registers the 'pll_languages_list' and 'pll_after_languages_cache' filters.
 	}
 
 	public function tearDown() {
--- a/tests/phpunit/tests/test-flags.php
+++ b/tests/phpunit/tests/test-flags.php
@@ -13,26 +13,28 @@ class Flags_Test extends PLL_UnitTestCase {
 		self::create_language( 'en_US' );
 		self::create_language( 'fr_FR' );
 
-		@mkdir( WP_CONTENT_DIR . '/polylang' );
-		copy( dirname( __FILE__ ) . '/../data/fr_FR.png', WP_CONTENT_DIR . '/polylang/fr_FR.png' );
-	}
-
-	static function wpTearDownAfterClass() {
-		parent::wpTearDownAfterClass();
-
-		unlink( WP_CONTENT_DIR . '/polylang/fr_FR.png' );
-		rmdir( WP_CONTENT_DIR . '/polylang' );
+		$wp_filter['pll_languages_list']->remove_all_filters();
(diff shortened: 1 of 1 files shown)


Chouby:
I believe that you spotted general issues that we should fix globally:
* Hooks are backuped by WordPress in `setUp()` https://github.com/WordPress/wordpress-develop/blob/5.5.0/tests/phpunit/includes/abstract-testcase.php#L115 and restored in `tearDown()`. That means that if some hooks are setup in `setUpBeforeClass()`, they won't be removed. That may the reason you were looking for.
* Could we not disable the cache or at least clean it globally in `PLL_UnitTestCaseTrait`?

raaaahman:
>     * Hooks are backuped by WordPress in `setUp()` https://github.com/WordPress/wordpress-develop/blob/5.5.0/tests/phpunit/includes/abstract-testcase.php#L115 and restored in `tearDown()`. That means that if some hooks are setup in `setUpBeforeClass()`, they won't be removed. That may the reason you were looking for.

Not really, since the filters were registered during the `setUp()` but... have you noticed this class `setUp()` method doesn't call to `parent::setUp()`?

raaaahman:
>     * Could we not disable the cache or at least clean it globally in `PLL_UnitTestCaseTrait`?

The idea of mocking the `PLL_Cache` [like you did in Switcher_Test::setUp()](https://github.com/polylang/polylang/blob/b5e0a0fce5412e1c2686f723bba4e3fdb504de9f/tests/phpunit/tests/test-switcher.php#L26-L27) is a clean way to do it, but we can't actually generalize this behavior:

- The reason I didn't use it in these tests is that [PLL_Model::get_language() will need to get the language from the cache](https://github.com/polylang/polylang/blob/bb5e7f55ab90277363d4077586bb9eaabe12f752/include/model.php#L183).
- [A new PLL_Cache is instantiated during PLL_Model::__construct()](https://github.com/polylang/polylang/blob/bb5e7f55ab90277363d4077586bb9eaabe12f752/include/model.php#L28), which means we have to replace it each time we use a new PLL_Model, which we should do because our production code sometimes use a `PLL_Model`, and sometimes a `PLL_Admin_Model` (which is not reflected in most of our test at the moment).

The cleanest solution would be to make our cache as a decorator of the model, overriding the methods which may pull results from the cache. So we could instantiate a model without its cache in our tests (and in production as well).

Cleaning the cache has to be done after all the instantiation have been made in the `setUp()`, given the test scenario, some instantiation could as well be done in the test itself. I don't really see a way to generalize this process.

manooweb (approved):

tests/phpunit/tests/test-flags.php
```diff
@@ -20,17 +22,24 @@ static function wpTearDownAfterClass() {
...
 	function setUp() {
-		self::$polylang = new PLL_Frontend( self::$polylang->links_model );
+		parent::setUp();
+
+		$options       = array_merge( PLL_Install::get_default_options(), array( 'default_lang' => 'en_US' ) );
+		$model         = new PLL_Model( $options );
+		$links_model   = new PLL_Links_Default( $model );
+		$this->pll_env = new PLL_Frontend( $links_model );
+		$this->pll_env->init();
+		$this->pll_env->model->cache->clean();
```
Chouby, line 32: Is it useful? You just instantiated the PLL_Model object. So I expect that the cache is clean. Am I Wrong?

raaaahman: In fact, if you remove this line, one test fails. Because any previous call to `PLL_Model::get_language()` would have set values in the cache, we would otherwise be checking values set from the `PLL_Frontend::init()` values above rather than those computed in the test themselves...

On another hand, it doesn't seem that instantiating and initializing a `PLL_Frontend` object is mandatory for these tests to pass. Although, I am not sure were the code that is supposed to make these tests pass resides, so I am not able to check whether setting a `PLL_Frontend` instance is relevant or not.

Marianne380 (approved):

tests/phpunit/tests/test-flags.php
Chouby: I thought that I already answered to this. I may have forgotten to push the green button :(

It looks to me that since the tests use only the `PLL_Model` instance, we have no need for the `PLL_Frontend` and `PLL_Links_Default` instances.

raaaahman: In fact, a lot of the logic tested here happens in [PLL_Links_Model::pll_languages_list()](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/links-model.php#L98-L104) and [PLL_Links_Model::pll_after_languages_cache()](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/links-model.php#L115-L126), which are triggered on [pll_languages_list](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/model.php#L89) and [pll_after_languages_cache](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/model.php#L116).

These filters are registered during the [instantiation of the PLL_Links_Model subclass](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/links-model.php#L29-L30) in [PLL_Model::get_links_model()](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/model.php#L594), that happens twice:

- During the tests bootstrap, in [Polylang::init()](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/include/class-polylang.php#L192), which is triggered by WordPRess `plugins_loaded` hook.
- In [PLL_UnitTestCase::wpSetUpBeforeClass()](https://github.com/polylang/polylang/blob/5d2a023bf06e0159022a1e716a4df5d027d880fc/tests/phpunit/includes/testcase-trait.php#L23)

Both filters are backed up in [WP_UnitTestCase_Base](https://github.com/WordPress/wordpress-develop/blob/3ebeaffe80d31511a0de7f0aa37fb181d9832e82/tests/phpunit/includes/abstract-testcase.php#L111), then triggered in each test.

The commit [c22d9990b](https://github.com/polylang/polylang/pull/566/commits/c22d90b41d428df0a1d1cacc4953b4bc426149ab) removes both of these hooks to use some set in the test `setUp()` method instead.

tests/phpunit/tests/test-flags.php
```diff
@@ -3,34 +3,48 @@
...
+
 		parent::wpSetUpBeforeClass();
 
 		self::create_language( 'en_US' );
 		self::create_language( 'fr_FR' );
 
-		@mkdir( WP_CONTENT_DIR . '/polylang' );
-		copy( dirname( __FILE__ ) . '/../data/fr_FR.png', WP_CONTENT_DIR . '/polylang/fr_FR.png' );
+		$wp_filter['pll_languages_list']->remove_all_filters();
+		$wp_filter['pll_after_languages_cache']->remove_all_filters();
```
raaaahman, line 14: This could be done in almost every test class.

raaaahman:
Note that tests for custom flags where performed with the `fr` language, whereas test for default flags where done with the `en`. I'd rather always use the same language for both custom and default flags, that way I could make sure that the tests pass because of their own code rather than because of the state set in a previous test.

raaaahman:
This PR is still needed, #742 didn't change much these files: https://github.com/polylang/polylang/pull/742/files#diff-3b8c2e71ce02459eb059613bfbf2c3751b12dbf54305c453fba9374d3003f43c

Chouby:
Closed in favor of #961

PR #567, opened 2020-08-28
Add test for not outputting inline style for custom flags

raaaahman:
## Before

#559 Print some CSS styles as inline in the admin dashboard when custom flags are set.

## Changes

Because our users may want to write their own CSS rules for their custom flags, we shouldn't add an inline style attribute unless the user specified some dimensions by hooking on the `pll_custom_flag` filter.

This PR adds tests that verify inline style will not be printed for custom flags on the fronted.


manooweb (approved):
Ok 👍

Marianne380 (approved):

PR #570, opened 2020-09-08
Fix PHP notice in sitemap module

Chouby:
Reported in wp.org forum: https://wordpress.org/support/topic/notice-undefined-variable-slug-in/

```
Notice: Undefined variable: slug in /path/wp-content/plugins/polylang/modules/sitemaps/sitemaps.php on line 133
```
when visiting Settings > Permalinks.
with WP 5.5 and Polylang 2.8.1

I did not test, but my guess is that it occurs when only one language is defined.

### DIFF
--- a/modules/sitemaps/sitemaps.php
+++ b/modules/sitemaps/sitemaps.php
@@ -129,7 +129,7 @@ class PLL_Sitemaps {
 		}
 
 		foreach ( $rules as $key => $rule ) {
-			if ( false !== strpos( $rule, 'sitemap=$matches[1]' ) ) {
+			if ( isset( $slug ) &&  false !== strpos( $rule, 'sitemap=$matches[1]' ) ) {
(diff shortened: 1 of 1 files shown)


manooweb:
> Reported in wp.org forum: https://wordpress.org/support/topic/notice-undefined-variable-slug-in/
> 
> ```
> Notice: Undefined variable: slug in /path/wp-content/plugins/polylang/modules/sitemaps/sitemaps.php on line 133
> ```
> 
> when visiting Settings > Permalinks.
> with WP 5.5 and Polylang 2.8.1
> 
I didn't have this notice by visiting this 🤔 (Polyalng 2.8.2)

> I did not test, but my guess is that it occurs when only one language is defined.

But I have it when I delete the only one language defined.

Otherwise the PR fix the issue.

manooweb (approved):
ok 👍

PR #573, opened 2020-09-18
Design selectmenu fields as other fields in settings languages pages

manooweb:
This PR propose to fix https://github.com/polylang/polylang-pro/issues/662

First
- width of fields is defined in WordPress edit styles https://github.com/WordPress/WordPress/blob/5.5.1/wp-admin/css/edit.css#L1092-L1102
- jQuery UI selectmenu widget is both used in the Polylang settings languages page and the wizard.
- in the wizard we need a 100% width for responsive reason
- width of jQuery UI selectmenu widget is managed in JS as an option of the widget to be able to use % unit and introduces a style attribute which isn't overloadable in external styles without the use of `!important` - which isn't a good practice
- notice jQuery UI selectmenu widget substract 2% of the original width defined 🤔

So for both the `#lang_list` and `#flag_list` I defined the jQuery UI selectmenu width to 97% for having the real value 95%.
For  the `#lang_list` in the wizard I forced the value to 100% by setting the width option of the  jQuery UI selectmenu to 102%.

The jQuery UI selectmenu seems to be aligned to the other fields now
image

For design issues, background color color and color are already the same as WordPress form's colors. I completed the selectmenu styles by adding missing properties (radius, border-color, box-shadow ...) which are come from WordPress form's styles
https://github.com/WordPress/WordPress/blob/5.5.1/wp-admin/css/forms.css#L38-L59
and
https://github.com/WordPress/WordPress/blob/5.5.1/wp-admin/css/forms.css#L86-L108
 and targeting only the settings languages page

Not focused field
image

Focused field
image


css/selectmenu.css
```diff
@@ -146,6 +146,18 @@
 	box-shadow: 0 1px 2px rgba(0, 0, 0, 0.07) inset;
 	color: #32373c;
 }
+/* Override to have same styles as WP form styles since WordPress 5.4 */
```
raaaahman, line 149: That's the jQuery UI styles that you are overriding, right?

raaaahman (approved):

css/selectmenu.css
manooweb: Specifically Polylang selectmenu styles overwrite form's WordPress styles because jQuery UI selectmenu widget generate a `span` HTML tag which isn't target by WordPress form's styles which target only HTML standard form tags (input... , select, textarea ... )
So we need to target specifically jQuery UI selectmenu widget in the Polylang settings languages  page.

Chouby: As long as WordPress doesn't use a select library. For the reference: https://core.trac.wordpress.org/ticket/31696 was planned for 4.1, then 4.4 then the project focused on other priorities ;-)

PR #577, opened 2020-09-27
Delay load_plugin_textdomain after plugin initialization

Chouby:
The best practice is to load the translation after all plugins are loaded to allow 3rd party plugins to optionnaly change the translation file loaded. This is already what we do and that's ok. However, during the developement of https://github.com/polylang/polylang-pro/pull/668, I noticed that our modules are loaded after the translation file is loaded. This means that it's impossible for our modules to modify the translation file for our own translations. 

This PR delays the call to `load_plugin_textdomain` after the modules are loaded, allowing them to filter the translation file loaded.


manooweb (approved):
ok 👍

PR #578, opened 2020-09-27
Merge instead of overriding locales for translations updates

Chouby:
The method `update_check_locales` hooked to `plugins_update_check_locales` and `themes_update_check_locales` completely overrides the locales passed to these filters. There is a risk to loose the locales if they are not in our languages list.

This PR keeps the locales passed and adds our own locales to the list instead of overriding the values.


manooweb (approved):
Ok 👍

PR #579, opened 2020-09-27
remove the "title" markup

sebastienserre:
solve the [#578](https://github.com/polylang/polylang-pro/issues/578) issue

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -233,7 +233,6 @@ class PLL_Language {
 			return '';
 		}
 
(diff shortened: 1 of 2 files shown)


manooweb (commented):
Well done ! 
However we removed the title attribute everywhere in WordPress admin - column of post list, settings languages page, languages legacy metabox ... - without taking care of the replacement of the attribute by another kind of tooltip.
We need to decide if we want to keep a tooltip and of course find a accessible solution for this tooltip.

Notice that in Gutenberg, one exist and could be used in the React metabox.

PR #583, opened 2020-10-05
Yoast SEO: Fix indexables

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/681

Since the version 14.0, Yoast SEO often gets the home url directly from the `yoast_indexable table`, thus bypassing the `home_url` filter. We worked around this in #564. 

However, the filter `wpseo_frontend_presentation` has no effect to the breadcrumbs. This PR uses the filter `wpseo_breadcrumb_indexables` to fix this specific case for the homepage and the post type archive link (visible when viewing a single post).

Also the PR reviews the ways the titles and descriptions are translated. Indeed, the previous method attempted to translate the value passed to the filter `wpseo_frontend_presentation` which could already be a translated string instead of the original string, making the translation mechanism inefficient. So we restart from the source string stored in options instead.

Finally, the PR also fixes the author archive link.

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -405,27 +405,33 @@ class PLL_WPSEO {
 	public function frontend_presentation( $presentation ) {
 		if ( is_front_page() ) {
 			$presentation->model->permalink = pll_home_url();
+			$presentation->model->title = WPSEO_Options::get( 'title-home-wpseo' );
+			$presentation->model->description = WPSEO_Options::get( 'title-home-wpseo' );
 		}
 
-		if ( is_post_type_archive() ) {
-			$presentation->model->permalink = get_post_type_archive_link( get_post_type() );
-		}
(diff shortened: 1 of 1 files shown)


integrations/wpseo/wpseo.php
```diff
@@ -404,25 +405,56 @@ public function wpseo_canonical( $url ) {
 	public function frontend_presentation( $presentation ) {
 		if ( is_front_page() ) {
 			$presentation->model->permalink = pll_home_url();
+			$presentation->model->title = WPSEO_Options::get( 'title-home-wpseo' );
+			$presentation->model->description = WPSEO_Options::get( 'title-home-wpseo' );
```
manooweb, line 409: As we see together during your explanations, only testing is_front_page is too restrictive because we make the replacement only for a static front page and not for the blog posts page.

manooweb (changes requested):
Otherwise it's ok 👍

manooweb (approved):

PR #584, opened 2020-10-06
Fix current language possibly assigned when editing untranslatable objects

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/684

Users can activate and deactivate the translation of post types and taxonomies. If a post is assigned a language and then the user deactivates the translation for its post type, the language is still stored in database and `PLL_Translated_Post::get_language()` still returns its language.

This means that the current language is assigned when editing this post: https://github.com/polylang/polylang/blob/2.8.2/admin/admin-base.php#L264

This PR first checks that the post type is translatable before querying the language. This fixes  https://github.com/polylang/polylang-pro/issues/684

Additionnally, the same check is added when editing terms.


manooweb (approved):
OK 👍

PR #586, opened 2020-10-08
Check if page_for_post is not false before assign it

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/656

This function checks if we have a data set in the current language `$this->curlang->page_for_posts` and if so, returns it.

In the case tested for the issue, the function did an `isset` on the property `$this->curlang->page_for_posts` which was indeed set, but was `false`.
_(because the secondary language doesn't have a blog page translation, the tested case was to not translate the blog page on the secondary language)._

So now we check that it also has a value different from false before returning it for `page_for_posts`.

Edit (stack trace debug break point PhpStorm) :

```
static-pages.php:108, PLL_Admin_Static_Pages->translate_page_for_posts()
class-wp-hook.php:289, WP_Hook->apply_filters()
plugin.php:206, apply_filters()
option.php:178, get_option()
static-pages.php:90, PLL_Static_Pages::pll_languages_list()
class-wp-hook.php:287, WP_Hook->apply_filters()
plugin.php:206, apply_filters()
model.php:89, PLL_Model->get_languages_list()
model.php:176, PLL_Model->get_language()
model.php:172, PLL_Model->get_language()
translated-post.php:68, PLL_Translated_Post->get_language()
crud-posts.php:82, PLL_CRUD_Posts->save_post()
class-wp-hook.php:289, WP_Hook->apply_filters()
class-wp-hook.php:311, WP_Hook->do_action()
plugin.php:478, do_action()
post.php:4296, wp_insert_post()
post.php:4390, wp_update_post()
post.php:3213, wp_trash_post()
post.php:260, {main}()
```

### DIFF
--- a/tests/phpunit/tests/test-static-pages.php
+++ b/tests/phpunit/tests/test-static-pages.php
@@ -312,6 +312,42 @@ class Static_Pages_Test extends PLL_UnitTestCase {
 		$this->assertEquals( array( get_post( $en ) ), $GLOBALS['wp_query']->posts );
 	}
 
+	function test_untranslated_page_for_posts_2() {
+		wp_delete_post( self::$posts_fr, true );
+
+		$en = $this->factory->post->create(
+			array(
+				'post_title' => 'english post',
+				'post_type'  => 'post',
+			)
+		);
+		self::$polylang->model->post->set_language( $en, 'en' );
+
+		$page_en = $en = $this->factory->post->create(
+			array(
+				'post_title' => 'page en',
+				'post_type'  => 'page',
+			)
+		);
+		self::$polylang->model->post->set_language( $en, 'en' );
+		$page_fr = $fr = $this->factory->post->create(
+			array(
+				'post_title' => 'page fr',
+				'post_type'  => 'page',
+			)
+		);
+		self::$polylang->model->post->set_language( $fr, 'fr' );
+		self::$polylang->model->post->save_translations( $en, compact( 'en', 'fr' ) );
+
+		wp_delete_post( $page_fr, true );
+
+		self::$polylang->curlang = self::$polylang->model->get_language( 'en' ); // brute force
+		$this->go_to( home_url( '/en/posts/' ) );
+
+		$this->assertEquals( array(), $GLOBALS['wp_query']->posts );
+	}
+
+
 	function test_paged_page_for_posts() {
 		update_option( 'posts_per_page', 2 ); // to avoid creating too much posts
 
--- a/tests/phpunit/tests/test-static-pages.php
+++ b/tests/phpunit/tests/test-static-pages.php
@@ -312,25 +312,6 @@ class Static_Pages_Test extends PLL_UnitTestCase {
 		$this->assertEquals( array( get_post( $en ) ), $GLOBALS['wp_query']->posts );
 	}
 
-	function test_get_option_return() {
-		wp_delete_post( self::$posts_fr, true );
-
-		update_option( 'page_for_posts', self::$posts_en );
-		self::$polylang = new PLL_Frontend( self::$polylang->links_model );
-		self::$polylang->init();
-
-		self::$polylang->curlang = self::$polylang->model->get_language( 'fr' );
-
-		add_filter( 'option_page_for_posts', array( self::$polylang->static_pages, 'translate_page_for_posts' ) );
-
-		$page_for_posts_fr = self::$polylang->model->post->get( get_option( 'page_for_posts' ), 'fr' );
-		$page_for_posts_en = self::$polylang->model->post->get( get_option( 'page_for_posts' ), 'en' );
-
-		$this->assertEquals( self::$posts_en, $page_for_posts_en );
-		$this->assertFalse( $page_for_posts_fr );
-	}
-
-
 	function test_paged_page_for_posts() {
 		update_option( 'posts_per_page', 2 ); // to avoid creating too much posts
 
--- a/tests/phpunit/tests/test-static-pages.php
+++ b/tests/phpunit/tests/test-static-pages.php
@@ -312,38 +312,22 @@ class Static_Pages_Test extends PLL_UnitTestCase {
 		$this->assertEquals( array( get_post( $en ) ), $GLOBALS['wp_query']->posts );
 	}
 
-	function test_untranslated_page_for_posts_2() {
+	function test_get_option_return() {
 		wp_delete_post( self::$posts_fr, true );
 
-		self::$polylang->model->clean_languages_cache();
+		update_option( 'page_for_posts', self::$posts_en );
+		self::$polylang = new PLL_Frontend( self::$polylang->links_model );
+		self::$polylang->init();
 
-		$en = $this->factory->post->create(
-			array(
-				'post_title' => 'english post',
-				'post_type'  => 'post',
-			)
-		);
-		self::$polylang->model->post->set_language( $en, 'en' );
+		self::$polylang->curlang = self::$polylang->model->get_language( 'fr' );
 
-		$page_en = $en = $this->factory->post->create(
-			array(
-				'post_title' => 'page en',
-				'post_type'  => 'page',
-			)
-		);
-		self::$polylang->model->post->set_language( $en, 'en' );
-		$page_fr = $fr = $this->factory->post->create(
-			array(
-				'post_title' => 'page fr',
-				'post_type'  => 'page',
-			)
-		);
-		self::$polylang->model->post->set_language( $fr, 'fr' );
-		self::$polylang->model->post->save_translations( $en, compact( 'en', 'fr' ) );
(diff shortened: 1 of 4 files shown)


tests/phpunit/tests/test-static-pages.php
```diff
@@ -312,6 +312,42 @@ function test_page_for_posts() {
 		$this->assertEquals( array( get_post( $en ) ), $GLOBALS['wp_query']->posts );
 	}
 
+	function test_untranslated_page_for_posts_2() {
+		wp_delete_post( self::$posts_fr, true );
```
raaaahman, line 316: I don't recall that we tested manually what happens when the page for posts translations is deleted. This might be different from what happens when it doesn't exists in the first place.

tests/phpunit/tests/test-static-pages.php
```diff
@@ -312,6 +312,42 @@ function test_page_for_posts() {
...
+		$page_fr = $fr = $this->factory->post->create(
+			array(
+				'post_title' => 'page fr',
+				'post_type'  => 'page',
+			)
+		);
+		self::$polylang->model->post->set_language( $fr, 'fr' );
+		self::$polylang->model->post->save_translations( $en, compact( 'en', 'fr' ) );
+
+		wp_delete_post( $page_fr, true );
```
raaaahman, line 342: The languages are set in the cache as soon as in the [wpSetUpBeforeClass()](https://github.com/polylang/polylang/blob/f58b73b8ac7f5b621c9feb516083b0629668a5e8/tests/phpunit/tests/test-static-pages.php#L27) uses them.

Here, it doesn't seem that the `wp_delete_post()` call impacted the cached English language in the same manner than in production. You might check if some filters / actions are missing or should not have been set.

tests/phpunit/tests/test-static-pages.php
```diff
@@ -312,6 +312,42 @@ function test_page_for_posts() {
...
+
+		self::$polylang->curlang = self::$polylang->model->get_language( 'en' ); // brute force
+		$this->go_to( home_url( '/en/posts/' ) );
+
+		$this->assertEquals( array(), $GLOBALS['wp_query']->posts );
```
raaaahman, line 347: Since you have fixed the issue, shouldn't it return a non empty array?

Marianne380: No, actually, I was testing with the old code (without the fix).

Indeed, since the fix, it should return a array with the English posts.

tests/phpunit/tests/test-static-pages.php
Marianne380: I tested both cases manually, when I didn't translate it and when I translated and then deleted it.

raaaahman:
### Details about the issue

- First the cache is cleaned up on page deletion:
 ```
model.php:141, PLL_Model->clean_languages_cache()
class-wp-hook.php:287, WP_Hook->apply_filters()
class-wp-hook.php:311, WP_Hook->do_action()
plugin.php:478, do_action()
taxonomy.php:3852, _update_post_term_count()
taxonomy.php:3266, wp_update_term_count_now()
taxonomy.php:3238, wp_update_term_count()
post.php:7242, _update_term_count_on_transition_post_status()
class-wp-hook.php:287, WP_Hook->apply_filters()
class-wp-hook.php:311, WP_Hook->do_action()
plugin.php:478, do_action()
post.php:4813, wp_transition_post_status()
post.php:4160, wp_insert_post()
post.php:4354, wp_update_post()
post.php:3209, wp_trash_post()
post.php:260, {main}()
```
- Because of this, when the page is updated, and `pll_save_post` is fired, cache is generated in [PLL_CRUD_Posts::save_post()](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/include/crud-posts.php#L82). This is where **a wrong value is computed** for `$page_for_posts` in the default language, based on the value in the secondary language as shown here (because `$this->curlang` is the language of the deleted page) :
image
- Then when the page for posts is requested again in the default language, because the cache **has been set** during the previous page deletion, it will use this wrong value that have been stored when any call to [PLL_Modell::get_language()](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/include/model.php#L175) is performed.
- If we clean the cache (by saving the languages or URL modifications for example), the next request to the page for posts will instead call the `PLL_Modell:get_languages_list()` to generate the languages again (with correct values).

### Root causes analysis

There are three factor that leads to this issue:

1. Duplicated operation: The [call to PLL_Translated_Post::get() in PLL_Static_Pages::pll_languages_list()](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/include/static-pages.php#L88-L89) should already translate the page on front and page for posts. By calling `get_option()` the values for this pages' ids are filtered through [PLL_Static_Pages::translate_page_for_posts()](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/include/static-pages.php#L104-L107), which is not only superfluous in this context, but also incorrect as it translates the pages on front and pages for posts of **every languages** in the current language, which has eventually lead to this issue.
2. The condition in [PLL_Static_Pages::translate_page_for_posts()](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/include/static-pages.php#L104-L107) is incomplete: which is what this PR fixes. Alas, the condition **has widely been used** throughout the plugin as it is triggered each time `get_option()` is called. A change here might have repercussion on unsuspected parts of the code. Alas again, since the `PLL_Static_Pages` filters are not set in every of our automated tests, we might have a hard time figuring out if it impact other part of the code.
3. The languages cache has been regenerated whereas no change have been made to the languages themselves. This might seem unrelated at first, but it leads the wrong value to be used in other parts of the code instead of being temporary.

### Proposal

For a quick fix of the issue, I would suggest we tackle root cause 1 (see above), as it is the part that has the less repercussions on the rest of the plugin.

Ultimately, I believe we should fix root causes 2 and 3 soon, before they do more harm to the whole plugin, but the solutions here might be more thoroughfully tested.

**Comment by Chouby** [comment]
The issue 1 described by @raaaahman may be due to the filter added in https://github.com/polylang/polylang/blob/2.8.3/admin/admin-static-pages.php#L57

File: tests/phpunit/tests/test-translate-page-for-posts.php [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+	public function test_translate_page_for_posts_on_default_language() {
+		self::$polylang->curlang = self::$polylang->model->get_language( 'en' );
+
+		$return = self::$polylang->static_pages->translate_page_for_posts( get_option( 'page_for_posts' ) );
+
+		$this->assertEquals(self::$posts_en, $return );
+	}
+
+	public function test_translate_page_for_posts_on_secondary_language() {
+		self::$polylang->curlang = self::$polylang->model->get_language( 'fr' );
```
raaaahman, line 56: I have this test failing in my environment.

In fact, all of your tests use the languages values that are stored in the cache during the [update_option( 'page_for_posts' )](https://github.com/polylang/polylang/blob/eb22bfdc585b9ee680e610b0715c12da77b3da8b/tests/phpunit/tests/test-translate-page-for-posts.php#L40) call, which happens before the [`PLL_Frontend_Static_Pages` is instantiated](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/frontend/frontend.php#L68), so the `PLL_Model::get_languages_list()` call is not filtered by [PLL_Static_Pages::pll_languages_list()](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/include/static-pages.php#L85-L94).

tests/phpunit/tests/test-translate-page-for-posts.php
```diff
@@ -0,0 +1,72 @@
...
+
+		$return = self::$polylang->static_pages->translate_page_for_posts( get_option( 'page_for_posts' ) );
+
+		$this->assertEquals(self::$posts_fr, $return );
+
+	}
+
+	public function test_translate_page_for_posts_when_page_for_posts_has_no_translations() {
+		wp_delete_post( self::$posts_fr, true );
```
raaaahman, line 65: Since you only use the `self::$post_fr` value in `Translate_Page_For_Posts::test_translate_page_for_posts_on_default_language()`, maybe you should create the page locally inside the mentioned test, instead of creating it globally and removing it afterward. (Tests should not change their results because of the order they are ran in).

raaaahman (changes requested):
The test are way easier to read this, and you are covering more usages. Their names are very explicit, and that's what we need, good job.

I only need that you fix the test that is failing, and update the `wpSetUpBeforeClass()` method. The rest is up to you to change it or keep it.

Chouby:
Looks good. Broken tests are not related to this PR.

PR #587, opened 2020-10-09
Improve how WP functions are checked

Chouby:
This PR adds two improvements for our phpunit tests:
1. It moves the methods to check if WP functions have been modified to a trait, to allow re-usability (with Polylang Pro in mind).
2. It allows to check functions during the WP development without including `alpha`, `beta` in the version passed to `check_method()`. 
For example, it allows to call:
```php
$this->check_method( '824f95fb45ff39f0a42556f93c259d3d', '5.6', 'get_comment_delimited_block_content' );
``` 
if the function has been modified during the development of WP 5.6 while we must currently use:
```php
$this->check_method( '824f95fb45ff39f0a42556f93c259d3d', '5.6-alpha', 'get_comment_delimited_block_content' );`
```


manooweb (approved):
ok 👍

PR #588, opened 2020-10-09
VIPWPCS 2.2

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/651 for Polylang.

The goal of this PR is to follow the move of VIPWPCS 2.2 which replaces `WordPressVIPMinimum.Variables.VariableAnalysis.UnusedVariable` by `VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable` and `WordPressVIPMinimum.Variables.VariableAnalysis.UndefinedVariable` by `VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable`.

Thus it restores the latest version of VIPWPCS in composer.json.
In the config file, it excludes the views and loaders for the undefined variables checks and accepts unused variables before require, include, etc... (NB: this property doesn't seem to work in all cases).

The updated sniff detects more cases of unused variables that are fixed in this PR, mainly in function definitions, but some are ignored, mainly to keep the same method signature between parent and child classes. Cases that were previously ignored with the old sniff have been updated.

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -25,7 +25,7 @@ class PLL_Admin_Classic_Editor {
 		$this->pref_lang = &$polylang->pref_lang;
 
 		// Adds the Languages box in the 'Edit Post' and 'Edit Page' panels
-		add_action( 'add_meta_boxes', array( $this, 'add_meta_boxes' ), 10, 2 );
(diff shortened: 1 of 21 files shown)


manooweb (approved):
ok 👍

PR #592, opened 2020-10-15
Fix tests for canonical redirect

raaaahman:
## Before

The `PLL_Canonical_Test::test_category()` throwed false positives regarding to the following issue: https://github.com/polylang/polylang-pro/issues/693

## Changes 

- All tests are now executed in a PLL_Frontend context
- Fix to avoid `PLL_Choose_Lang_Url::set_language_from_url()` believing we're calling the homepage
- Refresh Polylang between each assertion: Because [PLL_Choose_Lang_Url::init() checks if 'pll_language_defined' has already been triggered](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/frontend/choose-lang-url.php#L24), we need to refresh the hooks between each assertion, so there's only one assertion per test.

To enhance tests speed, WordPress chose to [set the fixtures (post, taxonomies, etc.) only once in the wpSetUpBeforeClass() method](https://github.com/WordPress/wordpress-develop/blob/701dee26f19b4bfdcbc1cc636f5fdecd8a753432/tests/phpunit/includes/testcase-canonical.php#L18-L20). These has two drawbacks:

- Because all data set during the `wpSetUpBeforeClass()` is backed up and shared, it needed to [mirror the `generate_shared_fixtures()` method with a `delete_shared_fixtures()`](https://github.com/WordPress/wordpress-develop/blob/701dee26f19b4bfdcbc1cc636f5fdecd8a753432/tests/phpunit/includes/testcase-canonical.php#L22-L24). This introduce a liability when adding new tests, as the developer should not forgot to erase the new fixtures he introduces for these tests.
- The test data is absent from the [test cases where it is tested](https://github.com/WordPress/wordpress-develop/blob/701dee26f19b4bfdcbc1cc636f5fdecd8a753432/tests/phpunit/tests/canonical.php#L21-L33), which makes tests less easy to read.

The solution in this PR then sacrifices performance in favor of readability. Although an optimization could be added in the future.

## Going further

- Optimize performance by generating the database fixtures only once
- As discussed during October 15, 2020 tech meeting, the modification to the test setup could (and should) be used in every test relying on `WP_UnitTest_Case::go_to()` method.

### DIFF
--- a/tests/phpunit/includes/testcase-canonical.php
+++ b/tests/phpunit/includes/testcase-canonical.php
@@ -38,22 +38,21 @@ class PLL_Canonical_UnitTestCase extends WP_Canonical_UnitTestCase {
 		// Needed by {@see pll_requested_url()}.
 		$_SERVER['REQUEST_URI'] = $test_url;
 
-		$model             = new PLL_Model( $this->options );
-		$links_model       = new PLL_Links_Directory( $model );
-		self::$polylang    = new PLL_Frontend( $links_model );
-		self::$polylang->init();
+		$model = new PLL_Model( $this->options );
 
 		// switch to pretty permalinks
 		$wp_rewrite->init();
 		$wp_rewrite->set_permalink_structure( $this->structure );
 
 		// register post types and taxonomies
-		self::$polylang->model->post->register_taxonomy(); // needs this for 'lang' query var
+		$model->post->register_taxonomy(); // needs this for 'lang' query var
 		create_initial_taxonomies();
 
 		// reset the links model according to the permalink structure
-		self::$polylang->links_model = self::$polylang->model->get_links_model();
-		self::$polylang->links_model->init();
+		$links_model    = $model->get_links_model();
+		self::$polylang = new PLL_Frontend( $links_model );
+		self::$polylang->init();
+		do_action_ref_array( 'pll_init', array( self::$polylang ) );
 
 		// flush rules
 		$wp_rewrite->extra_rules_top = array(); // brute force since WP does not do it :(
--- a/tests/phpunit/includes/testcase-canonical.php
+++ b/tests/phpunit/includes/testcase-canonical.php
@@ -35,6 +35,10 @@ class PLL_Canonical_UnitTestCase extends WP_Canonical_UnitTestCase {
 	public function assertCanonical( $test_url, $expected, $ticket = 0, $expected_doing_it_wrong = array() ) {
 		global $wp_rewrite;
 
+		if ( did_action( 'pll_language_defined' ) ) {
+			$this->fail( 'Canonical tests MUST have only one call to PLL_UnitTestCase_Canonical::assertCanonical() per test.' );
+		}
+
 		// Needed by {@see pll_requested_url()}.
 		$_SERVER['REQUEST_URI'] = $test_url;
 
--- a/tests/phpunit/includes/testcase-canonical.php
+++ b/tests/phpunit/includes/testcase-canonical.php
@@ -11,6 +11,55 @@ class PLL_Canonical_UnitTestCase extends WP_Canonical_UnitTestCase {
 	}
 
 	/**
+	 * Set up the Polylang environment before testing canonical redirects.
+	 *
+	 * @param string $test_url
+	 * @param string $expected
+	 * @param int    $ticket
+	 * @param array  $expected_doing_it_wrong
+	 */
+	public function assertCanonical( $test_url, $expected, $ticket = 0, $expected_doing_it_wrong = array() ) {
+		global $wp_rewrite;
+
+		// Needed by {@see pll_requested_url()}.
+		$_SERVER['REQUEST_URI'] = $test_url;
+
+		$options = array_merge(
+			PLL_Install::get_default_options(),
+			array(
+				'default_lang' => 'en',
+				'hide_default' => 0,
+				'post_types' => array(
+					'cpt' => 'cpt', // translate the cpt // FIXME /!\ 'after_setup_theme' already fired and the list of translated post types is already cached :(
+				),
+			)
+		);
+		$model = new PLL_Model( $options );
+		$links_model = new PLL_Links_Directory( $model );
+		self::$polylang = new PLL_Frontend( $links_model );
+		self::$polylang->init();
+
+		// switch to pretty permalinks
+		$wp_rewrite->init();
+		$wp_rewrite->set_permalink_structure( $this->structure );
+
+		// register post types and taxonomies
+		self::$polylang->model->post->register_taxonomy(); // needs this for 'lang' query var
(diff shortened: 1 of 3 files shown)


Chouby (changes requested):
Here are some suggestions:
1. Move the code from `_test_canonical_redirect()` to `assertCanonical()` in the test case, making `assertCanonical()` call this code and not the opposite, but leave the definition of options in the tests `setup()` .
2. Create `$links_model` from `PLL_Model::get_links_model()`
3. In all the tests, call the modified `assertCanonical()` instead of `_test_canonical_redirect()`

That way I expect that your fix can profit to all tests extending `PLL_Canonical_UnitTestCase`. But this probably requires to modify them including in Polylang Pro to use data providers as you did here.

4. Maybe check `did_action( 'pll_language_defined' )` in the `assertCanonical()` and send an error to avoid calling the assertion 2 times in a test.
5. Re-introduce the removed tests. I prefer to keep them ;-)
6. Fix the PHPCS errors: https://travis-ci.org/github/polylang/polylang/jobs/735980202

tests/phpunit/includes/testcase-canonical.php
```diff
@@ -1,54 +1,86 @@
...
+		create_initial_taxonomies();
 
-		$query_vars = array_diff( $wp->query_vars, $wp->extra_query_vars );
-		if ( ! empty( $parsed_can_url['query'] ) ) {
-			parse_str( $parsed_can_url['query'], $_qv );
+		// reset the links model according to the permalink structure
+		$links_model    = $model->get_links_model();
+		self::$polylang = new PLL_Frontend( $links_model );
+		self::$polylang->init();
+		do_action_ref_array( 'pll_init', array( self::$polylang ) );
```
raaaahman, line 59: This is looking more and more like `Polylang::init()`. What would be the issues if I wanted to replace the manual intialization by a call to `Polylang::init()`?

tests/phpunit/includes/testcase-canonical.php
```diff
@@ -1,54 +1,86 @@
...
-
-	// mainly copy paste from WP_Canonical_UnitTestCase::assertCanonical
-	public function assertCanonical( $test_url, $expected ) {
-		if ( is_string( $expected ) ) {
-			$expected = array( 'url' => $expected );
-		} elseif ( is_array( $expected ) && ! isset( $expected['url'] ) && ! isset( $expected['qv'] ) ) {
-			$expected = array( 'qv' => $expected );
-		}
+class PLL_Canonical_UnitTestCase extends WP_Canonical_UnitTestCase {
+	use PLL_UnitTestCase_Trait;
```
Chouby, line 4: You can't do this., as this is causing this error:
`PHP Warning:  Declaration of PLL_UnitTestCase_Trait::wpSetUpBeforeClass() should be compatible with WP_Canonical_UnitTestCase::wpSetUpBeforeClass($factory) in /home/travis/build/polylang/polylang/tests/phpunit/includes/testcase-canonical.php on line 17`

Chouby (changes requested):

tests/phpunit/includes/testcase-canonical.php
raaaahman: I think we should adapt our `PLL_UnitTestCase_Trait`, as all of our test cases will receive the `$factory` parameter: https://github.com/WordPress/wordpress-develop/blob/86fc3af215a61f042804e3feb57627976d2bb54d/tests/phpunit/includes/abstract-testcase.php#L76

raaaahman:
This PR now needs #595 to be merged. It may raise conflicts.

raaaahman:
Reverted 8633d41 so this could be merged with #596

tests/phpunit/includes/testcase-canonical.php
raaaahman: I've faced some technical difficulties here:

-  We can't statically call `PLL_UnitTestCase_Trait::wpSetUpBeforeClass()` directly, as the `self` keyword would not resolve to `PLL_Canonical_UnitTestCase`, so all the needed properties will be set to `null`.
-  It is not possible to use [PHP traits' conflict resolution mechanism](https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.traits.conflict) to override a method that doesn't have the same parameters.
- We can't use a `WP_Canonical_UnitTestCase` instance to delegate calls to `assertCanonical( $test_url, $expected_url )` as it would not let us override the `WP_Canonical_UnitTestCase::getCanonical( $test_url )` as [it is done here](https://github.com/polylang/polylang/blob/2565189a88c65ff30cb70890d50821148526be1b/tests/phpunit/includes/testcase-canonical.php#L77-L85)

This leaves us with the following solutions:

1. Adapting `PLL_UnitTestCase_Trait::wpSetUpBeforeClass()` to be compatible with `WP_Canonical_UnitTestCase::wpSetUpBeforeClass( $factory )` would force us to also adapt every classes that uses this trait, in `polylang`, `polylang-pro` and `polylang-woocommerce` as well.
2. Declaring a `PLL_Canonical_UnitTestCase::wpSetUpBeforeClass( $factory )` method to override `PLL_UnitTestCase::wpSetUpBeforeClass()` according to [PHP traits' precedence rules](https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.traits.precedence) would force us to duplicate the code of `PLL_UnitTestCase_Trait::wpSetUpBeforeClass()` inside the new method.
3. Do not extend the `WP_Canonical_UnitTestCase` class and duplicate its `assertCanonical( $test_url, $expected_url )` method [as it was done before this PR](https://github.com/polylang/polylang/blob/0c9fd403caa735875c2a5dad077948f8a03f1103/tests/phpunit/includes/testcase-canonical.php#L6-L53).

i would rather not to resort to duplication whenever possible.

tests/phpunit/includes/testcase-canonical.php
```diff
@@ -1,54 +1,86 @@
...
-			$this->assertEquals( $expected['url'], $parsed_can_url['path'] . ( ! empty( $parsed_can_url['query'] ) ? '?' . $parsed_can_url['query'] : '' ) );
-		}
+		$this->options = array_merge(
+			PLL_Install::get_default_options(),
+			array(
+				'default_lang' => 'en',
+				'hide_default' => 0,
+				'post_types'   => array(
+					'cpt' => 'pllcanonical',
+					// translate the cpt // FIXME /!\ 'after_setup_theme' already fired and the list of translated post types is already cached :(
```
raaaahman, line 21: Does this `FIXME` still apply?

Chouby: Yes probably. The list of translated post types is cached here https://github.com/polylang/polylang/blob/2.8.3/include/model.php#L241-L243

PR #595, opened 2020-10-20
Use WP_TestCase_Factory passed as parameter

raaaahman:
## Before

The [WP_UnitTest_Factory](https://github.com/WordPress/wordpress-develop/blob/master/tests/phpunit/includes/factory/class-wp-unittest-factory.php) used to create WordPress related fixtures in the database was retrieved from a call to `WP_UnitTestCase_Base::factory()` method call.

## Changes

- Pass the  `WP_UnitTest_Factory` as a parameter of `PLL_UnitTestCase_Trait::wpSetUpBeforeClass()` to be compatible with [the way it is called in `WP_UnitTestCase::setUpBeforeClass()`](https://github.com/WordPress/wordpress-develop/blob/123a965c70acee04b1cb7ab58599fb6284999835/tests/phpunit/includes/abstract-testcase.php#L76)
- Use the `WP_UnitTest_Factory` passed as parameter in tests

## Details

Calling `PLL_UnitTestCase_Trait::wpSetUpBeforeClass()` raised PHP warnings in #592 CI jobs.

### DIFF
--- a/tests/phpunit/includes/testcase-domain.php
+++ b/tests/phpunit/includes/testcase-domain.php
@@ -5,8 +5,11 @@ class PLL_Domain_UnitTestCase extends PLL_UnitTestCase {
 	protected $hosts;
 	protected static $server;
 
-	static function wpSetUpBeforeClass() {
-		parent::wpSetUpBeforeClass();
+	/**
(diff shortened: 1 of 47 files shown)


Chouby:
Rebase and force push to solve a conflict.

PR #597, opened 2020-10-21
Fix canonical redirect for taxonomy terms

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/667
Fixes https://github.com/polylang/polylang-pro/issues/693

Additionnally, the PR also adds more tests for plain permalinks which must be redirected to the canonical redirect and fixes the tests for the home page which were duplicating the tests for a standard page since #596.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -383,10 +383,14 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 			}
 		}
 
-		elseif ( is_category() || is_tag() || is_tax() ) {
-			$obj = $wp_query->get_queried_object();
-			if ( ! empty( $obj ) && $this->model->is_translated_taxonomy( $obj->taxonomy ) ) {
-				$language = $this->model->term->get_language( (int) $obj->term_id );
+		elseif ( is_category() || is_tag() || is_tax() || ( is_404() && ! empty( $wp_query->tax_query ) ) ) {
+			$queried_terms = $wp_query->tax_query->queried_terms;
+			unset( $queried_terms['language'] );
+			$taxonomy = key( $queried_terms );
+			if ( $this->model->is_translated_taxonomy( $taxonomy ) ) {
+				$term_id = $this->get_queried_term_id( $taxonomy, $queried_terms[ $taxonomy ] );
+				$language = $this->model->term->get_language( $term_id );
+				$redirect_url = get_term_link( $term_id );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


frontend/frontend-filters-links.php
```diff
@@ -445,4 +449,38 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
+	 * @return int
+	 */
+	protected function get_queried_term_id( $taxonomy, $queried_term ) {
+		$field = $queried_term['field'];
+		$term  = reset( $queried_term['terms'] );
+
+		// We can get a term_id when requesting a plain permalink, eg /?cat=1.
+		if ( 'term_id' === $field ) {
+			return $term;
+		}
```
raaaahman, line 473: You shouldn't need this if you follow the suggestions from the previous comment.

frontend/frontend-filters-links.php
```diff
@@ -445,4 +449,38 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
+
+		// We get a slug when requeting a pretty peramlin with the wrong language.
```
raaaahman, line 475: Some typos here.

tests/phpunit/tests/test-canonical.php
```diff
@@ -149,6 +157,10 @@ public function test_category_without_language() {
 		$this->assertCanonical( '/category/parent/', '/en/category/parent/' );
 	}
 
+	public function test_category_from_plain_permalink() {
+		$this->assertCanonical( '?cat=' . self::$term_en, '/en/category/parent/' );
+	}
+
```
raaaahman, line 163: This test seems relevant.

frontend/frontend-filters-links.php
```diff
@@ -383,10 +383,14 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
-			if ( ! empty( $obj ) && $this->model->is_translated_taxonomy( $obj->taxonomy ) ) {
-				$language = $this->model->term->get_language( (int) $obj->term_id );
+		elseif ( is_category() || is_tag() || is_tax() || ( is_404() && ! empty( $wp_query->tax_query ) ) ) {
+			$queried_terms = $wp_query->tax_query->queried_terms;
+			unset( $queried_terms['language'] );
+			$taxonomy = key( $queried_terms );
+			if ( $this->model->is_translated_taxonomy( $taxonomy ) ) {
+				$term_id = $this->get_queried_term_id( $taxonomy, $queried_terms[ $taxonomy ] );
+				$language = $this->model->term->get_language( $term_id );
+				$redirect_url = get_term_link( $term_id );
```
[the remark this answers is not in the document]
Chouby: I don't want to be so restrictive to create a `is_plain_permalink()`. We indeed detected the bug in this case. But WP iis fact set to use pretty permalinks (so we can't rely on this option) and it's perfectly possible to mix things such as `/en/?cat=1`.

frontend/frontend-filters-links.php
```diff
@@ -469,6 +469,11 @@ protected function get_queried_term_id( $tax_query ) {
 			return $term;
 		}
 
+		// Avoid a useless DB request if the taxonomy is not translatable.
+		if ( ! $this->model->is_translated_taxonomy( $taxonomy ) ) {
+			return $term;
+		}
```
raaaahman, line 475: You are changing the conditional logic from the previous version:

```php
$obj = $wp_query->get_queried_object();
if ( ! empty( $obj ) && $this->model->is_translated_taxonomy( $taxonomy ) ) {
    $language = $this->model->term->get_language( $obj->term_id );
}
```

The previous version would NOT call to `PLL_Translated_Term::get_language()` if the taxonomy is not translatable.

Here it would call to `PLL_Translated_Term::get_language()` with a slug parameter instead.

Chouby: Yes I know. But it has no impact. The final result should be the same.

frontend/frontend-filters-links.php
```diff
@@ -383,10 +383,10 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
 			}
 		}
 
-		elseif ( is_category() || is_tag() || is_tax() ) {
-			$obj = $wp_query->get_queried_object();
-			if ( ! empty( $obj ) && $this->model->is_translated_taxonomy( $obj->taxonomy ) ) {
-				$language = $this->model->term->get_language( (int) $obj->term_id );
+		elseif ( is_category() || is_tag() || is_tax() || ( is_404() && ! empty( $wp_query->tax_query ) ) ) {
```
manooweb, line 386: to make a explanation in a new PR

PR #600, opened 2020-10-22
Fix test_upload

Chouby:
The test used to compare 2 languages objects, one evaluated before the media upload, the other evaluated after the media upload. This is wrong as the 2 different objects must differ by the count. In this PR, I now compare the language codes instead.

Note: the test used to pass in WP < 5.6 (with a term count set to 0, for both languages). The different behavior may have been introduced by https://core.trac.wordpress.org/ticket/40351


raaaahman (approved):
Is the terms count an information useful to have in our `PLL_Language` instances though?

Chouby:
For posts, we display it in languages table. 
And we use it internally to clean unused translations groups: https://github.com/polylang/polylang/blob/2.8.3/include/translated-object.php#L145-L147

PR #602, opened 2020-10-23
Fix breadcrumb for untranslatable post type

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/696

Our compatibility code for Yoast indexables did not check if the post type was translatable before acting. This should break only the call to`pll__()`, but I added the test for all the modifications made as they don't make sense when the post type is not translatable.


manooweb (approved):
Ok 👍

sebastienserre:
This is fixing the bug

sebastienserre (approved):

PR #603, opened 2020-10-23
Check that the language is active

Marianne380:
Check that the language is active before returning it as the preferred language (to avoid 404).

Fix https://github.com/polylang/polylang-pro/issues/697


Marianne380:
This PR used Polylang Pro code in Polylang, see now : https://github.com/polylang/polylang-pro/pull/707

PR #604, opened 2020-10-23
Fix broken coverage for page for posts

raaaahman:
## Before

#592 introduced an error in tests which made this condition not tested anymore:

https://github.com/polylang/polylang/blob/c12f911078a5dc6b006e9f685b8aa64e3f4afeee/frontend/frontend-filters-links.php#L393-L396

## Changes

- Modifies some tests to correctly cover this conditional branch
- Explicit the conditions of the static front page tests 

Partially fixes https://github.com/polylang/polylang-pro/issues/701 Although some conditional branches have never been covered.


raaaahman:
Replaced by #633

PR #607, opened 2020-10-28
Fix deprecated function usage in WP 5.6

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/710

This removes the call to `_get_path_to_translation()` in WP 5.6+ as it sends a deprecated notice and is not useful anymore.


manooweb (commented):
ok 👍
However I can't reproduce the bug the line was fixing for WP < 5.6

manooweb (approved):

Chouby:
@Chrystll reproduced it with WP 5.5. I don't with WP 5.6. Hopefully I'm not doing wrong.

PR #608, opened 2020-10-28
Allow to translate 0 in string translations

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/712

This PR allows explicitely to register any scalar as string for translation, except empty strings.


manooweb (approved):
ok 👍

Chouby:
Not sure it is ok. I'll investigate the issue reported by @Chrystll with empty widget titles. Thus, I will not include this in 2.8.4.

PR #610, opened 2020-10-29
Fix the sticky posts cache never cleaned

### LINKED ISSUE #601: Polylang doesn't flush object cache for sticky posts
## Prerequisites

1. Setup WordPress and install and configure Polylang. 
2. Enable Memcache on your site.
3. Make a published post "sticky".
4. Validate that site is cached.
5. "Unsticky" the post.
6. Make a different published post "sticky".
7. The first published post appears sticky and the second does not.
8. Scratch your head in confusion. 
9. Disable Polylang and the correct sticky post appears.

The expected behavior is that the new sticky post will be the post that appears, and not the previously stickied post. 

I am using Polylang on a site hosted by WordPress VIP Go. Here is a description of how their object cache works. 

https://wpvip.com/documentation/vip-go/caching-on-vip-go/#object-cache

In working with one of the support engineers, they described the issue as thus, 

"It looks like there is a caching issue here, but in particular with the object cache (Memcached).

The sticky posts are saved in a WordPress option. When this is looked up, Polylang intercepts.
Polylang then does its own lookup to find the corresponding posts for other languages, and it caches those results separately from the original option.

Here's where this is happening:
https://github.com/wpcomvip/wikimedia-blog-wikimedia-org/blob/develop/plugins/polylang/frontend/frontend-filters.php#L85

The problem is that the cache that Polylang set doesn't have any built-in expiration or updates. So regardless of the value of the sticky posts option in the database, it is replaced with an unchanging cached value. (It might expire eventually and be refreshed, but there's no telling when. )

Polylang likely needs to add a bit of code to flush their sticky posts cache when sticky posts are changed."

For a work-around solution, site admins can clear the sticky_posts cache on command using wp-cli.

`wp cache delete sticky_posts options`

I'm enjoying using Polylang and VIP together. While I have a work-around I can use for now, I wanted to make you aware of  the issue.

Chouby:
Fixes #601 

This PR deletes our cache for the sticky posts when the option `sticky_posts` is added, updated or deleted.

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -24,6 +24,11 @@ class PLL_Filters {
 		$this->options = &$polylang->options;
 		$this->curlang = &$polylang->curlang;
 
+		// Deletes our cache for sticky posts when the list is updated.
+		add_action( 'update_option_sticky_posts', array( $this, 'delete_sticky_posts_cache' ) );
+		add_action( 'add_option_sticky_posts', array( $this, 'delete_sticky_posts_cache' ) );
+		add_action( 'delete_option_sticky_posts', array( $this, 'delete_sticky_posts_cache' ) );
+
 		// Filters the comments according to the current language
 		add_action( 'parse_comment_query', array( $this, 'parse_comment_query' ) );
 		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


manooweb (approved):
ok 👍

Issue #582, opened 2020-10-02
Default XML sitemap does not work without pll_language cookie


chesio:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest Polylang version.
- [x] The issue still exists against the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] This bug happens with a default WordPress theme active
- [x] I can reproduce this bug consistently

## Steps to reproduce the issue

0. Add `define('PLL_COOKIE', false);` to `wp-config.php`
1. Install Polylang 2.8.2 on WordPress 5.5.1, configure two languages, add some posts in both languages.
2. Visit http://www.example.com/wp-sitemap.xml
3. Click on posts sitemap in one language -> OK
4. Click on posts sitemap in second language -> observe that it contains posts in first language.

## Expected behavior and actual behavior

When I follow those steps, I see that posts sitemap always contains posts in one language.

I was expecting posts in other language(s).

## Observation

In the `wp-sitemap.xml`, all sitemaps have language slug in URL - for example:
* http://www.example.com/de/wp-sitemap-posts-post-1.xml
* http://www.example.com/en/wp-sitemap-posts-post-1.xml
* etc. for other types

When I go to http://www.example.com/de/wp-sitemap-posts-post-1.xml or http://www.example.com/en/wp-sitemap-posts-post-1.xml, I am always being redirected to http://www.example.com/wp-sitemap-posts-post-1.xml (= URL without language slug).

1. If cookies are not disabled, then http://www.example.com/wp-sitemap-posts-post-1.xml lists posts according to language saved in cookie.
2. If cookies are disabled, then http://www.example.com/wp-sitemap-posts-post-1.xml lists posts in the same language.

I would expect http://www.example.com/de/wp-sitemap-posts-post-1.xml to list posts in German and http://www.example.com/en/wp-sitemap-posts-post-1.xml to list posts in English.

## Note

I've investigated the issue with Polylang, but I have also installation with Polylang Pro with the same problem.

chesio:
Note: I previously assumed posts are always being displayed in default language, but this is not the case. They are always in the same language, but it's not always the default one.

Chouby:
[...]

PR #611, opened 2020-10-30
Fix sitemaps always redirected to the default language

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/703

* In the multilingual sitemap provider, add the current language slug in the sitemap url when it is not called from the index.
* Add phpunit tests to test the canonical sitemap url.
* For multiple domains or subdmomains, make sure that sitemap urls are always evalauted on the current domain.


chesio:
Does this (also) fix the issue described in #582 ?

Chouby:
@chesio I suppose it will. The multilingual sitemaps are not supposed to be dependent on the cookie. We are annoyed by the changes intoduced in WP 5.5.1 by this ticket https://core.trac.wordpress.org/ticket/50910.

PR #614, opened 2020-11-03
post.js code modernization

Chouby:
Closes https://github.com/polylang/polylang-pro/issues/733

### DIFF
--- a/js/post.js
+++ b/js/post.js
@@ -19,7 +19,7 @@
  * Quick edit
  */
 (function( $ ) {
-	$( document ).bind(
+	$( document ).on(
 		'DOMNodeInserted',
 		function( e ) {
 			var t = $( e.target );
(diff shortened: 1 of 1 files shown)


js/post.js
```diff
@@ -19,7 +19,7 @@
  * Quick edit
  */
 (function( $ ) {
-	$( document ).bind(
+	$( document ).on(
 		'DOMNodeInserted',
 		function( e ) {
```
manooweb, line 24: Ok for the bind jQuery shorthand however it seems that `DOMNodeInserted` mutation events is deprecated too
See https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events
and https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

Chouby:
Also replaces (function($) { /* ... */ })(jQuery); by jQuery(function( $ ) { /* ... */ });.

manooweb (approved):
Ok 👍
For `DOMNodeInserted` mutation events deprecation, we can wait because it works fine with WP 5.6 however we need to keep in mind somewhere not to forget it

PR #616, opened 2020-11-03
Fix jQuery.fn.change() event shorthand is deprecated

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/725


manooweb (approved):
Ok 👍

manooweb (approved):
Ok. latest modification replacement `jQuery.Ready()` syntax by `jQuery( function($) )` one.

PR #618, opened 2020-11-03
Externalize javascript for the language switcher widget form

Chouby:
Closes https://github.com/polylang/polylang-pro/issues/723
It looks like the reason for keeping the javascript associated to the language switcher widget form is not valid anymore.

This PR proposes to move this javascript to widget.js.


manooweb (approved):
ok 👍

PR #619, opened 2020-11-03
Add navigation markup to the language switcher widget

Chouby:
Lets follow the move initiated by WP 5.5. See https://make.wordpress.org/core/2020/07/09/accessibility-improvements-to-widgets-outputting-lists-of-links-in-5-5/
Closes https://github.com/polylang/polylang-pro/issues/590


manooweb (approved):
Ok 👍

PR #621, opened 2020-11-04
fix(admin-base): jQuery.parseJSON is deprecated :use JSON.parse

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/717


manooweb (approved):
OK 👍

PR #625, opened 2020-11-05
Fix jQuery.fn.change() event shorthand is deprecated

Marianne380:
Fixes the omission in the previous PR : https://github.com/polylang/polylang/pull/617
See https://github.com/polylang/polylang-pro/issues/726


manooweb (approved):
Ok 👍

PR #626, opened 2020-11-05
Don't attempt to modify rewrite rules when there is no language

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/744

In this PR, we make sure that the logic to modify the sitemap rewrite rules is not executed when there is no language. 
Indeed although the module is not loaded when there is no languages, this protection is not sufficient when deleting a language and there is none remaining.

### DIFF
--- a/modules/sitemaps/sitemaps.php
+++ b/modules/sitemaps/sitemaps.php
@@ -118,16 +118,19 @@ class PLL_Sitemaps {
 	public function rewrite_rules( $rules ) {
 		global $wp_rewrite;
 
-		$newrules = array();
-
 		$languages = $this->model->get_languages_list( array( 'fields' => 'slug' ) );
+
+		if ( empty( $languages ) ) {
+			return $rules;
+		}
+
 		if ( $this->options['hide_default'] ) {
 			$languages = array_diff( $languages, array( $this->options['default_lang'] ) );
 		}
 
-		if ( ! empty( $languages ) ) {
-			$slug = $wp_rewrite->root . ( $this->options['rewrite'] ? '^' : '^language/' ) . '(' . implode( '|', $languages ) . ')/';
-		}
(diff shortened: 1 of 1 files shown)


raaaahman:
- Don't we need to flush the rewrite rules when the last language is deleted (because we don't want localized sitemaps anymore)?
- Are the rewrite rules for sitemaps cached/stored?
- Do we need the `PLL_Sitemaps` class at all when the last language is deleted?

Chouby:
> * Don't we need to flush the rewrite rules when the last language is deleted (because we don't want localized sitemaps anymore)?
Yes and it is done in https://github.com/polylang/polylang/blob/2.8.4/admin/admin-model.php#L160
> * Are the rewrite rules for sitemaps cached/stored?
All rewrite rules are stored in the same WP option 'rewrite_rules'. We don't have tocare about that. We jus act on filtered applied when they are evaluated.
> * Do we need the `PLL_Sitemaps` class at all when the last language is deleted?
Not but it is there and you can't relaay change that. The steps are as follow:
1. `PLL_Sitemaps` is instantiated because we have one language.
2. The user deletes the language so we start processing the action:
  1. Delete the language.
  2. Flush the language cache.
  3. Flush the rewrite rules : at this point the filter in `PLL_Sitemaps` is called even if we don't need it.

raaaahman:
> Not but it is there and you can't really change that. The steps are as follow:
> 
>     1. `PLL_Sitemaps` is instantiated because we have one language.
> 
>     2. The user deletes the language so we start processing the action:
>        
>        1. Delete the language.
>        2. Flush the language cache.
>        3. Flush the rewrite rules : at this point the filter in `PLL_Sitemaps` is called even if we don't need it.

Wouldn't it be simpler to just remove the troublesome filter?

raaaahman (approved):

PR #633, opened 2020-11-09
Fix broken coverage for page for posts

raaaahman:
## Before

#592 introduced an error in tests which made this condition not tested anymore:

https://github.com/polylang/polylang/blob/c12f911078a5dc6b006e9f685b8aa64e3f4afeee/frontend/frontend-filters-links.php#L393-L396

## Changes

- Modifies some tests to correctly cover this conditional branch
- Instantiates the `PLL_Admin_Pages` module only in the tests that need it.

Partially fixes https://github.com/polylang/polylang-pro/issues/701 Although some conditional branches have never been covered. 

Replaces #604 (rebase)

### DIFF
--- a/tests/phpunit/tests/test-canonical.php
+++ b/tests/phpunit/tests/test-canonical.php
@@ -66,14 +66,9 @@ class Canonical_Test extends PLL_Canonical_UnitTestCase {
 
 		self::$polylang->model->post->save_translations( self::$page_for_posts_en, compact( 'en', 'fr' ) );
 
-		update_option( 'page_for_posts', self::$page_for_posts_fr );
-
 		self::$page_on_front_en = $factory->post->create( array( 'post_type' => 'page', 'post_title' => 'home' ) );
 
 		self::$polylang->model->post->set_language( self::$page_on_front_en, 'en' );
-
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/test-canonical.php
```diff
@@ -118,6 +109,8 @@ public function init_for_sitemaps() {
 
 	public static function wpTearDownAfterClass() {
 		_unregister_post_type( 'pllcanonical' );
+		_unregister_post_type( 'trcpt' );
+		_unregister_post_type( 'cpt' );
```
Marianne380, line 113: Do we need to unregistered taxonomies as well, because I've added some in the fixtures as well?

Marianne380 (approved):

tests/phpunit/tests/test-canonical.php
raaaahman: Good idea.

raaaahman: Although, it might already be the case, see the doc comment for this function: https://github.com/WordPress/wordpress-develop/blob/97b2f07d2eba6cdd88bb960ba539d4c07f84f71f/tests/phpunit/includes/utils.php#L377-L382

raaaahman:
Replaced by #634

PR #634, opened 2020-11-09
Fix canonical tests and page for posts not redirected

Chouby:
This PR fixes the canonical tests:
* Correctly applies the right option for each test with a page for posts.
* Rename tests which wrongly refered to a static front page.

This revealed that `test_page_for_post_option_should_be_translated_from_plain_permalink()` was wrongly passing. 
=> Fix the code accordingly.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -387,18 +387,18 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $wp_query->tax_query ) ) ) {
 				$term_id = $this->get_queried_term_id( $wp_query->tax_query );
 				$language = $this->model->term->get_language( $term_id );
+				$redirect_url = get_term_link( $term_id );
 			}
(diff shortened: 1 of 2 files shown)


raaaahman (approved):
Indeed, we needed a first test to run in order to reset the `$GLOBAL['wp_sitemaps']` variable: https://github.com/WordPress/wordpress-develop/blob/eb9fb00ee60ed392179f7f8148a2b4fcf6b05ff9/tests/phpunit/includes/abstract-testcase.php#L161

frontend/frontend-filters-links.php
```diff
@@ -387,20 +387,20 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
 			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $wp_query->tax_query ) ) ) {
 				$term_id = $this->get_queried_term_id( $wp_query->tax_query );
 				$language = $this->model->term->get_language( $term_id );
+				$redirect_url = get_term_link( $term_id );
```
manooweb, line 390: To make an explanation in a new PR

PR #638, opened 2020-11-10
Fix text direction field not populated when selecting a predefined language

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/748


manooweb (approved):
Ok 👍

PR #640, opened 2020-11-17
Deactivate Polylang as soon as possible

Chouby:
... when the site is queried with deactivate-polylang=1
See https://github.com/polylang/polylang-pro/pull/759

Up to now, integrations are loaded when the querying mysite.com/?deactivate-polylang=1. This can cause issues such as the one reported in https://github.com/polylang/polylang-pro/issues/755. Rather than protecting each integration one by one, I propose to stop loading the plugin very soon.


manooweb (approved):
OK 👍

PR #641, opened 2020-11-18
Fix a fatal error when the translation group is corrupted

Chouby:
A fatal error has been reported:
```
PHP Fatal error: Uncaught Error: Cannot unset string offsets in /home/wp/disk/wordpress/wp-content/plugins/polylang-pro/vendor/wpsyntex/polylang/include/translated-object.php:167
```
How to reproduce?
0. Activate Polylang 2.8.4
1. Create 2 posts translations of each other.
2. In the database, empty the description for the corresponding translations group term.
3. Attempt to delete permanently one of the 2 posts.

NB: Unfortunately, we don't know how the description could be removed on the customer's site.

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -163,13 +163,14 @@ abstract class PLL_Translated_Object {
 
 		if ( ! empty( $term ) ) {
 			$d = maybe_unserialize( $term->description );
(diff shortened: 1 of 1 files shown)


sebastienserre (approved):
Tested on the customer website, I can delete permanently without Fatale Error

manooweb (approved):
Ok 👍

PR #642, opened 2020-11-19
Insert WordPress core dynamic blocks doesn't work anymore in WP 5.6

manooweb:
WordPress 5.6-RC1-49652
Polylang 2.9-dev

By testing dynamic blocks - like Archives, Latest comments - for the [#760](https://github.com/polylang/polylang-pro/issues/760) issue with Polylang Pro, I noticed it doesn't work anymore with Polylang too and WP 5.6.

It's the same case we noticed with our language switcher block in the [#688](https://github.com/polylang/polylang-pro/issues/688) issue and fixed in the [#692](https://github.com/polylang/polylang-pro/pull/692) PR.

In fact it concerns all dynamic blocks and now WP 5.6 set a `data` property to `null` when it calls the REST API to render the blocks. When we corrected for our language switcher block we corrected for all the other ones in Polylang Pro. 

However WordPress core dynamic blocks can be used in block editor with Polylang and trigger the issue with the same javascript error and the block isn't rendered.

image

This PR propose to fix the issue the same way by testing if `options.data` is null too.


js/block-editor.js
```diff
@@ -11,7 +11,7 @@ wp.apiFetch.use(
 	function( options, next ) {
 		// If options.url is defined, this is not a REST request but a direct call to post.php for legacy metaboxes.
 		if ( 'undefined' === typeof options.url ) {
-			if ( 'undefined' === typeof options.data ) {
+			if ( 'undefined' === typeof options.data || null === options.data ) {
 				// GET
 				options.path += ( ( options.path.indexOf( '?' ) >= 0 ) ? '&lang=' : '?lang=' ) + getCurrentLanguage();
```
raaaahman, line 16: Shouldn't we create a `getRequestType()` function that we could use in both Polylang and Polylang Pro, so we will not forget to update both if it change in the future?

manooweb: 🤔 it isn't the same JS file enqueued. This one is enqueued only in Polylang. Polylang use another one build with webpack as the metabox React part

raaaahman (approved):
Tested with minified files, works as intended. :+1:

PR #644, opened 2020-11-22
Fix a fatal error when mixing multiple domains and plain permalinks

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/761

Although Polylang doesn't allow to use one domain per language when using plain permalinks, it's possible to come back to plain permalinks after the Polylang option has been selected. This causes a fatal error since Polylang 2.8. This PR proposes a quick fix to avoid this error.


manooweb (approved):
OK 👍

PR #645, opened 2020-11-22
Fix canonical redirect for paged page for posts

Chouby:
Since #634, we aim to correctly redirect a page for posts requested by its page id in plain permalinks to its correct pretty permalink url. The PR however introduced one bug and missed another one..

In plain permalinks, a paged page for post is not accessible i.e `https://mysite.com?paged=2&page_id=44` is always redirected to  `https://mysite.com?page_id=44`.

In pretty permalinks, `https://mysite.com?paged=2&page_id=44` is redirected to `https://mysite.com/en/posts/` instead of `https://mysite.com/en/posts/page/2/`.

This PR fixes the 2 issues and introduces a test for the second one.

PR #646, opened 2020-11-23
Fix get_canonical() in PLL_Canonical_UnitTestCase

Chouby:
The method `get_canonical()` in `PLL_Canonical_UnitTestCase` applies successively `PLL_Filters_Links::check_canonical_url()` and `redirect_canonical()`. This however doesn't reproduce the behavior in Polylang. Indeed `redirect_canonical()` is called only if `PLL_Filters_Links::check_canonical_url()` doesn't redirects itself.

This causes [this test](https://github.com/polylang/polylang/pull/645/files#diff-f788d849633ea312a1ed9d404920b14d83617efe27a56217db85c98f7390a713R225-R236) to always pass  even before the bug was fixed.

This PR fixes  `PLL_Canonical_UnitTestCase::get_canonical()` to better reproduce the Polylang behavior.


manooweb (approved):
Ok 👍

PR #647, opened 2020-11-23
Add tests for default category

Marianne380:
To cover : 

https://github.com/polylang/polylang/blob/2.8.4/admin/admin-filters-columns.php#L270-L273
https://github.com/polylang/polylang/blob/2.8.4/include/model.php#L398-L405

### DIFF
--- a/tests/phpunit/tests/test-admin-filters-term.php
+++ b/tests/phpunit/tests/test-admin-filters-term.php
@@ -414,47 +414,6 @@ class Admin_Filters_Term_Test extends PLL_UnitTestCase {
 		$this->assertEqualSets( array( 'en', 'fr', 'de', 'es' ), array_keys( $translations ) );
 	}
 
-	function test_new_default_category_with_empty_translations() {
-		$old = self::$polylang->model->term;
-		self::create_language( 'pt_PT' );
-		$default = (int) get_option( 'default_category' );
-
-		self::$polylang->model->term = $this->getMockBuilder( 'PLL_Translated_Term' )
-			->setMethods(
-				array(
-					'get_translations',
-				)
-			)
-			->setConstructorArgs(
-				array(
-					&self::$polylang->model,
-				)
-			)
-			->getMock();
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-admin-filters-term.php
```diff
@@ -414,6 +414,47 @@ function test_new_default_category() {
...
+
+		self::$polylang->model->create_default_category( self::$polylang->model->get_language( 'pt' )->slug );
+
+		self::$polylang->model->term = $old;
+
+		$translations = self::$polylang->model->term->get_translations( $default );
+
+		$this->assertEqualSets( array( 'en', 'pt' ), array_keys( $translations ) );
+	}
+
```
Chouby, line 457: It looks to me that https://github.com/polylang/polylang/blob/2.8.4/include/model.php#L398-L405 is dead code. Indeed `$translations = $this->term->get_translations( $default );` should at least return `$default`, due to https://github.com/polylang/polylang/blob/2.8.4/include/translated-object.php#L196-L198.

So you succeeded to execute these lines by mocking `PLL_Translated_Term::get_translations()` but this should not be representative of the behavior in production. I rather propose not to add this test and later remove the corresponding code in your refactoring.

tests/phpunit/tests/test-columns.php
```diff
@@ -145,6 +145,22 @@ function test_term_language() {
...
+		$default = (int) get_option( 'default_category' );
+
+		// with capability
+		$column = self::$polylang->filters_columns->term_column( '', 'language_en', $default );
+		$this->assertNotFalse( strpos( $column, 'default_cat' ) );
+
+		// without capability
+		wp_set_current_user( 0 );
+		$column = self::$polylang->filters_columns->term_column( '', 'language_en', $default );
+		$this->assertNotFalse( strpos( $column, 'default_cat' ) );
```
Chouby, line 161: I am not sure it's useful.

Chouby (changes requested):

PR #649, opened 2020-11-25
Fix PHP notice undefined index wp_the_query

Chouby:
Fix `PHP Notice:  Undefined index: wp_the_query in /polylang/frontend/choose-lang-content.php on line 92`
It may happen if a 3rd party plugin makes a WP_Query very soon in a function hooked to `plugins_loaded` and Polylang is set to get the language from the content. 

To test, add this in a plugin:
```php
add_action(
	'plugins_loaded',
	function() {
		get_posts();
	}
);
```


sebastienserre (approved):
This is fixing the issue

PR #650, opened 2020-11-25
Refresh Media Library frame after switching language

raaaahman:
## Goal

Find a way to refresh the attachments in the Media Libraries opened in the Classic Editor, in order to fetch attachments in the new language, after it has been switched from Polylang's metabox.

- Fixes https://github.com/polylang/polylang-pro/issues/756
- Fixes partially (Classic Editor only) https://github.com/polylang/polylang-pro/issues/737
- Fixes https://github.com/polylang/polylang-wc/issues/317

## Use case analysis

Polylang offers users to the possibility to change the language of the currently edited post through its _Language metabox_. However, when doing so, the different WordPress Media Library items are not refreshed to match the new language. This happens in:

- The _Set Featured Image_ action link in the post edition screen
- The  _Add Media_ button in the post edition screen

This Media Library frame is then reused in plugins including:

- WooCommerce
    - The _Set Product Image_ action link and _Add Media_ buttons in the product edition screen work similarly than those in the post edition screen
    - The _Set Category Image_ action link in the product category edition screen work similarly than the _Set Featured Image_ in the post edition screen
- ACF
    - The _Image_, _File_ and _Gallery_ fields extends the behavior of _Add Media_ button in the WordPress' post edition screen

## Code analysis

WordPress Media Library is built upon the Backbone.js framework. The top level view that contains the Media Library is an instance of `wp.media.view.MediaFrame.Select` or `wp.media.view.MediaFrame.Post` component. It is then subdivided into more specific components, down to a `wp.media.view.Attachments` view, displaying the different media attachments.

The `wp.media.view.Attachments` view is linked to a `wp.media.model.Attachments` collection that holds the data to be displayed by the view. When the collection is updated, the associated view updates itself to.

The `wp.media.view.Attachments` is 'mirrored' by a `wp.media.model.Query`, that will decide if it should fetch data from an AJAX request or not.

Class hierarchy of WordPress' Media Frame components

A `wp.media.frame` global variable will hold the instance of the last opened MediaFrame. **Warning:** this media frame might be a custom subclass of `wp.media.view.MediaFrame` from a plugin like ACF.

A `wp.media.model.Attachments.all` global holds one copy of each unique attachments displayed in any MediaFrame. However, I did not find the usage of these copies.

## Edge cases

- ~~Switching language to a language formerly used language will open an empty list. Example: opening media in English(US), switching to French(FR), opening medias in French, then switching again to English(US) and trying to open the media again.~~
- ~~Opening the Media Library from the _Add Media_ button, then opening it from the _Set Featured Image_ button will open two different MediaFrames. Switching languages will only refresh the last opened frame  (the featured image).~~

## Notes

- Underscore's `_.extend()` might be equivalent to native Javascript's `Object.assign()`.
- FIXME: Because we are now holding a reference to each `wp.media.model.Attachments` collections created, this will prevent the garbage collector to free them from memory when they're not used anywhere else.

### DIFF
--- a/js/classic-editor.js
+++ b/js/classic-editor.js
@@ -270,8 +270,8 @@ var media = _.extend(
 					 */
 					attachmentsCollection.reset();
 					if (attachmentsCollection.mirroring) {
-						  attachmentsCollection.mirroring._hasMore = true;
-						  attachmentsCollection.mirroring.reset();
+						attachmentsCollection.mirroring._hasMore = true;
+						attachmentsCollection.mirroring.reset();
 					}
 				}
 			);
--- a/js/classic-editor.js
+++ b/js/classic-editor.js
@@ -239,7 +239,7 @@ function resetCurrentMediaFrame() {
 	}
 	if (wp.media.editor) {
 		var editorFrame = wp.media.editor.get( wp.media.editor.activeEditor );
-		if (editorFrame) {
+		if (editorFrame && editorFrame !== wp.media.frame) {
 			resetMediaFrame( editorFrame );
 		}
 	}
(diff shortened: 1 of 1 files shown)


raaaahman:
Included in #677

PR #651, opened 2020-11-27
Handle SameSite parameter for language cookie

Chouby:
Since Chrome 80, we need to handle the SameSite parameter when setting the language cookie.

This parameter is available in PHP only since the version 7.3. We need to support previous versions which still concern 55% of WordPress installs at the time of writing this. A hack was possible in these previous versions to add the SameSite parameter, however it looks like this hack doesn't work anymore in PHP 7.3.

This PR adds a `PLL_Cookie` static class to wrap the `setcookie` function. The method `set()` uses one or the other method depending on the PHP version.

### DIFF
--- a/integrations/cache/cache-compat.php
+++ b/integrations/cache/cache-compat.php
@@ -34,19 +34,22 @@ class PLL_Cache_Compat {
 	 * @since 2.3
 	 */
 	public function add_cookie_script() {
-		$domain = ( 2 == PLL()->options['force_lang'] ) ? wp_parse_url( PLL()->links_model->home, PHP_URL_HOST ) : COOKIE_DOMAIN;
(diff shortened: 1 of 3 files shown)


raaaahman (approved):

manooweb (commented):
Nothing more to add to @raaaahman's comments

integrations/cache/cache-compat.php
```diff
@@ -34,19 +34,22 @@ public function init() {
 	 * @since 2.3
 	 */
 	public function add_cookie_script() {
-		$domain = ( 2 == PLL()->options['force_lang'] ) ? wp_parse_url( PLL()->links_model->home, PHP_URL_HOST ) : COOKIE_DOMAIN;
+		$domain   = ( 2 === PLL()->options['force_lang'] ) ? wp_parse_url( PLL()->links_model->home, PHP_URL_HOST ) : COOKIE_DOMAIN;
+		$samesite = ( 3 === $this->options['force_lang'] ) ? 'None' : 'Lax';
```
manooweb, line 38: Comment to reference in a new PR

PR #654, opened 2020-12-03
Fix update of option translations registered with a wildcard

Chouby:
This PR follows up #613 and aims to fix 2 bugs I discovered in `PLL_Translate_Option::check_value_recursive()`.
1. `array_keys()` was used for options saved as object.
2. A typo coming from a wrong copy paste with `$child` used instead of `$key`.
Tests are added to cover this part of the code.


manooweb (approved):
Ok 👍

PR #655, opened 2020-12-03
A big cleanup

Chouby:
Polylang 1.2 was released 7 years ago. Polylang 1.8 was released 5 years ago. It's now time to stop the backward compatibility.

This PR:
* Removes the possibility to update from versions < 1.8.
* Removes the backward compatibility with versions < 1.8

It also removes some compatibilty code with old versions of WP which are now unsupported.


manooweb (approved):
Ok 👍 It's indeed a huge cleanup 😮

PR #656, opened 2020-12-04
Define abstract methods in PLL_Links_Model

Chouby:
We already propose 4 links models, but our users may need to implement a different one. This requires to extend `PLL_Links_Model`. However it's currently not clear enough which methods need to be implemented, as our parent class "transparently" uses methods defined only in our child classes.

This PR adds abstract methods to `PLL_Links_Model` to clarify which methods child classes are required to implement.

### DIFF
--- a/include/links-model.php
+++ b/include/links-model.php
@@ -4,43 +4,130 @@
  */
 
 /**
- * Links model abstract class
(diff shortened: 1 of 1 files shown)


manooweb (approved):
Ok. 👍

PR #657, opened 2020-12-04
Define abstract methods in PLL_Translated_Object

Chouby:
This PR clarifies the methods used in `PLL_Translated_Object` that the child classes are required to implement by defining abstract methods. 

It also takes the opportunity to improve the inline documentation.


manooweb (approved):
Ok. 👍

PR #660, opened 2020-12-07
Use PHPdoc style to document PLL_Language properties and remove useless ones

Chouby:
- This PR replaces the old style documentation for `PLL_Language` properties by PHPDoc documentation.
- It also removes some useless properties which were automatically imported from the language `WP_Term` properties (for example, `taxonomy` or `filter`). This is done by defining the properties explicitely instead of defining them implicitely in a somewhat anonymous way in the `foreach` loop.
- Additionnally, it applies WPCS coding standards to comments through the complete `PLL_Language` class.

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -4,77 +4,210 @@
  */
 
 /**
- * A language object is made of two terms in 'language' and 'term_language' taxonomies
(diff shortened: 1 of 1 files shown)


manooweb (approved):
Ok 👍

PR #662, opened 2020-12-10
Link the post saving to the metabox refreshing

manooweb:
fixes [#772](https://github.com/polylang/polylang-pro/issues/772)

Before this PR:

- both `post saving` and `metabox refreshing` processes are triggered separately
- each one are bound separately to a language dropdown change event
- `metabox refreshing` also set the new post language during the process
- `post saving` triggers the page reloading if the `post saving` is complete and succeeds
- it seems that before WordPress 5.3 the `post saving` is slower and then leave time to `metabox refreshing` to be done before reloading the page. Since WordPress 5.3 `post saving` seems to run faster and then reloading the page before the `metabox refreshing` is done which causes the [#772](https://github.com/polylang/polylang-pro/issues/772) issue.

So this PR propose to link the `post saving` process after the `metabox refreshing` process is complete.
To do so the code was modified as follow:

- bind all the processes to only one language dropdown change event
- specific cases for empty new post is put at the beginning because direct URL changing triggers automatically the page reloading without doing anything else.
- the `metabox refreshing` which also set the new post language is processed first
- the `post saving` process is linked to the `metabox refreshing` by waiting its responses and that its processing is complete

This new implementation seems to work for both WordPress < 5.3 and WordPress >= 5.3 including 5.6


js/block-editor.js
```diff
@@ -153,8 +98,75 @@ jQuery(
...
+					);
+
+					// Triggers the post save.
+					dispatch( 'core/editor' ).savePost();
+
+					// Process
+					savePostIsDone.then(
+						function() {
+							// If the post is well saved, we can reload the page
+							unsubscribe();
```
raaaahman, line 156: Why do we need to unsubscribe if we reload the page anyway?

raaaahman (commented):
Since the whole process is triggered on a 'change' event, why didn't we retrieve the `lang` value from this event instead of parsing its current value in the form?

https://github.com/polylang/polylang/blob/2b69b4666b84143897f445a8d1ebbee566639916/js/block-editor.js#L74

What could happen if the user switches language again before the `savePostIsDone` Promise is resolved?

manooweb:
> Since the whole process is triggered on a 'change' event, why didn't we retrieve the `lang` value from this event instead of parsing its current value in the form?
> 
> https://github.com/polylang/polylang/blob/2b69b4666b84143897f445a8d1ebbee566639916/js/block-editor.js#L74
> 
> What could happen if the user switches language again before the `savePostIsDone` Promise is resolved?

It's something which is like that at the beginning and we never modify 
See
https://github.com/polylang/polylang/blob/2.9/js/classic-editor.js#L95-L97

BTW I notice we forgot to remove the deprecated way to bind the change event in the classic-editor.js script

js/block-editor.js
manooweb: Done!

js/block-editor.js
```diff
@@ -153,8 +98,74 @@ jQuery(
...
+											resolve();
+										}
+									}
+								}
+							);
+						}
+					);
+
+					// Triggers the post save.
+					dispatch( 'core/editor' ).savePost();
```
raaaahman, line 150: Do we need to save the post again, now that the two actions are done sequentially?

raaaahman:
> Done! I retrieve the language value from the event which is passed now as a parameter of the callback function.

Do you believe this could have been sufficient solve the issue ?

Also, we might propagate the change in the similar functions: 

https://github.com/polylang/polylang/blob/2cb0ad7277d47a224808bcedf0980852be738587/js/classic-editor.js#L97

https://github.com/polylang/polylang/blob/cd20867e94dcf6d023dfa7e886b99bd2b50d8fc9/js/term.js#L169

Or we could take advantage of the use of webpack to refactor these three resembling functions.

raaaahman (approved):

PR #663, opened 2020-12-10

raaaahman:
## Before

Javascript files where built as `<original_file_name>.min.js` whether they were actually minified or not.

## Changes

- Javascript files with the `.src` suffix will be built in both minified and non-minified versions.
- Javascript files without any suffix will be minified
- Javascript files with the `.dep` suffix will never be built nor minified. They are only meant to be imported into other files.

## Going further

This method forces us to manually add the files built from source to the `.gitignore` list, as there are actually no differences between `.js` files that are built from sources and those that doesn't need to be built at all.


raaaahman:
Replaced by #675

PR #664, opened 2020-12-11
Webpackization remove duplication

raaaahman:
## Changes

This PR builds upon #663 

It uses the webpack configuration to extract a javascript function shared by both `classic-editor.js` and `block-editor.js` (now turned into source files).


raaaahman:
Replaced by #678

PR #665, opened 2020-12-11
Translate title-search-wpseo option

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/780

Since this PR #583  which rewrote a part of the WordPress SEO compatibility, we missed one of the indexable option: title-search-wpseo which isn't translated as it's reported in the issue.


sebastienserre (approved):
On my test website, the Title is now well translated !

PR #670, opened 2020-12-15
Use PHPDoc to document the main Polylang object properties

Chouby:
This PR replaces the old documentation of the child classes of `PLL_Base` - main Polylang object accessible with `PLL()` - by PHPDoc.

### DIFF
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -4,7 +4,7 @@
  */
 
 /**
- * Frontend controller, accessible from PLL().
+ * Main Polylang class when on frontend, accessible from @see PLL().
  *
  * @since 1.2
  */
@@ -24,7 +24,7 @@ class PLL_Frontend extends PLL_Base {
 	public $auto_translate;
 
 	/**
-	 * Instance of the class selecting the current language.
+	 * The class selecting the current language.
 	 *
 	 * @var PLL_Choose_Lang
 	 */
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -4,30 +4,73 @@
  */
 
 /**
- * Frontend controller
- * accessible as $polylang global object
- *
- * Properties:
- * options        => inherited, reference to Polylang options array
- * model          => inherited, reference to PLL_Model object
- * links_model    => inherited, reference to PLL_Links_Model object
- * links          => reference to PLL_Links object
- * static_pages   => reference to PLL_Frontend_Static_Pages object
- * choose_lang    => reference to PLL_Choose_Lang object
- * curlang        => current language
- * filters        => reference to PLL_Frontend_Filters object
- * filters_links  => reference to PLL_Frontend_Filters_Links object
- * filters_search => reference to PLL_Frontend_Filters_Search object
- * posts          => reference to PLL_CRUD_Posts object
- * terms          => reference to PLL_CRUD_Terms object
- * nav_menu       => reference to PLL_Frontend_Nav_Menu object
- * auto_translate => optional, reference to PLL_Auto_Translate object
+ * Frontend controller, accessible from PLL().
  *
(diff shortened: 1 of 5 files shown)


frontend/frontend.php
```diff
@@ -4,30 +4,73 @@
...
- * choose_lang    => reference to PLL_Choose_Lang object
- * curlang        => current language
- * filters        => reference to PLL_Frontend_Filters object
- * filters_links  => reference to PLL_Frontend_Filters_Links object
- * filters_search => reference to PLL_Frontend_Filters_Search object
- * posts          => reference to PLL_CRUD_Posts object
- * terms          => reference to PLL_CRUD_Terms object
- * nav_menu       => reference to PLL_Frontend_Nav_Menu object
- * auto_translate => optional, reference to PLL_Auto_Translate object
+ * Frontend controller, accessible from PLL().
```
raaaahman, line 7: Mixed responsibilities between a Controller and a Service Locator.

include/rest-request.php
```diff
@@ -4,27 +4,49 @@
...
- * options        => inherited, reference to Polylang options array
- * model          => inherited, reference to PLL_Model object
- * links_model    => inherited, reference to PLL_Links_Model object
- * links          => reference to PLL_Admin_Links object
- * static_pages   => reference to PLL_Static_Pages object
- * filters        => reference to PLL_Frontend_Filters object
- * filters_links  => reference to PLL_Filters_Links object
- * posts          => reference to PLL_CRUD_Posts object
- * terms          => reference to PLL_CRUD_Terms object
+ * REST API controller, accessible from PLL().
```
raaaahman, line 7: No methods related to a Controller responsibility. 100% Service Locator.

include/rest-request.php
```diff
@@ -39,14 +61,14 @@ public function init() {
 			$this->filters_links = new PLL_Filters_Links( $this );
 			$this->filters = new PLL_Filters( $this );
 
-			// Static front page and page for posts
+			// Static front page and page for posts?
```
raaaahman, line 64: Comments should provide answers to the readers. They shouldn't raise more questions.

settings/settings.php
```diff
@@ -4,32 +4,21 @@
...
- * Properties:
- * options          => inherited, reference to Polylang options array
- * model            => inherited, reference to PLL_Model object
- * links_model      => inherited, reference to PLL_Links_Model object
- * links            => inherited, reference to PLL_Admin_Links object
- * static_pages     => inherited, reference to PLL_Admin_Static_Pages object
- * filters_links    => inherited, reference to PLL_Filters_Links object
- * curlang          => inherited, optional, current language used to filter admin content
- * pref_lang        => inherited, preferred language used as default when saving posts or terms
+ * A class for the Polylang settings pages, accessible from PLL().
```
raaaahman, line 7: This one is actually a Controller.

settings/settings.php
```diff
@@ -4,32 +4,21 @@
...
 	 *
 	 * @var string $active_tab
 	 */
 	protected $active_tab;
 
 	/**
-	 * Array of modules classes
+	 * Array of modules classes.
 	 *
 	 * @var array $modules
```
raaaahman, line 23: You could use the `PLL_Settings_Modules[]` syntax. It would mean: an array of `PLL_Settings_Modules` instances, or instances of `PLL_Settings_Modules` child classes.

raaaahman (commented):
I suggest you'd try to avoid duplication in comments. For example, if `@param` doc tag says its a `string`, there's no need to say it again in the short description. If we need to change the parameter type in the future, that would mean we have to change the documentation in two places.

This gets more problematic when you reference the child classes from their parent class: if you change the inheritance chain, would you remember to update the documentation in the parent class?

PR #672, opened 2020-12-16
Modernize CI

szepeviktor:
- put config in chronological order
- boost speed with caching
- name jobs
- be explicit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,64 +1,73 @@
-sudo: false
+# TravisCI configuration for polylang/polylang
 
-language: php
+if: "branch = master"
+
+language: "php"
+os:
+  - "linux"
+dist: "xenial"
+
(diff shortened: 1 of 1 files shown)


szepeviktor:
Looks good!

kép

szepeviktor:
kép

Please take action. 🚨

szepeviktor:
That is the standard way of writing YAML strings. **All** string are double-quoted currently. Non-string are not double-quoted.
Please see the mother of all GHA configs https://github.com/ergebnis/php-library-template/blob/main/.github/workflows/integrate.yaml

szepeviktor:
If one uses double-quotes when necessary one has to decide line-by-line which is

- time consuming 
- and error prone

szepeviktor:
Thank you.

PR #673, opened 2020-12-17
Replace a property by a local variable

Chouby:
In `PLL_Translate_Option`, the property `old_value` is not declared, not documented and used in only on method. I propose to replace it by a local variable.


raaaahman (approved):
The `unfiltered` part of the name adds helpful information. Nice.

PR #675, opened 2020-12-17

raaaahman:
## Before

- Only minified javascript files where built from sources. These didn't allow us to import a Js file into another, because the non-minfied version would be incorrect.
- Minified javascript and CSS files where built in the same folder than their sources, this forced us to ignore them in our VCS by their name suffix.

## Changes

Replaces #663 

- Javascript files ~~with the `.src` suffix~~ will be built in both minified and non-minified versions.
- ~~Javascript files without any suffix will be minified~~ Included in rule above.
- Javascript files ~~with the `.dep` suffix~~ in the `js/lib` folder will never be built nor minified. They are only meant to be imported into other files.
- Javascript and CSS files are now built inside specific build folders, respectively `js/build/` and `css/build/`
- CSS sources are copied as they are for non minified versions.
- Build scripts are exported into a specific npm package

## Going further

~~This method forces us to manually add the files built from source to the `.gitignore` list, as there are actually no differences between `.js` files that are built from sources and those that doesn't need to be built at all.~~ Fixed by https://github.com/polylang/polylang/pull/675/commits/2b7df6f8a19d82123b5a022bee0eb0eefc6d5d7f
- Will conflict with https://github.com/polylang/polylang/pull/724 (because both changes the `wp_enqueue_script()` and `wp_enqueue_style()` calls) but they can be merged either way, and the way to resolve the conflict will be the same.
- The dependency to a npm package might become unnecessary if the build folders are moved inside Polylang Pro: the context of the build will then become Polylang Pro root folder and the dependencies to external tools will then be resolved from inside it.
- I ran into trouble by instantiating `mini-css-extract-plugin` twice (once in Polylang Pro root folder and once in `vendor/wpsyntex/polylang`), so I needed to install the dependencies of the build scripts inside both Polylang and Polylang Pro packages instead of in the build script package itself. All-in-all, this external package for build scripts is more troublesome than it is helpful.


.gitignore
```diff
@@ -6,3 +6,4 @@ tmp/
 *.map
 *.min.js
 *.min.css
+*.map
```
manooweb, line 9: Rule already exists on master just above `*.min.js` rule

manooweb:
🤔 Strange bin/build.sh and bin/distribute.sh are commited as empty and they mustn't because they doesn't have to override ones on master branch

webpack/FileNames.js
```diff
@@ -0,0 +1,69 @@
...
+
+	function computeBuildFilename( filename, suffix ) {
+		const nameWithoutSuffix = path.parse( filename ).name.split( '.' )[0];
+		suffix = suffix ? '.' + suffix : '';
+
+		return `${ path.parse( filename ).dir }/${ nameWithoutSuffix + suffix }.js`; // phpcs:ignore WordPress.WhiteSpace.OperatorSpacing
+	}
+
+	// Prepare webpack configuration to minify js files to source folder as target folder and suffix file name with .min.js extension.
+	function mapJsFiles( jsFileNames, minimize = false ) {
```
manooweb, line 42: Shouldn't we define these functions outside the `getJsFileNamesEntries()` one ?

manooweb (changes requested):

raaaahman:
> Strange bin/build.sh and bin/distribute.sh are commited as empty and they mustn't because they doesn't have to override ones on master branch

That's the diff files that is empty, because the files content didn't change. But the files permissions did (from 774 to 644). Don't know if it would be an issue though...

manooweb:
> > Strange bin/build.sh and bin/distribute.sh are commited as empty and they mustn't because they doesn't have to override ones on master branch
> 
> That's the diff files that is empty, because the files content didn't change. But the files permissions did (from 774 to 644). Don't know if it would be an issue though...

🤔 OK. I launch this bash scripts in my linux WSL kernel and I haven't any problem with permission on them. Is it the case on your team's linux environment ?

raaaahman:
Well 644 is not executable, so this is a problem indeed...

raaaahman:
Replaced by #726

PR #676, opened 2020-12-17
Fix missing property declaration and documentation

Chouby:
All is in the tiltle ;-)

### DIFF
--- a/admin/admin-block-editor.php
+++ b/admin/admin-block-editor.php
@@ -9,7 +9,19 @@
  * @since 2.5
  */
 class PLL_Admin_Block_Editor {
-	public $model;
+	/**
+	 * Instance of PLL_Model.
+	 *
+	 * @var PLL_Model
+	 */
+	protected $model;
+
+	/**
+	 * Preferred language to assign.
+	 *
+	 * @var PLL_Language
+	 */
+	protected $pref_lang;
 
 	/**
 	 * Constructor: setups filters and actions
--- a/admin/admin-block-editor.php
+++ b/admin/admin-block-editor.php
@@ -17,7 +17,7 @@ class PLL_Admin_Block_Editor {
 	protected $model;
 
 	/**
-	 * Preferred language to assign.
+	 * Preferred language to assign to a new post.
 	 *
 	 * @var PLL_Language
 	 */
--- a/admin/admin-notices.php
+++ b/admin/admin-notices.php
@@ -12,6 +12,18 @@
  * @since 2.7 Dismissed notices are stored in an option instead of a user meta
  */
 class PLL_Admin_Notices {
+	/**
+	 * Stores the plugin options.
+	 *
+	 * @var array
+	 */
+	protected $options;
+
+	/**
+	 * Stores custom notices.
+	 *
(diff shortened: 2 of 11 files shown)


admin/admin-block-editor.php
```diff
@@ -9,7 +9,19 @@
...
-	public $model;
+	/**
+	 * Instance of PLL_Model.
+	 *
+	 * @var PLL_Model
+	 */
+	protected $model;
+
+	/**
+	 * Preferred language to assign.
```
raaaahman, line 20: To assign to what?

admin/admin-notices.php
```diff
@@ -12,6 +12,18 @@
...
+	 * Stores the plugin options.
+	 *
+	 * @var array
+	 */
+	protected $options;
+
+	/**
+	 * Stores custom notices.
+	 *
+	 * @var string[]
```
raaaahman, line 25: :+1:

include/crud-posts.php
```diff
@@ -4,12 +4,32 @@
  */
 
 /**
- * Adds actions and filters related to languages when creating, updating or deleting posts
- * Actions a filters used when reaing posts are handled separately
+ * Adds actions and filters related to languages when creating, updating or deleting posts.
+ * Actions an filters used when reading posts are handled separately.
```
raaaahman, line 8: - Probably 'Actions **and** filters'. 
- 'Used' is very generic. 'Triggered' maybe?
- I suppose that is the reading that **is** handled separately.
- Separately from what?

include/nav-menu.php
```diff
@@ -10,7 +10,33 @@
...
+
+	/**
+	 * Theme name.
+	 *
+	 * @var string
+	 */
+	protected $theme;
+
+	/**
+	 * Array of menus in a given language used when auto add pages to menus.
```
raaaahman, line 35: I don't really know how menu are stored. Nested associative arrays?

Maybe you could point to WordPress documentation or code to help visualize those menus.

modules/wizard/wizard.php
```diff
@@ -36,6 +36,13 @@ class PLL_Wizard {
 	 */
 	protected $steps = array();
 
+	/**
+	 * The current step.
+	 *
+	 * @var string
+	 */
+	protected $step;
```
raaaahman, line 44: I would say it's worth changing the variable's name in these cases. `$current_step` would help remind us what that variable holds throughout all the code base rather than needing to read it from the variable's declaration.

raaaahman (approved):

modules/wizard/wizard.php
Chouby: It's not my goal to rewrite the code, as it's too many changes for my PR which only focuses on declarations and doc.

include/crud-posts.php
manooweb: `Separately` from creating, updating or deleting posts events quoted above the comment. It seems clear to me.

raaaahman: Ah, you mean that this class does not handle reading posts, whereas its name 'CRUD' might make us think it does...

Okay, I get it now.

PR #678, opened 2020-12-17
Webpackization remove duplication 

raaaahman:
## Changes

Includes #675 
Replaces #664 

It uses the webpack configuration to extract a javascript function shared by both `classic-editor.js` and `block-editor.js` (now turned into source files).


manooweb (commented):
OK for me as a proof of concept and after the PR #675 will be approved.

I think it is preferable to use a distribution build folder to be able to exclude this folder from the building process and then avoid the use of `.src.` and `.dep.` suffix.
Then we can keep .js extension both for source files and build files. For code developed only to make it common we can imagine a specific folder as library to store all the common piece of code.

That should simplify the build process too by not differentiating scripts, which need to import code, from the other scripts which have no need to import code

However we will need to review all the wp_enqueue_script() calls to reference the new distribution build folder.

raaaahman:
Since #675 needs more work, this one will have to be rebased anyway.

PR #680, opened 2020-12-17
Improve project build for internal use

manooweb:
Exactly the same proposition as the Polylang Pro [PR #792](https://github.com/polylang/polylang-pro/pull/792) but for Polylang.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,6 +1,12 @@
 #!/bin/sh
+COMPOSER_COMMAND='install'
+
+if [ "$1" = "-u" ] || [ "$1" = "-update" ]; then
+    COMPOSER_COMMAND='update'
+fi
+
 echo "Installing PHP packages..."
-composer update # Need update to ensure to have the latest version of Polylang dependencies.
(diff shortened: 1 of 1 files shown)


manooweb:
Exactly the same modification as in [Polylang Pro PR](https://github.com/polylang/polylang-pro/pull/792) and described in its latest comment https://github.com/polylang/polylang-pro/pull/792#issuecomment-748101464

bin/build.sh
```diff
@@ -1,11 +1,17 @@
 #!/bin/sh
-echo "Installing PHP packages..."
-composer install
+COMPOSER_COMMAND='install'
+
+if [ "$1" = "-u" ] || [ "$1" = "-update" ]; then
```
raaaahman, line 4: With CLI programs, the long options names (more than one letter) always start by two dashes. That way, the program can differentiate the `--update` option from the concatenation of the `-u -p -d -a -t -e -s` options.

manooweb: Done!

PR #681, opened 2020-12-18
Add missing documentation for class properties

Chouby:
This adds all missing PHPDoc comments to all class properties and enables the sniff Squiz.Commenting.VariableComment in PHPCS config, except for tests.

### DIFF
--- a/admin/admin-strings.php
+++ b/admin/admin-strings.php
@@ -9,8 +9,19 @@
  * @since 1.6
  */
 class PLL_Admin_Strings {
-	protected static $strings = array(); // strings to translate
-	protected static $default_strings; // default strings to register
+	/**
+	 * Stores the strings to translate.
+	 *
+	 * @var array
+	 */
+	protected static $strings = array();
+
+	/**
+	 * The strings to register by default
+	 *
+	 * @var string[]
+	 */
+	protected static $default_strings;
 
 	/**
 	 * Add filters
--- a/admin/admin-strings.php
+++ b/admin/admin-strings.php
@@ -12,12 +12,17 @@ class PLL_Admin_Strings {
 	/**
 	 * Stores the strings to translate.
 	 *
-	 * @var array
+	 * @var array {
+	 *   @type string $name      A unique name for the string.
+	 *   @type string $string    The actual string to translate.
+	 *   @type string $context   The group in which the string is registered.
+	 *   @type bool   $multiline Whether the string table should display a multiline textarea or a single line input.
+	 * }
 	 */
 	protected static $strings = array();
 
 	/**
-	 * The strings to register by default
+	 * The strings to register by default.
 	 *
 	 * @var string[]
 	 */
--- a/include/links-permalinks.php
+++ b/include/links-permalinks.php
@@ -40,7 +40,7 @@ abstract class PLL_Links_Permalinks extends PLL_Links_Model {
 	protected $use_trailing_slashes;
 
 	/**
-	 * Rewrite rules to always filter
+	 * The name of the rewrite rules to always modify.
 	 *
 	 * @var string[]
 	 */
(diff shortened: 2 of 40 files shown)


admin/admin-strings.php
```diff
@@ -9,8 +9,19 @@
  * @since 1.6
  */
 class PLL_Admin_Strings {
-	protected static $strings = array(); // strings to translate
-	protected static $default_strings; // default strings to register
+	/**
+	 * Stores the strings to translate.
+	 *
+	 * @var array
```
raaaahman, line 15: `string[]` is more precise than `array`

include/links-permalinks.php
```diff
@@ -9,9 +9,41 @@
...
+	 * @var bool
+	 */
+	protected $use_trailing_slashes;
+
+	/**
+	 * Rewrite rules to always filter
+	 *
+	 * @var string[]
+	 */
 	protected $always_rewrite = array( 'date', 'root', 'comments', 'search', 'author' );
```
raaaahman, line 47: The comment says that these are rules to always _filter_ while the variable name says these are rules to always _rewrite_. Which one is more precise?

include/walker-dropdown.php
```diff
@@ -9,6 +9,12 @@
  * @since 1.2
  */
 class PLL_Walker_Dropdown extends Walker {
+	/**
+	 * Database fields to use.
+	 * See Walker::$db_fields
```
raaaahman, line 14: Maybe use a `@see` doc tag with a link to `Walker::$db_fields`?

phpcs.xml.dist
```diff
@@ -121,6 +119,14 @@
 		<exclude-pattern>*/tests/*</exclude-pattern>
 	</rule>
 
+	<rule ref="Squiz.Commenting.VariableComment.Missing" >
```
raaaahman, line 122: The following comment makes me believe that we could skip the variable short description when it doesn't add more information: https://github.com/squizlabs/PHP_CodeSniffer/blob/9bcb29cce831a8f9b2c8aee0bba1a57445f9dbf1/src/Standards/Squiz/Sniffs/Commenting/VariableCommentSniff.php#L83

I'm referring to documentation like:

```php
	/**
	 * Instance of PLL_Model.
	 *
	 * @var PLL_Model
	 */
	public $model;
```

settings/table-string.php
```diff
@@ -14,7 +14,33 @@
...
+
+	/**
+	 * The string groups.
+	 *
+	 * @var string[]
+	 */
+	protected $groups;
+
+	/**
+	 * The selected string group or -1 if none is selected.
```
raaaahman, line 39: Helpful documentation. :+1:

admin/admin-strings.php
Chouby: Well in fact it's not an array of strings, but an array of arrays.
See https://github.com/polylang/polylang/blob/20c00ffe116c4f203e5088178c4fede8cd1319e1/admin/admin-strings.php#L48

phpcs.xml.dist
Chouby: It seems that you are right. I wonder how it could look in a generated doc.

include/walker-dropdown.php
Chouby: Would
`@see https://developer.wordpress.org/reference/classes/walker/#properties Walker::$db_fields`
be ok for you?

include/links-permalinks.php
Chouby: I propose to change to:
`The name of the rewrite rules to always filter.`
This is maybe more explicit.

phpcs.xml.dist
sebastienserre: It seems the WP parser doesn't parse the class properties, but I maybe missed something ...

include/walker-dropdown.php
manooweb: Exactly! https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/see.html#see
Don't forget that you can use a description behind the URL if necessary

PR #683, opened 2020-12-21
Fix missing or useless isset

Chouby:
This PR fixes errors reported by PHPStan (level 1)
- Useless `isset()`.
- Variables which might not be defined due to missing `isset()`.

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -282,7 +282,7 @@ class PLL_Admin_Model extends PLL_Model {
 
 		// Validate slug is unique
 		foreach ( $this->get_languages_list() as $language ) {
-			if ( $language->slug === $args['slug'] && ( null === $lang || ( isset( $lang ) && $lang->term_id != $language->term_id ) ) ) {
+			if ( $language->slug === $args['slug'] && ( null === $lang || $lang->term_id !== $language->term_id ) ) {
 				$errors->add( 'pll_non_unique_slug', __( 'The language code must be unique', 'polylang' ) );
 			}
 		}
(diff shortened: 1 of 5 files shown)


include/translated-object.php
```diff
@@ -165,7 +165,7 @@ protected function should_update_translation_group( $id, $translations ) {
 	public function save_translations( $id, $translations ) {
 		$id = (int) $id;
 
-		if ( ( $lang = $this->get_language( $id ) ) && isset( $translations ) && is_array( $translations ) ) {
```
raaaahman, line 168: Beware that `is_array()` returns a PHP Notice when passed an undefined variable.

raaaahman (changes requested):
Some changes modify slightly the behavior here. I don't feel confident they will not introduce subtle bugs or notices.

admin/admin-model.php
```diff
@@ -282,7 +282,7 @@ protected function validate_lang( $args, $lang = null ) {
 
 		// Validate slug is unique
 		foreach ( $this->get_languages_list() as $language ) {
-			if ( $language->slug === $args['slug'] && ( null === $lang || ( isset( $lang ) && $lang->term_id != $language->term_id ) ) ) {
+			if ( $language->slug === $args['slug'] && ( null === $lang || $lang->term_id !== $language->term_id ) ) {
```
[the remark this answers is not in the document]
Chouby: I hust double checked. Both languages are `PLL_Language` objects in which we explicitely cast the property to `int`: See https://github.com/polylang/polylang/blob/22e2915b15b8b1abf70b8ca91b66d8ca8e723787/include/language.php#L199

include/translated-object.php
Chouby: I removed the `isset()` because `$translations` is passed to the function and is not optional so is always set.

PR #684, opened 2020-12-21
Fix some documentation errors and wrong default values

Chouby:
This PR fixes some documentation errors.
It also fixes some default values which were not of the correct type.

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -257,15 +257,15 @@ class PLL_Admin_Model extends PLL_Model {
 	}
 
 	/**
-	 * Validates data entered when creating or updating a language
+	 * Validates data entered when creating or updating a language.
 	 *
 	 * @see PLL_Admin_Model::add_language()
 	 *
 	 * @since 0.4
 	 *
-	 * @param array  $args
-	 * @param object $lang optional the language currently updated, the language is created if not set
-	 * @return bool true if success / false if failed
+	 * @param array  $args Parameters of {@see PLL_Admin_Model::add_language() or @see PLL_Admin_Model::update_language()}.
+	 * @param object $lang Optional the language currently updated, the language is created if not set.
+	 * @return WP_Error
 	 */
 	protected function validate_lang( $args, $lang = null ) {
 		$errors = new WP_Error();
@@ -428,12 +428,12 @@ class PLL_Admin_Model extends PLL_Model {
 	}
 
 	/**
-	 * Returns untranslated posts and terms ids ( used in settings )
+	 * Returns untranslated posts and terms ids ( used in settings ).
 	 *
 	 * @since 0.9
-	 * @since 2.2.6 Add the $limit argument
+	 * @since 2.2.6 Add the $limit argument.
 	 *
-	 * @param in $limit Max number of posts or terms to return. Defaults to -1 (no limit).
+	 * @param int $limit Max number of posts or terms to return. Defaults to -1 (no limit).
 	 * @return array Array made of an array of post ids and an array of term ids
 	 */
 	public function get_objects_with_no_lang( $limit = -1 ) {
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -434,7 +434,12 @@ class PLL_Admin_Model extends PLL_Model {
 	 * @since 2.2.6 Add the $limit argument.
 	 *
 	 * @param int $limit Max number of posts or terms to return. Defaults to -1 (no limit).
-	 * @return array Array made of an array of post ids and an array of term ids
+	 * @return array {
+	 *     Objects without language.
+	 *
+	 *     @type int[] $posts Array of post ids.
+	 *     @type int[] $terms Array of term ids.
+	 * }
 	 */
 	public function get_objects_with_no_lang( $limit = -1 ) {
 		global $wpdb;
--- a/frontend/frontend-static-pages.php
+++ b/frontend/frontend-static-pages.php
@@ -17,9 +17,9 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 	protected $links_model;
 
 	/**
-	 * Instance of PLL_Links.
+	 * Instance of PLL_Frontend_Links.
 	 *
-	 * @var PLL_Links
+	 * @var PLL_Frontend_Links
 	 */
 	protected $links;
 
@@ -174,9 +174,9 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 	 *
 	 * @since 1.8
 	 *
-	 * @param bool|object $lang
-	 * @param object      $query
-	 * @return bool|object
+	 * @param PLL_Language|false $lang  The current language, false if it is not set yet.
+	 * @param WP_Query           $query The main WP query.
+	 * @return PLL_Language|false
 	 */
 	public function page_on_front_query( $lang, $query ) {
 		if ( ! empty( $lang ) || ! $this->page_on_front ) {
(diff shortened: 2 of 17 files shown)


admin/admin-model.php
```diff
@@ -257,15 +257,15 @@ public function update_language( $args ) {
...
+	 * Validates data entered when creating or updating a language.
 	 *
 	 * @see PLL_Admin_Model::add_language()
 	 *
 	 * @since 0.4
 	 *
-	 * @param array  $args
-	 * @param object $lang optional the language currently updated, the language is created if not set
-	 * @return bool true if success / false if failed
+	 * @param array  $args Parameters of {@see PLL_Admin_Model::add_language() or @see PLL_Admin_Model::update_language()}.
```
raaaahman, line 266: The problem is: if we change only one doc block between `PLL_Admin_Model::add_language()` and `PLL_Admin_Model::update_language()`, which one should we trust?

Since we have a data structure to define languages, couldn't we pass `PLL_Language` instances to those three methods?

admin/admin-model.php
```diff
@@ -257,15 +257,15 @@ public function update_language( $args ) {
...
+	 * @param object $lang Optional the language currently updated, the language is created if not set.
+	 * @return WP_Error
```
raaaahman, line 268: That's strange to have a method that always returns an error. Especially when it's a validation method...

frontend/frontend-static-pages.php
```diff
@@ -17,9 +17,9 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 	protected $links_model;
 
 	/**
-	 * Instance of PLL_Links.
+	 * Instance of PLL_Frontend_Links.
 	 *
-	 * @var PLL_Links
+	 * @var PLL_Frontend_Links
```
raaaahman, line 22: It's good to be more specific :+1:

frontend/frontend-static-pages.php
```diff
@@ -174,9 +174,9 @@ protected function is_front_page( $query ) {
 	 *
 	 * @since 1.8
 	 *
-	 * @param bool|object $lang
-	 * @param object      $query
-	 * @return bool|object
+	 * @param PLL_Language|false $lang  The current language, false if it is not set yet.
+	 * @param WP_Query           $query The main WP query.
+	 * @return PLL_Language|false
```
raaaahman, line 179: Very helpful . :+1:

include/api.php
```diff
@@ -412,11 +412,13 @@ function pll_count_posts( $lang, $args = array() ) {
...
- * Allows to access the Polylang instance
- * It is always preferable to use API functions
- * Internal methods may be changed without prior notice
+ * Allows to access the Polylang instance.
+ * However, it is always preferable to use API functions
+ * as internal methods may be changed without prior notice.
  *
  * @since 1.8
+ *
+ * @return PLL_Frontend|PLL_Admin|PLL_Settings|PLL_REST_Request
```
raaaahman, line 421: This will need to be updated if we ever add a new child class of `PLL_Base`. Alternatively, using `PLL_Base` as the return type will always be true, although a little less specific. Do you believe we'll ever add a new Polylang context / service locator?

include/switcher.php
```diff
@@ -109,32 +109,32 @@ protected function get_elements( $links, $args ) {
...
+	 *   @type int    $force_home             Will always link to home in translated language if set to 1, defaults to 0.
+	 *   @type int    $hide_if_no_translation Hides the link if there is no translation if set to 1, defaults to 0.
+	 *   @type int    $hide_current           Hides the current language if set to 1, defaults to 0.
+	 *   @type int    $post_id                Returns links to the translations of the post defined by post_id if set, defaults not set.
+	 *   @type int    $raw                    Return a raw array instead of html markup if set to 1, defaults to 0.
+	 *   @type string $item_spacing           Whether to preserve or discard whitespace between list items, valid options are 'preserve' and 'discard', defaults to 'preserve'.
+	 *   @type int    $admin_render           Allows to force the current language code in an admin context if set, default to 0. Need to set the admin_current_lang argument below.
+	 *   @type string $admin_current_lang     The current language code in an admin context. Need to set the admin_render to 1, defaults not set.
+	 * }
```
raaaahman, line 134: Nice. I hope it gets parsed correctly.

admin/admin-model.php
```diff
@@ -428,12 +428,12 @@ public function set_translation_in_mass( $type, $translations ) {
...
-	 * Returns untranslated posts and terms ids ( used in settings )
+	 * Returns untranslated posts and terms ids ( used in settings ).
 	 *
 	 * @since 0.9
-	 * @since 2.2.6 Add the $limit argument
+	 * @since 2.2.6 Add the $limit argument.
 	 *
-	 * @param in $limit Max number of posts or terms to return. Defaults to -1 (no limit).
+	 * @param int $limit Max number of posts or terms to return. Defaults to -1 (no limit).
 	 * @return array Array made of an array of post ids and an array of term ids
```
raaaahman, line 437: Could you use the same syntax as `PLL_Switcher::the_languages()`?

include/translate-option.php
```diff
@@ -31,21 +31,21 @@ class PLL_Translate_Option {
...
+	 * @param array  $keys Recursive array of option keys to translate in the form:
+	 *    array(
+	 *      'option_key_to_translate_1' => 1,
+	 *      'option_key_to_translate_2' => 1,
+	 *      'my_group' => array(
+	 *        'sub_key_to_translate_1' => 1,
+	 *        'sub_key_to_translate_2' => 1,
+	 *      ),
+	 *    )
+	 *    Note: only keys are interpreted. Any scalar can be used as values.
```
raaaahman, line 43: This is not consistent with the parameter below.

include/translated-object.php
```diff
@@ -387,11 +387,11 @@ public function current_user_can_synchronize( $id ) {
...
 		 *
-		 * @param $check Null to enable the capability check,
-		 *               true to always allow the synchronization,
-		 *               false to always disallow the synchronization.
-		 *               Defaults to true.
-		 * @param $id    The synchronization source object id.
+		 * @param bool|null $check Null to enable the capability check,
+		 *                         true to always allow the synchronization,
+		 *                         false to always disallow the synchronization.
+		 *                         Defaults to true.
```
raaaahman, line 393: This would be clearer with class constants: `ALWAYS`, `NEVER`, `IF_USER_HAS_CAPABILITY`.

include/translated-object.php
```diff
@@ -387,11 +387,11 @@ public function current_user_can_synchronize( $id ) {
...
+		 * @param itn       $id    The synchronization source object id.
```
raaaahman, line 394: Typo.

admin/admin-model.php
Chouby: We can't because this method is called *before* we create the language. `PLL_Language` handles languages already stored in DB.

frontend/frontend-static-pages.php
Chouby: Thanks to PHPStan ;-)

include/api.php
Chouby: I have no idea. Initially `PLL_REST_Request` did not exist. `PLL_Base` would not work as we often access methods which exist in only one of the context, so not in `PLL_Base`.

include/translate-option.php
Chouby: Could you clarify?

admin/admin-model.php
manooweb: Why ? Maybe the function name should be verify_lang and not validate_lang. 
For me it's exactly the same, it's add_language which has to control the result of this function.

include/translate-option.php
manooweb: I understand $args could contain a $context and a $sanitize_callback value. Am I right ?

Issue #591, opened 2020-10-14
Wrong language detection with Accept-Language: zh-Hant-HK


alvinhochun:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [ ] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] This bug happens with a default WordPress theme active
- [x] I can reproduce this bug consistently

## Steps to reproduce the issue

1. Add `en`, `zh` and `zh_HK` languages to WP site with `en` being the default.
2. Set active Windows 10 UI language to "Chinese (Hong Kong)".
3. Open the site on Edge (old).

## Expected behavior and actual behavior

Expected: It should redirect to the `zh_HK` homepage.

Actual: It redirects to the `zh` homepage instead.

## Additional information

The `Accept-Language` header sent by Edge is `zh-Hant-HK, zh-Hant; q=0.8, en-US; q=0.6, en; q=0.4, ja; q=0.2` (I have multiple UI languages installed). It is obvious that the site picked up on the `zh` part because it chose `zh` over `en`, but failed to pick up on the `HK` region tag possibly due to the presence of the `Hant` script tag.

P.S. This is on Win 10 version 1909 (build 18363.1082) and old Edge 44.18362.449.0. On my other system with Windows 10 version 1903 (build 18362.1082) and the same Edge verion, the `Accept-Language` header is simply `zh-HK` and so does not show the issue.

The detection behaviour can be more directly tested by using `curl` on a Linux terminal, Git Bash can also work (replace URL with test site):

```
$ curl http://example.com/ -s -D - -o /dev/null | grep -i location:
location: http://example.com/en/

$ curl -H "Accept-Language: zh-HK" http://example.com/ -s -D - -o /dev/null | grep -i location:
location: http://example.com/zh-hk/

$ curl -H "Accept-Language: zh-Hant-HK" http://example.com/ -s -D - -o /dev/null | grep -i location:
location: http://example.com/zh/
```

alvinhochun:
I should add that Firefox on Android does the same (it also uses `zh-Hant-HK`).

PR #687, opened 2020-12-22
Parsing HTTP Accept-Language Header

raaaahman:
## Goal

[IETF BCP 47 Section 2.1](https://tools.ietf.org/html/bcp47#section-2.1) define the different subtags that might exists in a language code.

At the moment, Polylang Browser Preferred Language feature [can only support any language that is composed of a `language` and an optional `region` subtag](https://github.com/polylang/polylang/blob/bd741dc86c23336865fd01dcc247b32e8b65db05/frontend/choose-lang.php#L133-L137). Any other language code composition will cause bugs in the feature, like reported in #591 

### Region inferring

[RFC 723 Section 5.3.5](https://tools.ietf.org/html/rfc7231#section-5.3.5) seems warns that:
  > User agents ought to provide guidance to users when setting
      a preference, since users are rarely familiar with the details of
      language matching as described above.  For example, users might
      assume that on selecting "en-gb", they will be served any kind of
      English document if British English is not available.

But in the actual implementation of the feature, Polylang already provides fallback for the language if the asked region does not exists on the server (e.g.: Polylang will fallback to a base `en` language if the requested `en_GB` is not registered).

What happens if a user then request `zh-hant-hk;q=1,zh-hant-cn;q=0.8;zh-cn;q=0.5` and the website has configured languages `zh-Hant-CN`, `zh-HK`, `zh-CN`? Should Polylang serve the closest script (`zh-Hant-CN`) or the closest region (`zh-HK`) ?

### Scripts in locale

Actually, WordPress do not use any script subtags in its locales. If Polylang was to include script subtags in its locale code, this might cause incompatibilities.

### Confusion with Polylang Pro's Local Fallback feature

If a user defines `zh-Hant-CN` as its locale fallback for `zh-Hant-HK`, but our algorithm decides to match `zh-HK` for the missing `zh-Hant-HK` content (or vice versa), the content and theme / plugins strings translations might not end up in the same language.

## Hints

- The [Negotiation library](https://github.com/willdurand/Negotiation/blob/master/src/Negotiation/AcceptLanguage.php) seems to be a popular and well maintained tool. However, it does not infer closest languages when they are not provided in HTTTP headers (meaning that `Accept-Languages: 'en-CA;q=1.0,en-US;q=0.8'` could not fallback to any `en` default, opposed to the actual Polylang behavior). It also parse script subtags, but does not take it into account when negotiating content.
- [PHP's Locale](https://www.php.net/manual/en/class.locale.php) seems to have useful utilities for the tasks. although it `Locale::acceptFromHttp()` method seems to have bugs with ... chinese scripts...

### DIFF
--- a/frontend/accept-language.php
+++ b/frontend/accept-language.php
@@ -10,6 +10,29 @@
  */
 class PLL_Accept_Language {
 	/**
+	 * @var string[]
+	 */
+	protected $subtags;
+
+	/**
+	 * @var float
+	 */
+	protected $quality;
+
+	/**
+	 * @var string[] Regular expression patterns.
+	 */
+	public static $subtag_patterns = array(
+		'language' => '([a-z]{2,3}|[a-z]{4}|[a-z]{5-8})\b',
+		'language-extension' => '(?:-([a-z]{3}){1,3}\b)?',
+		'script' => '(?:-([a-z]{4})\b)?',
+		'region' => '(?:-([a-z]{2}|[0-9]{3})\b)?',
+		'variant' => '(?:-([0-9][a-z]{1,3}|[a-z][a-z0-9]{4,7})\b)?',
+		'extension' => '(?:-([a-wy-z]-[a-z0-9]{2,8})\b)?',
+		'private-use' => '(?:-(x-[a-z0-9]{1,8})\b)?',
+	);
+
+	/**
 	 * Parse Accept-Language HTTP header according to IETF BCP 47.
 	 *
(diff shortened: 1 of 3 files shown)


raaaahman:
Note that it is not actually possible to register a locale as `zh_Hant_HK`, but we can register locale slugs like `zh-hant-hk` and have several languages with the same local.

Will this be enough to allow for content with different scripts?

raaaahman:
Replaced by #711

PR #688, opened 2020-12-23
Fix usage of sanitize_user()

Chouby:
`sanitize_user()` accepts only 2 parameters. See https://developer.wordpress.org/reference/functions/sanitize_user/.


manooweb (approved):
Ok 👍

PR #689, opened 2020-12-23
Fix documentation and undefined variables

Chouby:
This PR fixes several inline docs, undefined properties, undefined variables and a mispelled variable.


manooweb (approved):
Ok 👍

PR #690, opened 2020-12-23
Clarify the code of  `PLL_Language::get_flag_informations()`

Chouby:
PHPStan returns an error: Variable `$file` might not be defined in `PLL_Language::get_flag_informations()`.

This PR aims to clarify the code of this method in 3 successive steps:
1. Handle the default flag
2. Apply the filter
3. Define the flag src if missing

Whereas previously the first step was handled in 2 parts before and after applying the filter.

This change slightly impacts the function performance as the default flag is always base64 encoded, which wasn't the case previously when not using the default flag. As the function output is finally cached in a transient, I believe that this performance loss is acceptable, compared to the clarity gain.


manooweb (approved):
Ok. 👍

PR #691, opened 2020-12-23
Fix Variable $tr_parent in isset() always exists and is not nullable.

Chouby:
PHPStan reports this error as `PLL_Translated_Term::get_translation()` returns `false` and not `null` when the translation doesn't exist. In consequence, `isset()` is not the right way to test the result.


manooweb (approved):
Ok 👍

PR #695, opened 2020-12-24
Fix the visibility of PLL_Query properties

Chouby:
In #676, we declared the `PLL_Query` properties which were previously undeclared as protected. This now causes a fatal error in Polylang Pro. So this PR restores the visibility to public.

### DIFF
--- a/include/query.php
+++ b/include/query.php
@@ -12,12 +12,12 @@ class PLL_Query {
 	/**
 	 * @var PLL_Model
 	 */
-	protected $model;
+	public $model;
 
 	/**
 	 * @var WP_Query
 	 */
-	protected $query;
+	public $query;
 
 	/**
 	 * Constructor


include/query.php
```diff
@@ -12,12 +12,12 @@ class PLL_Query {
 	/**
 	 * @var PLL_Model
 	 */
-	protected $model;
+	public $model;
```
Marianne380, line 15: I'm not sure if this is useful for the `$model` property but okay for me.

Marianne380 (approved):

manooweb (approved):
Ok.

include/query.php
manooweb: > I'm not sure if this is useful for the `$model` property but okay for me.

I asked myself the same question beacause it seems that $model property isn't accessed externaly.

Chouby: I am not sure either. But I did not want to search. Maybe some day, we will take the time to look at all properties that are public and should not.

PR #700, opened 2020-12-31
Improve the doc for the api.

Chouby:
Everything is in the title.

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -4,30 +4,29 @@
  */
 
 /**
- * Template tag: displays the language switcher
- *
- * List of parameters accepted in $args:
- *
- * - dropdown               => displays a dropdown if set to 1, defaults to 0
- * - echo                   => echoes the switcher if set to 1 ( default )
- * - hide_if_empty          => hides languages with no posts ( or pages ) if set to 1 ( default )
- * - show_flags             => shows flags if set to 1, defaults to 0
- * - show_names             => shows languages names if set to 1 ( default )
- * - display_names_as       => whether to display the language name or its slug, valid options are 'slug' and 'name',
- * defaults to name
- * - force_home             => forces linking to the home page is set to 1, defaults to 0
- * - hide_if_no_translation => hides the link if there is no translation if set to 1, defaults to 0
- * - hide_current           => hides the current language if set to 1, defaults to 0
- * - post_id                => if not null, link to translations of post defined by post_id, defaults to null
- * - raw                    => set this to true to build your own custom language switcher, defaults to 0
- * - item_spacing           => whether to preserve or discard whitespace between list items, valid options are
- * 'preserve' and 'discard', defaults to preserve
+ * Template tag: displays the language switcher.
+ * The function does nothing if used outside the frontend.
  *
  * @api
  * @since 0.5
  *
- * @param array $args optional
- * @return null|string|array null if displaying, array if raw is requested, string otherwise
+ * @param array $args {
+ *   Optional array of arguments.
+ *
+ *   @type int    $dropdown               The list is displayed as dropdown if set to 1, defaults to 0.
+ *   @type int    $echo                   Echoes the list if set to 1, defaults to 1.
+ *   @type int    $hide_if_empty          Hides languages with no posts ( or pages ) if set to 1, defaults to 1.
+ *   @type int    $show_flags             Displays flags if set to 1, defaults to 0.
+ *   @type int    $show_names             Shows language names if set to 1, defaults to 1.
+ *   @type string $display_names_as       Whether to display the language name or its slug, valid options are 'slug' and 'name', defaults to name.
+ *   @type int    $force_home             Will always link to the homepage in the translated language if set to 1, defaults to 0.
+ *   @type int    $hide_if_no_translation Hides the link if there is no translation if set to 1, defaults to 0.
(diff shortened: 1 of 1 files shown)


include/api.php
```diff
@@ -38,14 +37,14 @@ function pll_the_languages( $args = array() ) {
...
- * Returns the language set in admin language filter on backend ( false if set to all languages )
+ * Returns the current language on frontend.
+ * Returns the language set in admin language filter on backend ( false if set to all languages ).
  *
  * @api
  * @since 0.8.1
  *
- * @param string $field Optional, the language field to return ( see PLL_Language ), defaults to 'slug', pass OBJECT constant to get the language object.
- * @return string|PLL_Language|bool The requested field for the current language
+ * @param string $field Optional, the language field to return ( @see PLL_Language ), defaults to 'slug'. Pass OBJECT constant to get the language object.
```
raaaahman, line 46: We should choose between using two of our own constants, or two strings.

include/api.php
```diff
@@ -122,15 +121,17 @@ function pll_home_url( $lang = '' ) {
...
  * @since 0.6
  *
- * @param string $name      a unique name for the string
- * @param string $string    the string to register
- * @param string $context   optional the group in which the string is registered, defaults to 'polylang'
- * @param bool   $multiline optional whether the string table should display a multiline textarea or a single line input, defaults to single line
+ * @param string $name      A unique name for the string.
+ * @param string $string    The string to register.
+ * @param string $context   Optional, the group in which the string is registered, defaults to 'polylang'.
+ * @param bool   $multiline Optional, true if the string table should display a multiline textarea,
```
raaaahman, line 132: Don't we use only multiline textareas now?

include/api.php
```diff
@@ -337,74 +344,88 @@ function pll_save_post_translations( $arr ) {
...
- * Returns an array of translations of a post
+ * Returns an array of translations of a post.
  *
  * @api
  * @since 1.8
  *
- * @param int $post_id
- * @return array an associative array of translations with language code as key and translation post_id as value
+ * @param int $post_id Post id.
+ * @return int[] An associative array of translations with language code as key and translation post id as value.
```
raaaahman, line 389: So it is not really an `int[]`, is it?

include/api.php
```diff
@@ -337,74 +344,88 @@ function pll_save_post_translations( $arr ) {
...
- * Returns an array of translations of a term
+ * Returns an array of translations of a term.
  *
  * @api
  * @since 1.8
  *
- * @param int $term_id
- * @return array an associative array of translations with language code as key and translation term_id as value
+ * @param int $term_id Term id.
+ * @return int[] An associative array of translations with language code as key and translation term id as value.
```
raaaahman, line 402: Same as above.

raaaahman (commented):
What does 'language code' refers to?

- A W3C locale?
- A WordPress locale?
- A slug?

include/api.php
Chouby: Values are only integers. This is waht the static analysis checks wehne accessing a value.

include/api.php
Chouby: In `PLL_Translated_Option` yes. But this function still offers the possibility to use a single line input.

include/api.php
Chouby: Sorry, I don't understand.

Chouby:
> What does 'language code' refers to?
> 
> * A W3C locale?
> * A WordPress locale?
> * A slug?

The "slug" is the word used by WordPress for taxonomy terms. I often used "language code" as used in the interface allowing to create or edit the languages.

include/api.php
```diff
@@ -77,40 +76,40 @@ function pll_default_language( $field = 'slug' ) {
...
+ * Among the post and its translations, returns the id of the post which is in the language represented by $slug.
  *
  * @api
  * @since 0.5
  *
- * @param int    $post_id post id
- * @param string $slug    optional language code, defaults to current language
- * @return int|false|null post id of the translation if exists, false otherwise, null if the current language is not defined yet
+ * @param int    $post_id Post id.
+ * @param string $slug    Optional language code, defaults to the current language.
```
raaaahman, line 85: `$slug` might get confused with the post's slug.

include/api.php
```diff
@@ -77,40 +76,40 @@ function pll_default_language( $field = 'slug' ) {
...
+ * Among the term and its translations, returns the id of the term which is in the language represented by $slug.
  *
  * @api
  * @since 0.5
  *
- * @param int    $term_id term id
- * @param string $slug    optional language code, defaults to current language
- * @return int|false|null term id of the translation if exists, false otherwise, null if the current language is not defined yet
+ * @param int    $term_id Term id.
+ * @param string $slug    Optional language code, defaults to the current language.
```
raaaahman, line 99: `$slug` might get confused with the term's slug.

raaaahman (changes requested):
> The "slug" is the word used by WordPress for taxonomy terms. I often used "language code" as used in the interface allowing to create or edit the languages.

Okay if its the same terminology than the user interface.

include/api.php
Chouby: This is in fact the language slug, used in the url, but you are right on the risk of confusion. I changed this.

raaaahman (approved):

PR #703, opened 2021-01-04
Stop preventing the usage of an untranslated front page

Chouby:
We use to prevent users to select a page on front if it is not translated in all languages to avoid support requests. This sometimes disturbs users an it is possible to work around this measure by temporarily deactivating Polylang. Since then, we have introduce the wizard which helps the  user to create the translated front pages and an admin notice which informs the user that the front pae is not translated in all languages.

I propose to remove the feature and let users make their choice.


manooweb (approved):
OK 👍
In fact I missed that there is an admin notice on the pages list screen 😁

PR #705, opened 2021-01-05
Add array keys in list to be properly displayed in doc.polylang.pro

sebastienserre:


Chouby:
Coul you rather use this syntax? https://github.com/polylang/polylang/blob/2.9.1/include/cookie.php#L49-L57

sebastienserre:
In fact my goal is to have an unordered list to avoid https://doc.polylang.pro/hook/pll_flag/
list-key

sebastienserre:
Yes we can !

Issue #349, opened 2019-03-30
PHPStan suggests these on Level 2


szepeviktor:
```
Note: Using configuration file /home/viktor/tmp/polylang/phpstan.neon.

 ------ -----------------------------------------------------------
  Line   admin/admin-base.php
 ------ -----------------------------------------------------------
  35     Instantiated class PLL_Pro not found.
  55     Access to an undefined property PLL_Admin_Base::$notices.
  56     Access to an undefined property PLL_Admin_Base::$links.
  70     Instantiated class PLL_Translate_Slugs_Model not found.
  71     Instantiated class PLL_Translate_Slugs not found.
 ------ -----------------------------------------------------------

 ------ --------------------------------------------------------------------------------
  Line   admin/admin-block-editor.php
 ------ --------------------------------------------------------------------------------
  17     Constructor of class PLL_Admin_Block_Editor has an unused parameter $polylang.
 ------ --------------------------------------------------------------------------------

 ------ -------------------------------------------------
  Line   admin/admin-filters-columns.php
 ------ -------------------------------------------------
  312    Cannot call method single_row() on bool|object.
  355    Cannot call method single_row() on bool|object.
 ------ -------------------------------------------------

 ------ ----------------------------------------------
  Line   admin/admin-filters-post-base.php
 ------ ----------------------------------------------
  42     Variable $translations might not be defined.
  43     Variable $translations might not be defined.
 ------ ----------------------------------------------

 ------ ------------------------------------------------
  Line   admin/admin-filters-term.php
 ------ ------------------------------------------------
  264    Variable $translations might not be defined.
  353    Cannot access property $assign_terms on array.
  353    Cannot access property $edit_terms on array.
  460    Cannot access property $edit_terms on array.
  613    Variable $translations might not be defined.
 ------ ------------------------------------------------

 ------ -----------------------------------------------------------------
  Line   admin/admin-filters.php
 ------ -----------------------------------------------------------------
  246    Function sanitize_user invoked with 3 parameters, 1-2 required.
 ------ -----------------------------------------------------------------

 ------ ----------------------------------------------
  Line   admin/admin-links.php
 ------ ----------------------------------------------
  151    Cannot access property $edit_terms on array.
 ------ ----------------------------------------------

 ------ --------------------------------------------------------------------------------------------------------
  Line   admin/admin-model.php
 ------ --------------------------------------------------------------------------------------------------------
  32     Cannot call method get_error_code() on bool.
[...]

PR #709, opened 2021-01-08
Raise PHPStan checks at level 3

Chouby:
This PR refactors `PLL_Frontend::switch_blog()` and its parent method to avoid checking in a hacky way if the parent method has done something. The initial test which was done only once in the parent method has now been moved to a new method `PLL_Base::is_active_on_new_blog()` which is called twice, once in parent method, once in the child method.

This PR also replaces several direct accesses to WordPress global variables by function calls when possible. For the global `$wp_query`, it is accessed through a new method to correctly document the type.

A return value is added to `PLL_Widget_Languages::form()` to match the parent method (even if this return value is not used).

The inline doc was improved in several places to pass the test.

The level is raised to 3 in the PHPStan config file for CI.

See #349

### DIFF
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -345,7 +345,7 @@ class PLL_Admin_Filters_Columns {
 			wp_die( 0 );
 		}
 
-		global $wp_list_table;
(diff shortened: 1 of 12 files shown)


manooweb (approved):
Ok. 👍

PR #710, opened 2021-01-11
Fix usage of wp_localize_script()

Chouby:
It is not documented that the 3rd parameter of `wp_localize_script()` accepts a string. 
This PR replaces it by an array in every places where we used a string, as per documetnation.
Additionnaly, it makes sure to always use same handle when enqueuing `admin.js` and always enqueue it in the footer.
It also removes the string `pll_flag_base_url` which seems unused.

### DIFF
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -345,8 +345,9 @@ class PLL_Wizard {
 		// Add ajax action on deactivate button in licenses step.
 		add_action( 'wp_ajax_pll_deactivate_license', array( $this, 'deactivate_license' ) );
 
-		wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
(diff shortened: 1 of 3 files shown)


manooweb (approved):
Ok 👍

modules/wizard/wizard.php
```diff
@@ -345,8 +345,9 @@ public function add_step_licenses( $steps ) {
 		// Add ajax action on deactivate button in licenses step.
 		add_action( 'wp_ajax_pll_deactivate_license', array( $this, 'deactivate_license' ) );
 
-		wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
-		wp_localize_script( 'pll_admin', 'pll_dismiss_notice', esc_html__( 'Dismiss this notice.', 'polylang' ) );
+		wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'wp-ajax-response', 'postbox', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
```
manooweb, line 348: Comment to link to this line

PR #711, opened 2021-01-11
Browser preferred language update

### LINKED ISSUE #591: Wrong language detection with Accept-Language: zh-Hant-HK

raaaahman:
## Before

Our Browser Preferred Language detection could only work with language and region information transmitted from WordPress visitors' browser agents, like 'en-GB' or 'zh-CN'.

It used to detect incomplete matches where either the 'language' or 'region' information matched one of the languages sent in the Accept-Language HTTP Header, given that the websites used **only one tag** in their slugs, like 'gb' or 'zh'.

Fixes #591

## Changes

- Updates 'Accept-Language' HTTP Header parsing to be compatible with [BCP 47](https://tools.ietf.org/html/bcp47#section-2.1)
- Allows our detection to match website's languages with close Accept-Languages values, including:
  - A 'language' only match
  - A 'language-region' match
  - A 'region' only match
  - A 'language-variant' match
 - Improve consistency of Preferred Language detection with custom slugs.

## Going further

Should we only match languages that are set as 'actives' when used in Polylang Pro?


Marianne380 (approved):

frontend/choose-lang.php
```diff
@@ -130,50 +138,14 @@ public function maybe_setcookie() {
...
+		$accept_langs = new PLL_Accept_Languages_Collection();
 
 		if ( isset( $_SERVER['HTTP_ACCEPT_LANGUAGE'] ) ) {
-			// Break up string into pieces ( languages and q factors )
-			preg_match_all(
-				'/([a-z]{1,8}(-[a-z]{1,8})?)\s*(;\s*q\s*=\s*((?>1|0)(?>\.[0-9]+)?))?/i',
-				sanitize_text_field( wp_unslash( $_SERVER['HTTP_ACCEPT_LANGUAGE'] ) ),
-				$lang_parse
-			);
+			$accept_langs = PLL_Accept_Languages_Collection::from_accept_language_header( sanitize_text_field( $_SERVER['HTTP_ACCEPT_LANGUAGE'] ) );
```
Chouby, line 144: Add `wp_unslash()`.

Chouby (changes requested):

alvinhochun:
Hi, thanks for making the changes to fix the issue. I believe our site currently has Polylang 3.0.6 installed, however which should include the changes here, but for some reason `zh-Hant-HK` in accept-language still redirects me to the `zh_CN` version of the site, any idea on how it could be happening?

This is our language settings:
圖片


```console
$ curl -H "Accept-Language: zh-Hant-HK" https://krita.org/ -s -D - -o /dev/null | grep -i location:
location: https://krita.org/zh/

$ curl -H "Accept-Language: zh-HK" https://krita.org/ -s -D - -o /dev/null | grep -i location:
location: https://krita.org/zh-hk/
```

raaaahman:
Indeed, the results seems to be dependent on the order in which the languages are queried from the database:

https://github.com/polylang/polylang/blob/ce90849b24b459b4d0589a215b2b2eae75cf900d/frontend/accept-languages-collection.php#L138

We should perform a 'weighted comparison' instead. Example: matching language is worth 1000, matching region 100, matching script 10, etc. And the highest value is returned.

PR #714, opened 2021-01-12
Translation's language should match object's language when saving translations group

raaaahman:
## Before

There is no verification of the object language when it is assigned to a translation group. This can lead to translations group having several time the same object with different languages like: 

```php
$polylang->model->post->save_translations( 1, array( 'en' => 1, 'fr' => 1, 'es' => 1 );
```

Although, this was verified when saving a post or media from the classic editor in `PLL_Admin_Filters_Post` and `PLL_Admin_Filters_Media`.

## Changes

Move the protection against incorrectly constituted translations group in the `PLL_Translated_Object::save_translations()` method to improve robustness.

## Related issues 

- https://github.com/polylang/polylang-pro/issues/808
- https://github.com/polylang/polylang-pro/pull/810

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -188,9 +188,9 @@ abstract class PLL_Translated_Object {
 	 * @since 0.5
 	 *
 	 * @param int   $id           Object id ( typically a post_id or term_id ).
-	 * @param array $translations An associative array of translations with language code as key and translation id as value.
+	 * @param int[] $translations An associative array of translations with language code as key and translation id as value.
 	 *
-	 * @return int[] An associative array with locales as key and post ids as values.
+	 * @return int[] An associative array with language codes as key and post ids as values.
 	 */
 	public function save_translations( $id, $translations ) {
 		$translations = $this->validate_translations( $translations );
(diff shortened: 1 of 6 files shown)


Marianne380 (approved):

include/translated-object.php
```diff
@@ -188,13 +188,16 @@ protected function should_update_translation_group( $id, $translations ) {
 	 * @since 0.5
 	 *
 	 * @param int   $id           Object id ( typically a post_id or term_id ).
-	 * @param int[] $translations An associative array of translations with language code as key and translation id as value.
-	 * @return void
+	 * @param array $translations An associative array of translations with language code as key and translation id as value.
```
Chouby, line 191: Why did you change this doc line?

include/translated-object.php
```diff
@@ -236,9 +239,29 @@ public function save_translations( $id, $translations ) {
...
+			return array();
 		}
 	}
 
+	/**
+	 * Returns translations after checking the translated post is in the right language
+	 *
+	 * @param array $translations Locales as keys and post ids as values.
+	 *
+	 * @return array
```
Chouby, line 253: Could you add `@since` and precise the array to `int[]` ? Also the keys are not locales but language codes.

include/translated-object.php
raaaahman: I did a mistake while resolving a merge conflict.

PR #715, opened 2021-01-12
Use the filter get_object_terms instead of wp_get_object_terms

Chouby:
Since WP 4.2, the filter `get_object_terms` is available as an alternative to `wp_get_object_terms` and is more convenient to use for us as it provides directly the array of taxonomies as 3rd parameter instead of a comma separated list of taxnomies that we need to explode.

I propose to take profit of this "new" filter and slighly simplify the code. The inline doc is updated accordingly.


manooweb (approved):
Ok 👍

PR #718, opened 2021-01-14
Improve inline doc

Chouby:
This PR provides more precision for some methods, replacing `object` by `PLL_Language`. It also fixes several doc issues reported by PHPStan at levels 4 to 6.


manooweb (approved):
Ok 👍

PR #720, opened 2021-01-18
Increase PHPStan checks to level 6

Chouby:
This PR fixes errors reported by PHPStan at level 6 and includes checks at this level in CI.

A few errors are ignored for now:

- Usage of `wp_die()` which doesn't officially accept `int` although WordPress itself often pass `0` or `-1` to this function as we do.
- The 2 errors in `admin/admin-base.php` should be removed once we will refactor script enqueueing as already planned (not fixed here to avoid conflicts with the ongoing work).
- Errors reported in `include/api.php` and `include/translated-object.php` are about checks that should be useless but which are protecting us from 3rd party code doing things wrong. I prefer to keep the situation as is for now.
- The error reported in `include/functions.php` about `wpcom_vip_get_page_by_path()` seems to be a false positive.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -134,7 +134,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 			$page = 'lang' === $tab ? 'mlang' : "mlang_$tab";
 			if ( empty( $parent ) ) {
 				$parent = $page;
-				add_menu_page( $title, __( 'Languages', 'polylang' ), 'manage_options', $page, null, 'dashicons-translation' );
+				add_menu_page( $title, __( 'Languages', 'polylang' ), 'manage_options', $page, '__return_null', 'dashicons-translation' );
 				$admin_page_hooks[ $page ] = 'languages'; // Hack to avoid the localization of the hook name. See: https://core.trac.wordpress.org/ticket/18857
 			}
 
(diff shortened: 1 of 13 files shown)


manooweb:
> * Errors reported in `include/api.php` and `include/translated-object.php` are about checks that should be useless but which are protecting us from 3rd party code doing things wrong. I prefer to keep the situation as is for now.

You talked about 

```
 ------ -------------------------------------------------------------------------------
  Line   include/api.php
 ------ -------------------------------------------------------------------------------
  152    Else branch is unreachable because ternary operator condition is always true.
 ------ -------------------------------------------------------------------------------

```
and

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------
  Line   include/translated-object.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------
  105    Call to function is_wp_error() with int will always evaluate to false.
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your phpstan.neon.dist.
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------

```

?
It could be interesting to explain why with a comment in the code because we make PHPstan ignore it.

manooweb (changes requested):
See my [comment](https://github.com/polylang/polylang/pull/720#issuecomment-762761774) otherwise it's ok 👍

PR #721, opened 2021-01-18
Protect against chained redirects.

manooweb:
After analysis about the issue reported in [#13419](https://secure.helpscout.net/conversation/1391585616/13419?folderId=927655) this PR propose to improve the strenghness of our PLL_Frontend_Filters_Links::check_canonical_url() method to protect against chained redirects as it is done in WordPress.

See https://github.com/WordPress/WordPress/blob/5.6/wp-includes/canonical.php#L761-L769

In fact:

- Polylang is hooked on `template_redirect` with a higher priority https://github.com/polylang/polylang/blob/2.9.1/frontend/frontend-filters-links.php#L73
 than WordPress `redirect_canonical()` function https://github.com/WordPress/WordPress/blob/5.6/wp-includes/default-filters.php#L538 to be able to catch wrong behaviours before WordPress canonical URL calculation and then redirect if necessary
- redirect_canonical() function should always return the same URL
- GN Publisher plugin is a new feed which is unknown from WordPress espacially here https://github.com/WordPress/WordPress/blob/5.6/wp-includes/canonical.php#L422-L432 and the `/feed/gn` is never removed from the URL
-  WordPress redirect_canonical() function then adds `/feed/gn` https://github.com/WordPress/WordPress/blob/5.6/wp-includes/canonical.php#L450 and https://github.com/WordPress/WordPress/blob/5.6/wp-includes/canonical.php#L510-L514 in the URL each time it is called, and then also here https://github.com/WordPress/WordPress/blob/5.6/wp-includes/canonical.php#L762. So WordPress never redirects to the canonical URL even if Polylang is deactivated.
- So Polylang redirects before WordPress which will never redirect because the URL changes everytime.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -464,9 +464,9 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 		$redirect_url = apply_filters( 'pll_check_canonical_url', $redirect_url, $language );
 
 		// The language is not correctly set so let's redirect to the correct url for this object
-		if ( $do_redirect && $redirect_url && $requested_url != $redirect_url ) {
+		if ( $do_redirect ) {
 			// Protect against chained redirects.
-			if ( $redirect_url === $this->check_canonical_url( $redirect_url, false ) ) {
+			if ( $redirect_url && $requested_url != $redirect_url && $redirect_url === $this->check_canonical_url( $redirect_url, false ) ) {
 				wp_safe_redirect( $redirect_url, 301, POLYLANG );
 				exit;
 			} else {
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -465,8 +465,13 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 
 		// The language is not correctly set so let's redirect to the correct url for this object
 		if ( $do_redirect && $redirect_url && $requested_url != $redirect_url ) {
-			wp_safe_redirect( $redirect_url, 301, POLYLANG );
-			exit;
+			// Protect against chained redirects.
+			if ( ! redirect_canonical( $redirect_url, false ) ) {
+				wp_safe_redirect( $redirect_url, 301, POLYLANG );
+				exit;
+			} else {
+				return;
+			}
 		}
 
 		return $redirect_url;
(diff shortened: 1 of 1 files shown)


raaaahman:
I don't get what you are trying to solve here. From WordPress' `redirect_canonical()` doc block:

> * Prevents redirection for feeds, trackbacks, searches, and
> * admin URLs.

If WordPress never redirects feeds, shall we?

Could you write down the cases you tested? Something like:

- requested url
- expected WordPress redirect
- WordPress + Polylang redirect
- WordPress + GN Publisher redirect
- WordPress + Polylang + GN Publisher redirect

raaaahman:
I don't get what you are trying to solve here. From WordPress' `redirect_canonical()` doc block:

> * Prevents redirection for feeds, trackbacks, searches, and
> * admin URLs.

If WordPress never redirects feeds, shall we?

Could you write down the cases you tested? Something like:

- requested url
- expected WordPress redirect
- WordPress + Polylang redirect
- WordPress + GN Publisher redirect
- WordPress + Polylang + GN Publisher redirect

raaaahman:
I don't get what you are trying to solve here. From WordPress' `redirect_canonical()` doc block:

> * Prevents redirection for feeds, trackbacks, searches, and
> * admin URLs.

If WordPress never redirects feeds, shall we?

Could you write down the cases you tested? Something like:

- requested url
- expected WordPress redirect
- WordPress + Polylang redirect
- WordPress + GN Publisher redirect
- WordPress + Polylang + GN Publisher redirect

raaaahman (approved):

frontend/frontend-filters-links.php
```diff
@@ -465,8 +465,13 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
 		if ( $do_redirect && $redirect_url && $requested_url != $redirect_url ) {
-			wp_safe_redirect( $redirect_url, 301, POLYLANG );
-			exit;
+			// Protect against chained redirects.
+			if ( $redirect_url === $this->check_canonical_url( $redirect_url, false ) ) {
+				wp_safe_redirect( $redirect_url, 301, POLYLANG );
+				exit;
+			} else {
+				return;
+			}
```
Chouby, line 474: The `return` at this position looks strange to me.

In the current situation, when `$do_redirect` is true, if we don't redirect, we always return the requested url (although the variable is named `$redirect_url`).
In your proposal, if we dont redirect, either we return nothing or we return the requested url, depending of the failing test. I would prefer only one case, for example, always return nothing:
```php
if ( $do_redirect ) {
	if ( $redirect_url && $requested_url != $redirect_url && $redirect_url === $this->check_canonical_url( $redirect_url, false ) ) {
		wp_safe_redirect( $redirect_url, 301, POLYLANG );
		exit;
	} else {
		return;
	}
}

```
and keep the return valiue only when `$do_redirect` is false. What do you think?

Chouby (changes requested):

frontend/frontend-filters-links.php
manooweb: Indeed closer than WordPress code. Done !

PR #723, opened 2021-01-19
Fix argument of an invalid type supplied for foreach

Chouby:
Some functions expecting to return an array may also return something else, for example if an error occurred. This PR adds checks before entering `foreach()` to improve the code robustness. Errors reported by PHPStan at level 7.


manooweb (approved):
Ok. 👍

PR #725, opened 2021-01-20
Use correct handle for admin script in Wizard's languages step

raaaahman:
## Before

The handles in the `wp_enqueue_script()` call didn't match the one returned in the `$step` array, resulting in the `admin.js` script not being loaded for the languages step page.

## Going further

Does the `wp_enqueue_script()` actually have any effect?

### DIFF
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -493,7 +493,7 @@ class PLL_Wizard {
 			'name'    => esc_html__( 'Languages', 'polylang' ),
 			'view'    => array( $this, 'display_step_languages' ),
 			'handler' => array( $this, 'save_step_languages' ),
-			'scripts' => array( 'pll-wizard-languages', 'pll-wizard-language-choice' ),
+			'scripts' => array( 'pll-wizard-languages', 'pll_admin' ),
 			'styles'  => array( 'pll-wizard-selectmenu' ),
 		);
 		return $steps;
--- a/modules/wizard/wizard.php
+++ b/modules/wizard/wizard.php
@@ -365,7 +365,7 @@ class PLL_Wizard {
 		// Add ajax action on deactivate button in licenses step.
 		add_action( 'wp_ajax_pll_deactivate_license', array( $this, 'deactivate_license' ) );
 
-		wp_enqueue_script( 'pll_admin', plugins_url( '/js/build/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'wp-ajax-response', 'postbox', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
+		wp_enqueue_script( 'pll_admin', plugins_url( '/js/build/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery' ), POLYLANG_VERSION, true );
 		wp_localize_script( 'pll_admin', 'pll_admin', array( 'dismiss_notice' => esc_html__( 'Dismiss this notice.', 'polylang' ) ) );
 
 		if ( $this->is_licenses_step_displayable() ) {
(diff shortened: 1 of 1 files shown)


manooweb:
Handle name has been changed in PR #710 and I missed it during the review.
Handle names are used in the HTML wizard page template here https://github.com/polylang/polylang/blob/master/modules/wizard/view-wizard-page.php#L36

manooweb:
Now I notice a javascript error too.
image

It's because we reintroduced in PR #710 dependency on `poxtbox` script which is useless because there is no metabox to manage in the wizard steps.
See https://github.com/polylang/polylang/pull/710/files#r561803332

**Note 1** that pagenow variable is not defined because the wizard doesn't use the WordPress `admin-header.php` file to build the HTML page but its own one.
See https://github.com/WordPress/WordPress/blob/5.6/wp-admin/admin-header.php#L86

**Note 2** we reintroduced dependency on `wp-ajax-response` script too which is also useless because not used in the jQuery UI selectmenu widget part. It is really used in the Polylang settings page.

So I removed the 2 depencencies too in this PR

modules/wizard/wizard.php
```diff
@@ -365,7 +365,7 @@ public function add_step_licenses( $steps ) {
 		// Add ajax action on deactivate button in licenses step.
 		add_action( 'wp_ajax_pll_deactivate_license', array( $this, 'deactivate_license' ) );
 
-		wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'wp-ajax-response', 'postbox', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
+		wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery' ), POLYLANG_VERSION, true );
```
raaaahman, line 368: It seems this script will always be dequeued at line 466, because both methods are always executed.

modules/wizard/wizard.php
```diff
@@ -652,14 +653,14 @@ public function save_step_media() {
 	 */
 	public function add_step_untranslated_contents( $steps ) {
 		if ( ! $this->model->get_languages_list() || $this->model->get_objects_with_no_lang( 1 ) ) {
-			wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'wp-ajax-response', 'postbox', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
+			wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
```
raaaahman, line 656: It seems this script will never be enqueued because a script with the same handle will always be enqueued at line 467.

raaaahman (commented):
It seems that some calls to enqueue have no effect.

Maybe we could:

- Only add the step corresponding to the current one, so we only enqueue the needed scripts
OR
- Enqueues all the scripts from the `PLL_Wizard::enqueue_scripts()` method, so we only enqueue each script once

modules/wizard/wizard.php
manooweb: Exactly! It isn't necessary because we are sure the Languages step does the job just before.
However we have to see each step independent of the others because one step could be not added to the wizard.

modules/wizard/wizard.php
manooweb: Yes see my comment in the PR description.
The second enqueue need a dependency which doesn't exist in the first one.
The only way to update a dependency on a script already enqueued is to dequeue it before.

I could put the first enqueue in the condition when we add the license step which will work with Polylang which doesn't have a license step but won't work with Polylang Pro without dequeuing.

I also could keep as before with the same dependency in the first enqueue however it's strange to say we have a dependency which isn't needed. It's mainly why I decided to do like that.

As we discussed, It's preferable in the future

manooweb (approved):
OK 👍

raaaahman (commented):
Okay.

PR #726, opened 2021-01-20

raaaahman:
## Before

- Only minified javascript files where built from sources. These didn't allow us to import a Js file into another, because the non-minfied version would be incorrect.
- Minified javascript and CSS files where built in the same folder than their sources, this forced us to ignore them in our VCS by their name suffix.

## Changes

Replaces #663 
Replaces #675 

- Javascript files ~~with the `.src` suffix~~ will be built in both minified and non-minified versions.
- ~~Javascript files without any suffix will be minified~~ Included in rule above.
- Javascript files ~~with the `.dep` suffix~~ in the `js/lib` folder will never be built nor minified. They are only meant to be imported into other files.
- Javascript and CSS files are now built inside specific build folders, respectively `js/build/` and `css/build/`
- CSS sources are copied as they are for non minified versions.
- Build scripts are exported into a specific npm package

## Going further

~~This method forces us to manually add the files built from source to the `.gitignore` list, as there are actually no differences between `.js` files that are built from sources and those that doesn't need to be built at all.~~ Fixed by https://github.com/polylang/polylang/pull/675/commits/2b7df6f8a19d82123b5a022bee0eb0eefc6d5d7f
- Will conflict with https://github.com/polylang/polylang/pull/724 (because both changes the `wp_enqueue_script()` and `wp_enqueue_style()` calls) but they can be merged either way, and the way to resolve the conflict will be the same.
- The dependency to a npm package might become unnecessary if the build folders are moved inside Polylang Pro: the context of the build will then become Polylang Pro root folder and the dependencies to external tools will then be resolved from inside it.
- I ran into trouble by instantiating `mini-css-extract-plugin` twice (once in Polylang Pro root folder and once in `vendor/wpsyntex/polylang`), so I needed to install the dependencies of the build scripts inside both Polylang and Polylang Pro packages instead of in the build script package itself. All-in-all, this external package for build scripts is more troublesome than it is helpful.


manooweb (changes requested):
As we discussed this morning, I'm ok on the principle.
However see the @wpsyntex/polylang-build-scripts package as a simple way to share common functions without any dependencies. The main advantage is that we just have to maintain these functions only once - in the package.
In fact dependencies is managed by the plugins themseleves (Polylang or Polylang Pro). So as the functions is used in the building process in each plugin, we doesn't need to make them depending on other packages.

manooweb (approved):
Ok 👍

modules/wizard/wizard.php
```diff
@@ -652,9 +652,9 @@ public function save_step_media() {
 	 */
 	public function add_step_untranslated_contents( $steps ) {
 		if ( ! $this->model->get_languages_list() || $this->model->get_objects_with_no_lang( 1 ) ) {
-			wp_enqueue_script( 'pll_admin', plugins_url( '/js/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'wp-ajax-response', 'postbox', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
-			wp_localize_script( 'pll_admin', 'pll_admin', array( 'dismiss_notice' => esc_html__( 'Dismiss this notice.', 'polylang' ) ) );
-			wp_enqueue_style( 'pll-wizard-selectmenu', plugins_url( '/css/selectmenu' . $this->get_suffix() . '.css', POLYLANG_FILE ), array( 'dashicons', 'install', 'common' ), POLYLANG_VERSION );
+			wp_enqueue_script( 'pll_admin', plugins_url( '/js/build/admin' . $this->get_suffix() . '.js', POLYLANG_FILE ), array( 'jquery', 'wp-ajax-response', 'postbox', 'jquery-ui-selectmenu' ), POLYLANG_VERSION, true );
+			wp_localize_script( 'pll_admin', 'pll_admin', array( 'pll_dismiss_notice' => esc_html__( 'Dismiss this notice.', 'polylang' ) ) );
```
Chouby, line 656: Isn't it a mistake to change this?  The js uses `pll_admin.dismiss_notice`.

Chouby (changes requested):

Chouby (approved):

PR #727, opened 2021-01-20
Improve inline doc

Chouby:
This PR improves the precision of the inline doc for objects. This fixes errors reported by PHPStan at level 7 (call to undefined methods and access to undefiend properties).

After this doc change I had to modify one line of code to convert a `WP_Term` to a `PLL_Language` to pass the current level of PHPStan tests. This is also included in this PR. See:  https://github.com/polylang/polylang/pull/727/files#diff-273f6e535c932726956ef771d69f9e76fdf21035b75a9f2b7b510aa055cce60fR165-R166. Everything else is only doc.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -401,11 +401,11 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	}
 
 	/**
-	 * Adds the languages list in admin bar for the admin languages filter
+	 * Adds the languages list in admin bar for the admin languages filter.
 	 *
 	 * @since 0.9
 	 *
-	 * @param object $wp_admin_bar
+	 * @param WP_Admin_Bar $wp_admin_bar WP_Admin_Bar global object.
 	 * @return void
 	 */
 	public function admin_bar_menu( $wp_admin_bar ) {
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -313,13 +313,13 @@ class PLL_Admin_Classic_Editor {
 	}
 
 	/**
-	 * Filters the pages by language in the parent dropdown list in the page attributes metabox
+	 * Filters the pages by language in the parent dropdown list in the page attributes metabox.
 	 *
 	 * @since 0.6
 	 *
-	 * @param array  $dropdown_args Arguments passed to wp_dropdown_pages
-	 * @param object $post
-	 * @return array Modified arguments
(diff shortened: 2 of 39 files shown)


Chouby:
Strange. Travis reports this error:
```
------ -------------------------------------------------------------- 
  Line   install/upgrade.php                                           
 ------ -------------------------------------------------------------- 
  142    Parameter #1 $lang of static method PLL_MO::get_id() expects  
         PLL_Language, WP_Term given.                                  
 ------ -------------------------------------------------------------- 
```
which is correct. But It's not reported on my locla computer.

manooweb:
> Strange. Travis reports this error:
> 
> ```
> ------ -------------------------------------------------------------- 
>   Line   install/upgrade.php                                           
>  ------ -------------------------------------------------------------- 
>   142    Parameter #1 $lang of static method PLL_MO::get_id() expects  
>          PLL_Language, WP_Term given.                                  
>  ------ -------------------------------------------------------------- 
> ```
> 
> which is correct. But It's not reported on my local computer.

Not reported on my local environments both ubuntu and git bash

Otherwise, as usual, I have these ones on git bash with `implode` function

```
 ------ ----------------------------------------------------------------------- 
  Line   frontend\frontend-auto-translate.php
 ------ -----------------------------------------------------------------------
  186    Parameter #2 $pieces of function implode expects array, string given.
 ------ -----------------------------------------------------------------------

 ------ -----------------------------------------------------------------------
  Line   include\model.php
 ------ -----------------------------------------------------------------------
  551    Parameter #2 $pieces of function implode expects array, string given.
 ------ -----------------------------------------------------------------------


```
On ubuntu no errors is reported 🤔

Chouby:
9f2fa80188db482951535230eb3a94924ff96bbc solves the issue raised by PHPStan in `PLL_Upgrade::upgrade_2_1()` avoiding a call to `PLL_MO::get_id()`. The idea is to rely only on WordPress functions for upgrades. Here the queries made by `PLL_MO::get_id()`, and `get_post()` in the loop, are replaced by a unique call to `get_posts()`. The code is not tested, but the risk is very low as the modification is simple and hopefully almost nobody use versions < 2.1.

+ Rebase and force push to fix a merge conflict.

Chouby:
For the errors with `implode()`, they look as false positive to me. I propose to keep things as is for now.

manooweb (approved):
Ok 👍

PR #728, opened 2021-01-22
More precision for objects inline doc

Chouby:
Only doc. However, this caused a new PHPStan error that I chose to ignore due to https://wordpress.org/support/topic/detect-browser-language-sometimes-setting-null-language


manooweb (approved):
Ok 👍

PR #729, opened 2021-01-25
Fix property access errors reported by PHPStan

Chouby:
Often our code doesn't check for errors. For example, when we can write:
```php
$term = get_term( $term_id );
$slug = $term->slug
```
So we suppose that we got a term, which is true in most cases, but may be wrong.

This PR adds several tests to protect our code from fatals when something went wrong.
Errors reported by PHPStan at level 7.


manooweb (approved):
Ok 👍
We will improve the robustness of the code.

PR #733, opened 2021-01-28
Hide license keys

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/812

- Hides the license keys by a using password input
- Links to account instead of checkout to renew existing license keys (not to show the license key in the url).


manooweb (approved):
Ok 👍

PR #735, opened 2021-01-29
Create default term class

Marianne380:
Moving everything related to the default categories into a class dedicated to that ( `PLL_Admin_Default_Term` )

Edit: As discussed with the developers, `add_hooks` is now a static method that instantiates a `PLL_Admin_Default_Term` class per taxonomy.


tests/phpunit/tests/test-sync.php
```diff
@@ -281,6 +281,7 @@ function test_save_post_with_sync() {
 
 		$this->assertEquals( 'fr', self::$model->post->get_language( $to )->slug );
 		$this->assertEqualSetsWithIndex( array( 'en' => $from, 'fr' => $to ), self::$model->post->get_translations( $from ) );
+		wp_remove_object_terms( $to, (int) get_option( 'default_category' ), 'category' );
```
raaaahman, line 284: Isn't it an issue if the default category is added on each synchronized post?

tests/phpunit/tests/test-default-term.php
```diff
@@ -0,0 +1,143 @@
...
+
+		$term_id = $this->factory->term->create( array( 'taxonomy' => 'category', 'name' => 'new-default' ) );
+		update_option( 'default_category', $term_id );
+
+		$this->assertEquals( $term_id, get_option( 'default_category' ) );
+		$translations = self::$model->term->get_translations( $term_id );
+		$this->assertEqualSets( array( 'en', 'fr', 'de', 'es' ), array_keys( $translations ) );
+	}
+
+	function test_notice_for_objects_with_no_lang() {
```
raaaahman, line 105: I don't believe we are testing the default category here. It would be more robust to create a new category and test with it.

Then if we need to test a **special case** with the default category, we can add a new test for it.

tests/phpunit/tests/test-sync.php
```diff
@@ -263,8 +271,10 @@ function test_save_post_with_sync() {
 		$key = add_post_meta( $from, 'key', 'value' );
 		$metas[ $key ] = array( 'key' => 'key', 'value' => 'value' );
 
+		$this->pll_admin->default_term = new PLL_Admin_Default_Term( $this->pll_admin );
```
raaaahman, line 274: This is already done earlier in the test. You are setting the filters and action a second time here.

tests/phpunit/tests/test-sync.php
```diff
@@ -281,7 +291,7 @@ function test_save_post_with_sync() {
 
 		$this->assertEquals( 'fr', self::$model->post->get_language( $to )->slug );
 		$this->assertEqualSetsWithIndex( array( 'en' => $from, 'fr' => $to ), self::$model->post->get_translations( $from ) );
-		wp_remove_object_terms( $to, (int) get_option( 'default_category' ), 'category' );
+//		wp_remove_object_terms( $to, (int) get_option( 'default_category' ), 'category' );
```
raaaahman, line 294: Commented out code.

admin/admin-default-term.php
```diff
@@ -45,11 +45,12 @@ public function __construct( &$polylang ) {
 		foreach ( $taxonomies as $taxonomy ) {
 			if ( 'category' === $taxonomy ) {
 				// Allows to get the default categories in all languages
-				add_filter( 'option_default_category', array( $this, 'option_default_category' ) );
-				add_action( 'update_option_default_category', array( $this, 'update_option_default_category' ), 10, 2 );
+				add_filter( 'default_option_' . $taxonomy, array( $this, 'option_default_category' ) );
+				add_action( 'update_option_default_' . $taxonomy, array( $this, 'update_option_default_category' ), 10, 2 );
```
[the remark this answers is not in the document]
manooweb: @raaaahman 🤔 IMO I think it is for the following. In fact to solve this issue https://github.com/polylang/polylang-pro/issues/625 i.e. default term on all taxonomies

tests/phpunit/tests/test-sync.php
Marianne380: Oops, I hadn't noticed.

admin/admin-filters-term.php
```diff
@@ -85,10 +95,6 @@ public function __construct( &$polylang ) {
 		add_action( 'wp_ajax_term_lang_choice', array( $this, 'term_lang_choice' ) );
 		add_action( 'wp_ajax_pll_terms_not_translated', array( $this, 'ajax_terms_not_translated' ) );
 
-		// Allows to get the default categories in all languages
-		add_filter( 'option_default_category', array( $this, 'option_default_category' ) );
```
manooweb, line 89: Just to make a link to this old line of code

manooweb (changes requested):

admin/admin-default-term.php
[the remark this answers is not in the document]
manooweb: Yes! See the old code here
https://github.com/polylang/polylang/pull/735#discussion_r585662910 the name of the filter @line 48 is wrong

raaaahman (approved):

admin/admin-default-term.php
```diff
@@ -0,0 +1,242 @@
...
+
+			// Adds the language column in the 'Post Tags'table.
+			if ( 'post_tag' === $taxonomy ) {
+				add_filter( 'manage_' . $taxonomy . '_custom_column', array( $this, 'term_column' ), 10, 3 );
+			}
+		}
+
+		// Prevents deleting all the translations of the default category
+		add_filter( 'map_meta_cap', array( $this, 'fix_delete_default_category' ), 10, 4 );
+	}
```
raaaahman, line 67: Hook action and filters in an `add_hooks()` method.

tests/phpunit/tests/test-settings.php
```diff
@@ -62,6 +62,13 @@ function test_edit_language() {
 	}
 
 	function test_notice_for_objects_with_no_lang() {
+		$links_model = self::$model->get_links_model();
+		$this->pll_admin = new PLL_Admin( $links_model );
+		self::delete_all_languages();
+		$this->pll_admin->default_term = new PLL_Admin_Default_Term( $this->pll_admin );
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR' );
```
raaaahman, line 70: Not needed with previously mentioned changes.

tests/phpunit/tests/test-sync.php
```diff
@@ -235,7 +238,12 @@ function test_create_page_translation() {
 		$this->assertEquals( 'full-width.php', get_page_template_slug( $to ) );
 	}
 
-	function test_save_post_with_sync() {
+	function test_save_post_with_syncd() {
+		self::delete_all_languages();
+		$this->pll_admin->default_term = new PLL_Admin_Default_Term( $this->pll_admin );
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR' );
+
```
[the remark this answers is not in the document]
raaaahman: Will be gone with modification of the `PLL_UnitTestCase_Trait` as well.

raaaahman (changes requested):
Use the `PLL_Admin_Default_Term::handle_default_category_on_create_language()` in the `PLL_UnitTestCase_Trait::create_language()` to assign the language to the default category.

tests/phpunit/includes/testcase-trait.php
```diff
@@ -61,6 +61,12 @@ static function create_language( $locale, $args = array() ) {
 		$values['term_group'] = 0; // Default term_group.
 
 		$args = array_merge( $values, $args );
+
+		$links_model     = self::$model->get_links_model();
+		$pll_admin = new PLL_Admin( $links_model );
+		$admin_default_term = new PLL_Admin_Default_Term( $pll_admin );
+		add_action( 'pll_add_language', array( $admin_default_term, 'handle_default_category_on_create_language' ) );
```
raaaahman, line 68: The goal to put the actions in a dedicated function was not to hook it in this method.

Here's what happens with this code:

- Each time we call `PLL_UnitTestCase_Trait::create_language()` in a `wpSetUpBeforeClass()` method, we are going to register one new 'pll_add_language' action per call.
- Before each test, `WP_UnitTestCase_Base::setUp()`will [back up these actions](https://wordpress.org/plugins/unagi/).
- If we call `PLL_Admin_Default_Term::add_hooks()` during the test ([like it is done here](https://github.com/polylang/polylang/blob/306ae6240e8dcc63a3202d04f92c9169394612a0/tests/phpunit/tests/test-default-term.php#L54-L55)), we will add one more 'pll_add_language' action.
- If we call `PLL_UnitTestCase_Trait::create_language()` during the test ([like it is done here](https://github.com/polylang/polylang/blob/306ae6240e8dcc63a3202d04f92c9169394612a0/tests/phpunit/tests/test-default-term.php#L57-L60)), we will add one more 'pll_add_language' action per call.
- After the test, `WP_UnitTestCase_Base::tearDown()` will replace all the hooks registered during the tests by the hooks backed up before the test: https://github.com/WordPress/wordpress-develop/blob/3d8a5fd839d3eb1642a38f25e4d2901b009371e2/tests/phpunit/includes/abstract-testcase.php#L168 (this means that it keeps the 'pll_add_language' action we hooked during `wpSetUpBeforeClass()`)

So we will end up with too duplicated hooks in every scenario. Couldn't you call directly the `handle_default_term_on_create_language()` method with the correct argument directly?

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+
+	/**
+	 * Setups filters and actions needed.
+	 *
+	 * @return void
+	 */
+	public function add_hooks() {
+		$taxonomies = get_taxonomies();
+		foreach ( $taxonomies as $taxonomy ) {
+			if ( 'category' === $taxonomy ) {
```
Chouby, line 53: This is the only line where I should read `category`. Everywhere else (function names, filters, actions, options), you need to remove the reference to the category to mak this class fully generic.

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * Manages filters and actions related to default terms.
+ *
+ * @since 3.0
```
Chouby, line 9: Change to `@since 3.1`.

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+ */
+class PLL_Admin_Default_Term {
+
+	/**
+	 * A reference to the PLL_Model instance.
+	 *
+	 * @since 2.8
```
Chouby, line 16: We generally don't have `@since` for variables (and this one is wrong).

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+	 * Preferred language to assign to new contents.
+	 *
+	 * @var PLL_Language
+	 */
+	protected $pref_lang;
+
+	/**
+	 * Reference to Polylang options array
+	 *
+	 * @var array $options
```
Chouby, line 32: Remove `$options` in comment. This is redundant.

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+	 */
+	protected $options;
+
+	/**
+	 * @param object $polylang
```
Chouby, line 37: Please add comments in all your methods (description, param short description, `@since`).

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+		$this->model = &$polylang->model;
+		$this->pref_lang = &$polylang->pref_lang;
+		$this->options = &$polylang->options;
+	}
+
+	/**
+	 * Setups filters and actions needed.
+	 *
+	 * @return void
+	 */
```
Chouby, line 49: Please add missing `@since` for your new functions.

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+				add_action( 'pll_update_default_lang', array( $this, 'update_default_category_language' ) );
+
+				// Adds the language column in the 'Categories' table.
+				add_filter( 'manage_' . $taxonomy . '_custom_column', array( $this, 'term_column' ), 10, 3 );
+			}
+
+			// Adds the language column in the 'Post Tags'table.
+			if ( 'post_tag' === $taxonomy ) {
+				add_filter( 'manage_' . $taxonomy . '_custom_column', array( $this, 'term_column' ), 10, 3 );
+			}
```
Chouby, line 69: Why would you need this? This taxonomy has no default term.

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+		// Prevents deleting all the translations of the default category
+		add_filter( 'map_meta_cap', array( $this, 'fix_delete_default_category' ), 10, 4 );
+	}
+
+	/**
+	 * Filters the default category in note below the category list table and in settings->writing dropdown
+	 *
+	 * @since 1.2
+	 *
+	 * @param int $value
```
Chouby, line 81: Ad short descriptions to all the function parameters.

admin/admin-default-term.php
```diff
@@ -0,0 +1,249 @@
...
+
+	/**
+	 * Checks if the new default category is translated in all languages
+	 * If not, create the translations
+	 *
+	 * @since 1.7
+	 *
+	 * @param int $old_value
+	 * @param int $value
+	 *
```
Chouby, line 99: No empty line between `@param` and `@return`.

Chouby (changes requested):

admin/admin-default-term.php
raaaahman: What about we don't force ourselves on documenting everything even we have no good information to add?

Let's imagine we rewrite this one like the following:

```php
 * @param int $value An integer value.
```

What would be the benefits?

admin/admin-default-term.php
raaaahman: Could we add it as a sniff in PHPCS ?

admin/admin-default-term.php
manooweb: > What about we don't force ourselves on documenting everything even we have no good information to add?
> 
> Let's imagine we rewrite this one like the following:
> 
> ```
>  * @param int $value An integer value.
> ```
> 
> What would be the benefits?

Obviously said like that it doesn't make sense 😁
However as the `option_{$option}` filter could pass a $value `mixed` variable
See https://developer.wordpress.org/reference/hooks/option_option/
 it could be interesting to precise in the case of taxonomy that this integer value corresponds on the taxonomy term id for the option `default_{$taxonomy}`

raaaahman: Indeed, the fact that the parameter is a taxonomy term id is not obvious in this case. So what about renaming the variable to a more fitting name like `$taxonomy_term_id` ?

manooweb: It could 👍 What is the opinion of each other ? 
Preferable to document by naming variable or by commenting or both 🙂 ?

Marianne380: We might as well do both.

admin/admin-default-term.php
```diff
@@ -0,0 +1,268 @@
...
+
+	/**
+	 * Setups filters and actions needed.
+	 *
+	 * @since 3.1
+	 *
+	 * @return void
+	 */
+	public function add_hooks() {
+		$taxonomies = get_taxonomies();
```
Marianne380, line 55: Should we store the registered taxonomies in a class property or call this method each time we need it ?

tests/phpunit/tests/test-default-term.php
```diff
@@ -0,0 +1,143 @@
...
+		ob_start();
+		include ABSPATH . 'wp-admin/edit-tag-form.php';
+
+		return ob_get_clean();
+	}
+
+	function test_default_category_in_edit_tags() {
+		$this->pll_admin->links = new PLL_Admin_Links( $this->pll_admin );
+
+
```
manooweb, line 56: One useless blankk line

manooweb (changes requested):

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -437,18 +437,20 @@ function test_categories_script_data_in_footer() {
...
 
-		// All categories we have i.e. default categories
-		foreach ( self::$model->get_languages_list() as $lang ) {
-			$terms[ $lang->slug ] = array( 'category' => array( self::$model->term->get( 1, $lang->slug ) ) );
-		}
+		$terms['fr'] = array( 'category' => array( self::$model->term->get( $term_id, 'fr' ) ) );
 
-		$this->assertNotFalse( strpos( $footer, 'var pll_term_languages = ' . wp_json_encode( $terms ) ) );
+		// Deletion of the last three characters of the string we are looking for, because we also have the default category ( {"fr":{"category":[term_fr]}} )
+		$this->assertNotFalse( strpos( $footer, substr( 'var pll_term_languages = ' . wp_json_encode( $terms ), 0, -3 ) ) );
```
manooweb, line 453: Maybe rewrite what we test in the assertion. It isn't clear now we have to test all the javascript variable and it strange to me to truncate the returned value.

manooweb (changes requested):

manooweb (approved):
OK 👍

PR #736, opened 2021-02-01
Legacy widgets filters

raaaahman:
WIP fixes https://github.com/polylang/polylang-pro/issues/836

To do:

- Instantiate just in time, in hope this would allow us not to override an existing instance.
- Shouldn't we check for not being in a Page Builder on the front-end, but in a different manner than how it is done in the admin context?

### DIFF
--- a/admin/widgets-options-filters-admin.php
+++ b/admin/widgets-options-filters-admin.php
@@ -5,10 +5,10 @@
 
 /**
  * Class PLL_Widgets_Filters
+ *
+ * Adds new options to {@see https://developer.wordpress.org/reference/classes/wp_widget/ WP_Widget} and saves them.
  */
-class PLL_Widgets_Filters_Admin extends PLL_Widgets_Filters {
-
-
+class PLL_Widgets_Options_Filters_Admin extends PLL_Widgets_Options_Filters {
 	/**
 	 * Modifies the widgets forms to add our language dropdown list.
 	 *
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -36,6 +36,11 @@ class PLL_REST_Request extends PLL_Base {
 	public $static_pages;
 
 	/**
+	 * @var PLL_Widgets_Filters
+	 */
+	public $widgets_filters;
+
+	/**
 	 * Setup filters.
 	 *
 	 * @since 2.6
@@ -52,6 +57,7 @@ class PLL_REST_Request extends PLL_Base {
 
 			$this->filters_links = new PLL_Filters_Links( $this );
 			$this->filters = new PLL_Filters( $this );
+			$this->widgets_filters = new PLL_Widgets_Filters( $this );
 
 			// Static front page and page for posts.
 			if ( 'page' === get_option( 'show_on_front' ) ) {
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -36,7 +36,7 @@ class PLL_REST_Request extends PLL_Base {
 	public $static_pages;
 
 	/**
-	 * @var PLL_Widgets_Options_Filters
+	 * @var PLL_Filters_Widgets_Options
 	 */
 	public $widgets_filters;
 
@@ -57,7 +57,7 @@ class PLL_REST_Request extends PLL_Base {
 
 			$this->filters_links = new PLL_Filters_Links( $this );
 			$this->filters = new PLL_Filters( $this );
-			$this->widgets_filters = new PLL_Widgets_Options_Filters( $this );
+			$this->widgets_filters = new PLL_Filters_Widgets_Options( $this );
 
 			// Static front page and page for posts.
 			if ( 'page' === get_option( 'show_on_front' ) ) {
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -36,7 +36,7 @@ class PLL_REST_Request extends PLL_Base {
 	public $static_pages;
 
 	/**
-	 * @var PLL_Widgets_Filters
+	 * @var PLL_Widgets_Options_Filters
 	 */
 	public $widgets_filters;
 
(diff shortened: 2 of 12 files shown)


admin/widgets-options-filters-admin.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @package Polylang
+ */
+
+/**
+ * Class PLL_Widgets_Filters
+ *
+ * Adds new options to {@see https://developer.wordpress.org/reference/classes/wp_widget/ WP_Widget} and saves them.
+ */
+class PLL_Widgets_Options_Filters_Admin extends PLL_Widgets_Options_Filters {
```
manooweb, line 11: Shouldn't we name it PLL_Admin_Widgets_Options_Filters ? To respect how we name all the other class

admin/widgets-options-filters-admin.php
```diff
@@ -0,0 +1,51 @@
...
+	 * @param array     $instance The current Widget's options.
+	 * @param array     $new_instance The new Widget's options.
+	 * @param array     $old_instance Not used.
+	 * @param WP_Widget $widget The Widget object.
+	 * @return array The processed Widget options.
+	 * @since 3.0
+	 */
+	public function widget_update_callback( $instance, $new_instance, $old_instance, $widget ) {
+		$key = $this->get_language_key( $widget );
+		if ( ! empty( $_POST[ $key ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
manooweb, line 45: 👍 Note in the context of the legacy widget screen, it seems that the $new_instance doesn't contains the $key element - name of the languages field. Its for that we have to retrieve the value for the $_POST global variable.

manooweb (changes requested):
Just rename PLL_Widgets_Options_Filters_Admin maybe ?

**/!\ Note** that in the new widget screen when add a legacy widget the first time, languages field - if modified - isn't save correctly due to the issue on widget form fields reported in https://github.com/WordPress/gutenberg/issues/28668

We need to refresh the new widget screen, then change the languages and update again to really take the language in account.

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
...
+	 * Called when widget options are saved.
+	 * Saves the language associated to the widget.
+	 *
+	 * @param array     $instance The current Widget's options.
+	 * @param array     $new_instance The new Widget's options.
+	 * @param array     $old_instance Not used.
+	 * @param WP_Widget $widget WP_Widget object.
+	 * @return array Widget options.
+	 * @since 0.3
+	 * @since 3.0 Moved from PLL_Admin_Filters
```
Chouby, line 75: Same as above for `@since`.

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		return $instance;
+	}
+
+	/**
+	 * Returns the key used by Polylang to pass language data.
+	 *
+	 * @param WP_Widget $widget
+	 * @return string
```
Chouby, line 93: Please add `@since` and description for the param.

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * Class PLL_Widgets_Filters
+ *
+ * Add new options to {@see https://developer.wordpress.org/reference/classes/wp_widget/ WP_Widget} and saves them.
+ */
```
Chouby, line 10: Please add `@since`

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	/**
+	 * @param WP_Widget $widget
+	 * @param null      $return
+	 * @param array     $instance
+	 * @return void
+	 *
+	 * @since 3.0 Moved PLL_Admin_Filters
+	 */
```
Chouby, line 40: Same as above for `@since`. Also please add the removed short description for parameters.

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @package Polylang
+ */
+
+/**
+ * Class PLL_Widgets_Filters
+ *
+ * Add new options to {@see https://developer.wordpress.org/reference/classes/wp_widget/ WP_Widget} and saves them.
+ */
+class PLL_Widgets_Options_Filters {
```
Chouby, line 11: For consistency with other class names, I suggest to use `PLL_Filters_Widgets_Options`.

include/rest-request.php
```diff
@@ -35,6 +35,11 @@ class PLL_REST_Request extends PLL_Base {
 	 */
 	public $static_pages;
 
+	/**
+	 * @var PLL_Widgets_Options_Filters
+	 */
+	public $widgets_filters;
```
Chouby, line 41: Same as above for the property name.

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * Class PLL_Widgets_Filters
```
Chouby, line 7: Not the right class name.

include/widgets-options-filters.php
```diff
@@ -0,0 +1,98 @@
...
+ */
+class PLL_Widgets_Options_Filters {
+
+	/**
+	 * @var PLL_Model
+	 */
+	public $model;
+
+	/**
+	 * PLL_Widgets_Filters constructor.
```
Chouby, line 19: Not the right class name.

Chouby (changes requested):

manooweb (approved):

PR #737, opened 2021-02-03
Test widgets filter is not displayed for page builders

raaaahman:
## Goal

Some page builders, like SiteOrigin Page Builder, might use the `post` screen to edit the page. This test verifies that we do not output our language filters for widgets in this unrelated screen.

Edit based on #736


manooweb (approved):
Ok 👍

PR #738, opened 2021-02-03
Fix wrong action url in search forms

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/829

The `PLL_Frontend_Filters_Search::get_search_form()` method used to replace the home url without language code by the search url in the current language, for example  `https://mysite.com`, replaced by  `https://mysite.com/fr/`.

If the filter `get_search_form` is applied directly by a theme, the url is already filtered by `PLL_Frontend_Filters_Links::home_url()`. Thus the search form passed to `PLL_Frontend_Filters_Search::get_search_form()` may contain a url looking like  `https://mysite.com/fr/accueil/` which was transformed to the wrong url  `https://mysite.com/fr/fr/accueil`.

This PR replaces any url in the action attribute instead of looking for the home url.

### DIFF
--- a/frontend/frontend-filters-search.php
+++ b/frontend/frontend-filters-search.php
@@ -49,13 +49,14 @@ class PLL_Frontend_Filters_Search {
 	}
 
 	/**
-	 * Adds the language information in the search form
+	 * Adds the language information in the search form.
+	 *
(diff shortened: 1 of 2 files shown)


manooweb (approved):
Ok 👍

PR #739, opened 2021-02-04
Use two classes for switcher

Marianne380:
Use of two classes to manage the Switcher, one in Frontend and the other in Admin.

See : https://github.com/polylang/polylang-pro/pull/851

Fix https://github.com/polylang/polylang-pro/issues/835

### DIFF
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -115,7 +115,7 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 				/** This filter is documented in include/switcher.php */
 				$options = apply_filters( 'pll_the_languages_args', $options ); // Honor the filter here for 'show_flags', 'show_names' and 'dropdown'.
 
-				$switcher = PLL_Switcher::create( $this->polylang );
+				$switcher = new PLL_Frontend_Switcher();
 				$args = array_merge( array( 'raw' => 1 ), $options );
 
 				/** @var array */
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -17,6 +17,13 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 	public $curlang;
 
 	/**
+	 * Reference to the Polylang object.
+	 *
+	 * @var object
+	 */
+	protected $polylang;
+
+	/**
 	 * Constructor
 	 *
 	 * @since 1.2
@@ -26,6 +33,7 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 	public function __construct( &$polylang ) {
 		parent::__construct( $polylang );
 
+		$this->polylang = &$polylang;
 		$this->curlang = &$polylang->curlang;
 
 		// Split the language switcher menu item in several language menu items
@@ -107,7 +115,7 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 				/** This filter is documented in include/switcher.php */
 				$options = apply_filters( 'pll_the_languages_args', $options ); // Honor the filter here for 'show_flags', 'show_names' and 'dropdown'.
 
-				$switcher = new PLL_Switcher();
+				$switcher = PLL_Switcher::create( $this->polylang );
 				$args = array_merge( array( 'raw' => 1 ), $options );
 
 				/** @var array */
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -17,13 +17,6 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 	public $curlang;
 
 	/**
-	 * Reference to the Polylang object.
-	 *
-	 * @var object
-	 */
-	protected $polylang;
-
-	/**
 	 * Constructor
 	 *
 	 * @since 1.2
@@ -33,7 +26,6 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 	public function __construct( &$polylang ) {
 		parent::__construct( $polylang );
 
-		$this->polylang = &$polylang;
 		$this->curlang = &$polylang->curlang;
 
 		// Split the language switcher menu item in several language menu items
--- a/include/admin-switcher.php
+++ b/include/admin-switcher.php
@@ -30,21 +30,7 @@ class PLL_Admin_Switcher extends PLL_Switcher {
 	 * @return string|array either the html markup of the switcher or the raw elements to build a custom language switcher
 	 */
 	public function the_languages( $links, $args = array() ) {
-		$defaults = $this->get_default_the_languages();
-
-		$args = $this->filter_arguments_pll_languages( $args, $defaults );
-
-		$args['hide_if_empty'] = 0;
-		// Force not to hide the language for the block preview even if the option is checked.
-		$args['hide_if_no_translation'] = 0;
-
-		$elements = $this->get_elements( $links, $args );
-
-		if ( $args['raw'] ) {
-			return $elements;
-		}
-
-		list( $out, $args ) = $this->prepare_pll_walker( $this->get_current_language( $links ), $args, $elements );
+		list( $out, $args ) = $this->get_the_languages( $links, $args );
 
 		if ( $args['echo'] ) {
 			echo $out; // phpcs:ignore WordPress.Security.EscapeOutput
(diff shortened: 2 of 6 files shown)


include/admin-switcher.php
```diff
@@ -0,0 +1,88 @@
...
+	/**
+	 * Get the language elements for use in a walker
+	 *
+	 * @since 1.2
+	 *
+	 * @param PLL_Frontend_Links $links Instance of PLL_Frontend_Links.
+	 * @param array              $args  Arguments passed to {@see PLL_Switcher::the_languages()}.
+	 * @return array Language switcher elements.
+	 */
+	protected function get_elements( $links, $args ) {
```
raaaahman, line 15: Much of `PLL_fronted_Switcher::get_elements()` and `PLL_Admin_Switcher::get_elements()` is quite similar, we could share this method from the parent class. See proposed changes below.

include/admin-switcher.php
```diff
@@ -0,0 +1,88 @@
...
+		$first = true;
+		$out   = array();
+
+		foreach ( $links->model->get_languages_list( array( 'hide_empty' => $args['hide_if_empty'] ) ) as $language ) {
```
raaaahman, line 19: There is two cases here:

- `PLL_Admin_Switcher` will never hide an empty language (see line 69).
- `PLL_Frontend_Switcher` would also like to exclude language for posts with no translations

So we could extract `$links->model->get_languages_list( array( 'hide_empty' => $args['hide_if_empty'] )` into a method that will change between `PLL_Admin_Switcher` and `PLL_Frontend_Switcher`.

Read the following comments to see how it could change the code.

include/admin-switcher.php
```diff
@@ -0,0 +1,88 @@
...
+			list( $id, $order, $slug, $locale, $classes, $url ) = $this->init_foreach_language( $language );
+
+			$curlang = $links->options['default_lang'];
```
raaaahman, line 22: As this is one of the only lines who changes between `PLL_Admin_Switcher` and `PLL_Frontend_Switcher`, this would need to be extracted into two methods with the same names in order to reuse the `get_elements()` method.

include/admin-switcher.php
```diff
@@ -0,0 +1,88 @@
...
+	 *   @type string $admin_current_lang     The current language code in an admin context. Need to set the admin_render to 1, defaults not set.
+	 * }
+	 * @return string|array either the html markup of the switcher or the raw elements to build a custom language switcher
+	 */
+	public function the_languages( $links, $args = array() ) {
+		$defaults = $this->get_default_the_languages();
+
+		$args = $this->filter_arguments_pll_languages( $args, $defaults );
+
+		$args['hide_if_empty'] = 0;
```
raaaahman, line 69: We wouldn't need this if we make the suggested change on line 19.

include/admin-switcher.php
```diff
@@ -0,0 +1,88 @@
...
+		// Force not to hide the language for the block preview even if the option is checked.
+		$args['hide_if_no_translation'] = 0;
```
raaaahman, line 71: This is now only used in `PLL_Fronted_Switcher`.

include/frontend-switcher.php
```diff
@@ -0,0 +1,146 @@
...
+
+			$curlang = $links->curlang->slug;
+
+			$current_lang = $curlang == $slug;
+
+			if ( $this->manage_url( $classes, $args, $links, $language, $url ) ) {
+				list( $url, $no_translation, $classes ) = $this->manage_url( $classes, $args, $links, $language, $url );
+			} else {
+				continue;
+			}
```
raaaahman, line 30: If the `PLL_Admin_Switcher` had the `manage_url()` method as well, we could share the whole `get_elements()` method in their parent class.

include/frontend-switcher.php
```diff
@@ -0,0 +1,146 @@
...
+
+		$args = $this->filter_arguments_pll_languages( $args, $defaults );
+
+		$elements = $this->get_elements( $links, $args );
+
+		if ( $args['raw'] ) {
+			return $elements;
+		}
+
+		list( $out, $args ) = $this->prepare_pll_walker( $links->curlang->slug, $args, $elements );
```
raaaahman, line 80: If we make the previously suggested changes, we could share all this code inside the parent class.

include/frontend-switcher.php
```diff
@@ -0,0 +1,146 @@
...
+		/**
+		 * Filter the link in the language switcher
+		 *
+		 * @since 0.7
+		 *
+		 * @param string $url    the link
+		 * @param string $slug   language code
+		 * @param string $locale language locale
+		 */
+		$url = apply_filters( 'pll_the_language_link', $url, $language->slug, $language->locale );
```
raaaahman, line 135: I believe both of these were executed in the Admin context before, as the `$url` variable would always be `null` then.

This could go in a `PLL_Switcher::manage_url()` method.

include/switcher.php
```diff
@@ -10,6 +10,16 @@
  */
 class PLL_Switcher {
 
+	public static function create( $polylang ) {
+		if ( $polylang instanceof PLL_Frontend ) {
+			return new PLL_Frontend_Switcher();
+		} else if ( $polylang instanceof  PLL_Admin ) {
+			return new PLL_Admin_Switcher();
+		} else {
+			return '';
```
raaaahman, line 19: A string is not a return value that I would expect here.

I would suggest we throw an exception at this point, as this case could only happen if we made an error during development.

raaaahman (commented):
This is definitely an improvement.

I believe we could reuse more code between the two classes by doing the changes below.

raaaahman (approved):

include/switcher.php
```diff
@@ -4,11 +4,28 @@
...
+	 */
+	public static function create( $polylang ) {
+		if ( $polylang instanceof PLL_Frontend ) {
+			return new PLL_Frontend_Switcher();
+		} elseif ( $polylang instanceof PLL_Admin ) {
+			return new PLL_Admin_Switcher();
+		} else {
+			return null;
+		}
+	}
```
manooweb, line 28: Couldn't we only pass the $links object and test its kind of instance instead of the Polylang object ?

manooweb (changes requested):

frontend/frontend-nav-menu.php
```diff
@@ -107,7 +115,7 @@ public function wp_get_nav_menu_items( $items ) {
 				/** This filter is documented in include/switcher.php */
 				$options = apply_filters( 'pll_the_languages_args', $options ); // Honor the filter here for 'show_flags', 'show_names' and 'dropdown'.
 
-				$switcher = new PLL_Switcher();
+				$switcher = PLL_Switcher::create( $this->polylang );
```
raaaahman, line 118: Do we really need to call on the factory method here? Since we are in the `PLL_Frontend_Nav_Menu`, couldn't we always instantiate a `PLL_Frontend_Switcher` instead?

manooweb: Good question 🤔 Indeed we are supposed to be already in a frontend context.

manooweb (changes requested):

include/switcher.php
raaaahman: Isn't there a REST case where we need the Front-End Switcher? Are the links also an instance of `PLL_Admin_Links` in this case?

manooweb: Yes! In REST context we are using `PLL_Admin_Links`. We manage the detection by the use of a `is_block_editor` parameter passed to the request.
Otherwise we decided all together to keep passing the Polylang full object.

manooweb (approved):
Ok 👍 with the latest commit.

frontend/frontend-nav-menu.php
```diff
@@ -115,7 +115,7 @@ public function wp_get_nav_menu_items( $items ) {
 				/** This filter is documented in include/switcher.php */
 				$options = apply_filters( 'pll_the_languages_args', $options ); // Honor the filter here for 'show_flags', 'show_names' and 'dropdown'.
 
-				$switcher = PLL_Switcher::create( $this->polylang );
+				$switcher = new PLL_Frontend_Switcher();
```
raaaahman, line 118: It seems we do not need the `$polylang` property anymore then:

https://github.com/polylang/polylang/blob/2e0eb9c1ea34e74678329a722ea1840c617821e0/frontend/frontend-nav-menu.php#L24-L36

raaaahman (changes requested):

frontend/frontend-nav-menu.php
manooweb: Done in the latest commit 👍

manooweb (approved):
Changes requested by @raaaahman  is done in the latest commit.

raaaahman (approved):
:+1:

Chouby:
Closed due to conflict with #769

PR #743, opened 2021-02-05
Remove unused filter for WP_Widget::display_callback()

raaaahman:
## Changes

~~It appears that this filter is not used anymore, because the widgets are already filtered on `sidebars_widgets`.~~

EDIT: Widgets can be displayed as template tag by [the_widget()](https://developer.wordpress.org/reference/functions/the_widget/) function.

In this case, we will need the display callback.


Chouby:
I prefer not to include this in 3.0 as we are too close from the beta. I want to be sure that we have tests to replace the removed ones before merging this.

manooweb (approved):
ok 👍

Issue #741, opened 2021-02-05
Deprecated jQuery functions


mrleemon:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [ ] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] This bug happens with a default WordPress theme active
- [x] I can reproduce this bug consistently

## Steps to reproduce the issue

According to the jQuery Migrate Helper your plugin (most recent version) throws some warnings:

```
/wp-content/plugins/polylang/js/classic-editor.min.js: jQuery.fn.change() event shorthand is deprecated
/wp-content/plugins/polylang/js/classic-editor.min.js: jQuery.unique is deprecated; use jQuery.uniqueSort
/wp-content/plugins/polylang/js/classic-editor.min.js: jQuery.isArray is deprecated; use Array.isArray
/wp-content/plugins/polylang/js/classic-editor.min.js: jQuery.isFunction() is deprecated

```

Chouby:
We forgot the first one in our jQuery migration project in 2.9. The 3 others ar not coming directly from our code but from WodPress.

mrleemon:
Thanks!

mrleemon:
There's still a deprecated `click()` shorthand function in there:

https://github.com/polylang/polylang/blob/1a4dea94613a7daaf171ef79eb7b3421fc6a6f67/js/classic-editor.js#L49

And a deprecated `bind('click')` in `nav-menu.js`:

https://github.com/polylang/polylang/blob/81fcebcc5dc12f9133c419ebb2da31f548a7d944/js/nav-menu.js#L9

mrleemon:
Also, as of jQuery 3.0, the syntax `jQuery( document ).ready( handler )` is deprecated and the `jQuery( handler )` syntax is the only recommended one: [https://api.jquery.com/ready/](https://api.jquery.com/ready/)

There are two instances in:

https://github.com/polylang/polylang/blob/81fcebcc5dc12f9133c419ebb2da31f548a7d944/js/user.js#L7

and:

[...]

PR #746, opened 2021-02-08
Add a dialog box to ask a confirmation about a language changing.

manooweb:
This PR propose to fix https://github.com/polylang/polylang-pro/issues/240 part needed by the classic editor:
- classic editor
- legacy metabox in block editor ( Polylang or by using PLL_USE_BLOCK_EDITOR_PLUGIN constant)

To display the dialog box we're using the jQuery UI dialog widget with its WordPress default style.
As we can take advantage of how our javascript code is now built, the part of the dialog management is developed in its own script and shared in the the 2 cases quoted above.

This PR also fix https://github.com/polylang/polylang/issues/741 because we forgot one bind `change` function during our [jQuery migration project](https://github.com/orgs/polylang/projects/1)

This PR doesn't look at the media case. In fact, it seems to me that nothing is done for this metabox. Indeed, the metabox isn't refreshed after a language changing because the `change` handler is attached to `post_lang_choice` id instead of `attachments[n][language]` where `n` is the post id

### DIFF
--- a/js/block-editor.js
+++ b/js/block-editor.js
@@ -2,6 +2,12 @@
  * @package Polylang
  */
 
+import {
+	initializeLanguageOldValue,
+	initializeConfimationModal,
+	bypassConfirmation
+} from './lib/confirmation-modal';
+
 /**
  * Filter REST API requests to add the language in the request
  *
@@ -50,6 +56,10 @@ function getCurrentLanguage() {
  */
 jQuery(
 	function( $ ) {
+		// Initialize current language to be able to compare if it changes.
+		initializeLanguageOldValue();
+
+
 		// Ajax for changing the post's language in the languages metabox
 		$( '.post_lang_choice' ).on(
 			'change',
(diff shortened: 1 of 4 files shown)


js/block-editor.js
```diff
@@ -50,56 +56,77 @@ function getCurrentLanguage() {
...
+										}
+									}
+								);
+								blockEditorSavePostAndReloadPage();
 							}
 						);
-						blockEditorSavePostAndReloadPage();
-					}
+					},
+					() => {} // Do nothing when promise is rejected by clicking the Cancel dialog button.
```
raaaahman, line 129: The Promise's reject callback does nothing both in block and classic editors. Do we need a reject callback at all?

js/lib/confirmation-modal.js
```diff
@@ -0,0 +1,101 @@
...
+	let content = jQuery( 'textarea#content' ).val();
+	let excerpt = jQuery( 'textarea#excerpt' ).val();
+	// Block editor case
+	if ( document.getElementById( 'editor' ) && 'undefined' !== typeof wp.data ) {
+		const editor = wp.data.select( 'core/editor' );
+		title = editor.getEditedPostAttribute( 'title' ).trim();
+		content = editor.getEditedPostAttribute( 'content' ).trim();
+		excerpt = editor.getEditedPostAttribute( 'excerpt' ).trim();
+	}
+	return ! title && ! content && ! excerpt;
```
raaaahman, line 22: Since this function behaves differently whether it is called from the block or classic editor, I don't see benefits from  sharing it.

js/lib/confirmation-modal.js
```diff
@@ -0,0 +1,101 @@
...
+// Dialog box for alerting the user about a risky changing.
+export const initializeConfimationModal = () => {
+	// Create dialog container.
+	const dialogContainer = jQuery(
+		'<div/>',
+		{
+			id: 'pll-dialog',
+			style: 'display:none;'
+		}
+	).text( __( 'Are you sure you want to change the language of the current content?', 'polylang' ) );
```
raaaahman, line 34: Thankfully, this looks pretty similar to the [React.createElement()](https://reactjs.org/docs/react-without-jsx.html) signature. So sharing the internationalized strings should be simple when we decide to do it.

js/lib/confirmation-modal.js
```diff
@@ -0,0 +1,101 @@
...
+					maxWidth: '100%',
+					open: function( event, ui ) {
+						// Change dialog box position for rtl language
+						// if ( $( 'body' ).hasClass( 'rtl' ) ) {
+						// 	$( this ).parent().css(
+						// 		{
+						// 			right: $( this ).parent().css( 'left' ),
+						// 			left: 'auto'
+						// 		}
+						// 	);
```
raaaahman: Why is this code commented out?

js/lib/confirmation-modal.js
```diff
@@ -0,0 +1,101 @@
...
+				}
+			);
+		}
+	);
+	return { dialogContainer, dialogResult };
+}
+
+export const initializeLanguageOldValue = () => {
+	// Keep the old language value to be able to compare to the new one and revert to it if necessary.
+	languagesList.attr( 'data-old-value', languagesList.children( ':selected' )[0].value );
```
raaaahman, line 101: Wouldn't it be simpler to return the old language value from this function, and only manipulate it inside the javascript?

js/block-editor.js
manooweb: Yes! It's just to avoid a javascript error that says when uncatch a promise.

js/lib/confirmation-modal.js
manooweb: No! You need to have the old value before you change it. Indeed when you enter in the `change` handler the value of the select HTML tag has been already changed and you don't know the old one

js/lib/confirmation-modal.js
manooweb: Sorry, I replied on slack when I processed with coding standards I saw it too.
It was something I forgot. It's coded to adapt the dialog box to rtl language. 
I adapted the code and pushed it again.

js/lib/confirmation-modal.js
manooweb: I prefer to have at the same place what is an empty post in the 2 cases.
That allows to have a similar test in classic and block-editor.js script which are enqueued.
It seems to me to be clearer.

js/block-editor.js
```diff
@@ -50,56 +56,77 @@ function getCurrentLanguage() {
...
-					lang:       event.target.value,
-					post_type:  $( '#post_type' ).val(),
-					post_id:    $( '#post_ID' ).val(),
-					_pll_nonce: $( '#_pll_nonce' ).val()
+				if ( $( this ).data( 'old-value' ) !== selectedOption.value && ! emptyPost ) {
+					dialog.dialog( 'open' );
+				} else {
+					// Block editor doesn't allow to save an empty post.
+					// So we must revert the language to the old one because the browser asks if we want to quit the page if we trigger its reloading.
+					selectedOption.value = $( this ).data( 'old-value' );
```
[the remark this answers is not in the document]
manooweb: I modified to accept the language changing in case 2 - `Promise.resolve()`. But I think we need to reload the page especially because the block editor don't save an empty post. 

However it's a limit case when the legacy metabox is used in Polylang Pro by using the PLL_USE_BLOCK_EDITOR_PLUGIN constant and to refresh correctly categories, tags or parent page. It's impossible now with Polylang which doesn't implement REST API to filter categories, tags and parent page by language.

This reload can't be reload as a new post because we don't need to inject the `new_lang` paremeter.

In the other hand if we reload the page in this case and accept the language changing, it's because the user emptied his post. The browser warns about that and asks if he want to reload the page. So we can't do the reloading programmatically without a user confirmation.

So we can consider acceptable that just change the language without reloading the page is enough.

raaaahman (approved):
Good job. :+1:

PR #756, opened 2021-02-15
Add Valencià

mmozos:


Chouby:
Thank you for your pull request. I am sorry but I won't merge it. The predeifned languages list includes only languages for which a language pack exists in for WordPress and we want ot keep this policy.

For your own needs, you can use the filter avaliable to add your language to your custom installs.

PR #760, opened 2021-02-22
Update phpstan-wordpress to 0.7.3 and fix new reported errors

Chouby:
phpstan-wordpress 0.7.3 allowed to detect a few more errors.
One of them is ignored. I believe that it's a false positive due to https://core.trac.wordpress.org/ticket/52603


manooweb (approved):
Ok 👍

PR #761, opened 2021-02-22
Yoast: fix post type archive title and metadesc not translated

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/855

1. The Yoast SEO options are cached after they are read from `get_option()`.
2. We wait for `wp_loaded` to determine the post type archives for which we need to translat the title and metadesc. This is too late as the options are already in the cache and the `option_{$option_name}` filters are not executed anymore.

Fortunately, Yoast SEO allows to clear the options cache which allows us to execute our filters once done.


raaaahman (approved):
Okay.

PR #762, opened 2021-02-23
Extract language selection to context classes.

raaaahman:
## Before

`PLL_CRUD_Terms::get_queried_language()` checked the value of `$GLOBALS['pagenow']`, which is a global variable available in the admin context only, whereas `PLL_CRUD_Terms` is used in both admin and frontend contexts (REST?).

`PLL_CRUD_Terms` did also read the `$filter_lang` property of whichever context class was passed, but this property only exists in `PLL_Admin_Base` and its child classes.

## Changes

Extract the method that returns the correct language to use, and move it to the context classes.

Fixes [Polylang Pro #884](https://github.com/polylang/polylang-pro/issues/884)

## Going Further

- The `get_requested_language()` method could belong to a specialized class, `PLL_Request`, that we could use to carry other information from the global variables to the modules.
- The `get_requested_language()` method could belong to the `PLL_Query` class instead.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -467,7 +467,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	 * @param array $args
 	 * @return PLL_Language
 	 */
-	public function get_requested_language( $args ) {
+	public function get_requested_language( $args = array() ) {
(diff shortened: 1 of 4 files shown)


manooweb (approved):
Ok. 👍

Chouby:
Closed in favor of #766

PR #763, opened 2021-02-23
Fix site title not translated in email change confirmation email

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/92

Strings translations are usually not loaded on admin side. We however need to make an exception when emails are sent. This PR loads the strings translations when the user profile is saved in case the email change confirmation email is sent.


manooweb (approved):
Ok 👍
Is this comment always useful ?
https://github.com/polylang/polylang/blob/8e5da70c51c716de4064a3419bc3b2b24a38333f/tests/phpunit/tests/test-filters.php#L357

PR #766, opened 2021-02-25
Fix undefined index 'pagenow'

Chouby:
This PR is an alternative to #762, fixing the issue in a more simple way.
Fixes https://github.com/polylang/polylang-pro/issues/884


manooweb (approved):
Ok 👍

PR #767, opened 2021-02-25
Add a travis job to test the deve version of WP with PHP 5.6

Chouby:
After #764, it seems useful not to only run our tests for the nightly version of WP with the latest PHP version but also with the oldest, in the hope that this will help to catch more issues.


manooweb (approved):
Ok 👍

PR #768, opened 2021-02-25
Rewrite how to add polylang parameters in all admin ajax requests

### LINKED ISSUE #744: $.ajaxPrefilter .data should be string
This is a copy of an issue reported on wp.org: https://wordpress.org/support/topic/ajaxprefilter-data-should-be-string/

We have a compatibility issue with Polylang. The problem caused by the following code in polylang/admin/admin-base.php. It seems like Polylang hooks itself in the admin area to the $.ajaxPrefilter to place its query variable to every request. It seems like that problem is that jQuery (probably is happens since WordPress started to use jQuery 3) expects the data to be string instead of object.

The error can be simple reproduced in the admin area with the following code [only Polylang plugin required]:

```js
var undef;
jQuery.ajax({
    type: 'POST',
    url: ajaxurl,
    data: undef
});
```

Original code:

```php
<script type="text/javascript">
    if (typeof jQuery != 'undefined') {
        jQuery(
            function($){
                $.ajaxPrefilter(function (options, originalOptions, jqXHR) {
                    if ( -1 != options.url.indexOf( ajaxurl ) || -1 != ajaxurl.indexOf( options.url ) ) {

                        if ( 'undefined' === typeof options.data ) {
                            options.data = ( 'get' === options.type.toLowerCase() ) ? '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>' : <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>;
                        } else {
                            if ( 'string' === typeof options.data ) {
                                if ( '' === options.data && 'get' === options.type.toLowerCase() ) {
                                    options.url = options.url+'&<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>';
                                } else {
                                    try {
                                        var o = JSON.parse(options.data);
                                        o = $.extend(o, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>);
                                        options.data = JSON.stringify(o);
                                    }
                                    catch(e) {
                                        options.data = '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>&'+options.data;
                                    }
                                }
                            } else {
                                options.data = $.extend(options.data, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>);
                            }
                        }
                    }
                });
            }
        );
    }
</script>
```

I propose the following fix for this issue:

```php
<script type="text/javascript">
    if (typeof jQuery != 'undefined') {
        jQuery(
            function($){
                $.ajaxPrefilter(function (options, originalOptions, jqXHR) {
                    if ( -1 != options.url.indexOf( ajaxurl ) || -1 != ajaxurl.indexOf( options.url ) ) {

                        if ( 'undefined' === typeof options.data ) {
                            options.data = '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>';
                        } else {
                            if ( 'string' === typeof options.data ) {
                                if ( '' === options.data && 'get' === options.type.toLowerCase() ) {
                                    options.url = options.url+'&<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>';
                                } else {
                                    try {
                                        var o = JSON.parse(options.data);
                                        o = $.extend(o, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>);
                                        options.data = JSON.stringify(o);
                                    }
                                    catch(e) {
                                        options.data = '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>&'+options.data;
                                    }
                                }
                            } else {
                                options.data = JSON.stringify($.extend(options.data, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>));
                            }
                        }
                    }
                });
            }
        );
    }
</script>
```

manooweb:
fixes #744 and fixes https://github.com/polylang/polylang-pro/issues/821

To correctly understand, refer to my latest comment on https://github.com/polylang/polylang-pro/issues/821 https://github.com/polylang/polylang-pro/issues/821#issuecomment-779915122 and we need to know how [jQuery ajax](https://api.jquery.com/jquery.ajax/#jQuery-ajax-settings) function works. Especially that `processData` option is set to true by default and then `ajaxPrefilter` is done after [jQuery.param()](https://api.jquery.com/jQuery.param/#jQuery-param-obj) which converts data as an object into a string.
https://github.com/jquery/jquery/blob/main/src/ajax.js#L566-L571

So `processData` option is really the right method to know what kind of data we receive in `ajaxPrefilter` function:

- `true`: always a string, set by the user or convert by` jQuery.param()`
- `false`: could also be a string  set by the user and need to be processed as when `processData` is set to `true` or anything else probably an object

Thus, in most cases, we have a string to process and we want to add the Polylang parameters.
So we have to test to cases:
- undefined or empty data string: fully override the value by the Polylang parameters.
- otherwise concatenate the Polylang parameters to the data string

If processData is enforce to false and data send as an object the polylang parameters are added as properties of this data object.

We consider that the user know what it does and how the data need to be retrieved server side to be process. 
Any way it is a borderline case.

## Further more
I tested the cases that I thought about and described above on codepen https://codepen.io/manooweb/pen/abBENbx and by using the webservice https://httpbin.org/

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -269,8 +269,6 @@ abstract class PLL_Admin_Base extends PLL_Base {
 					function( $ ){
 						$.ajaxPrefilter( function ( options, originalOptions, jqXHR ) {
 							if ( -1 != options.url.indexOf( ajaxurl ) || -1 != ajaxurl.indexOf( options.url ) ) {
-								console.log( 'pll', '$str = <?php echo $str; ?>');
-								console.log( 'pll', '$arr <?php echo $arr; ?>');
 
 								function addStringParamaters() {
 									if ( 'undefined' === typeof options.data || '' === options.data ) {
@@ -279,6 +277,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 										options.data = options.data + '&<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>';
 									}
 								}
+
 								// options.processData set to true is the default jQuery process where the data is converted in a query string by using jQuery.param().
 								// This step is done before applying filters. Thus here the options.data is already a string in this case.
 								// @See https://github.com/jquery/jquery/blob/3.5.1/src/ajax.js#L563-L569 jQuery ajax function.
@@ -298,7 +297,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 						});
 					}
 				);
-			}
+			}q
 		</script>
 		<?php
 	}
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -297,7 +297,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 						});
 					}
 				);
-			}q
+			}
 		</script>
 		<?php
 	}
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -284,7 +284,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 								 * @See https://github.com/jquery/jquery/blob/3.5.1/src/ajax.js#L563-L569 jQuery ajax function.
 								 */
 								if ( options.processData ) {
-									addStringParamaters();
+									addStringParameters();
 								} else {
 									/*
 									 * If options.processData is set to false data could be undefined or pass as a string.
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -270,7 +270,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 						$.ajaxPrefilter( function ( options, originalOptions, jqXHR ) {
 							if ( -1 != options.url.indexOf( ajaxurl ) || -1 != ajaxurl.indexOf( options.url ) ) {
 
-								function addStringParamaters() {
+								function addStringParameters() {
 									if ( 'undefined' === typeof options.data || '' === options.data ) {
 										options.data = '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>';
 									} else {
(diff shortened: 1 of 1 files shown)


admin/admin-base.php
```diff
@@ -266,27 +266,31 @@ public function admin_print_footer_scripts() {
 		<script type="text/javascript">
 			if (typeof jQuery != 'undefined') {
 				jQuery(
-					function($){
-						$.ajaxPrefilter(function (options, originalOptions, jqXHR) {
+					function( $ ){
+						$.ajaxPrefilter( function ( options, originalOptions, jqXHR ) {
 							if ( -1 != options.url.indexOf( ajaxurl ) || -1 != ajaxurl.indexOf( options.url ) ) {
```
Chouby, line 271: Is this test still useful?  The original version used to modify `options.url` but that's not the case anymore.

admin/admin-base.php
```diff
@@ -266,27 +266,31 @@ public function admin_print_footer_scripts() {
...
-								if ( 'undefined' === typeof options.data ) {
-									options.data = ( 'get' === options.type.toLowerCase() ) ? '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>' : <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>;
+
+								function addStringParamaters() {
```
Chouby, line 273: Did you mean `addStringParam`**e**`ters()`?

admin/admin-base.php
```diff
@@ -266,27 +266,31 @@ public function admin_print_footer_scripts() {
...
-												o = $.extend(o, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?>);
-												options.data = JSON.stringify(o);
-											}
-											catch(e) {
-												options.data = '<?php echo $str; // phpcs:ignore WordPress.Security.EscapeOutput ?>&'+options.data;
-											}
-										}
+									// If options.processData is set to false data could be undefined or pass as a string.
+									// So data as to be processed as if options.processData is set to true.
+									if ( 'undefined' === typeof options.data || 'string' === typeof options.data ) {
```
Chouby, line 289: What would you think to combine all the tests in one?

```php
if ( options.processData || 'undefined' === typeof options.data || 'string' === typeof options.data ) {
addStringParameters();
} else {
options.data = Object.assign( options.data, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?> );
}
```

It would even allow us to remove the function `addStringParameters()` if we want to reduce the js size, as it woud be called only once (maybe with a price on the readability though).

May I suggest to use the syntax` /* */` for multiline comments as in https://github.com/polylang/polylang/blob/3.0-beta1/include/class-polylang.php#L55-L58. I know we have more comments with the single line syntax in Polylang but I started to be more consistent in the recent past and use single line syntax for single line comments and multiline syntax for multiline comments.

Chouby (changes requested):

admin/admin-base.php
manooweb: Yes! typo

admin/admin-base.php
manooweb: The first condition, maybe to test only URL which concerns WordPress ajax request. The second one I'm not sure to understand what it tests.

admin/admin-base.php
manooweb: Ok for multiline comments.
As we discussed in live together, it seems preferable for readibility - correctly understand each case - to keep the code as this.

PR #769, opened 2021-02-26
Fix switcher block not working in a post retrieved in REST API

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/892

The PR clarifies the `PLL_Switcher` by introducing a constant for the default arguments of `the_languages()` and 2 new methods `get_current_language()` and `get_link()`.

The parameter passed as current language is prioritized over the gloabal current language. It defaults to the default language in case no current language is found.

The `get_link()` method now checks the instance of `$this->links` to avoid a fatal error in case it is not a `PLL_Frontend_Links`.
A new way to get the link, based on the global post is introduced to make the language switcher functionnal in REST API.

The parameter `admin_render` is kept to control the output of the javascript for dropdowns.

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -29,11 +29,8 @@
  * @return string|array Either the html markup of the switcher or the raw elements to build a custom language switcher.
  */
 function pll_the_languages( $args = array() ) {
-	if ( PLL() instanceof PLL_Frontend ) {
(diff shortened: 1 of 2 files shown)


manooweb:
Also fixes https://github.com/polylang/polylang-pro/issues/835

PR #773, opened 2021-03-05
Bump min WP version to 5.4 and remove useless checks in tests

Chouby:
~~Following our policy of support of ol WP versions, I propose to bump the minimum required WP version to 5.3.1.
I propose 5.3.1 instead of 5.3 due to the functions that were introduced in this version and that we are using in Polylang Pro.~~
The PR also removes some checks that are now useless for all supported versions.

Edit: I updated the check for min WP version with Jetpack which just bumped its min WP version requirement to 5.6. Done in this PR to avoid a merge conflict.

Edit: We have finally decided to bump the min version to 5.4

### DIFF
--- a/tests/phpunit/tests/plugins/test-jetpack.php
+++ b/tests/phpunit/tests/plugins/test-jetpack.php
@@ -1,6 +1,6 @@
 <?php
 
-if ( version_compare( $GLOBALS['wp_version'], '5.2', '>=' ) && file_exists( DIR_TESTROOT . '/../jetpack/jetpack.php' ) ) {
+if ( file_exists( DIR_TESTROOT . '/../jetpack/jetpack.php' ) ) {
 
 	require_once DIR_TESTROOT . '/../jetpack/functions.opengraph.php';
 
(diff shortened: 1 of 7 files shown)


tests/phpunit/tests/plugins/test-jetpack.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-if ( version_compare( $GLOBALS['wp_version'], '5.2', '>=' ) && file_exists( DIR_TESTROOT . '/../jetpack/jetpack.php' ) ) {
+if ( version_compare( $GLOBALS['wp_version'], '5.6', '>=' ) && file_exists( DIR_TESTROOT . '/../jetpack/jetpack.php' ) ) {
```
raaaahman, line 3: What happens with WordPress 5.4 and 5.5 here?

Chouby: Jetpack doesn't load.

raaaahman (approved):

PR #774, opened 2021-03-05
Add post types without translation in Site Health

sebastienserre:
- refresh Branch
- Display 5 ID of each Post Type without Languages
site-health-no-tr


modules/site-health/admin-site-health.php
```diff
@@ -82,24 +83,30 @@ protected function exclude_language_keys() {
...
+				} else {
+					$value = "$key => $value";
+				}
 			}
 		);
 
 		return implode( ' | ', $array );
 	}
 
+
```
manooweb, line 109: 2nd blank line not necessary here.

modules/site-health/admin-site-health.php
```diff
@@ -172,6 +187,43 @@ public function info_options( $debug_info ) {
...
 
+	/**
+	 * Get an array with post_type as key and post IDs as value
+	 *
+	 * @param int $limit  Nb of post max to show per post type.
+	 * @return array
+	 *
+	 * @since   3.0
+	 */
+	public function get_post_ids_without_lang( $limit = 10 ) {
```
manooweb, line 198: Why not set the default value to 5 as it is called with this value in `info_options()` method ?

manooweb (changes requested):

modules/site-health/admin-site-health.php
```diff
@@ -172,6 +187,43 @@ public function info_options( $debug_info ) {
...
+		$posts = array();
+		$languages                  = pll_languages_list();
```
manooweb, line 200: Strange alignment of the assigment

manooweb (changes requested):

modules/site-health/admin-site-health.php
sebastienserre: Yes you're right

modules/site-health/admin-site-health.php
```diff
@@ -163,6 +168,13 @@ public function info_options( $debug_info ) {
 			}
 		}
 
+		$post_no_lang = $this->get_post_ids_without_lang( );
```
manooweb, line 171: Oops! no space here when empty brackets

manooweb (changes requested):

manooweb (approved):
Ok 👍 with the latest changes and rebase on master

include/model.php
```diff
@@ -634,4 +634,75 @@ public function get_links_model() {
...
+
+		/**
+		 * Filters the list of untranslated posts ids and terms ids
+		 *
+		 * @since 0.9
+		 *
+		 * @param array|false $objects false if no ids found, list of post and/or term ids otherwise.
+		 */
+		return apply_filters( 'pll_get_objects_with_no_lang', empty( $posts ) && empty( $terms ) ? false : array( 'posts' => $posts, 'terms' => $terms ) );
+	}
```
Chouby, line 707: It looks lile you finally don't use this method, so I suggest not to move it;

modules/site-health/admin-site-health.php
```diff
@@ -172,6 +184,43 @@ public function info_options( $debug_info ) {
...
 	}
 
+	/**
+	 * Get an array with post_type as key and post IDs as value
+	 *
+	 * @param int $limit  Nb of post max to show per post type.
+	 * @return array
+	 *
+	 * @since   3.0
+	 */
```
Chouby, line 194: Could you use the same format as usual `@since` above `@param`. Also provide more precision for the type of array `int[][]`.

modules/site-health/admin-site-health.php
```diff
@@ -172,6 +184,43 @@ public function info_options( $debug_info ) {
...
+	public function get_post_ids_without_lang( $limit = 5 ) {
+		$posts     = array();
+		$languages = pll_languages_list();
```
Chouby, line 197: Don't use API functions internally (to avoid relying on global in phunit tests). Here you can access to the internal method `$this->model->get_languages_list()`.

modules/site-health/admin-site-health.php
```diff
@@ -163,6 +168,13 @@ public function info_options( $debug_info ) {
 			}
 		}
 
+		$post_no_lang = $this->get_post_ids_without_lang();
+
+		if ( ! empty( $post_no_lang ) ) {
+			$fields['post-no-lang']['label'] = __( 'Posts without language', 'polylang' );
+			$fields['post-no-lang']['value'] = $this->format_array( $post_no_lang );
+		}
+
```
Chouby, line 177: The information is displayed in the Polylang options section. I would prefer that we create a new section (displayed only if there are posts without language).

Chouby (changes requested):

modules/site-health/admin-site-health.php
sebastienserre: I'm not sure about how to write the type of the array ...

modules/site-health/admin-site-health.php
sebastienserre: What do you think about creating a "Polylang Warning" section with the Post without ID inside and maybe other informations in the future ?

modules/site-health/admin-site-health.php
manooweb: As for @param, I think. The returned array is an array of int: post IDs

modules/site-health/admin-site-health.php
manooweb: Why not! Keep in mind that if this section is empty it mustn't be displayed. So if all its subsections are empty the section musn't be displayed.

PR #780, opened 2021-03-09
Fix greedy regex which may match all search <form> attributes

marian-kadanka:
As reported here https://wordpress.org/support/topic/search-form-impact/ , search form is messed up because &lt;form&gt; element is stripped of all attributes except of action attribute, in get_search_form hook.

To fix this issue, replace greedy regex which may match all search &lt;form&gt; attributes with non-greedy regex matching only the action attribute.

### DIFF
--- a/frontend/frontend-filters-search.php
+++ b/frontend/frontend-filters-search.php
@@ -64,7 +64,7 @@ class PLL_Frontend_Filters_Search {
 				// Take care to modify only the url in the <form> tag.
 				preg_match( '#<form.+>#', $form, $matches );
 				$old = reset( $matches );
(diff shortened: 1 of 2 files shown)


Chouby:
@marian-kadanka Thanks for the PR. We noticed that there were still some edge cases that could be missed by the regex, so we further improved it and added a phpunit test for the initial bug reported.

raaaahman (approved):
Okay. Although it would be nice to add the cases discussed (single quotes, no quotes, and empty `action` attribute) in our tests, so we will prevent regression whenever we need to change it in the future.

PR #787, opened 2021-03-10
Fix custom flags not working

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/910

In #660, custom flag properties of `PLL_Language` were assigned the `protected` visibility. This means that thy are not saved in the transient in https://github.com/polylang/polylang/blob/3.0.1/include/model.php#L118 as `get_object_vars()` doesn't return protected properties outside the object context.


manooweb (approved):
Ok 👍

PR #789, opened 2021-03-12
Yoast SEO: Fix sitemap.xml not redirected on secondary domains

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/914

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -106,7 +106,7 @@ class PLL_WPSEO {
 			$path = ltrim( wp_parse_url( pll_get_requested_url(), PHP_URL_PATH ), '/' );
 		}
 
-		if ( 'sitemap_index.xml' === $path || preg_match( '#([^/]+?)-sitemap([0-9]+)?\.xml|([a-z]+)?-?sitemap\.xsl#', $path ) ) {
+		if ( preg_match( '#sitemap(_index)?\.xml|([^\/]+?)-?sitemap([0-9]+)?\.xml|([a-z]+)?-?sitemap\.xsl#', $path ) ) {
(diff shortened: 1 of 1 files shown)


raaaahman (commented):
[This filter](https://github.com/polylang/polylang/blob/57c5184ed39e671e0776c8c20d251fe5b76f0a88/integrations/wpseo/wpseo.php#L32) is hooked on the same priority than [PLL_Sitemaps::sitemap_url()](https://github.com/polylang/polylang/blob/02ebae9725b358e0beb5cc038d085c4e5a1794ab/modules/sitemaps/sitemaps-domain.php#L41). Which one should trigger first?

This regex seems to handle several cases. How could we know which ones are handled?

Chouby:
> [This filter](https://github.com/polylang/polylang/blob/57c5184ed39e671e0776c8c20d251fe5b76f0a88/integrations/wpseo/wpseo.php#L32) is hooked on the same priority than [PLL_Sitemaps::sitemap_url()](https://github.com/polylang/polylang/blob/02ebae9725b358e0beb5cc038d085c4e5a1794ab/modules/sitemaps/sitemaps-domain.php#L41). Which one should trigger first?

I believe it doesn't matter
> This regex seems to handle several cases. How could we know which ones are handled?

The goal is to have the right domain for all sitemap urls handled in Yoast SEO. The initial version was written based on the rewrite rules added by Yoast SEO. Unfortunately this did not include `sitemap.xml` itself. I don't know how to make sure that we don't miss another case.

raaaahman:
> The goal is to have the right domain for all sitemap urls handled in Yoast SEO. The initial version was written based on the rewrite rules added by Yoast SEO. Unfortunately this did not include `sitemap.xml` itself. I don't know how to make sure that we don't miss another case.

Could we write a test each time we encounter a new case?

raaaahman (approved):

PR #791, opened 2021-03-12
Fix warning on frontend-filters-links.php

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/912

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -507,7 +507,7 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 		$taxonomy = $this->get_queried_taxonomy( $tax_query );
 
 		$field = $queried_terms[ $taxonomy ]['field'];
-		$term  = is_array( $queried_terms[ $taxonomy ]['terms'] )  ? reset( $queried_terms[ $taxonomy ]['terms'] ) : false;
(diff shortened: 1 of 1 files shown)


frontend/frontend-filters-links.php
```diff
@@ -507,7 +507,7 @@ protected function get_queried_term_id( $tax_query ) {
 		$taxonomy = $this->get_queried_taxonomy( $tax_query );
 
 		$field = $queried_terms[ $taxonomy ]['field'];
-		$term  = reset( $queried_terms[ $taxonomy ]['terms'] );
+		$term  = $queried_terms[ $taxonomy ]['terms'] ? reset( $queried_terms[ $taxonomy ]['terms'] ) : '';
```
manooweb, line 510: I think is sufficient to protect this `reset` call.
But what about `$queried_terms[ $taxonomy ]` $queried_terms is really null ? 
Is there no notice when we try to access it `$queried_terms[ $taxonomy ]['field']` or `$queried_terms[ $taxonomy ]['terms']`

What will happen behind with the get_terms() call where we use the `$field` and `$term` values ?

What about the `return reset( $terms )` at the end ? Because get_terms() can return another thing than an array.
https://developer.wordpress.org/reference/functions/get_terms/

manooweb (changes requested):

Chouby (approved):

PR #792, opened 2021-03-12
Change constant to enqueue scripts from

raaaahman:
## Before

Scripts were enqueued from a relative file path that was correctly resolved when Polylang was located in the `plugins` directory, but incorrectly resolved when Polylang was located in the `mu-plugins` directory.

## Changes

Use another constant to enqueue scripts from the current file path.
Fixes https://github.com/polylang/polylang-pro/issues/916


manooweb (approved):
Ok 👍

PR #794, opened 2021-03-13
Yoast : fix cpt breadcrumb title when the option is empty

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/909


manooweb (approved):
OK 👍

PR #797, opened 2021-03-15
Move all js files contained in js/ folder to the js/src subfolder and CSS files from css/ to css/src

manooweb:
After analysis of the https://github.com/polylang/polylang-pro/issues/908 issue, it seems that strings from lib/confirmation-modal.js are extracted when they aren't needed because they are also in the final built files block-editor.js and classic-editor.js.

The goal is to be able not to extract translatable strings from the js source files but only from the built ones.

So, we exclude js/lib and js/src subfolders to make the polylang distribution release.
Then translate.wordpress.org tools will only know the js/build/ folder to extract translatable strings from the js code.


Chouby:
May I suggest to move CSS too?

manooweb:
> Wouldn't it be more consistent to move `js/lib` to `js/src/lib`?
> May I suggest to move CSS too?

Both are done 😉

raaaahman (approved):

PR #801, opened 2021-03-17
Use the language custom block attribute to be able to filter blocks on frontend

Marianne380:
- Displays the block widget only if the current language is the selected language for displaying the widget.
- Adds a test that checks if the widget would display in the sidebar if the current language is the selected language for displaying the widget.

Fixes https://github.com/polylang/polylang-pro/issues/860


Marianne380:
Replaced by : https://github.com/polylang/polylang-pro/pull/922

PR #807, opened 2021-03-19
Correctly update translations group when saving media language

### LINKED ISSUE #913: Polylang not translating default title and meta description with using Yoast's Meta_Surface
## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate. Related to #505 

1. Install and enable Polylang 3.11 & Yoast 15.9.2
2. Translate post title
Captura de pantalla 2021-09-08 a las 11 40 07
3. Use `YoastSEO()->meta->for_post(1)->title;` (Spanish post). 

When I follow those steps, I see English title.

I was expecting Spanish title.

raaaahman:
Fixes https://github.com/polylang/polylang-pro/issues/913

## Before

Changing a media's language from the pop up in the grid view for WordPress' Media Library would send an AJAX request that will trigger the `attachment_fields_to_save`. 

This would call the `PLL_Admin_Filters_Media::save_media()` function, just as it would for an update request from the Classic Editor. 

However:

- The translations group data was not included when changing the media's language from the grid view, which caused a bug (see https://github.com/polylang/polylang-pro/issues/913 )
- There was no need to update a media's language from the update request (classic editor), because it would always have been previously changed by an AJAX request.

## Changes

- Extract one single method to update a post's language, which changes the translations group based on the previously saved translations group.
- Trigger this update method each time a media is modified. Fixes #913
- Prevent update when language didn't change.

Fixes https://github.com/polylang/polylang-pro/issues/913

## Questions

Post type is retrieved from $_POST when saved through AJAX request, but couldn't we retrieve it from a `get_post_type()` call instead? https://github.com/polylang/polylang/compare/fix-913?expand=1#diff-a3220b8ef2b2eea37d45f314a0292d4f68a9457274f3d3cb0e7960aecb3e6687R119

If so, we could remove the third parameter of the `PLL_Translated_Post::update_language( $post_id, $lang, $post_type )` method and query the post type from the inside of this method instead.


tests/phpunit/tests/test-media.php
```diff
@@ -79,27 +79,35 @@ function test_attachment_fields_to_edit() {
 		$this->assertFalse( isset( $fields['language'] ) );
 	}
 
+	/**
+	 * @since 3.1 Since the language and translations are updated through a previous AJAX call, we'd rather not perform an unnecessary update now.
```
manooweb, line 83: Maybe comment a little bit more inside the test how it works.
Mock of the hook `set_object_terms` triggers by the` edit_post()` function call and then test that there is no language term modification.

manooweb (changes requested):

admin/admin-classic-editor.php
```diff
@@ -171,23 +171,18 @@ public function post_lang_choice() {
...
-
-		if ( empty( $post_type_object ) ) {
-			wp_die( 0 );
-		}
-
-		if ( ! current_user_can( $post_type_object->cap->edit_post, $post_ID ) ) {
-			wp_die( -1 );
+		$return = $this->model->post->update_language( $post_ID, $lang, $post_type );
+		if ( is_wp_error( $return ) ) {
+			// phpcs:disable PSR2.ControlStructures.SwitchDeclaration.TerminatingComment
```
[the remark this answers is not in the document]
manooweb: > Should it be better to use a break; statement in the first case even if we use wp_die() ?
> It makes possible not to use a PHPCS ignore rule.

Revert because of a PHPStan analysis detect that break statement is unreachable due to wp_die() call.

manooweb (approved):
Ok 👍

include/translated-post.php
```diff
@@ -265,4 +265,42 @@ public function get_untranslated( $type, $untranslated_in, $lang, $search = '' )
...
+	 * @return WP_Error|true
+	 */
+	public function update_language( $post_id, $lang, $post_type ) {
+		$post_type_object = get_post_type_object( $post_type );
+
+		if ( empty( $post_type_object ) ) {
+			return new WP_Error( 'invalid_argument', "{$post_type} is not a valid Post Type." );
+		}
+
+		if ( ! current_user_can( $post_type_object->cap->edit_post, $post_id ) ) {
```
raaaahman, line 287: Capability should not be checked inside the model, but on a higher level.

admin/admin-classic-editor.php
```diff
@@ -171,23 +171,18 @@ public function post_lang_choice() {
...
+			switch ( $return->get_error_code() ) {
+				case 'missing_capability':
+					wp_die( -1 );
+				default:
+					wp_die( 0 );
+			}
```
raaaahman, line 182: `wp_die()` should return an error message instead of an int.

include/translated-post.php
```diff
@@ -265,4 +265,42 @@ public function get_untranslated( $type, $untranslated_in, $lang, $search = '' )
...
+	 *
+	 * @since 3.1
+	 *
+	 * @param int          $post_id Id of the post to change the language of.
+	 * @param PLL_Language $lang New language to set to the post.
+	 * @param string       $post_type Post type of the post to change language of.
+	 *
+	 * @return WP_Error|true
+	 */
+	public function update_language( $post_id, $lang, $post_type ) {
```
raaaahman, line 280: This method could be pulled up in the `PLL_Translatated_Object` class.

Hint: `PLL_Translated_Term` should have a similar 'inline edit' method.

admin/admin-filters-media.php
```diff
@@ -116,11 +116,7 @@ public function save_media( $post, $attachment ) {
 		// Language is filled in attachment by the function applying the filter 'attachment_fields_to_save'
 		// All security checks have been done by functions applying this filter
 		if ( ! empty( $attachment['language'] ) ) {
-			$this->model->post->set_language( $post['ID'], $attachment['language'] );
-		}
-
-		if ( isset( $_POST['media_tr_lang'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$this->save_translations( $post['ID'], array_map( 'absint', $_POST['media_tr_lang'] ) ); // phpcs:ignore WordPress.Security.NonceVerification
+			$this->model->post->update_language( $post['ID'], $this->model->get_language( $attachment['language'] ), $post['post_type'] );
```
raaaahman, line 119: Should probably add a `current_user_can( $translated_post_type->edit_post )`.

admin/admin-filters-term.php
```diff
@@ -340,23 +340,8 @@ protected function save_language( $term_id, $taxonomy ) {
...
-				}
-
-				elseif ( array_key_exists( $old_lang->slug, $translations ) ) {
-					unset( $translations[ $old_lang->slug ] );
-					$this->model->term->save_translations( $term_id, $translations );
-				}
-			}
-
-			$this->model->term->set_language( $term_id, $lang ); // Set new language
+			$lang = $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) );
```
raaaahman, line 343: Do we need to check capability here as well? If so, which is the correct capability?

admin/admin-filters-media.php
```diff
@@ -116,11 +116,18 @@ public function save_media( $post, $attachment ) {
...
+			$post_type = get_post_type( $post['ID'] );
+			$post_type_object = get_post_type_object( $post_type );
+
+			if ( empty( $post_type_object ) ) {
+				return $post;
+			}
+
+			if ( ! current_user_can( $post_type_object->cap->edit_post, $post['ID'] ) ) {
+				return $post;
+			}
```
Chouby, line 128: You can simplify all this with:
```php
if ( ! current_user_can( 'edit_post', $post['ID'] ) ) {
	return $post;
}
```

admin/admin-filters-post.php
```diff
@@ -211,7 +167,7 @@ public function bulk_edit_posts() {
 			if ( $lang = $this->model->get_language( sanitize_key( $_GET['inline_lang_choice'] ) ) ) {
 				$post_ids = array_map( 'intval', (array) $_REQUEST['post'] );
 				foreach ( $post_ids as $post_id ) {
-					$this->inline_save_language( $post_id, $lang );
+					$this->model->post->update_language( $post_id, $lang, get_post_type( $post_id ) );
```
Chouby, line 170: You kept the old version. I suggest to restore the method `inline_save_language()`, keep the validation in this function and call the new method `update_language()` with 2 parameters there.

admin/admin-filters-post.php
```diff
@@ -231,7 +187,7 @@ public function inline_edit_post() {
 			$post_id = (int) $_POST['post_ID'];
 			$lang = $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) );
 			if ( $post_id && $lang ) {
-				$this->inline_save_language( $post_id, $lang );
+				$this->model->post->update_language( $post_id, $lang, get_post_type( $post_id ) );
```
Chouby, line 190: Same as above.

admin/admin-filters-term.php
Chouby: Yes. `current_user_can( 'edit_term', $term_id )`

admin/admin-classic-editor.php
```diff
@@ -159,34 +159,30 @@ public function post_lang_choice() {
...
 		$post_type_object = get_post_type_object( $post_type );
 
 		if ( empty( $post_type_object ) ) {
-			wp_die( 0 );
+			wp_die( esc_html( "{$post_type} is not a valid Post Type." ) );
 		}
 
 		if ( ! current_user_can( $post_type_object->cap->edit_post, $post_ID ) ) {
-			wp_die( -1 );
+			wp_die( esc_html( "User does not have {$post_type_object->cap->edit_post} capability." ) );
```
Chouby, line 182: I suggest: `You are not allowed to edit this post`.

admin/admin-classic-editor.php
```diff
@@ -159,34 +159,30 @@ public function post_lang_choice() {
...
-			wp_die( 0 );
+		if ( empty( $lang ) ) {
+			wp_die( esc_html( "{$lang_slug} is not a valid language." ) );
 		}
 
 		$post_type_object = get_post_type_object( $post_type );
 
 		if ( empty( $post_type_object ) ) {
-			wp_die( 0 );
+			wp_die( esc_html( "{$post_type} is not a valid Post Type." ) );
```
Chouby, line 178: Please remove the capitals for "**p**ost **t**ype".

Chouby (changes requested):

PR #810, opened 2021-03-22
Refactor method sidebars_widgets 

Marianne380:
Refactor method `sidebars_widgets` to separate it into several methods in order to re-use them in `widget-editor-language-attribute.php` to avoid duplicated code.

See : https://github.com/polylang/polylang-pro/pull/922#discussion_r597571113

Link to this PR : https://github.com/polylang/polylang-pro/pull/922 and so issue https://github.com/polylang/polylang-pro/issues/860

### DIFF
--- a/frontend/frontend-filters.php
+++ b/frontend/frontend-filters.php
@@ -237,9 +237,9 @@ class PLL_Frontend_Filters extends PLL_Filters {
 	 *
 	 * @since 3.1
 	 *
-	 * @param array $sidebars_widgets An associative array of sidebars and their widgets
+	 * @param array $sidebars_widgets        An associative array of sidebars and their widgets
 	 * @param  array $wp_registered_widgets  Array of all registered widgets.
-	 * @return array|null             An associative array of sidebars and their widgets or nothing.
+	 * @return array|null                    An associative array of sidebars and their widgets or nothing.
 	 */
 	public function init_sidebars_widgets( $sidebars_widgets, $wp_registered_widgets ) {
 		if ( empty( $wp_registered_widgets ) ) {
--- a/frontend/frontend-filters.php
+++ b/frontend/frontend-filters.php
@@ -17,6 +17,11 @@ class PLL_Frontend_Filters extends PLL_Filters {
 	public $cache;
 
 	/**
+	 * @var string $cache_key Cache key.
+	 */
+	public $cache_key;
+
+	/**
 	 * Constructor: setups filters and actions
 	 *
 	 * @since 1.2
@@ -161,16 +166,7 @@ class PLL_Frontend_Filters extends PLL_Filters {
 	public function sidebars_widgets( $sidebars_widgets ) {
 		global $wp_registered_widgets;
 
-		if ( empty( $wp_registered_widgets ) ) {
-			return $sidebars_widgets;
-		}
-
-		$cache_key         = md5( maybe_serialize( $sidebars_widgets ) );
-		$_sidebars_widgets = $this->cache->get( "sidebars_widgets_{$cache_key}" );
-
-		if ( false !== $_sidebars_widgets ) {
-			return $_sidebars_widgets;
-		}
+		$this->init_sidebars_widgets( $sidebars_widgets );
 
 		foreach ( $sidebars_widgets as $sidebar => $widgets ) {
 			if ( 'wp_inactive_widgets' === $sidebar || empty( $widgets ) ) {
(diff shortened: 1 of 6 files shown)


frontend/frontend-filters.php
```diff
@@ -161,44 +166,105 @@ public function widget_display_callback( $instance ) {
...
 
 		return $sidebars_widgets;
 	}
 
+	/**
+	 * Init widgets sidebars, checks if they already exist in the cache to use them later.
+	 *
+	 * @param array $sidebars_widgets An associative array of sidebars and their widgets
+	 *
+	 * @return mixed|void
```
manooweb, line 200: sure of the @return documentation ?

manooweb (changes requested):

frontend/frontend-filters.php
```diff
@@ -161,42 +166,141 @@ public function widget_display_callback( $instance ) {
...
+		}
 
-		$this->cache->set( "sidebars_widgets_{$cache_key}", $sidebars_widgets );
+		$this->cache_key         = md5( maybe_serialize( $sidebars_widgets ) );
+		$_sidebars_widgets = $this->cache->get( "sidebars_widgets_{$this->cache_key}" );
 
-		return $sidebars_widgets;
+		if ( false !== $_sidebars_widgets ) {
+			return $_sidebars_widgets;
+		}
```
raaaahman, line 250: Now the method has no default return value, which could be troublesome for filters set by others plugins on the same hook.

raaaahman (changes requested):
Explicit methods names and helpful documentation.

However, I'm worried that changing the return value of a filter might cause bugs in third-party plugins hooked on the same hooks...

frontend/frontend-filters.php
Marianne380: See the following commits.

raaaahman: Since we are supposed to receive an array, maybe we should return an array as well? That way, plugins that hook up later on will not risk to break because of an incorrect type.

frontend/frontend-filters.php
```diff
@@ -161,42 +166,146 @@ public function widget_display_callback( $instance ) {
 	public function sidebars_widgets( $sidebars_widgets ) {
 		global $wp_registered_widgets;
 
-		if ( empty( $wp_registered_widgets ) ) {
-			return $sidebars_widgets;
+		$existing_sidebars_widgets = $this->init_sidebars_widgets( $sidebars_widgets, $wp_registered_widgets );
+		if ( $existing_sidebars_widgets ) {
+			return $existing_sidebars_widgets;
 		}
```
raaaahman, line 172: Should this be different than https://github.com/polylang/polylang-pro/blob/bcd43b2ebde21d6d8f34483e2cd2ff2ab4651887/modules/block-editor/widget-editor-language-attribute.php#L79 ?

frontend/frontend-filters.php
Marianne380: This part of the code was supposed to quit the method : https://github.com/polylang/polylang/blob/master/frontend/frontend-filters.php#L164-L173

frontend/frontend-filters.php
Marianne380: Not really, it's an omission, as this is the part that is used to exit the method.

frontend/frontend-filters.php
```diff
@@ -237,17 +237,20 @@ public function sort_widgets_sidebars( $sidebars_widgets, $wp_registered_widgets
...
 	 */
-	public function init_sidebars_widgets( $sidebars_widgets, $wp_registered_widgets ) {
+	public function init_sidebars_widgets( $sidebars_widgets, $wp_registered_widgets, $instance = null ) {
 		if ( empty( $wp_registered_widgets ) ) {
 			return $sidebars_widgets;
 		}
 
-		$this->cache_key         = md5( maybe_serialize( $sidebars_widgets ) );
-		$_sidebars_widgets = $this->cache->get( "sidebars_widgets_{$this->cache_key}" );
+		$instance = $instance ? $instance : $this;
```
raaaahman, line 250: This is needed only because of `PLL_Widgets_Editor_Frontend_Fitlers`.

Since we have decided to use one cache per class, maybe this separation into function should be thought out again?

raaaahman (approved):

manooweb (approved):
Ok. 👍

raaaahman:
> The `sidebars_widgets()` method in Polyang Pro is a copy of the method with the same name in this class, except for the call to the callback `handle_widget_in_sidebar_callback()` so I suggest to move all the code of the method in `sort_widgets_sidebars()` to avoid repeating yourself.

This has previously provoked a cache issue:

- The `PLL_Frontend_Filters::sort_widgets_sidebars()` is called from the same instance in Polylang Pro: https://github.com/polylang/polylang-pro/blob/4c7570afc67753d1ba1fd2766c7d5075058996bd/modules/block-editor/widget-editor-frontend-filters.php#L79
- So when the cache is checked, it already contains the md5 of the sidebars widgets: https://github.com/polylang/polylang/blob/d5c0d8bc97af144892ee7cdb4f4c4d19fe56a32a/frontend/frontend-filters.php#L173
- Thus the widgets blocks are not filtered.

This is solvable by either:

1. Changing the value of the cache key to be different in each call.
2. Extracting the `sort_widgets_sidebars()` and related methods to a `PLL_Frontend_Filters_Widgets` class that is subclassed in Polylang Pro: the two classes will then ben separatedly instantiated, and won't share their respective caches.
3. Eventually do both.

PR #820, opened 2021-04-07
Add a new method to check the modification of function copied from WordPress

raaaahman:
## Before

We used to verify that the WordPress functions we have copied didn't change from one WordPress version to another.

However, we could inadvertedly change from within our files a function that have been copied.

## Changes

Add a test method to check if our copied methods / functions have been changed or not. If we need to indeed change one of these functions, we would then have to update the md5 in the tests.

Need to be merged before https://github.com/polylang/polylang-pro/pull/931


manooweb (approved):
OK 👍

PR #821, opened 2021-04-08
Change Admin menu Background Color

sebastienserre:
Add a background color on "Show All Languages" admin menu if it's filtered on a language.
It will maybe avoid some support request from customers who do not find their content.

Edit >For the latest modifications See [the comment](https://github.com/polylang/polylang/pull/821#issuecomment-854504920) below

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -456,22 +456,22 @@ abstract class PLL_Admin_Base extends PLL_Base {
 		 * @param array $items The admin languages filter submenu items.
 		 */
 		$items = apply_filters( 'pll_admin_languages_filter', array_merge( array( $all_item ), $this->model->get_languages_list() ) );
+
 		$menu = array(
 			'id'    => 'languages',
 			'title' => $selected->flag . $title,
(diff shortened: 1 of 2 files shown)


sebastienserre:
All Languages shown:
show-all-languages
Language filtered:
languages-filtered

sebastienserre:
Effectively, the color is not OK with the "Sunrise" Admin Colour Scheme.
sunrise
it's OK with the Warning Orange color:
screenshot-2021-04-15-16-52

manooweb:
I think the red color code you use first wasn't the Polylang red backgroud color `a03f3f` because it works pretty well with it

image

In your last proposition with the sunrise theme, why didn't you use the WordPress orange sunrise background color `dd823b` ?

image

It seems to me we should keep the white text color whatever the background we will choose espcially if we use the light theme

image

image

manooweb:
The Polylang red `a03f3f` seems to be the right choice.
image

See https://wpsyntex.slack.com/archives/C017GJQBZP1/p1620050707007400

For the light I decided to also override the text color with white instead of grey
image

image

For all the other themes it's already in white

raaaahman (approved):
Tested with [Wave tool](https://wave.webaim.org/), no low contrast reported.

PR #824, opened 2021-04-13
Insert pll_ajax_backend parameter in a new empty object

manooweb:
This PR fixes the [HS #14681 ticket](https://secure.helpscout.net/conversation/1481323694/14681?folderId=860979) (private)

options.data could be undefined when it come from [jQuery Form plugin](https://github.com/jquery-form/form) espacially when it is used to upload a file because [processData](https://github.com/jquery-form/form/blob/v4.3.0/src/jquery.form.js#L396) is set to false and [data to null](https://github.com/jquery-form/form/blob/v4.3.0/src/jquery.form.js#L424)

So we need to insert pll_ajax_backend parameter in a new empty object instead to avoid the javascript error which blocks the uploading process.

**Note** that by [jQuery Form plugin implementation](https://github.com/jquery-form/form/blob/v4.3.0/src/jquery.form.js#L430-L434) the pll_ajax_backend parameter will be never sent to the server

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -317,7 +317,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 										addPolylangParametersAsString();
 									} else {
 										// Otherwise options.data is probably an object.
-										options.data = Object.assign( options.data, <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?> );
+										options.data = Object.assign( options.data || {} , <?php echo $arr; // phpcs:ignore WordPress.Security.EscapeOutput ?> );
 									}
 								}
 							}


raaaahman (approved):

Chrystll:
There is no regression with the fix that we done in the 3.0.3 "Fix an ajax conflict with WooCommerce Tree Table Rate Shipping and HubSpot All-In-One Marketing" https://github.com/polylang/polylang/pull/803. It doesn't break the registration and the connection with the hubspot plugin.
As a reminder here is how to tested:
1. create an account on https://www.hubspot.fr/
2. go to Dashboard > Plugins > activated Polylang and HubSpot All-In-One Marketing - Forms, Popups, Live Chat https://wordpress.org/plugins/leadin/
3. go to Dashboard > Hubspot > User guide > select your account > click on "Continue with this account" > you get "Your website is now connected to hubspot" as shown below 
screenshot-trunk-test com-2021 04 23-09_16_30 (1)
and no more a 400 bad request in the console as shown here:
111620065-8ba4e700-87e6-11eb-888f-fe749668bf72

PR #830, opened 2021-05-03
Prevent race condition when deleting images

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/853

The goal of `PLL_CRUD_Posts::wp_delete_file()` is to prevent WP to delete image files when only a media is deleted and at least one media translation is kept.

The previous method was taken from https://wordpress.org/plugins/file-gallery/. The process was as follows:
1. WP deletes the attachment in database.
2. WP deletes all resized files.
3. WP attempts to deletes the main file.
4. Polylang looks for the attachments using the main file. If it find ones, it re-generates the resized files and tells WP not to delete the main file.

The issue is that when deleting the files from the grid view, media are deleted individually with an ajax request per media. This causes a race condition with a request regenerating the resized files while anoth would have delete them if they had existed.

The new method proposed in this PR is as follows:
1. WP deletes the attachment in database.
2. WP attempts to delete the resized files.
3. Polylang gets the main file name from the resized file name and looks for the attachments using the main file. If it find ones it tells WP not to delete the file.
4. WP attempts to deletes the main file.
5. Polylang looks for the attachments using the main file. If it find ones it tells WP not to delete the file.

### DIFF
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -228,8 +228,7 @@ class PLL_CRUD_Posts {
 	}
 
 	/**
-	 * Prevents WP deleting files when there are still media using them
-	 * Thanks to Bruno "Aesqe" Babic and its plugin file gallery in which I took all the ideas for this function
+	 * Prevents WP deleting files when there are still media using them.
 	 *
 	 * @since 0.9
 	 *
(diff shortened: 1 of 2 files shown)


raaaahman (approved):
After discussion:

- We couldn't find the WordPress function that assigns names to resized files.
- There is possibility to query the complete resized file names with: `SELECT post_id FROM $wpdb->postmeta WHERE meta_key = '_wp_attached_file_metadata' and meta_value LIKE %s"`. Although this may be less performant.
- The code was already tested by https://github.com/polylang/polylang/blob/214d980f353aa51424923f67d531ca390889ce5a/tests/phpunit/tests/test-media.php#L39

manooweb (approved):
OK 👍

PR #831, opened 2021-05-04
Set curlang as current post language when cleaning the post cache

### LINKED ISSUE #828: During cron, use most recently used language for post type archive links
## Prerequisites

1. Create a custom post type in WordPress
2. In Polylang, set the URL modifications setting to "_The language is set from the directory name in pretty permalinks_" and uncheck the "_Hide URL language information for default language_" option. Then enable translation for the new post type in the "_Custom post types and Taxonomies_" setting.
3. Create a post with the new post type and publish it.
4. Create a WP Cron job that prints the URL of the new post (`get_permalink($post)`) and the URL of the corresponding post type archive (`get_post_type_archive_link($post->post_type)`) to the error_log.

When I follow those steps, I see the correct URL for the custom post in the error log, but the language folder is missing from the URL of the corresponding post type archive.

I'd like to see the localized URL of the post type archive that belongs to the custom post.

### Use Case

A lot of sites are using caching plugins like Cache Enabler, WP Rocket, or WP Super Cache to improve their [Core Web Vitals](https://web.dev/vitals/). When a new blog post is published, these plugins offer the option to either clear the whole cache or only associated pages, like previous and next posts as well as term and post type archive pages. The latter option is preferrable on larger sites that are updated several times a day, since it's computationally expensive to regenerate the whole cache and difficult to keep the cache warm if it is wiped every time a new blog post gets pulished.

### Root Problem

Unfortunately, it's difficult for the caching plugins to determine which language's post type archive should actually be cleared if a new post is published. Everything still works if you publish the post right away. In that case, `PLL_Filters_Links->post_type_archive_link()` can determine the language from the current post or the current user's session. But if you schedule the post to be published later, `PLL_Filters_Links->post_type_archive_link()` will be executed in the context of a WP Cron job with no language attached to it (`PLL_Filters_Links->$curlang` will be empty). Consequently, `get_post_type_archive_link()` cannot be localized anymore, and the caching plugins get the wrong URL to purge.

### Suggested Solution

I don't see a clean, straightforward solution to this fundamental problem, but a workaround that would at least serve as a remedy for the abovementioned caching scenario. My suggestion is to keep track of the language that was last used in the context of a post type and to fall back to it if no current language can be determined. 

To give you a better idea, here's what I'm currently using to work around the post type archive link issue on my system (in combination with Cache Enabler):

```php
add_filter( 'pll_post_type_link', function( $link, $lang ) {
    global $polylang_recent_lang;
    if ( ! empty( $lang) && wp_doing_cron() ) {
        $polylang_recent_lang = $lang;
    }
    return $polylang_recent_lang;
}, 10, 2 );


add_filter( 'post_type_archive_link', function( $link ) {
    global $polylang, $polylang_recent_lang;
    if (       ! empty( $polylang ) 
            &&   empty( $polylang->curlang ) 
            && ! empty( $polylang_recent_lang ) 
            &&   wp_doing_cron() ) {
        $polylang->curlang = $polylang_recent_lang;
    }
    return $link;
}, 19);
```
## Thank You

Since this is my first comment on the Polylang project on GitHub, let me use this opportunity to thank you for your work and to commend you on Polylang's exemplary code quality. It's a pleasure to work with! 

I'm responsible for a fairly large site (10k daily visits, 12k posts) and we gladly pay the annual Pro license even though we also have a lifetime WPML license, which we used in the past. We switched because of reliability issues and we wanted the same slugs across all languages (not sure right now, whether WPML has (had) limitations in this regard). We also need per-language user rights and looked into MultilingualPress for this purpose, but didn't switch since that plugin seems to have lower market adoption and we feared that we'd run into more compatibility issues because of this.

Chouby:
Fixes #828 

When a custom post is published, cache plugins purge the post cache but also the post type archive cache.
See for examples:
* WP Rocket: https://github.com/wp-media/wp-rocket/blob/v3.8.8/inc/common/purge.php#L63
* Cache enabler: https://github.com/keycdn/cache-enabler/blob/1.7.1/inc/cache_enabler.class.php#L1217 

Our [`post_type_archive_link` filter](https://github.com/polylang/polylang/blob/3.0.4/include/filters-links.php#L198) uses `$curlang` to evaluate the language of the link. So we thus must ensure that `PLL()->curlang` is equal to the language of the post being published.

In this PR, I propose to set `PLL()->curlang` in a method hooked to `clean_post_cache`. WP Rocket uses this hook and it is expected that other cache plugins uses this hook too. Note that Cache enabler uses `save_post`, but this shouldn't be an issue as this hooked is fired after `clean_post_cache`.

### DIFF
--- a/integrations/cache/cache-compat.php
+++ b/integrations/cache/cache-compat.php
@@ -24,6 +24,8 @@ class PLL_Cache_Compat {
 		if ( ! defined( 'WP_ROCKET_VERSION' ) || version_compare( WP_ROCKET_VERSION, '3.0.5', '<' ) ) {
 			add_action( 'wp', array( $this, 'do_not_cache_site_home' ) );
 		}
+
+		add_action( 'clean_post_cache', array( $this, 'clean_post_cache' ), 1 );
 	}
 
 	/**
@@ -68,6 +70,22 @@ class PLL_Cache_Compat {
 	public function do_not_cache_site_home() {
 		if ( ! defined( 'DONOTCACHEPAGE' ) && PLL()->options['browser'] && PLL()->options['hide_default'] && is_front_page() && pll_current_language() === pll_default_language() ) {
 			define( 'DONOTCACHEPAGE', true );
+		}
+	}
+
+	/**
+	 * Defines the current language as the current post language when cleaning the post cache.
+	 *
+	 * This allows cache plugins to clean the right post type archive cache.
+	 *
+	 * @since 3.0.5
+	 *
+	 * @param int $post_id Post id.
+	 */
+	public function clean_post_cache( $post_id ) {
+		$lang = PLL()->model->post->get_language( $post_id );
+		if ( $lang ) {
+			PLL()->curlang = $lang;
 		}
 	}
 }


manooweb (approved):
Ok 👍

PR #835, opened 2021-05-06
Support session cookie with the pll_cookie_expiration filter

### LINKED ISSUE #834: Set session coookie
## Prerequisites

- [ ] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [ ] This bug happens with a default WordPress theme active

Not necessary an issue

Be able to set a sesison cookie

I was expecting...

To be able to set a session cookie :) 

Hey, it will be great if we can pass a `-1` arg or smtg like this to `add_filter( 'pll_cookie_expiration', function() { return -1; } );` and be able to set a session cookie 

then, in `includes/cookies.php` i suppose it will be smtg like:

```
$expiration = apply_filters( 'pll_cookie_expiration', YEAR_IN_SECONDS );

$defaults = array(
        'expires'  => ($expiration > 0) ? time() + $expiration : '',
        'path'     => COOKIEPATH,

```

and ... that's all

currently, we update this file in place, at every update :) 

thanks

Chouby:
Fixes #834

This PR allows to pass 0 to the `pll_cookie_expiration` filter, thus setting a session cookie instead of persistent cookie.

### DIFF
--- a/include/cookie.php
+++ b/include/cookie.php
@@ -10,7 +10,7 @@
  */
 class PLL_Cookie {
 	/**
-	 * Parses the cookie parameters
+	 * Parses the cookie parameters.
 	 *
 	 * @since 2.9
 	 *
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👍

PR #836, opened 2021-05-06
Pass custom classes to the link elements in PLL_Walker_List

raaaahman:
## Before

`PLL_Walker_List::start_el()` could set custom classes on the `li` elements but not its inner `a` element.

## Changes

The `$arg['classes']` parameter could contain nested arrays. These array are used to pass classes to the child elements.

## Going further

- Is the nested array a good idea? Should we use a `'child_classes'` argument instead?
- Should we implement that behavior for `PLL_Walker_Dropdown` as well?


raaaahman:
Replaced by #846

PR #837, opened 2021-05-11
Avoid javascript errors when a plugin define its own editor for translated post types

raaaahman:
## Before

Polylang only loads javascript related to the metabox on an a translated post type editor. However, if a plugin overrides WordPress editors for a translated post type, our script would be loaded, even if the target HTML elements are missing. That's what happened in: https://secure.helpscout.net/conversation/1509819229/15063?folderId=860976

## Changes

Chain jQuery methods instead of accessing directly the properties. If the selected DOM node does not exists at some point of the chain, these methods will return an empty jQuery object, so the chain will continue and only the final result will be `undefined`.

## Going further

- [x] ~~Do we need to adress this in the `initializeConfirmationModal()` function as well? And in other scripts?~~

### DIFF
--- a/js/src/lib/confirmation-modal.js
+++ b/js/src/lib/confirmation-modal.js
@@ -94,6 +94,9 @@ export const initializeConfimationModal = () => {
 }
 
 export const initializeLanguageOldValue = () => {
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👍

PR #840, opened 2021-05-17
Allow to query comments by multiple languages

Chouby:
Unlike posts and terms, it was not possible to query comments by multiple languages. This PR solves this using the same way used for terms.

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -98,7 +98,7 @@ class PLL_Filters {
 	 * @since 2.0
 	 *
 	 * @param WP_Comment_Query $query  WP_Comment_Query object.
-	 * @return PLL_Language|false The language to use in the filter, false otherwise.
+	 * @return PLL_Language|string|false The language to use in the filter, false otherwise.
 	 */
 	protected function get_comments_queried_language( $query ) {
 		// Don't filter comments if comment ids or post ids are specified.
@@ -113,7 +113,12 @@ class PLL_Filters {
 			return false;
 		}
 
-		return empty( $query->query_vars['lang'] ) ? $this->curlang : $this->model->get_language( $query->query_vars['lang'] );
+		// If comments are queried with a 'lang' parameter.
+		if ( isset( $query->query_vars['lang'] ) ) {
+			return $query->query_vars['lang'];
+		}
+
+		return $this->curlang;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


include/filters.php
```diff
@@ -98,7 +98,7 @@ public function delete_sticky_posts_cache() {
 	 * @since 2.0
 	 *
 	 * @param WP_Comment_Query $query  WP_Comment_Query object.
-	 * @return PLL_Language|false The language to use in the filter, false otherwise.
+	 * @return PLL_Language|string|false The language to use in the filter, false otherwise.
```
raaaahman, line 101: We should always return an array, either empty, with one or several `PLL_Language` instances.

This would avoid type checking each time we want to use these values, and thus simplify the code. It would also improve PHPStan capability to highlight errors we could write in our code.

tests/phpunit/tests/test-filters.php
```diff
@@ -143,16 +143,24 @@ function test_get_comments() {
...
 
 		// don't use the same default args as above to avoid hitting the cache
 		$comments = get_comments( array( 'fields' => 'ids', 'lang' => 'en' ) );
 		$this->assertCount( 1, $comments );
-		$this->assertEquals( get_comment( $en )->comment_ID, reset( $comments ) );
+		$this->assertEquals( $en, reset( $comments ) );
+
+		$comments = get_comments( array( 'fields' => 'ids', 'lang' => 'en,fr' ) );
+		$this->assertCount( 2, $comments );
+		$this->assertEqualSets( array( $en, $fr ), $comments );
```
raaaahman, line 163: I don't believe the order is important here. Would two 'assertContains()` calls be more fitting?

raaaahman (commented):
Do we want to start using more `assertSame()` assertions?

tests/phpunit/tests/test-filters.php
Chouby: `assertEqualSets()` is not supposed to check the order. `assertEqualSetsWithIndex()` does.

include/filters.php
Chouby: I'ill review that.

tests/phpunit/tests/test-filters.php
raaaahman: Good to know. :)

raaaahman (approved):
As a general remark: beware of Primitive Obsession.

raaaahman (approved):

PR #841, opened 2021-05-17
enable post search box when create_post capability is disabled

Onatcer:
Currently it is not possible to link different language versions of Custom Post Types via the interface on the post itself when `create_posts` capabilities are not present for the custom post type. 

The input field is disabled when the add post link is present. That does make sense when you want to add a new translation from inside a post, however if `create_posts` capabilities are disabled, it's likely that they are only added programatically. However it is currently not possible to link the two posts. 

I hope I didn't overlook anything by just removing the disabled option, I tested that the creation of new posts via the interface is still not possible when  `create_posts` capabilities are not present and that the edit links work as expected.


Chouby:
I agree the this disabled field was a mistake. I will remove it for terms too.

PR #842, opened 2021-05-19
Adapt widgets script to work both with legacy widgets screen and block-based widgets screen

manooweb:
Closes https://github.com/polylang/polylang-pro/issues/838 and https://github.com/polylang/polylang-pro/issues/840

After the Gutenberg issue [#28668](https://github.com/WordPress/gutenberg/issues/28668) has been solved by the PR [#31485](https://github.com/WordPress/gutenberg/pull/31485) we'are able to use correctly the widget fields by their ids and then we need to adapt the `widgets.js` script to the new block-based widgets screen.

This PR is here to propose this adaptation.

## `Dropdown`, `Show names` and `Show flags` fields
In fact the legacy widget form is displayed when we click on the legacy widget block preview and thus the HTML DOM is accessible by the jQuery script. Nothing really change in this form and then the script can work as in the legacy widgets screen.

One thing change is the HTML structure of the page and the we need to adapt the code to point to the correct wrapper container to attach correctly the event listeners when we change the checkboxes value.
We need to choose a HTML element we're sure it is displayed at the very beginning of the page render: HTML element with the `edit-widgets-main-block-list` CSS class seems to the only choice.

So the behaviours are exactly the same as in the legacy widgets screen

When `Dropdown` is chosen, `Show names`, `Show flags` and `Hide current` fields aren't displayed
image

`Show names` and `Show Flags` fields can't be unchecked together
These two options are only possible

image
image

## Language field
At the opposite of the legacy widgets screen, it isn't the widgets forms which are displayed at the beginning when the page is loaded.

image

These previews are done for each widget in their own `iframe` element. So it's a bit complex to wait for the iframe loading and manipulate its HTML DOM to insert the flag near the widget title. In addition, the preview is done to have a render as close as possible of the frontend render (WYSYWYG). So it seems to me it wasn't the right solution.

So I decided to insert the flag near the widget title when its form is displayed. However the widget form isn't ready yet when the page is loaded as the legacy widgets screen.
So I decided to initialize the flag, if necessary, when we click on the legacy widget block preview which triggers the widget form display.

image

When we change the `Language choice` field we need to update the flag near the widget title. As for `Dropdown`, `Show names` and `Show flags` fields described above we need to adapt the script to point to the right wrapper container and the right title HTML element which isn't set at the same place as in the legacy widgets screen - it is now outside the HTML `widget` element.


js/src/widgets.js
```diff
@@ -73,19 +76,35 @@ jQuery(
...
+			);
+		} else {
+			// Update flags when we click on the legacy widget to display its form.
+			widgets_container.on(
+				'click',
+				'.wp-block-legacy-widget',
+				function() {
+					add_flag( $( this ).find( '.widget' ) );
+				}
+			);
```
Marianne380, line 106: Couldn't this be done here : https://github.com/polylang/polylang/blob/f5afeec0da1faf6b530f8dca50027a9099a74509/js/src/widgets.js#L79-L84

Marianne380 (approved):

PR #844, opened 2021-05-19
Improve settings descriptions and change default values

Chouby:
- Improves the description for the url modifications.
- Improves the description fr the Media option. The default value is now not to translate the media.
- Improves the detection for the browser language detection. The option is now deactivated by default.

### DIFF
--- a/settings/settings-browser.php
+++ b/settings/settings-browser.php
@@ -30,7 +30,7 @@ class PLL_Settings_Browser extends PLL_Settings_Module {
 			array(
 				'module'        => 'browser',
 				'title'         => __( 'Detect browser language', 'polylang' ),
-				'description'   => __( 'When the front page is visited, set the language according to the browser preference', 'polylang' ),
+				'description'   => __( 'When the front page is visited, redirects to the homepage in the browser preferred language. As this doesn\'t work if the front page is cached, we will attempt to disable the front page cache for known cache plugins.', 'polylang' ),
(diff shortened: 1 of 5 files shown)


settings/settings-browser.php
```diff
@@ -30,7 +30,7 @@ public function __construct( &$polylang ) {
 			array(
 				'module'        => 'browser',
 				'title'         => __( 'Detect browser language', 'polylang' ),
-				'description'   => __( 'When the front page is visited, set the language according to the browser preference', 'polylang' ),
+				'description'   => __( 'When the front page is visited, redirects to the homepage in the browser preferred language. As this doesn\'t work if the front page is cached, we will attempt to disable the front page cache for known cache plugins.', 'polylang' ),
```
raaaahman, line 33: We should probably word it like `When the front page is visited, redirects to the front page...`, since 'front page' and 'home page' may refers to different pages when 'Static pages' are selected in _Settings > Reading_

manooweb: It seems we are all agree with the sentences

> When the front page is visited, redirects to itself in the browser preferred language. As this doesn't work if it is cached, Polylang will attempt to disable the front page cache for known cache plugins.
 
See https://wpsyntex.slack.com/archives/C017GJQBZP1/p1622733983070000?thread_ts=1622551579.029600&cid=C017GJQBZP1

PR #846, opened 2021-05-21
Pass custom classes to the link elements in PLL_Walker_List 

raaaahman:
## Before

`PLL_Walker_List::start_el()` could set custom classes on the `li` elements but not its inner `a` element.

## Changes

- Adds a `link_classes` entry in the `$args` array to pass classes to the `a` element in the output list.

Replaces #836 

## Going further

- We could use the recursive behavior of WordPress `Walker` to handle separately the `li` from the nested `a` elements. This could allow to nest more elements if needed.


Marianne380 (approved):

include/switcher.php
```diff
@@ -126,7 +126,14 @@ protected function get_elements( $args ) {
...
 			$locale = $language->get_locale( 'display' );
-			$classes = array( 'lang-item', 'lang-item-' . $id, 'lang-item-' . esc_attr( $slug ) );
+			$item_classes = array( 'lang-item', 'lang-item-' . $id, 'lang-item-' . esc_attr( $slug ) );
+			$classes = isset( $args['classes'] ) && is_array( $args['classes'] ) ?
+				array_merge(
+					$item_classes,
+					$args['classes']
+				) :
+				$item_classes;
+			$link_classes = isset( $args['link_classes'] ) ? $args['link_classes'] : array();
```
raaaahman, line 136: This new parameter needs to be documented.

include/switcher.php
```diff
@@ -169,7 +176,7 @@ protected function get_elements( $args ) {
 				$first = false;
 			}
 
-			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes' );
+			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
```
raaaahman, line 179: This new output field needs to be documented.

PR #848, opened 2021-05-22
Image optimization with ImageAlpha (png8+alpha) and ImageOptim.

lowwebtech:
Hi,

I just did an image optimization to reduce the weight of the images and the plugin.
It reduces images weights by 276kB.

PNG images are converted to PNG8 (256 colors often less), alpha is preserved, also optimized with ImageOptim.

Hope it helps


Chouby:
Thank you. Yes it helps! Even for base 64 encoded flags, the improvement is significant.

PR #850, opened 2021-05-26
Don't add language columns when editing a term

sebastienserre:
Return empty columns to avoid Warning (on WC Brands)

### DIFF
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -247,10 +247,13 @@ class PLL_Admin_Filters_Columns {
 	 * @return string[] modified List of columns.
 	 */
 	public function add_term_column( $columns ) {
-		if ( empty( $columns ) ) {
-			return $columns;
+		$screen = get_current_screen();
+
+		if ( 'edit-tags' === $screen->base ) { // Don't add our columns when editing a term.
+			return $this->add_column( $columns, 'posts' );
 		}
-		return $this->add_column( $columns, 'posts' );
+
+		return $columns;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


Chouby:
The issue is that the screen to add a term (with the terms list table) and the screen to edit a term both have the same screen id and thus both apply the same filter `manage_edit-{$taxonomy}_columns`. We want to add the filter on the first screen and not in the second one. Indeed this leads to a warning with WC Brands (due to a misuse of their own filter returning nothing) but it also adds useless languages checkboxes in the screen option for all taxonomies. I udpated the PR title accordingly.

manooweb (approved):
OK 👍

Chrystll:
Without Polylang:

screenshot-trunk-test com-2021 06 02-11_57_46

Before the PR with Polylang:
screenshot-trunk-test com-2021 06 02-11_59_36



With the PR, we no longer have the "screen options" containing the languages and we no longer have the warning below.

```
PHP Warning:  array_keys() expects parameter 1 to be array, null given in /home/christelle/Documents/git/polylang-pro/vendor/wpsyntex/polylang/admin/admin-filters-columns.php on line 78
PHP Stack trace:
PHP   1. {main}() /wp-admin/term.php:0
PHP   2. require_once() /wp-admin/term.php:73
PHP   3. WP_Screen->render_screen_meta() /wp-admin/admin-header.php:266
PHP   4. WP_Screen->show_screen_options() /wp-admin/includes/class-wp-screen.php:956
PHP   5. get_column_headers() /wp-admin/includes/class-wp-screen.php:995
PHP   6. apply_filters() /wp-admin/includes/screen.php:37
PHP   7. WP_Hook->apply_filters() /wp-includes/plugin.php:212
PHP   8. PLL_Admin_Filters_Columns->add_term_column() /wp-includes/class-wp-hook.php:292
PHP   9. PLL_Admin_Filters_Columns->add_column() /polylang-pro/vendor/wpsyntex/polylang/admin/admin-filters-columns.php:25
```

PR #852, opened 2021-05-28
Update Composer dependencies

Chouby:
This PR updates Composer dependencies.
- It fixes a PHPStan false positive just by changing the order of instructions.
- It fixes a wrong usage of `esc_html_e()` reported by VIP WPCS.
- It ignores a VIP WPCS recommendation.


manooweb (approved):
Ok 👍
Should we consider get_pages() as a get_posts() ?

For information
https://github.com/Automattic/VIP-Coding-Standards/issues/369
https://github.com/Automattic/VIP-Coding-Standards/issues/672

Chouby:
> Should we consider get_pages() as a get_posts() ?

Generally no. These are 2 completely different functions. `get_posts()` uses `WP_Query`. `get_pages()` does not. Unfortunately... See also https://core.trac.wordpress.org/ticket/12821. But we can query pages with `get_posts()`.

PR #853, opened 2021-06-02
Allow to have multiple language switchers in dropdown mode for nav menu

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/899

### Problem :

When there is multiple language switchers, the first one that is displayed miss one language.

The menu order wasn't incremented for the parent item. So it erased  a submenu item in case of several dropdown during the rendering of the menu by the `wp_nav_menu()` [function](https://github.com/WordPress/WordPress/blob/fde119d446e5680e49e024195db10a0d7ffd6991/wp-includes/nav-menu-template.php#L201).

### Solution :

Increment the parent item menu order to avoid that.

### DIFF
--- a/frontend/frontend-nav-menu.php
+++ b/frontend/frontend-nav-menu.php
@@ -135,12 +135,10 @@ class PLL_Frontend_Nav_Menu extends PLL_Nav_Menu {
 					$lang_item->classes = $lang['classes'];
 					if ( ! empty( $options['dropdown'] ) ) {
 						$lang_item->menu_order = $item->menu_order + $i;
-					} else {
-						$lang_item->menu_order += $offset;
-					}
-					if ( ! empty( $options['dropdown'] ) ) {
 						$lang_item->menu_item_parent = $item->db_id;
 						$lang_item->db_id = 0; // to avoid recursion
+					} else {
+						$lang_item->menu_order += $offset;
 					}
 					$new_items[] = $lang_item;
 					$offset++;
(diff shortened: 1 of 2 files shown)


Screenfeed (approved):
I didn't understood the thing about `$offset` and `$i` but since:
- 2 menu items don't have the same `menu_order` and,
- WP doesn't need consecutive `menu_order` numbers,

It's fine.

Hug0-Drelon:
Edit to get the correct `menu_order`, otherwise it doesn't display items correctly as noticed @Chouby.

frontend/frontend-nav-menu.php
```diff
@@ -119,26 +118,34 @@ public function wp_get_nav_menu_items( $items ) {
...
-					$lang_item->menu_order += $offset + $i++;
+					if ( ! empty( $options['dropdown'] ) ) {
+						$lang_item->menu_order = $item->menu_order + $i;
+					} else {
+						$lang_item->menu_order += $offset;
+					}
 					if ( ! empty( $options['dropdown'] ) ) {
 						$lang_item->menu_item_parent = $item->db_id;
 						$lang_item->db_id = 0; // to avoid recursion
 					}
```
Screenfeed, line 144: We can make it a bit simpler ;)

```suggestion
					if ( ! empty( $options['dropdown'] ) ) {
						$lang_item->menu_item_parent = $item->db_id;
						$lang_item->menu_order = $item->menu_order + $i;
						$lang_item->db_id = 0; // to avoid recursion
					} else {
						$lang_item->menu_order += $offset;
					}
```

Screenfeed (changes requested):

frontend/frontend-nav-menu.php
Hug0-Drelon: Yes indeed ! Thanks @Screenfeed , I committed your suggestion :)

Screenfeed (approved):

PR #854, opened 2021-06-03
Fix block editor preload path compatibility with WordPress 5.8

raaaahman:
## Before

We hooked a filter on `block_editor_preload_paths` to preload languages data during Block Editor loading. This hook has been deprecated in WordPress 5.8 alpha.

## Changes

- Create a wrapper for the preload paths filters that check on WordPress version and read the arguments accordingly.
    - Register our filter on the correct hook depending on WordPress version.
    - Transform the arguments received by the filters from a `WP_Block_Editor_Context` (new argument) into a `WP_Post` (used in our legacy filters).
    - Is used several times inside Polylang Pro (see https://github.com/polylang/polylang-pro/pull/1005)
- Replace filter registration in `PLL_Admin_Block`
- Add tests for the `PLL_Admin_Block::preload_paths()` method.
- Tests the `PLL_Block_Editor_Preload_Paths` class independently, so we don't have to change the tests in Polylang Pro.

### DIFF
--- a/admin/admin-block-editor.php
+++ b/admin/admin-block-editor.php
@@ -22,11 +22,6 @@ class PLL_Admin_Block_Editor {
 	protected $pref_lang;
 
 	/**
-	 * @var PLL_Block_Editor_Filter_Preload_Paths
-	 */
-	private $block_editor_filter_preload_paths;
-
-	/**
 	 * Constructor: setups filters and actions
 	 *
 	 * @since 2.5
@@ -37,7 +32,7 @@ class PLL_Admin_Block_Editor {
 		$this->model     = &$polylang->model;
 		$this->pref_lang = &$polylang->pref_lang;
 
-		$this->block_editor_filter_preload_paths = new PLL_Block_Editor_Filter_Preload_Paths( array( $this, 'preload_paths' ), 10, 2 );
+		new PLL_Block_Editor_Filter_Preload_Paths( array( $this, 'preload_paths' ), 10, 2 );
 	}
 
 	/**
--- a/admin/admin-block-editor.php
+++ b/admin/admin-block-editor.php
@@ -32,7 +32,7 @@ class PLL_Admin_Block_Editor {
 		$this->model     = &$polylang->model;
 		$this->pref_lang = &$polylang->pref_lang;
 
-		add_filter( 'block_editor_preload_paths', array( $this, 'preload_paths' ), 10, 2 );
+		$this->block_editor_filter_preload_paths = new PLL_Block_Editor_Filter_Preload_Paths( array( $this, 'preload_paths' ), 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


admin/admin-block-editor.php
```diff
@@ -32,7 +37,7 @@ public function __construct( &$polylang ) {
 		$this->model     = &$polylang->model;
 		$this->pref_lang = &$polylang->pref_lang;
 
-		add_filter( 'block_editor_preload_paths', array( $this, 'preload_paths' ), 10, 2 );
+		$this->block_editor_filter_preload_paths = new PLL_Block_Editor_Filter_Preload_Paths( array( $this, 'preload_paths' ), 10, 2 );
```
raaaahman, line 40: Don't need to store this instance in a variable.

admin/block-editor-filter-preload-paths.php
```diff
@@ -0,0 +1,62 @@
...
+	 *
+	 * @param callable $callback         Function or method to be executed when the filter is triggered.
+	 * @param int      $priority         Priority of execution for the given callback. Default 10.
+	 * @param int      $arguments_number Number of arguments to pass to the callback. Default 1.
+	 */
+	public function __construct( $callback, $priority = 10, $arguments_number = 1 ) {
+		$this->callback = $callback;
+		$this->arguments_number = $arguments_number;
+
+		if ( version_compare( $GLOBALS['wp_version'], '5.8-alpha', '<' ) ) {
```
raaaahman, line 35: We should check for the existence of `WP_Block_Editor_Context` class instead of WordPress versions (the change of filter names was introduced in an alpha version, but not present in the previous alphas).

admin/block-editor-filter-preload-paths.php
```diff
@@ -0,0 +1,62 @@
...
+	 * @param WP_Block_Editor_Context $block_editor_context The post resource data.
+	 * @return array|mixed|string[] (string|string[])[]
+	 */
+	public function block_editor_rest_api_preload_paths( $preload_paths, $block_editor_context ) {
+		if ( $this->arguments_number > 1 && null === $block_editor_context->post ) {
+			return $preload_paths;
+		}
+
+		// Only two cases: one or two arguments
+		$args = 1 === $this->arguments_number ? array( $preload_paths ) : array( $preload_paths, $block_editor_context->post );
```
raaaahman, line 58: We should set the `$block_editor_context` to `null` by default and not check on the number of arguments.

tests/phpunit/tests/test-block-editor-preload-paths.php
```diff
@@ -0,0 +1,83 @@
...
+			->with(
+				$this->isType( 'array' )
+			);
+
+		if ( version_compare( $GLOBALS['wp_version'], '5.8-alpha', '<' ) ) {
+			apply_filters( 'block_editor_preload_paths', array( '/' ), new WP_Post( new stdClass() ) );
+		} else {
+			block_editor_rest_api_preload( array( '/' ), new WP_Block_Editor_Context( array( 'post' => new WP_Post( new stdClass() ) ) ) );
+		}
+	}
```
raaaahman, line 33: Test that the second argument is always null.

Issue #493, opened 2020-04-07
After updating to 2.7 lang is forced for custom post type API endpoints


bartczyz:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] I can reproduce this bug consistently

## Problem

After updating to **Polylang PRO 2.7** the plugin stopped respecting "Custom post types and Taxonomies" settings. For example, if the setting for our custom post type was un-checked in **2.6.10** then the call would return the response, even tho the lang parameter was specified in the call:
![image](https://user-images.githubusercontent.com/1191500/78672156-8e157600-78e0-11ea-84ce-1e664211d07e.png)

And when I changed the setting and checked that I would like the custom post type to be translatable:
![image](https://user-images.githubusercontent.com/1191500/78672301-c026d800-78e0-11ea-9535-a8a1e7dc35b4.png)

Then the call is not returning anything as expected, since the content was not translated or added to any language:
![image](https://user-images.githubusercontent.com/1191500/78672419-e77da500-78e0-11ea-9115-eea4ee6ec262.png)

But after updating to **Polylang PRO 2.7** this option is ignored completely. And when I do API call after I unchecked the box, I get no content in return:

![image](https://user-images.githubusercontent.com/1191500/78672692-49d6a580-78e1-11ea-8732-59eca50a3e52.png)

~~Also, why when I'm using **"Show all languages"** to add content, only content for the default language is shown (Media Library, Latest Posts widget etc.)? This should show content for all languages. This is confusing, especially to the editors that don't know which language is a default one.~~

Chrystll:
Regarding your REST API issue it seems that it has been introduced by this commit "Set query vars lang value with the REST API lang parameter" https://github.com/polylang/polylang-pro/commit/bc77510ed2f13ea5b3dee05526312253550c7704. We are investigating it. Regarding your issue second issue about the "Show all languages" filte, could you please open a ticket on our [helpdesk](https://polylang.pro/support/)?  I would need that you show me your issue thanks to a video so that I understand exactly what is happening.

PR #855, opened 2021-06-04
Prevent setting a language filter when we should not

Chouby:
Related #493, https://github.com/polylang/polylang-pro/pull/486, https://github.com/polylang/polylang-pro/pull/985

This PR introduced `PLL_Query::can_filter_query()` to check if we can add a language filter in the query. This new method is used internally by `set_language()` (new) and `filter_query()` (which was already using the code moved to the new method).

This allows to use boths methods without prior checks and is a better fix for #493.

### DIFF
--- a/include/query.php
+++ b/include/query.php
@@ -105,6 +105,10 @@ class PLL_Query {
 	 * @return void
 	 */
 	public function set_language( $lang ) {
+		if ( ! $this->can_filter_query() ) {
+			return;
+		}
+
 		// Defining directly the tax_query ( rather than setting 'lang' avoids transforming the query by WP )
 		$lang_query = array(
 			'taxonomy' => 'language',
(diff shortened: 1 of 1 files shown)


include/query.php
```diff
@@ -105,6 +105,10 @@ public function get_queried_taxonomies() {
 	 * @return void
 	 */
 	public function set_language( $lang ) {
+		if ( ! $this->can_filter_query() ) {
+			return;
+		}
```
raaaahman, line 110: With this return here, it doesn't seem worth going into the `! $this->is_already_fitlered( $qvars )` condition in `PLL_Query::filter_query()` when `PLL_Query::can_filter_query()` returns `false`.

raaaahman (commented):
I believe that is simpler solution would be to have one class for filtered queries and another for unfiltered queries.

- `PLL_Filtered_Query::filter_query()` method would set the taxonomy query.
- `PLL_Unfiltered_Query::filter_query()` method would remove the `lang` parameter from WordPress query

What is puzzling me at the moment is the reason we sometimes call to the `filter_query()` method and sometimes bypass it to call it directly to the `set_language` method (from the `PLL_Choose_Lang` child classes).

Moreover, it seems that both calls could happen for a same query, and I don't see a reason for that...

PR #860, opened 2021-06-09
Let filter block_editor_rest_api_preload_paths unchanged when no post is provided

Chouby:
in #858, we decided not to propagate the filter when `block_editor_rest_api_preload_paths` is used with 2 parameters and no post is provided in the context. This introduces a limitation for https://github.com/polylang/polylang-pro/pull/1010 where we are not able to correctly differentiate the widget screen and the post editor with an untranslated post type.

I then appears better to propagate the filter with all the parameters allowing us to test the context.

### DIFF
--- a/admin/block-editor-filter-preload-paths.php
+++ b/admin/block-editor-filter-preload-paths.php
@@ -39,8 +39,7 @@ class PLL_Block_Editor_Filter_Preload_Paths {
 	/**
 	 * Filters the array of REST API paths that will be used to preloaded common data to use with the block editor.
 	 *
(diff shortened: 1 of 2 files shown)


manooweb (approved):
Ok 👍
with some doubts about what the test PLL_Block_Editor_Filter_Preload_Paths_Test::test_invoke_filter_with_one_argument is for

Chouby:
The test fails when you comment out this line: https://github.com/polylang/polylang/blob/425c6b88ae450cdda3665bf6f808e7f10551b461/admin/block-editor-filter-preload-paths.php#L52

It checks that when you require only one parameter (by default or in the 3rd param of the constructor), you receive only one parameter in the function hooked to the filter.

PR #861, opened 2021-06-10
PHP Notice: Trying to get property 'post_type' of non-object

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1012
Related #854, #858, #860

With the deprecation of the filter `block_editor_preload_paths`, we must now check the instance of the object received in the second parameter of `PLL_Admin_Block_Editor::preload_paths()`


manooweb (approved):
Ok. 👍

PR #867, opened 2021-06-14
Fix switcher preview in widget screen

Marianne380:
See : https://github.com/polylang/polylang-pro/issues/987#issuecomment-859352976

closes https://github.com/polylang/polylang-pro/issues/987

### DIFF
--- a/tests/phpunit/tests/test-switcher.php
+++ b/tests/phpunit/tests/test-switcher.php
@@ -25,6 +25,9 @@ class Switcher_Test extends PLL_UnitTestCase {
 		$this->frontend = new PLL_Frontend( $links_model );
 		$this->frontend->init();
 
+		$this->admin = new PLL_Admin( $links_model );
+		$this->admin->init();
+
 		// De-activate cache for links
 		$this->frontend->links->cache = $this->getMockBuilder( 'PLL_Cache' )->getMock();
 		$this->frontend->links->cache->method( 'get' )->willReturn( false );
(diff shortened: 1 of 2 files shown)


manooweb (changes requested):
LGTM! However could we have a unit test?

tests/phpunit/tests/test-switcher.php
```diff
@@ -25,6 +25,9 @@ function setUp() {
 		$this->frontend = new PLL_Frontend( $links_model );
 		$this->frontend->init();
 
+		$this->admin = new PLL_Admin( $links_model );
+		$this->admin->init();
```
Chouby, line 29: May I suggest to add this only in the relevant test?

tests/phpunit/tests/test-switcher.php
```diff
@@ -186,4 +189,28 @@ function test_dropdown() {
...
+
+	function test_with_hide_if_no_translation_option_in_admin_context() {
+		$en = $this->factory->post->create();
+		self::$model->post->set_language( $en, 'en' );
+
+		$fr = $this->factory->post->create();
+		self::$model->post->set_language( $fr, 'fr' );
+
+		self::$model->clean_languages_cache(); // FIXME for some reason, I need to clear the cache to get an exact count
+		$this->admin->links->curlang = self::$model->get_language( 'en' );
```
Chouby, line 201: Why do you need a `$curlang` property?

Chouby (changes requested):

tests/phpunit/tests/test-switcher.php
Marianne380: We do have a `curlang` in the widget editor which is the default language, but this is not needed here.

tests/phpunit/tests/test-switcher.php
```diff
@@ -186,4 +186,31 @@ function test_dropdown() {
...
+
+		$fr = $this->factory->post->create();
+		self::$model->post->set_language( $fr, 'fr' );
+
+		self::$model->clean_languages_cache(); // FIXME for some reason, I need to clear the cache to get an exact count
+
+		$args['hide_if_no_translation'] = 1;
+		$switcher = $this->switcher->the_languages( $this->admin->links, $args );
+
+		$doc = new DomDocument();
```
manooweb, line 206: May I suggest to add an `assertNotEmpty` assertion to test `$switcher` to make the test fail and not to an error

```
F                                                                   1 / 1 (100%)

Time: 9.58 seconds, Memory: 46.00MB

There was 1 failure:

1) Switcher_Test::test_with_hide_if_no_translation_option_in_admin_context
Failed asserting that a string is not empty.
```

instead of
```
E                                                                   1 / 1 (100%)

Time: 10.8 seconds, Memory: 46.00MB

There was 1 error:

1) Switcher_Test::test_with_hide_if_no_translation_option_in_admin_context
DOMDocument::loadHTML(): Empty string supplied as input
```

manooweb (changes requested):

manooweb (approved):
ok 👍

manooweb (approved):
Ok 👍

PR #870, opened 2021-06-15
Don't disable the translations input fields for terms

Chouby:
This PR is the equivalent of #841 for terms.

Let's imagine that a user doesn't have the right to edit the content in German. This doesn't mean that he/she should be unable to link an existing German term with the English term being edited.


Chouby:
There is no need to change anything. The input field is disabled only or default terms. User capabilities have no impact.

PR #872, opened 2021-06-16
Fixing multiple dropdown switchers rendering both for blocks and widget

Hug0-Drelon:
Fix issues https://github.com/polylang/polylang-pro/issues/894 and https://github.com/polylang/polylang-pro/issues/895

## Problem 
The Javascript code included with each switcher was flawed because of the non-unique ids attributes.

## Solution
We chose to isolate the Javascript code and loop through the different switcher to listen to each one of them individually. We also pass the urls in value attribute instead of a slug.

## To go further
As mentionned before, ids aren't unique (see bellow). Wich doesn't affect our cases but maybe bring problems if someone wants to use it in the future (for CSS styling for instance).

block-switchers

Also, id and for attributes aren't matching each other in the widget switcher (see bellow). In term of accessibility, screen reader won't be able to read correctly.

widget-switchers

We can note that we doesn't manage accessibility in the rendering of a block switcher. Should we implement it ?


admin/admin-classic-editor.php
```diff
@@ -110,10 +110,11 @@ public function post_language() {
...
 			array(
-				'name'     => $id,
-				'class'    => 'post_lang_choice tags-input',
-				'selected' => $lang ? $lang->slug : '',
-				'flag'     => true,
+				'name'       => $id,
+				'class'      => 'post_lang_choice tags-input',
+				'selected'   => $lang ? $lang->slug : '',
+				'flag'       => true,
+				'value' => 'slug',
```
Marianne380, line 117: Indent here.

Marianne380 (commented):
LGTM

Marianne380 (approved):

Hug0-Drelon:
After a talk with @Chouby, it appears that we shouldn't use a JS file that meant to be rendered on frontend. We noticed that the initial problem was due to a non unique ID, so we have to resolve this at first.
Also, there is no label for accessibility for the switcher block and this need to be done for consistency with the widget.

manooweb (approved):
Ok 👍

admin/admin-classic-editor.php
```diff
@@ -110,10 +110,11 @@ public function post_language() {
...
 			array(
-				'name'     => $id,
-				'class'    => 'post_lang_choice tags-input',
-				'selected' => $lang ? $lang->slug : '',
-				'flag'     => true,
+				'name'       => $id,
+				'class'      => 'post_lang_choice tags-input',
+				'selected'   => $lang ? $lang->slug : '',
+				'flag'       => true,
+				'value'      => 'slug',
```
Chouby, line 117: You can rever this change. The default value for `value` is already `slug`.
See: https://github.com/polylang/polylang/blob/3.1/include/walker-dropdown.php#L103

include/walker-dropdown.php
```diff
@@ -31,12 +31,12 @@ class PLL_Walker_Dropdown extends Walker {
 	 * @return void
 	 */
 	public function start_el( &$output, $element, $depth = 0, $args = array(), $current_object_id = 0 ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
-		$value = $args['value'];
+		$value_type = $args['value'];
 		$output .= sprintf(
 			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
-			esc_attr( $element->$value ),
+			'slug' === $value_type ? esc_attr( $element->$value_type ) : esc_url( $element->$value_type ),
```
Chouby, line 37: You should check for `url` instead, as this is the new type of values that you are introducing which requires a specific escaping function `esc_url()`. As is, all types of values would be escaped by `esc_url()` except slugs. Here is an example of usage non-slug non-url value: https://github.com/polylang/polylang/blob/3.1/admin/admin-filters-term.php#L137

Chouby (changes requested):

include/walker-dropdown.php
Hug0-Drelon: Thanks, I didn't thought about other value than slug or url...

PR #879, opened 2021-06-22
Validate the redirect links before redirecting on frontend

Chouby:
In 58dd54e4b3112141ce8c028e348349e0c74fb190 (version 2.6) we replaced all usages of `wp_redirect()` by `wp_safe_redirect()` as per WPCS recommendation (`WordPress.Security.SafeRedirect` sniff).

On the other hand,  we sometimes have reports of Polylang redirecting to `/wp-admin/`. It's quite difficult to explain to users that the root cause of this issue is not Polylang redirecting to `/wp-admin/` (because of our usage of `wp_safe_redirect()`) but the underlying redirect link which was not validated by WordPress because of the usage of an unknown host.

My first intention was to ignore the WordPress sniff and use `wp_redirect()` on frontend. I however now propose to validate the redirect link before calling `wp_safe_redirect()`. That way, if the redirect link is not valid, then no redirects is made by Polylang. And we still use `wp_safe_redirect()` to avoid open redirects and please WPCS.

The only downside I see in this proposal is that `wp_validate_redirect()` is called twice, once directly and once by `wp_safe_redirect()`.

### DIFF
--- a/frontend/choose-lang.php
+++ b/frontend/choose-lang.php
@@ -261,7 +261,7 @@ abstract class PLL_Choose_Lang {
 			if ( $redirect = apply_filters( 'pll_redirect_home', $redirect ) ) {
 				$this->maybe_setcookie();
 				header( 'Vary: Accept-Language' );
-				wp_safe_redirect( $redirect, 302, POLYLANG );
+				wp_redirect( $redirect, 302, POLYLANG );
 				exit;
 			}
 		}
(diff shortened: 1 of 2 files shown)


manooweb (approved):
Ok 👍
As I said on slack https://wpsyntex.slack.com/archives/C017SSTK4KY/p1624352970057200 this solution looks good to me.

PR #881, opened 2021-06-22
Define a language in REST context

Screenfeed:
Whenever a `lang` key is sent in `$_REQUEST`, define a current language as a `PLL_REST_Request->curlang` public property, like in other contexts.
The property can be an instance of `PLL_Language`, or `false` if something goes wrong, or `null` if no `lang` arg is sent.
Fixes [PLL Pro 949](https://github.com/polylang/polylang-pro/issues/949).

### DIFF
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -71,8 +71,6 @@ class PLL_REST_Request extends PLL_Base {
 		}
 
 		if ( ! empty( $this->curlang ) ) {
-			$GLOBALS['text_direction'] = $this->curlang->is_rtl ? 'rtl' : 'ltr';
-
 			/** This action is documented in frontend/choose-lang.php */
 			do_action( 'pll_language_defined', $this->curlang->slug, $this->curlang );
 		} else {
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -9,34 +9,39 @@
  * @since 2.6
  */
 class PLL_REST_Request extends PLL_Base {
+	/**
+	 * @var PLL_Language|false|null A `PLL_Language` when defined, `false` otherwise. `null` until the language
+	 *                              definition process runs.
+	 */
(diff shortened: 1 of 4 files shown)


manooweb (approved):
Ok 👍
Notice that:
- add scripts in composer.json could be in conflict with the PR #898 
- the page list block is correctly filtered by the REST request language when it is added however when we change the language in the metabox the block isn't filtered by the new language because the block render isn't rerun when the post is saved. We need to refresh the admin edit post page.

Screenfeed:
@manooweb Should I remove the composer scripts from this PR? They are useless now with #898 and may prevent a conflict.

Screenfeed:
Done.

manooweb (approved):

include/rest-request.php
```diff
@@ -50,23 +55,41 @@ class PLL_REST_Request extends PLL_Base {
...
-				$this->static_pages = new PLL_Static_Pages( $this );
+			if ( empty( $this->curlang ) && ! empty( $this->options['default_lang'] ) && is_string( $this->options['default_lang'] ) ) {
+				// A lang has been requested but it is invalid, let's fall back to the default one.
+				$this->curlang = $this->model->get_language( sanitize_key( $this->options['default_lang'] ) );
 			}
+		}
 
-			$this->links = new PLL_Admin_Links( $this );
+		if ( ! empty( $this->curlang ) ) {
+			$GLOBALS['text_direction'] = $this->curlang->is_rtl ? 'rtl' : 'ltr';
```
Chouby, line 74: As we don't flter the locale, we should not change the text direction.

Chouby (changes requested):

include/rest-request.php
Screenfeed: Done in a6700ee.

PR #883, opened 2021-06-24
Display translation icon in old widget editor when no language is selected for a widget

Hug0-Drelon:
Related to https://github.com/polylang/polylang-pro/issues/926

## Enhancement 
Add a translation icon when no language is selected. See bellow.
Capture d’écran 2021-06-24 à 14 54 04

## Process
Add the dashicon for translation when no language is selected in `widget.js`. It needs a bit of margin in `admin.css` to be consistent with the flags.

*Note that is works with live preview in WP customizer with those modifications*

### DIFF
--- a/css/src/admin.css
+++ b/css/src/admin.css
@@ -463,3 +463,8 @@
 		line-height: 20px !important;
 	}
 }
+
+/* manage translation icon in widget title for old screen */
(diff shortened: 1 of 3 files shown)


manooweb:
As we decided in DM today the icon pushed on the widget title bar can be confusing.
So we give up this PR.

PR #885, opened 2021-06-28
Delay the copy of terms and post metas

Chouby:
This is a follow-up of #884 and necessary for https://github.com/polylang/polylang-pro/pull/1040 to have all actions fired in the correct order.


manooweb (approved):
Ok. 👍

PR #886, opened 2021-06-28
FIx the default category displayed in the Categories

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/1036

There was a **missing** `_`.

So the filter called was `option_defaultcategory` instead of `option_default_category`.
So we never pass in `option_default_term` method which return the default category in the filtered language.

### DIFF
--- a/tests/phpunit/tests/test-default-term.php
+++ b/tests/phpunit/tests/test-default-term.php
@@ -142,7 +142,6 @@ class Default_Term_Test extends PLL_UnitTestCase {
 
 	function test_get_option_default_category() {
 		$this->pll_admin->pref_lang = self::$model->get_language( 'es' );
-		$this->pll_admin->set_current_language();
 
 		$option = get_option( 'default_category' );
 		$option_lang = self::$model->term->get_language( $option );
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-default-term.php
```diff
@@ -139,4 +139,14 @@ function test_default_category_in_list_table() {
 		$this->assertFalse( strpos( $list, 'edit-tags.php?action=delete&amp;taxonomy=category&amp;tag_ID=' . $fr . '&amp;' ) );
 		$this->assertNotFalse( strpos( $list, 'edit-tags.php?action=delete&amp;taxonomy=category&amp;tag_ID=' . $id . '&amp;' ) );
 	}
+
+	function test_get_option_default_category() {
+		$this->pll_admin->pref_lang = self::$model->get_language( 'es' );
+		$this->pll_admin->set_current_language();
```
manooweb, line 145: This line seems useless to me.
In fact we never have the condition to set the current language in this test and we always fall in the `pll_no_language_defined` case.

See https://github.com/polylang/polylang/blob/master/admin/admin-base.php#L339-L377

manooweb (changes requested):

manooweb (approved):
Ok 👍

PR #887, opened 2021-06-28
FIx category feed not redirected when the language code is wrong

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/819

## Problem 
*Context : two language created (`en` and `fr`) with one category in english but not translated in french, containing some posts*

When requesting a untranslated category feed (for instance `/fr/category/uncategorized/feed/` ), no redirection is made toward the correct feed (i.e. with the correct language slug, like `/en/category/uncategorized/feed/` ).

## Solution
If no post is found for the wrong URL and if the query is for a category feed (`is_category()` and `is_feed()`), we can set the language to the language corresponding to requested category. So, later in `PLL_Frontend_Filters_Links::check_canonical_url()`, the `pll_check_canonical_url` filter will set the correct language slug in the redirect url.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -392,6 +392,11 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 			}
 		}
 
+		elseif ( $this->links_model->using_permalinks && is_category() && ! empty( $this->wp_query()->query['cat'] ) ) {
+			// When we receive a plain permaling with a cat query var, we need to redirect to the pretty permalink.
+			extract( $this->get_queried_term_language( true ) );
+		}
+
 		elseif ( is_category() || is_tag() || is_tax() ) {
 			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
 				if ( $this->links_model->using_permalinks && ( ! empty( $this->wp_query()->query['cat'] ) || ! empty( $this->wp_query()->query['tag'] ) ) ) {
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👍

frontend/frontend-filters-links.php
```diff
@@ -407,6 +407,13 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
 			if ( ! empty( $obj ) && $this->model->is_translated_taxonomy( $obj->taxonomy ) ) {
 				$language = $this->model->term->get_language( (int) $obj->term_id );
 			}
+
+			if ( is_feed() && empty( $obj ) ) { // Allows to replace the language correctly in a category feed query.
+				if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
+					$term_id = $this->get_queried_term_id( $this->wp_query()->tax_query );
+					$language = $this->model->term->get_language( $term_id );
+				}
```
Chouby, line 415: I suggest to avoid the repetition of the same code below.

tests/phpunit/tests/test-canonical.php
```diff
@@ -459,4 +459,26 @@ public function test_sitemap_with_category() {
...
+		$post_id = $this->factory->post->create(
+			array(
+				'post_title' => 'test',
+				'post_type'  => 'page',
+				'category'   => $cat_id,
+			)
+		);
+		self::$model->post->set_language( $post_id, 'en' );
+
+		$this->assertCanonical( '/fr/category/test-category/feed/', '/en/category/test-category/feed/' );
```
Chouby, line 482: You can  use the common posts and categories and replace everything by:
```php
$this->assertCanonical( '/fr/category/parent/feed/', '/en/category/parent/feed/' );
```

Chouby (changes requested):

frontend/frontend-filters-links.php
Hug0-Drelon: Indeed, I will refactor.

PR #889, opened 2021-06-30
Add test to improve coverage for PLL_Admin_Default_Term

Marianne380:
This method https://github.com/polylang/polylang/blob/3.1-beta1/admin/admin-default-term.php#L250-L267 wasn't covered by a test.

### DIFF
--- a/tests/phpunit/tests/test-default-term.php
+++ b/tests/phpunit/tests/test-default-term.php
@@ -157,6 +157,7 @@ class Default_Term_Test extends PLL_UnitTestCase {
 	function test_update_term_when_updating_default_language() {
 		$option = get_option( 'default_category' );
 		$option_lang = self::$model->term->get_language( $option );
+		$es_option = self::$model->term->get_translation( $option, 'es' );
 
 		$this->assertEquals( 'en', $option_lang->slug );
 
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/test-default-term.php
```diff
@@ -139,4 +139,19 @@ function test_default_category_in_list_table() {
...
 	}
+
+	function test_update_term_when_updating_default_language() {
+		$option = get_option( 'default_category' );
+		$option_lang = self::$model->term->get_language( $option );
+
+		$this->assertEquals( 'en', $option_lang->slug );
+
+		$es = self::$model->get_language( 'es' );
+		self::$model->update_default_lang( $es->slug );
```
Screenfeed, line 150: Wouldn't it be simpler to do this?
```suggestion
		self::$model->update_default_lang( 'es' );
```

Screenfeed (approved):

Chouby (changes requested):
The method `update_default_lang()` is doing more than what you check. Your test would passes with any Spanish category becoming default. In fact, the new default category is the Spanish translation of the previous default category. Could you check that too?

Chouby (approved):

PR #892, opened 2021-07-02
Consistent margin between flag and select for post and media

Hug0-Drelon:
## Problem

The margin between the flag and the `select` wasn't consistent in post edit and media edit. The theses [lines](https://github.com/polylang/polylang/blob/2dab3ad9997a5d1f302078d535ce5fecfed08fcc/css/src/admin.css#L229).

## Changes

Set the same margin for both post and media. Note that I added margin for the `p` inside the translation title because it wasn't handle automatically by Wordpress (like in Classic Editor, [see](https://github.com/WordPress/WordPress/blob/473295ab123b35ed54104b599f12f9e8b5e97283/wp-admin/css/common.css#L307)) in the block editor. See bellow the original display :

Capture d’écran 2021-07-02 à 15 20 07

### DIFF
--- a/css/src/admin.css
+++ b/css/src/admin.css
@@ -214,16 +214,16 @@
 	padding: 2px;
 }
 
+#post-translations p {
+	margin: 1em 0;
+}
+
 #post-translations .spinner,
 #term-translations .spinner {
 	float: none;
 	margin: 0;
 	background-position: center;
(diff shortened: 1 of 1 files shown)


manooweb (approved):
Ok after discussion 👍

PR #894, opened 2021-07-06
Fix RTL display in admin sidebar for translations table title

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1042

## Problem
RTL wasn't handled in the sidebar. Translations table title was displayed on the left for RTL languages.

## Fix
Modify CSS to handle this.

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -121,7 +121,7 @@ class PLL_Admin_Classic_Editor {
 
 		// NOTE: the class "tags-input" allows to include the field in the autosave $_POST ( see autosave.js )
 		printf(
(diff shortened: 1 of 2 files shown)


manooweb:
Could we fix other style small bugs?

- classic editor with RTL language, the margins between the dropdown and the flag aren't correct.
image
- legacy metabox in post block editor, as above and the top margin between the translations title and the dropdown isn't correct  too.
image

Notice it is already the case for LTR language
image

It seems there is a specific post block editor style which set this margin to 0: `.interface-complementary-area p`

Hug0-Drelon:
Margin is now correct for the flags as well as the margin for titles.

manooweb (approved):
OK 👍

PR #896, opened 2021-07-07
Fix radio button display for small screen

Hug0-Drelon:
## Problem
Radio button wasn't well displayed on small screen. See :
screenshot-trunk-test com-2021 07 07-11_15_50

## Solution
Remove the rules concerning radio button. Also, WP seems to have a default style for [them](https://github.com/WordPress/WordPress/blob/473295ab123b35ed54104b599f12f9e8b5e97283/wp-admin/css/forms.css#L1417).

### DIFF
--- a/css/src/admin.css
+++ b/css/src/admin.css
@@ -30,11 +30,6 @@
 	content: "\25cf";
 }
 
-.form-field input[type="radio"] {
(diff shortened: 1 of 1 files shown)


manooweb (approved):
OK 👍

PR #897, opened 2021-07-08
Fix margin for labels in add language form

Hug0-Drelon:
## Problem
Lack of margin for the labels.
Capture d’écran 2021-07-08 à 11 15 49

## Changes
For consistency with other forms (export string translations form for instance), add the same margin that WP apply to its forms ([cf](https://github.com/WordPress/WordPress/blob/473295ab123b35ed54104b599f12f9e8b5e97283/wp-admin/css/forms.css#L777)) . See the display now :
Capture d’écran 2021-07-08 à 11 46 24

*Note that the margin is applied to all `<label>` in the form and not only these with a radio button.*

### DIFF
--- a/css/src/admin.css
+++ b/css/src/admin.css
@@ -8,8 +8,12 @@
 }
 
 .pll-legend {
-	margin: 0.35em 0 0.5em;
+	display: block;
 	padding: 2px 0;
+	color: #1d2327;
+	font-weight: 400;
+	text-shadow: none;
+	margin: 0.35em 0 0.5em;
 }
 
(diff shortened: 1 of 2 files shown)


Hug0-Drelon:
As discussed with @Chouby, I applied the margin to `<legend>`tag too. I used the same class as in PLL-Pro for the export form ([cf](https://github.com/polylang/polylang-pro/blob/d5601b13f1240b81f9b9215c9077ced921fca511/modules/export/view-tab-export-strings.php#L63))

PR #898, opened 2021-07-08
Add composer commands to launch all our tools the same way

manooweb:
The goal of this PR is to propose composer commands to launch each of our tools the same way:

| Tool | Command | Description
| -------- | ---- | ----- |
| PHPUnit | composer test | Runs PHPUnit tests. |
| PHPCS | composer cs | Runs PHPCS linter. |
| PHPStan | composer stan | Runs PHPStan analysis. |
| PHPCS & PHPStan | composer lint | Runs both PHPCS and PHPStan. |
| Build project | composer build | Builds the project. |
| Distribute project | composer dist | Make the zip file to distibute the project release. |

**Note** 
1. It is possible to pass parameters to the tool by using the `--` separator between the composer command and the tool parameter.

For example:

- to run PHPUnit only on one class or method:
  - `composer test -- --filter Admin_Test`
  - `composer test -- --filter Admin_Test::test_admin_bar_menu`
- to run PHPCS only on a folder or file
  -  `composer cs -- admin/`
  - `composer cs -- admin/admin-base.php`

2. the `readme.md` file has been updated to take in account the new `composer build` command.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,9 +28,9 @@
 		]
 	},
 	"scripts": {
-		"test":"phpunit",
-		"cs":"phpcs",
-		"stan": "phpstan analyze --memory-limit=400M",
+		"test":"vendor/bin/phpunit",
+		"cs":"vendor/bin/phpcs",
+		"stan": "vendor/bin/phpstan analyze --memory-limit=400M",
 		"lint": [
 			"@cs",
 			"@stan"
--- a/composer.json
+++ b/composer.json
@@ -26,5 +26,24 @@
 			"modules/",
 			"settings/"
 		]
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -26,5 +26,24 @@
 			"modules/",
 			"settings/"
 		]
+	},
+	"scripts": {
+		"test":"vendor/bin/phpunit",
```
Screenfeed, line 31: What do you think of this?

```suggestion
		"test":"vendor/bin/phpunit --colors=always",
```

Screenfeed (approved):

PR #901, opened 2021-07-15
Improve rss feed support with plain permalinks

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1031

## Changes

In a feed request context, set the language to retrieve the archive of the translated taxonomy. If pretty permalinks option is used, then redirect to the pretty link corresponding to the desired taxonomy feed. Also redirect to pretty link for tag request but no feed.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -392,8 +392,18 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 			}
 		}
 
-		elseif ( $this->links_model->using_permalinks && is_category() && ! empty( $this->wp_query()->query['cat'] ) ) {
-			// When we receive a plain permaling with a cat query var, we need to redirect to the pretty permalink.
+		elseif ( is_feed() && ( is_category() || is_tag() || is_tax() ) ) {
+			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
+				$term_id = $this->get_queried_term_id( $this->wp_query()->tax_query );
+				$language = $this->model->term->get_language( $term_id );
+				if ( $this->links_model->using_permalinks ) {
+					$redirect_url = $this->maybe_add_page_to_redirect_url( get_term_feed_link( $term_id, '' ) );
+				}
+			}
+		}
+
+		elseif ( $this->links_model->using_permalinks && ( is_category() || is_tag() || is_tax() ) && ( ! empty( $this->wp_query()->query['cat'] ) || ! empty( $this->wp_query()->query['tag'] )) ) {
+			// When we receive a plain permalink with a cat or tag query var, we need to redirect to the pretty permalink.
 			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
 				$term_id = $this->get_queried_term_id( $this->wp_query()->tax_query );
 				$language = $this->model->term->get_language( $term_id );
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -402,7 +402,7 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


frontend/frontend-filters-links.php
```diff
@@ -392,8 +392,18 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
+			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
+				$term_id = $this->get_queried_term_id( $this->wp_query()->tax_query );
+				$language = $this->model->term->get_language( $term_id );
+				if ( $this->links_model->using_permalinks ) {
+					$redirect_url = $this->maybe_add_page_to_redirect_url( get_term_feed_link( $term_id, '' ) );
+				}
+			}
+		}
+
+		elseif ( $this->links_model->using_permalinks && ( is_category() || is_tag() || is_tax() ) && ( ! empty( $this->wp_query()->query['cat'] ) || ! empty( $this->wp_query()->query['tag'] ) ) ) {
```
Screenfeed, line 405: Is there a reason to use `$this->wp_query()->query['cat']` and `$this->wp_query()->query['tag']` here instead of `$this->wp_query()->tax_query`?

Screenfeed, line 405: Is there a reason to use `$this->wp_query()->query['cat']` and `$this->wp_query()->query['tag']` here instead of `$this->wp_query()->tax_query`?

Screenfeed: The advantage of `tax_query` is that it will work with things like `category_name` and `category__in`, etc. But I don't know if this is useful in this case.

Hug0-Drelon: In this case, we need to check if there is `cat` or `tag` in the query parameters (e.g. `www.example.com/?cat=foo&...`. As I understood, `tax_query` provides more information (as you said `category_name` for instance). So I think this is out of our concern here.

Screenfeed: OK, that’s what I thought. Thanks.

Screenfeed (approved):

frontend/frontend-filters-links.php
```diff
@@ -392,8 +392,18 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
-			// When we receive a plain permaling with a cat query var, we need to redirect to the pretty permalink.
+		elseif ( is_feed() && ( is_category() || is_tag() || is_tax() ) ) {
+			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
+				$term_id = $this->get_queried_term_id( $this->wp_query()->tax_query );
+				$language = $this->model->term->get_language( $term_id );
+				if ( $this->links_model->using_permalinks ) {
+					$redirect_url = $this->maybe_add_page_to_redirect_url( get_term_feed_link( $term_id, '' ) );
+				}
+			}
+		}
```
Chouby, line 403: We should be able to merge most of teh code of this test with the test below as the difference is only the link evaluated with `get_term_feed_link()` instead of `get_term_link()`.

tests/phpunit/tests/test-canonical.php
```diff
@@ -459,4 +459,48 @@ public function test_sitemap_with_category() {
...
+				'post_title' => 'test',
+				'post_type'  => 'page',
+				'category'   => $cat_id,
+			)
+		);
+		self::$model->post->set_language( $post_id, 'en' );
+
+		$this->assertCanonical( '/?cat=' . $cat_id . '&feed=rss2', '/en/category/test-category/feed/' );
+	}
+
```
Chouby, line 484: You can re-use the shared fixtures to simplify the test.

tests/phpunit/tests/test-canonical.php
```diff
@@ -459,4 +459,48 @@ public function test_sitemap_with_category() {
...
+			array(
+				'post_title' => 'test',
+				'post_type'  => 'page',
+				'tag'        => $tag_id,
+			)
+		);
+		self::$model->post->set_language( $post_id, 'en' );
+
+		$this->assertCanonical( '/?tag=test-tag&feed=rss2', '/en/tag/test-tag/feed/' );
+	}
```
Chouby, line 505: You can add the tag to the shared fixtures to and use it to keep only the assertion in the test.

Chouby (changes requested):

PR #905, opened 2021-08-10
Allow to remove the cookie with the pll_cookie expiration filter

### LINKED ISSUE #877: Allow cookie to be unset: follow-up to #834
## Prerequisites

- [X] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [ ] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [ ] This bug happens with a default WordPress theme active

This is a **feature request** and is a follow-up to #834.

It would be nice if `PLL_Cookie::parse_args()` allowed a negative integer returned from the `pll_cookie_expiration` filter to unset the cookie.

My use case for this is as follows:

We recently changed our Cookie Consent management solution.  Previously, the `pll_language` cookie was considered a **Strictly Necessary** cookie, so no consent was required.  Now, it is considered a **Functional** cookie (it probably should have been considered **Functional** all along, but that's besides the point).

So, we know have a lot of users who have had the cookie set (with the 1 year expiration) who haven't consented to the cookie and I need a way to unset it for those users.

Note that this same situation arises when a user has consented, but then _revokes_ that consent at some point...I need an _easy_ way to unset the cookie.

I have implemented a solution (in one my plugins) that works, but it's kind of complicated.   If it detects that consent has not been given, then it hooks into `wp` (such that my callback is called after `PLL_Choose_Lang::maybe_set_cookie()` is) and calls `setcookie()` with a timestamp in the past.

It would be much easier if I could just hook into `pll_cookie_expiration` and return a negative number, so that when Polylang calls `PLL_Cookie::set()`, Polylang itself will unset the cookie.

From the PHP side, the change would be as simple as changing `PLL_Cookie::parse_args()` to:

```php
$defaults = array(
	'expires'  => $expiration !== 0 ? time() + $expiration : 0,
	...
);
```

I haven't dug into the Polylang code that sets the cookie from JS, so there might be changes necessary there as well.

Hopefully, this makes sense.

Chouby:
Closes #877

### DIFF
--- a/include/cookie.php
+++ b/include/cookie.php
@@ -22,6 +22,7 @@ class PLL_Cookie {
 		 * Filters the Polylang cookie duration.
 		 *
 		 * If a cookie duration of 0 is specified, a session cookie will be set.
+		 * If a negative cookie duration is specified, the cookie is removed.
 		 * /!\ This filter may be fired *before* the theme is loaded.
 		 *
 		 * @since 1.8
@@ -31,7 +32,7 @@ class PLL_Cookie {
 		$expiration = apply_filters( 'pll_cookie_expiration', YEAR_IN_SECONDS );
 
 		$defaults = array(
-			'expires'  => $expiration > 0 ? time() + $expiration : 0,
+			'expires'  => $expiration !== 0 ? time() + $expiration : 0,
 			'path'     => COOKIEPATH,
 			'domain'   => COOKIE_DOMAIN, // Cookie domain must be set to false for localhost ( default value for COOKIE_DOMAIN ) thanks to Stephen Harris.
 			'secure'   => is_ssl(),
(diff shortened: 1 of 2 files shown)


include/cookie.php
```diff
@@ -31,7 +32,7 @@ protected static function parse_args( $args ) {
 		$expiration = apply_filters( 'pll_cookie_expiration', YEAR_IN_SECONDS );
```
Screenfeed, line 32: Since we're doing a strict comparison now instead of `>`, it would be good to typecast the filter's output:

```suggestion
		$expiration = (int) apply_filters( 'pll_cookie_expiration', YEAR_IN_SECONDS );
```

Screenfeed (changes requested):

Screenfeed (approved):

pbiron:
thank you!

PR #907, opened 2021-08-23
WPSEO: Use dynamic permalinks

### LINKED ISSUE #882: wpseo schema wrong url
## Prerequisites

1.Install polylang pro and wpseo last version and yoast test plugin
2.setup multilanguages with different domain (but maybe happen also with subdir conf)
3. monitor schema url

When I follow those steps, "sometimes" schema url of domain are wrong. It happen ONLY for item that are stored in indexable tables of yoast (recently intrduced), if I reset indexable tables (using yoast test plugin) they fix, but randomly they still appear wrong (maybe because putted in indexable tables from some async process) 

I was expecting the right schema url based on post languages

Chouby:
I luckily discovered the filter `wpseo_dynamic_permalinks_enabled` introduced in Yoast SEO 15.1. This is more or less what I asked for just before the indexables were introduced in the version 14.0.

This PR asks Yoast SEO to use dynamic permalinks. Overriding them should thus not be useful anymore. Hopefully using this filter should fix all permalinks issues once for all.

Unfortunately Yoast SEO doesn't honor the filter for breadcrumb links, so we still need to override them.

Moreover more tests demonstrate a bug for the static front page and blog page breadcrumbs which are also fixed by this PR.

Fixes #882 
Fixes https://github.com/polylang/polylang-pro/issues/1039

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -423,6 +423,10 @@ class PLL_WPSEO {
 				$breadcrumb_title = $breadcrumb_title ? $breadcrumb_title : $indexable->breadcrumb_title; // The option may be empty.
 				$indexable->breadcrumb_title = pll__( $breadcrumb_title );
 			}
+
+			if ( 'term' === $indexable->object_type && pll_is_translated_taxonomy( $indexable->object_sub_type ) ) {
+				$indexable->permalink = get_term_link( $indexable->object_id );
+			}
 		}
 
 		return $indexables;
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -17,6 +17,7 @@ class PLL_WPSEO {
 	 */
 	public function init() {
 		add_action( 'wp_loaded', array( $this, 'wpseo_translate_options' ) );
+		add_filter( 'wpseo_dynamic_permalinks_enabled', '__return_true', 999 );
 
 		if ( PLL() instanceof PLL_Frontend ) {
 			// Filters sitemap queries to remove inactive language or to get
@@ -366,7 +367,6 @@ class PLL_WPSEO {
 	public function frontend_presentation( $presentation ) {
 		switch ( $presentation->model->object_type ) {
 			case 'home-page':
-				$presentation->model->permalink = pll_home_url();
 				$presentation->model->title = WPSEO_Options::get( 'title-home-wpseo' );
 				$presentation->model->description = WPSEO_Options::get( 'metadesc-home-wpseo' );
 				$presentation->model->open_graph_title = WPSEO_Options::get( 'og_frontpage_title' );
@@ -375,14 +375,9 @@ class PLL_WPSEO {
 
 			case 'post-type-archive':
 				if ( pll_is_translated_post_type( $presentation->model->object_sub_type ) ) {
-					$presentation->model->permalink = get_post_type_archive_link( $presentation->model->object_sub_type );
 					$presentation->model->title = WPSEO_Options::get( 'title-ptarchive-' . $presentation->model->object_sub_type );
 					$presentation->model->description = WPSEO_Options::get( 'metadesc-ptarchive-' . $presentation->model->object_sub_type );
 				}
-				break;
-
-			case 'user':
-				$presentation->model->permalink = get_author_posts_url( $presentation->model->object_id );
 				break;
 
 			case 'system-page':
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -407,24 +407,18 @@ class PLL_WPSEO {
 	 */
 	public function breadcrumb_indexables( $indexables ) {
 		foreach ( $indexables as &$indexable ) {
-			// Handles both when the front page displays the list of posts or a static page.
 			if ( 'home-page' === $indexable->object_type || ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_on_front' ) === $indexable->object_id ) ) {
+				// Handles both when the front page displays the list of posts or a static page.
 				$indexable->permalink = pll_home_url();
 				$indexable->breadcrumb_title = pll__( WPSEO_Options::get( 'breadcrumbs-home' ) );
-			}
-
-			if ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_for_posts' ) === $indexable->object_id ) {
+			} elseif ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_for_posts' ) === $indexable->object_id ) {
 				$indexable->permalink = get_permalink( $indexable->object_id );
-			}
-
-			if ( 'post-type-archive' === $indexable->object_type && pll_is_translated_post_type( $indexable->object_sub_type ) ) {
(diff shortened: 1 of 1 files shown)


integrations/wpseo/wpseo.php
```diff
@@ -401,29 +396,36 @@ public function frontend_presentation( $presentation ) {
...
+			if ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_for_posts' ) === $indexable->object_id ) {
+				$indexable->permalink = get_permalink( $indexable->object_id );
+			}
+
+			if ( 'post-type-archive' === $indexable->object_type && pll_is_translated_post_type( $indexable->object_sub_type ) ) {
+				$indexable->permalink = get_post_type_archive_link( $indexable->object_sub_type );
+				$breadcrumb_title = WPSEO_Options::get( 'bctitle-ptarchive-' . $indexable->object_sub_type );
+				$breadcrumb_title = $breadcrumb_title ? $breadcrumb_title : $indexable->breadcrumb_title; // The option may be empty.
+				$indexable->breadcrumb_title = pll__( $breadcrumb_title );
+			}
```
Screenfeed, line 425: I'd suggest to use `break` at the end of each `if` statement, like this:
```suggestion
			if ( 'home-page' === $indexable->object_type || ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_on_front' ) === $indexable->object_id ) ) {
				$indexable->permalink = pll_home_url();
				$indexable->breadcrumb_title = pll__( WPSEO_Options::get( 'breadcrumbs-home' ) );
				break;
			}

			if ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_for_posts' ) === $indexable->object_id ) {
				$indexable->permalink = get_permalink( $indexable->object_id );
				break;
			}

			if ( 'post-type-archive' === $indexable->object_type && pll_is_translated_post_type( $indexable->object_sub_type ) ) {
				$indexable->permalink = get_post_type_archive_link( $indexable->object_sub_type );
				$breadcrumb_title = WPSEO_Options::get( 'bctitle-ptarchive-' . $indexable->object_sub_type );
				$breadcrumb_title = $breadcrumb_title ? $breadcrumb_title : $indexable->breadcrumb_title; // The option may be empty.
				$indexable->breadcrumb_title = pll__( $breadcrumb_title );
				break;
			}
```

integrations/wpseo/wpseo.php
```diff
@@ -401,29 +396,36 @@ public function frontend_presentation( $presentation ) {
...
+
+			if ( 'term' === $indexable->object_type && pll_is_translated_taxonomy( $indexable->object_sub_type ) ) {
+				$indexable->permalink = get_term_link( $indexable->object_id );
 			}
```
Screenfeed, line 429: What about the other cases, don't we need to do something? I see there are at least 3 other cases: `'system-page'` (404 and search results), `'user'`, and `'date-archive'`.

integrations/wpseo/wpseo.php
Chouby: I don't believe that we should break the loop, as we may have, for example, a link to the homepage and a link to a CPT archive in the same breadcrumb. Adding `continue` instead may be a slight optimzation to avoid further tests.

Screenfeed: You're right, that's what I wanted to do. Monday morning after holidays... 😑

integrations/wpseo/wpseo.php
Chouby: 'system-page' , 'user', and 'date-archive' should not appear as links in breadcrumbs. At least @Chrystll and I did not find options to add them (as we have for CPT archives or taxonomies).

Screenfeed (approved):

integrations/wpseo/wpseo.php
Chouby: I am proposing two optimizations.
1. The first one uses `elseif` instead of `continue` initially proposed, allowing to decrease teh number of tests in some cases.
```php
	public function breadcrumb_indexables( $indexables ) {
		foreach ( $indexables as &$indexable ) {
			// Handles both when the front page displays the list of posts or a static page.
			if ( 'home-page' === $indexable->object_type || ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_on_front' ) === $indexable->object_id ) ) {
				$indexable->permalink = pll_home_url();
				$indexable->breadcrumb_title = pll__( WPSEO_Options::get( 'breadcrumbs-home' ) );
			}

			elseif ( 'post' === $indexable->object_type && 'page' === $indexable->object_sub_type && get_option( 'page_for_posts' ) === $indexable->object_id ) {
				$indexable->permalink = get_permalink( $indexable->object_id );
			}

			elseif ( 'post-type-archive' === $indexable->object_type && pll_is_translated_post_type( $indexable->object_sub_type ) ) {
				$indexable->permalink = get_post_type_archive_link( $indexable->object_sub_type );
				$breadcrumb_title = WPSEO_Options::get( 'bctitle-ptarchive-' . $indexable->object_sub_type );
				$breadcrumb_title = $breadcrumb_title ? $breadcrumb_title : $indexable->breadcrumb_title; // The option may be empty.
				$indexable->breadcrumb_title = pll__( $breadcrumb_title );
			}

			elseif ( 'term' === $indexable->object_type && pll_is_translated_taxonomy( $indexable->object_sub_type ) ) {
				$indexable->permalink = get_term_link( $indexable->object_id );
			}
		}

		return $indexables;
	}
```
2. The second one returns to a `switch` as in the current version and is maybe the solution with the smallest number of tests. This comes with some price for the readability.
```php
	public function breadcrumb_indexables( $indexables ) {
		foreach ( $indexables as &$indexable ) {
			switch ( $indexable->object_type ) {
				case 'post':
					if ( 'page' !== $indexable->object_sub_type ) {
						break;
					}
					if ( get_option( 'page_for_posts' ) === $indexable->object_id ) {
						$indexable->permalink = get_permalink( $indexable->object_id );
						break;
					}
					if ( get_option( 'page_on_front' ) !== $indexable->object_id ) {
						break;
					}
					// Intentionally no break here as we are handling a front page (displaying posts).
				case 'home-page':
					$indexable->permalink = pll_home_url();
					$indexable->breadcrumb_title = pll__( WPSEO_Options::get( 'breadcrumbs-home' ) );
					break;
				case 'post-type-archive':
					if ( pll_is_translated_post_type( $indexable->object_sub_type ) ) {
						$indexable->permalink = get_post_type_archive_link( $indexable->object_sub_type );
						$breadcrumb_title = WPSEO_Options::get( 'bctitle-ptarchive-' . $indexable->object_sub_type );
						$breadcrumb_title = $breadcrumb_title ? $breadcrumb_title : $indexable->breadcrumb_title; // The option may be empty.
						$indexable->breadcrumb_title = pll__( $breadcrumb_title );
					}
					break;
				case 'term':
					if ( pll_is_translated_taxonomy( $indexable->object_sub_type ) ) {
						$indexable->permalink = get_term_link( $indexable->object_id );
					}
					break;
			}
		}

		return $indexables;
	}
```
What do you think?

Screenfeed: Even if I like `switch/case` statements in general, I don't think it's the right tool to use here, it decreases readability for the `'post'` case indeed.
Personally I prefer the first choice with `if/elseif`.

PR #908, opened 2021-08-26
Remove Yoast SEO backward compatibility code

Chouby:
Yoast SEO 17.0 has been released, When we will release Polylang 3.2, Yoast SEO 14.0 will be more than 18 months old. It's time to remove specific legacy code supporting older versions.

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -38,11 +38,7 @@ class PLL_WPSEO {
 			}
 
 			add_filter( 'pll_home_url_white_list', array( $this, 'wpseo_home_url_white_list' ) );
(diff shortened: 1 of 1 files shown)


Screenfeed (approved):
For future code removal, maybe we should deprecate old functions before removing them.

Chouby:
Sometimes I do it for internal functions when I expect that they could have been called directly by 3rd party developers. For example, here:https://github.com/polylang/polylang/blob/3.1.1/modules/sync/admin-sync.php#L189-L232.

Here I don't believe that there is any risk as I expect that the function should be used only by our hook in old versions of Yoast SEO.  However, if a user kept Yoast < 14.0 with the future Polylang 3.2, then he will lost the feature.

PR #917, opened 2021-09-21
Remove duplicated code

Chouby:
I noticed that these actions are already removed by `remove_all_meta_actions()`, so it's useless to remove them once again.


manooweb (approved):
Ok 👍

PR #923, opened 2021-10-03
Improve some PHPDocs

Chouby:
All is in the title.

### DIFF
--- a/settings/table-string.php
+++ b/settings/table-string.php
@@ -211,12 +211,12 @@ class PLL_Table_String extends WP_List_Table {
 	}
 
 	/**
-	 * Sort items
+	 * Sort registered string items.
 	 *
 	 * @since 0.6
 	 *
-	 * @param object $a The first object to compare
-	 * @param object $b The second object to compare
+	 * @param array $a The first item to compare.
+	 * @param array $b The second item to compare.
 	 * @return int -1 or 1 if $a is considered to be respectively less than or greater than $b.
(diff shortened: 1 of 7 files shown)


settings/table-string.php
```diff
@@ -211,12 +211,12 @@ protected function search_in_translations( $mos, $s ) {
 	}
 
 	/**
-	 * Sort items
+	 * Sort registered string items.
```
Hug0-Drelon, line 214:
```suggestion
	 * Sorts registered string items.
```

Hug0-Drelon (commented):
https://github.com/polylang/polylang/blob/1efdc84691bbccddcfdf98f4cb53af5e806dcf07/include/olt-manager.php#L73 
https://github.com/polylang/polylang/blob/1efdc84691bbccddcfdf98f4cb53af5e806dcf07/integrations/integrations.php#L34
https://github.com/polylang/polylang/blob/1efdc84691bbccddcfdf98f4cb53af5e806dcf07/modules/wpml/wpml-config.php#L40

Maybe use `accesses` instead of `access` on these previous lines if it's not a noun.

Chouby:
> Maybe use `accesses` instead of `access` on these previous lines if it's not a noun.
In this case the noun is ok for me.

Hug0-Drelon (approved):
OK 👍 
But the tests are failing, probably not because of this PR.

PR #932, opened 2021-11-01
Add list of wpml-config.xml in the site health

sebastienserre:
fix https://github.com/polylang/polylang-pro/issues/1122

@ reviewer, Can I avoid doing `new PLL_WPML_Config();` ?

### DIFF
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -171,7 +171,7 @@ class PLL_Admin_Site_Health {
 
 		$debug_info['pll_options'] = array(
 			/* translators: placeholder is the plugin name */
-			'label'  => sprintf( esc_html__( '%s Options', 'polylang' ), POLYLANG ),
+			'label'  => sprintf( esc_html__( '%s options', 'polylang' ), POLYLANG ),
 			'fields' => $fields,
 		);
 
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -171,7 +171,7 @@ class PLL_Admin_Site_Health {
 
 		// does a wpml-config.xml exists ?
 		$wpml = new PLL_WPML_Config();
-		$wpml_files = $wpml->files;
+		$wpml_files = $wpml::instance()->get_files();
 		if ( ! empty( $wpml_files ) ) {
 			$fields['wpml']['label'] = 'wpml-config.xml';
 			$fields['wpml']['value'] = $wpml_files;
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -320,7 +320,8 @@ class PLL_Admin_Site_Health {
 		// Create the section.
 		if ( ! empty( $fields ) ) {
 			$debug_info['pll_warnings'] = array(
-				'label'  => sprintf( esc_html__( 'Polylang informations', 'polylang' ), POLYLANG ),
+				/* translators: placeholder is the plugin name */
+				'label'  => sprintf( esc_html__( '%s informations', 'polylang' ), POLYLANG ),
 				'fields' => $fields,
 			);
 		}
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -317,7 +317,7 @@ class PLL_Admin_Site_Health {
 			$fields['wpml']['value'] = $wpml_files;
 		}
 
-		// Create the section
+		//Create the section
 		if ( ! empty( $fields ) ) {
 			$debug_info['pll_warnings'] = array(
 				'label'  => sprintf( esc_html__( 'Polylang informations', 'polylang' ), POLYLANG ),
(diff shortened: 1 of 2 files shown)


sebastienserre:
Result
screenshot-2021-11-01-14-32
:

manooweb:
> @ reviewer, Can I avoid doing `new PLL_WPML_Config();` ?

No real other choice because the Polylang object has no property which contains this PLL_WPML_Config object.

It's seems to me this is a singleton maybe we should use `PLL_WPML_Config::instance()` instead, AMHA.
See https://github.com/polylang/polylang/blob/master/modules/wpml/load.php#L15

modules/site-health/admin-site-health.php
```diff
@@ -168,6 +168,15 @@ public function info_options( $debug_info ) {
 				}
 			}
 		}
+
+		// does a wpml-config.xml exists ?
+		$wpml = new PLL_WPML_Config();
+		$wpml_files = $wpml->files;
```
Chouby, line 174:
```suggestion
		$wpml_files = PLL_WPML_Config::instance()->get_files();
```
This will be ok if you follow my suggestion in the other comment.

Chouby (changes requested):

Hug0-Drelon:
Should we add tests for this feature ?

Chouby:
We don't have any for the site health. That's somethint that would be nice to have in the future, although low priority compared to our current other topics.

Hug0-Drelon:
I've created an issue about the tests as a reminder https://github.com/polylang/polylang-pro/issues/1125.

modules/wpml/wpml-config.php
```diff
@@ -25,6 +25,13 @@ class PLL_WPML_Config {
 	 */
 	protected $xmls;
 
+	/**
+	 * Array of wpml-config.xml files.
+	 *
+	 * @var string[]
+	 */
+	public $files;
+
```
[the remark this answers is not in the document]
sebastienserre: I think I'm misunderstanding this `Don't add a public property. Switch the visibility of this property to private`. I've removed `public $files;` so I do not have something to put Private :pensive:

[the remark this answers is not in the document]
sebastienserre: I think I miss the `Use this property` as I've deleted it :pensive:

modules/wpml/wpml-config.php
```diff
@@ -25,6 +25,13 @@ class PLL_WPML_Config {
 	 */
 	protected $xmls;
 
+	/**
+	 * The list xml files.
```
Hug0-Drelon, line 29:
```suggestion
	 * The list of xml files.
```

modules/wpml/wpml-config.php
```diff
@@ -59,6 +66,7 @@ public static function instance() {
 	 */
 	public function init() {
 		$this->xmls = array();
+
```
Hug0-Drelon, line 69:
```suggestion
```

Hug0-Drelon: Just to avoid adding new line for nothing.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):
Ok 👍

Chouby (changes requested):
The PR looks good, however I I would move this new information to another section are these are not "options".

My proposal is to rename the section "Polylang warnings" to "Polylang informations" which should include both the wpml-config.xml informations and the current warnings.

I also prpose to change the lable from "wpml-config.xml" to "wpml-config.xml files". 

Finally I would change the name "Polylang Options" to "Polylang options". I missed that previously but I don't like capital letters at the middle of strings.

modules/site-health/admin-site-health.php
```diff
@@ -308,9 +310,17 @@ public function info_warning( $debug_info ) {
...
+		if ( ! empty( $wpml_files ) ) {
+			$fields['wpml']['label'] = 'wpml-config.xml files';
+			$fields['wpml']['value'] = $wpml_files;
+		}
+
+		// Create the section
 		if ( ! empty( $fields ) ) {
 			$debug_info['pll_warnings'] = array(
-				'label'  => sprintf( esc_html__( 'Polylang warnings', 'polylang' ), POLYLANG ),
+				'label'  => sprintf( esc_html__( 'Polylang informations', 'polylang' ), POLYLANG ),
```
Chouby, line 323:
```suggestion
				/* translators: placeholder is the plugin name */
				'label'  => sprintf( esc_html__( '%s informations', 'polylang' ), POLYLANG ),
```

Chouby (changes requested):

modules/site-health/admin-site-health.php
Chouby: I had previously missed that `sprintf()` was used without placeholder.

sebastienserre: I've added a Placeholder and the comment

Chouby (approved):

PR #938, opened 2021-11-22
Adjust query properties for empty search requests to avoid PHP notices

### LINKED ISSUE #937: Avoid PHP notices for invalid search requests
## Prerequisites

1. Create a language and assign it as a default for all existing posts
2. Use "The language is set from the directory name in pretty permalinks" setting and uncheck "Hide URL language information for default language"
3. Now open `example.com/de/randomstring/?s=` where `de` is the slug of the language

When I follow those steps, I see several PHP notices. See the following screenshots from Query Monitor:
php-notices

This is because the wrong template is loaded:
template

I was expecting no PHP notices and the 404 template to be loaded.

This seems to be caused by this code: https://github.com/polylang/polylang/blob/6ce5b2844d52e57d764764e6441fcd839dafc7c5/frontend/choose-lang.php#L310-L318

ocean90:
Fixes #937.

This avoids the PHP notices and ensures that the 404 template is loaded.


Chouby:
Closing in favor of #940.

PR #939, opened 2021-11-24
Fix Canonical_Test::test_plain_tag_feed 

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1132

## Problem

Since WordPress 5.9-alpha, `get_term_feed_link()` first parameter type changed. Prior to 5.9-alpha, the previous function accept only the `term_id`. Now it allow both `term_id` and `WP_Term`. BUT when passing a `term_id`, the `taxonomy` is set to a default value `category`.

See https://github.com/WordPress/wordpress-develop/commit/a8485376d27b5b3348a68963f6fd38200b6e64cf  

## Changes

Check the WordPress version before getting the term feed link.
If WP > 5.9, then use a `WP_Term` object.

*Note : the use of `WP_term` seems to increase performance. See the link above.*

⚠️  *This [PR](https://github.com/WordPress/wordpress-develop/pull/1945) should fix the broken backward compatibility in WordPress, if so this PR would be useless.*
See https://core.trac.wordpress.org/ticket/50225#comment:27

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ js/build/
 *.min.js
 *.min.css
 composer.lock
+.phpunit.result.cache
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👍

Chouby:
The issue will probably be solved in WP. See https://core.trac.wordpress.org/ticket/50225#comment:27

Chouby:
Closing as it was fixed by https://core.trac.wordpress.org/changeset/52255.

PR #940, opened 2021-11-30
An empty search should not display the homepage anymore

### LINKED ISSUE #937: Avoid PHP notices for invalid search requests

Chouby:
Fix #937 

WordPress used to display the homepage as the result of an empty search. We aimed to reproduce this behavior since  https://wordpress.org/support/topic/plugin-polylang-polylang-breaks-search-in-spun-theme. However, the behavior of WordPress has been modified since https://core.trac.wordpress.org/ticket/11330 has been fixed in WP 4.0.

Moreover our fix didn't handle correctly empty searches when the query had additional query vars (see #937).

This PR removes the "fix" added in v1.1.1 to display the homepage for an empty search. It adds tests to check that we reproduce the current behavior for empty searches with or without additional query vars.

### DIFF
--- a/frontend/choose-lang.php
+++ b/frontend/choose-lang.php
@@ -305,16 +305,13 @@ abstract class PLL_Choose_Lang {
 		if ( $lang = apply_filters( 'pll_set_language_from_query', false, $query ) ) {
 			$this->set_language( $lang );
 			$this->set_curlang_in_query( $query );
-		}
-
-		// sets is_home on translated home page when it displays posts
-		// is_home must be true on page 2, 3... too
-		// as well as when searching an empty string: http://wordpress.org/support/topic/plugin-polylang-polylang-breaks-search-in-spun-theme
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👍

Chrystll:
OK +1 Now we get as expected the same behavior of WordPress => is_404, see:
Capture du 2021-12-13 18-20-00

PR #942, opened 2021-12-08
WPML Compatibility: Check the type of the string before registering it

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1144
This is to avoid warnings in the case that a 3rd party plugin would register for example an array.
See also f96d4c4d0c314ab6b72b7f89b8ab748fe85955dc which was the same for strings registered with `pll_register_string()`.


Chrystll:
I still get the same issue with your PR. I get the same warning when visiting the Strings Translations where is registered an option containing an array. See:
Capture du 2021-12-09 09-49-45

Marianne380 (approved):

Issue #928, opened 2021-10-12
Term (Category/Tag) insertion removes list table columns


sybrew:
## Prerequisites

- [x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x] This bug happens with only Polylang plugin active
- [x] This bug happens with a default WordPress theme active
- [x] I can reproduce this bug consistently

## Steps to reproduce the issue

1. Activate Polylang and any other plugin that works with Term insertion (Yoast SEO, The SEO Framework, etc.).
2. Insert a term.
3. Data gets overwritten

## Expected behavior and actual behavior

It's a grey because I ran `debugger`...

When I follow those steps, I see...
![image](https://user-images.githubusercontent.com/7662236/136924345-6252ceaa-5bc3-41e9-a0e6-78a31e4b8a82.png)

I was expecting...
![image](https://user-images.githubusercontent.com/7662236/136924326-14797e1f-9a6e-4e9c-990e-60890e333689.png)

This line is too greedy with its replacement: https://github.com/polylang/polylang/blob/3.1.2/js/src/term.js#L78

I don't think that line does anything useful since Polylang already correctly supplies the output via WordPress's List Table.

Chouby:
Sorry, I thought I had already answered to this issue. Are you sure that the issue is with *any* plugin inserting a column? I have (re-)tested with the plugins I usually use for this feature (WooCommerce, Yoast SE0) and their columns are correctly filled when I insert a new term.

sybrew:
Hi Chouby,

This issue on term-adding occurs with any column added _after_ the flag selection. Perhaps Yoast SEO has changed the order of columns recently, so they are no longer affected.

Yoast SEO is still affected, though, when you use post/page quick edit.

Please also test with other plugins, such as The SEO Framework, or even add a custom column after it so you can control the variables. Regardless, the code I advise removing doesn't seem to do anything useful (anymore).

Chouby:
I tested again with Yoast SEO and its columns are displayed before the Polylang columns so if the order has an importance, I can't test it with this plugin.

[...]

PR #943, opened 2021-12-09
Add Yoast SEO columns when updating the post list table rows

Chouby:
When quick editing a post, the columns added by Yoast SEO are not drawn correctly. It's because the relevant hooks are not added for ajax actions, except the `inline-save` action used by WordPress. We need to add the hooks for our action `pll_update_post_rows` too.

See also https://github.com/polylang/polylang/issues/928#issuecomment-989883330

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -46,6 +46,10 @@ class PLL_WPSEO {
 		} else {
 			add_filter( 'pll_copy_post_metas', array( $this, 'copy_post_metas' ), 10, 2 );
 			add_filter( 'pll_translate_post_meta', array( $this, 'translate_post_meta' ), 10, 3 );
+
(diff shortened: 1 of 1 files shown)


manooweb (approved):
OK 👍

sybrew:
These lines should've been deleted for it overrides default WordPress behavior, which already updates rows: https://github.com/polylang/polylang/blob/7900a51b24fd435d06d361f912714d98d101f669/js/src/term.js#L71-L123.

Instead, this pull's merge added more code for the currently implemented redundant workaround fixing a bug in WordPress 3.1; your team must maintain this redundancy by manually including support for every existing plugin. Remove this redundancy by removing the lines above, and every plugin built according to WordPress's List Table API will work again without any future bespoke and time-consuming intervention necessary. 

As a bonus: nothing in Polylang itself will break; just a lot of code won't be called anymore (which you can subsequently delete as well). It'll take a load off.

PR #947, opened 2021-12-13
Improve language definition process in REST requests

Screenfeed:
This should not change much the current behavior, this is only a starting point for what is to come.

- Move the language definition process into a separated method `set_current_language()`.
- Fire a filter after defining the language. This allows more flexibility during the process, by allowing 3rd party code to define the current language at this point.
- Create a dummy REST request based on the current REST route. This allows to use `->get_param()` for example, to help defining the current language. This request object is passed to the hook from the previous point.
- Use the new request object instead of relying on `$_REQUEST`.


manooweb:
Introducing `antecedent/patchwork` in the tests suite make the `Ajax_On_Front_Test::test_locale_for_logged_in_user` fail. So this PR also rewrite this test to use the spanish .mo file provided by the WP tests suite instead of using our own .mo file.

manooweb (approved):
OK 👍

manooweb (approved):
OK 👍
My previous [comment](https://github.com/polylang/polylang/pull/947#issuecomment-998700930) is thus useless because patchwork has been removed and also the modification of the concerned test.

Screenfeed:
We decided to not "rush" the language definition, and do not act before the real API is loaded.

PR #949, opened 2021-12-13
PHP 8.1: Fix deprecated notices

Chouby:
**Fixes in Polylang code:**
- `wp_parse_url()` may return null. Ensure that we cast the return value to a string when used in a function expecting a string.
- `wp_cache_get()` may return false. The conversion to an array must nok be explicit.

**Fixes in PHPUnit tests:**
- Ensure that `Accept_Languages_Collection_Test::get_known_language()` always returns a language with the `w3c` property.
- Avoid loading remote block patterns which is useless and currently fires a bunch of notices.
- Specify an author for posts in the Ephemera widget test to avoid `get_the_author()` to return null which in the end causes a notice in `ent2ncr()`. 
- Specify `$plugin_page` in Settings test to avoid a notice in `get_plugin_page_hookname()`.
- Make sure to define `$options['default_lang']` in tests as, [when not defined, the home and search urls are not set in `PLL_Language` objects](https://github.com/polylang/polylang/blob/3.1.3/include/links-model.php#L169), causing a notice in `set_url_scheme()`. 
- To remove relationships, pass an empty array to `wp_set_object_terms()` instead of null.

**Fired by WP independantly from Polylang:**
- ```PHP Deprecated:  rtrim(): Passing null to parameter #1 ($string) of type string is deprecated in /app/polylang/tmp/wordpress/wp-includes/formatting.php on line 2760```
This is due to `load_script_textdomain()` having always an empty path for WP scripts. See proposed patch: https://core.trac.wordpress.org/attachment/ticket/53635/53635-09-script-translations.patch

**Not fixed in this PR:**
- ```PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in /app/polylang/tmp/wordpress/wp-includes/link-template.php on line 3786```. I propose to fix this in a separate PR.

### DIFF
--- a/frontend/choose-lang-url.php
+++ b/frontend/choose-lang-url.php
@@ -50,7 +50,7 @@ class PLL_Choose_Lang_Url extends PLL_Choose_Lang {
 
 		$requested_url   = pll_get_requested_url();
 		$requested_host  = str_replace( 'www.', '', wp_parse_url( $requested_url, PHP_URL_HOST ) ); // Remove www. for the comparison
-		$requested_path  = rtrim( str_replace( $this->index, '', wp_parse_url( $requested_url, PHP_URL_PATH ) ), '/' ); // Some PHP setups turn requests for / into /index.php in REQUEST_URI
+		$requested_path  = rtrim( str_replace( $this->index, '', (string) wp_parse_url( $requested_url, PHP_URL_PATH ) ), '/' ); // Some PHP setups turn requests for / into /index.php in REQUEST_URI
 		$requested_query = wp_parse_url( $requested_url, PHP_URL_QUERY );
 
 		// Home is requested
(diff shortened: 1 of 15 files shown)


Screenfeed:
I see several of this:
```php
$options = array_merge( PLL_Install::get_default_options(), array( 'default_lang' => 'en' ) );
```
Perhaps we should create a method:
```php
protected function get_default_options( array $options = array() ) {
	return array_merge( PLL_Install::get_default_options(), array( 'default_lang' => 'en' ), $options );
}
```
What do you think?

Chouby:
I don't like the idea to hardcode the default language value, when the languages themselves are not.
```php
protected function get_default_options_wih_default_lang( string $default_lang ) {
	return array_merge( PLL_Install::get_default_options(), array( 'default_lang' => $default_lang) );
}
```
Or:
```php
protected function get_options( array $options = array() ) {
	return array_merge( PLL_Install::get_default_options(), $options );
}
```
would be acceptable for me. The added value is however not that high.

Screenfeed:
We probably need to work on this on another PR then: I wanted to factorize some code but it needs to be more "global".

Screenfeed (approved):

PR #950, opened 2021-12-16
Update switcher.php

marcomarasco:
Prevent fatal errors in missing model definition due multisite site clone procedures.


Chouby:
Thanks for your PR. However it's causing several tests to error or fail. See https://app.travis-ci.com/github/polylang/polylang/jobs/552567834. Could you please describe the error tha you attempting to fix as well as the steps to reproduce it, starting from a fresh install?

PR #956, opened 2021-12-24
Add hook in PLL_Admin_Model to update translation group 

Marianne380:
Link to https://github.com/polylang/polylang-pro/pull/1168

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -463,9 +463,12 @@ class PLL_Admin_Model extends PLL_Model {
 					}
 					unset( $tr[ $old_slug ] );
 
-					// Delete sync between translations when the old slug doesn't exist anymore.
+					// Delete sync between translations when deleting a language or update slug.
 					if ( isset( $tr['sync'] ) ) {
 						if ( array_key_exists( $old_slug, $tr['sync'] ) ) {
+							if ( $new_slug ) {
+								$tr['sync'][ $new_slug ] = $tr['sync'][ $old_slug ];
+							}
 							unset( $tr['sync'][ $old_slug ] );
 						}
 					}
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -461,10 +461,15 @@ class PLL_Admin_Model extends PLL_Model {
 				 *
 				 * @since 3.2
 				 *
-				 * @param array   $tr       The translations group.
-				 * @param string  $old_slug The old language slug.
-				 * @param string  $new_slug The new language slug.
-				 * @param WP_Term $term     The term containing the post or term translations.
+				 * @param array<int|array<string>> $tr {
+				 *     List of translations with lang codes as array keys and IDs as array values.
+				 *     Also in this array:
+				 *
+				 *     @type array<string> $sync List of synchronized translations with lang codes as array keys and array values.
+				 * }
+				 * @param string                   $old_slug The old language slug.
+				 * @param string                   $new_slug The new language slug.
+				 * @param WP_Term                  $term     The term containing the post or term translations.
 				 */
 				$tr = apply_filters( 'update_translation_group', $tr, $old_slug, $new_slug, $term );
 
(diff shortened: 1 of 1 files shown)


Chouby:
Some remarks:
- You are modifying Polylang Pro data from Polylang, which doesn't look logical to me. We can probably do it from Polylang Pro (we may need to add hooks in Pollan though). Let's discuss the best way.
- Modified slugs can alos be found in array values, not ony keys.
- I expect Phpunit tests.

Chouby (changes requested):

admin/admin-model.php
```diff
@@ -454,6 +454,18 @@ public function update_translations( $old_slug, $new_slug = '' ) {
...
+				/**
+				 * Filters the post translations group.
+				 *
+				 * @since 3.2
+				 *
+				 * @param array  $tr       The post translations group.
+				 * @param string $old_slug The old language slug.
+				 * @param string $new_slug The new language slug.
+				 */
+				$tr = apply_filters( 'update_post_translations_group', $tr, $old_slug, $new_slug );
```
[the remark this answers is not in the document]
Marianne380: Knowing that `$tr` is used in the preparation SQL query here : https://github.com/polylang/polylang/blob/a754f33dceb84b55cc451e6f2462e8b9ba4b396a/admin/admin-model.php#L470
I suggest to apply the filter before.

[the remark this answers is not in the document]
Screenfeed: >Although, this is not important for our current case, shouldn't we add a 4th parameter to pass the taxonomy.

Agree. Adding `$term` would be better though imho.

>Shouldn't we apply the filter after the modifications rather than before?

I don't think so, since the values of `$dr`, `$dt`, and `$ut` depend on the values in `$tr`.

admin/admin-model.php
```diff
@@ -454,6 +454,18 @@ public function update_translations( $old_slug, $new_slug = '' ) {
 			foreach ( $terms as $term ) {
 				$term_ids[ $term->taxonomy ][] = $term->term_id;
 				$tr = maybe_unserialize( $term->description );
+
+				/**
+				 * Filters the translation group before updating it in database.
```
Chouby, line 459:
```suggestion
				 * Filters the unserialized translation group description before it is
				 * updated when a language is deleted or a language slug is changed.
```

Chouby (changes requested):

admin/admin-model.php
```diff
@@ -454,6 +454,20 @@ public function update_translations( $old_slug, $new_slug = '' ) {
...
 				$term_ids[ $term->taxonomy ][] = $term->term_id;
 				$tr = maybe_unserialize( $term->description );
+
+				/**
+				 * Filters the unserialized translation group description before it is
+				 * updated when a language is deleted or a language slug is changed.
+				 *
+				 * @since 3.2
+				 *
+				 * @param array   $tr       The translations group.
```
Screenfeed, line 464: What about using something similar to your other PR?

```php
				 * @param array<int|array<string>> $tr       {
				 *     List of translations with lang codes as array keys and IDs as array values.
				 *     Also in this array:
				 *
				 *     @type array<string> $sync List of synchronized translations with lang codes as array keys and array values.
				 * }
```

Marianne380: Good idea :+1:

Screenfeed (approved):

admin/admin-model.php
```diff
@@ -454,6 +454,25 @@ public function update_translations( $old_slug, $new_slug = '' ) {
...
+				 *
+				 * @param array<int|array<string>> $tr {
+				 *     List of translations with lang codes as array keys and IDs as array values.
+				 *     Also in this array:
+				 *
+				 *     @type array<string> $sync List of synchronized translations with lang codes as array keys and array values.
+				 * }
+				 * @param string                   $old_slug The old language slug.
+				 * @param string                   $new_slug The new language slug.
+				 * @param WP_Term                  $term     The term containing the post or term translations.
```
Chouby, line 472:
```suggestion
				 * @param WP_Term                  $term     The term containing the post or term translation group.
```

PR #958, opened 2022-01-04
Introduced function pll_filter_input()

Screenfeed:
This function is needed in future developments.
`pll_filter_input()` is basically a wrapper for the native php function `filter_input()`: wrapping it make it stubable for phpunit tests.

Note: the PR [PLLPro#1167](https://github.com/polylang/polylang-pro/pull/1167) introduces a new (undeclared) class property `$polylang->full_site_editing`, maybe this PR should also declare it.


Screenfeed:
Closing because it won't be used.
See [this commit](https://github.com/polylang/polylang-pro/pull/1167/commits/076d0a4fb4931b6f481f84494f8a230aa3465bfb).

PR #959, opened 2022-01-04
Remove navigation role from nav element

Chouby:
For references:
https://make.wordpress.org/core/2022/01/04/miscellaneous-core-changes-in-wordpress-5-9/
https://core.trac.wordpress.org/ticket/54054


manooweb (approved):
OK 👍

PR #960, opened 2022-01-11
Yoast SEO: Test the existence of current language before using it

### LINKED ISSUE #954: PHP Warning:  Attempt to read property "slug" on null in /integrations/wpseo/wpseo.php on line 305
## Prerequisites

- [X] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [X] This bug happens with only Polylang plugin active
- [X] This bug happens with a default WordPress theme active

I just update to php 8.0 and in my php error logs all the time returns:

23-Dec-2021 10:27:14 UTC] PHP Warning:  Attempt to read property "slug" on null in /xxx/wp-content/plugins/polylang/integrations/wpseo/wpseo.php on line 305

Thxs

Chouby:
Fixes #954 

Unfortunately, the details provided in #954 did not allow @Chrystll to reproduce the issue. Althougth the settings indicate that the language is set from the content (the current language definition is thus delayed to the `wp` action), the tests demonstrate that the current language is already set when the filter `wpseo_frontend_presenters` is applied by Yoast SEO.

I would have prefer to understand, but anyway the warning itself is easy to fix. 

EDIT: Failing tests are not related to this PR (but to a change in Jetpack 10.5).

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -288,17 +288,17 @@ class PLL_WPSEO {
 	}
 
 	/**
-	 * Get alternate language codes for Opengraph
+	 * Get alternate language codes for Opengraph.
 	 *
 	 * @since 2.7.3
 	 *
-	 * @return array
+	 * @return string[]
 	 */
 	protected function get_ogp_alternate_languages() {
 		$alternates = array();
 
 		foreach ( PLL()->model->get_languages_list() as $language ) {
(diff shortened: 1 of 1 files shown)


manooweb (approved):
OK 👍

PR #961, opened 2022-01-14
Make flag tests independent of other tests

Chouby:
In my investigations to fix deprecated notices fired by PHP 8.1, I came back to the issues spotted by @raaaahman in #566. This PR replaces it with the same goals.

The Flag tests rely on PLL_Links_Model hooks setup globally. As we need to remove these global hooks (planned in a future PR), we nee to remove this dependence. The links model in now instantiated in `set_up()`.

Cleaning up of `$_SERVER['HTTPS']` was made at the end of the test. This is correct when the test passes, but is not done if the test fails. This ends up breaking more unrelated tests. Cleaning up has been moved to `tear_down()`.

### DIFF
--- a/tests/phpunit/tests/test-flags.php
+++ b/tests/phpunit/tests/test-flags.php
@@ -22,6 +22,19 @@ class Flags_Test extends PLL_UnitTestCase {
 		rmdir( WP_CONTENT_DIR . '/polylang' );
 	}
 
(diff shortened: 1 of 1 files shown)


Chouby:
The failing test is not related to this PR.

Screenfeed (approved):

PR #965, opened 2022-01-17
Trick Jetpack to avoid a doing_it_wrong notice in test

Chouby:
The issue occurs because we load Jetpack after the `wp_default_scripts` action (unlike a standard WP install).
This PR proposes to trick Jetpack to bypass the doing_it_wrong notice.
Fix https://github.com/polylang/polylang-pro/issues/1179


manooweb (approved):
OK 👍

PR #966, opened 2022-01-18
PHP 8.1: Fix deprecated notice fired in set_url_scheme()

Chouby:
**The goal of this PR is to fix the deprecated notice with PHP 8.1:**
```
PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in /home/travis/build/polylang/polylang/tmp/wordpress/wp-includes/link-template.php on line 3786
```
This notice is fired in `set_url_scheme()` called from `PLL_Language::set_url_scheme()`.
As far as I know, this notice should never occur in production but only in phpunit tests. This is due to ghost filters `PLL_Links_Model::pll_after_languages_cache()` acting on incomplete `PLL_Language` objects (without the home or flag urls).

Indeed apart from the links model used in a test, we could have other links model instantiated in the class `Polylang` when the plugin is loaded initially before all tests, and also each time a language is created with the `create_language()` helper method.

**Changes introduced by this PR:**
- In PHPUnit tests, stop instantiating a `PLL_Links_Model` object in `create_language()` helper method. This means that the default category in the corresponding language is not created anymore. This creation is thus made only in the test classes needing it.
- In PHPUnit tests, we need to clear the languages cache when actions done in `wpSetUpBeforeClass()` call `PLL_Model::get_languages_list()` in a way or another. This is to avoid to use a truncated languages list (without the home or flag urls) in the following tests.
- In `PLL_Canonical_UnitTestCase::set_up()`, get options from the database (instead of using the default options) to ensure that we get the default language inside.
- In the class `Polylang`. The links model object was always instantiated. It is now instantiated only when the main Polylang obejct is instantiated. This avoids to have a ghost `PLL_Links_Model` object and its active filters when the plugin is initially loaded from the tests bootstrap.
- In `PLL_Links_Models::set_home_url()`, I remove a test which was added specifically to hide notices occuring in PHPUnit tests (a way to hide things under the rug), as this is now fixed by this PR.

### DIFF
--- a/include/class-polylang.php
+++ b/include/class-polylang.php
@@ -193,7 +193,6 @@ class Polylang {
 		 */
 		$class = apply_filters( 'pll_model', PLL_SETTINGS || self::is_wizard() ? 'PLL_Admin_Model' : 'PLL_Model' );
 		$model = new $class( $options );
-		$links_model = $model->get_links_model();
 
 		if ( ! $model->get_languages_list() ) {
 			/**
@@ -227,7 +226,8 @@ class Polylang {
 		$class = apply_filters( 'pll_context', $class );
 
 		if ( ! empty( $class ) ) {
-			$polylang = new $class( $links_model );
+			$links_model = $model->get_links_model();
+			$polylang    = new $class( $links_model );
 
 			/**
 			 * Fires after the $polylang object is created and before the API is loaded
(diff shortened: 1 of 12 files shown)


Screenfeed (approved):
@Chouby Do you think the same thing will impact PLL Pro?

PR #970, opened 2022-01-27
Change ajax-response id not to target default ajax-response HTML tag which is used by WordPress

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/1202

Continuation of https://github.com/polylang/polylang/pull/936

This PR propose to target an unexisting pll-ajax-response id not to overwrite the default WordPress message notice.

### DIFF
--- a/js/src/admin.js
+++ b/js/src/admin.js
@@ -292,7 +292,9 @@ jQuery(
 					ajaxurl,
 					data,
 					function( response ) {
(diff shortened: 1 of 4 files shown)


Chrystll:
Your PR doesn't fix the dimissable notice that we get in WooCommerce Order when switching the language, see:
screenshot-trunk-test com-2022 01 29-10_07_38

PR #972, opened 2022-01-31
Improve compatibility of wpml_object_id

Chouby:
Although this was not the root cause of the issue, working on https://github.com/polylang/polylang-pro/issues/1204, I noticed small differences between WPML and our implementation of the `wpml_object_id` filter.

The goals of this PR are:
- Improve the readability of the `icl_object_id()` function.
- Improve the validation.
- Support the [undocumented](https://wpml.org/documentation/support/wpml-coding-api/wpml-hooks-reference/#hook-605256) `'any'` element type.
- Remove the support of the `'term'` element type which is not supported by WPML.
- Stop honoring `$return_original_if_missing` when the element type is not translatable, as WPML does.

### DIFF
--- a/modules/wpml/wpml-legacy-api.php
+++ b/modules/wpml/wpml-legacy-api.php
@@ -168,9 +168,24 @@ if ( ! function_exists( 'icl_object_id' ) ) {
 	 * @return int|null The object id of the translation, null if the translation is missing and $return_original_if_missing set to false
 	 */
 	function icl_object_id( $id, $type = 'post', $return_original_if_missing = false, $lang = '' ) {
-		$lang = $lang ? $lang : pll_current_language();
+		if ( empty( $id ) ) {
+			return null;
+		}
+
+		if ( 'any' === $type ) {
+			$type = get_post_type( $id );
+		}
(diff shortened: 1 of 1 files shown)


Screenfeed (changes requested):
One thing: the function signature is different.
This function is declared like this:
```php
function icl_object_id( $id, $type = 'post', $return_original_if_missing = false, $lang = '' ) {
```
But the original function is different:
```php
function icl_object_id( $element_id, $element_type = 'post', $return_original_if_missing = false, $ulanguage_code = null ) {
```
First, we can see that the default value for the 4th parameter is not the same, but there's more than that: the parameter names are not the same. Why is it important? [Named arguments in PHP 8.0](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments).

Also, personally I would make sure that the value returned by the function is actually a real integer, and the 1st argument too (I've seen too many "integers in string"):
- `$id = (int) $id;` before `if ( 'any' === $type ) {`.
- `return (int) $tr_id;`

Screenfeed (approved):

PR #973, opened 2022-02-02
remove the last slash on $root that causes the pattern replace fails …

mybooking-es:
Our setup:

- Multi language site
- WP ultimo to manage plans and custom domains
- Polylang to translate web pages

When $root ends with two slashes the replacement the path with language does not work and the links does not include the language path. 

This line can return a $root with two ending slashes if home ends by / and $this->root is empty.

$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '#^https?://#', '://', $this->home . '/' . $this->root );

I know it can be improved but a modification has been required to make our setup work

### DIFF
--- a/include/links-directory.php
+++ b/include/links-directory.php
@@ -71,6 +71,10 @@ class PLL_Links_Directory extends PLL_Links_Permalinks {
 			$slug = $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : $base . $lang->slug . '/';
 			$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '#^https?://#', '://', $this->home . '/' . $this->root );
 
+			if (preg_match( '/\/\/$/', $root)) {
+				  $root = substr( $root, 0, -1 );
+			}
+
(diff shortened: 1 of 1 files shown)


Chouby:
It looks like there is something wrong in your install. The home url never ends with a `/` in WordPress. If present, it's explicitely removed by WordPress when retrieving the home option: https://github.com/WordPress/WordPress/blob/5.9/wp-includes/option.php#L211-L213

`'home_url()` should always return the home url without trailing slash. This is what we use in https://github.com/polylang/polylang/blob/3.1.4/include/links-model.php#L49. If there is a trailing slash in your install, it's probably added by a filter. I dont think that we should fix this issue on Polylang side.

mybooking-es:
Thank you for the details @Chouby. I understand that is a problem with our install. We finally got it working using define('PLL_CACHE_LANGUAGES', false); We found that when we accessed the dashboard the value of _transient_pll_languages_list was updated with the double slash. It was restored once we access Polylang settings and updated the value. Removing the use of the cache seems that makes our install to work. I want to check it more deeply.

PR #978, opened 2022-02-07
Add Lower and Upper Sorbian flags

sebastienserre:
https://secure.helpscout.net/conversation/1778921054/18182
In same time, these are both languages spoken in Germany, that's why we currently have DE flag


Chouby:
This flag is not mentionned on Wikipedia for the regions where these languages are spoken: https://de.wikipedia.org/wiki/Niederlausitz, https://de.wikipedia.org/wiki/Oberlausitz

PR #984, opened 2022-02-17
Improve sanitization and validation for translated objects

Screenfeed:
This introduces 2 new methods for `PLL_Translated_Object`: `sanitize_int_id()` and `sanitize_int_ids_list()`.

This also changes the behavior of the method `validate_translations()` (previous behavior is kept as default one). It can now be used for both saving to the DB and also when retrieving:
- A new `$context` parameter has been added to differentiate the contexts.
- In both contexts: a new `$id` parameter has been added: if the translations list is empty, this ID is added to it.
- In 'save' context: unchanged behavior, check that the IDs and languages correspond.
- In 'display' context: make sure the languages exist.

More precisely in the code of `validate_translations()`:
1. The first part is only for the 'save' context: it checks that the IDs are in the right language. This is the same thing as previously.
1. The 2nd part is common to both contexts: if the `$id` argument is provided and the list is empty, it is added to the list.
1. The last part is only for the 'display' context: it makes sure that the languages are valid. Why not in the 'save' context? Because this is already done in point 1.

Use all these everywhere in `PLL_Translated_Object`, `PLL_Translated_Post`, and `PLL_Translated_Term`, to sanitize/validate entries and outputs.

Also, various small improvements and fixes:
- Use phpStan at max level on the 3 classes. Only 2 errors remain.
- Improved some DocBlocks.
- Early returns.
- Use `===` instead of `==`.
- Use `! empty()` instead of `count() > 0` (better perfs).
- `get_object_term_cache()` and `get_term()` can return an instance of `WP_Error`.
- Var name: use `$descr` instead of `$d`.
- `current_user_can_synchronize()` now always return a boolean.
- In child classes, use `$this->tax_language` and `$this->tax_translations` instead of hard-coded values.
- Remove some var assignation from function arguments (`foo( $bar = bar() )`) and `if` statements (`if ( $bar = bar() ) {`).

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -205,9 +205,7 @@ abstract class PLL_Translated_Object {
 		}
 
 		// Sanitize and validate the translations array.
-		$translations = $this->sanitize_int_ids_list( $translations );
-		$translations = array_merge( array( $lang->slug => $id ), $translations ); // Make sure this object is in translations.
-		$translations = $this->validate_translations( $translations );
+		$translations = $this->validate_translations( $translations, $id );
 
 		// Unlink removed translations.
 		$old_translations = $this->get_translations( $id );
@@ -314,7 +312,6 @@ abstract class PLL_Translated_Object {
 
 		$term         = $this->get_object_term( $id, $this->tax_translations );
 		$translations = empty( $term->description ) ? array() : maybe_unserialize( $term->description );
-		$translations = $this->sanitize_int_ids_list( $translations );
 
 		return $this->validate_translations( $translations, $id, 'display' );
 	}
@@ -491,11 +488,11 @@ abstract class PLL_Translated_Object {
 	}
 
 	/**
-	 * Validates translations.
-	 * Make sure to sanitize arguments passed to this method.
+	 * Validates and sanitizes translations.
 	 * This will:
 	 * - Make sure to return only translations in existing languages (and only translations).
-	 * - Add the provided translation (`$id`) if the list is empty.
+	 * - Sanitize the values.
+	 * - Make sure the provided translation (`$id`) is in the list.
 	 * - Check that the translated objects are in the right language, if `$context` is set to 'save'.
 	 *
 	 * @since 3.1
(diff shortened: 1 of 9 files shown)


manooweb (approved):
ok 👍

include/translated-object.php
```diff
@@ -189,79 +189,74 @@ protected function should_update_translation_group( $id, $translations ) {
...
+			wp_set_object_terms( $p, $group, $this->tax_translations );
+		}
+
+		if ( ! is_array( $terms ) ) {
+			return $translations;
+		}
+
+		// Clean now unused translation groups.
+		foreach ( $terms as $term ) {
+			// Get fresh count value.
```
Chouby, line 251: That was indeed a missing comment!

include/translated-object.php
```diff
@@ -398,14 +414,48 @@ public function where_clause( $lang ) {
...
+		if ( false !== $cache ) {
+			return $this->sanitize_int_ids_list( $cache );
+		}
+
+		$object_ids = $wpdb->get_col( $wpdb->prepare( "SELECT object_id FROM $wpdb->term_relationships WHERE term_taxonomy_id = %d", $lang->$tt_id ) );
+		$object_ids = $this->sanitize_int_ids_list( $object_ids );
+		wp_cache_set( $cache_key, $object_ids, 'terms' );
+
+		return $object_ids;
+	}
```
Chouby, line 442: This method was removed in #980. It's coming back probably because a merge conflict. I propose to remove it again.

Chouby (changes requested):

include/translated-object.php
Screenfeed: Done in [19b73bf](https://github.com/polylang/polylang/pull/984/commits/19b73bfc175a4559d04798eb348343a61feadcea).

include/translated-post.php
```diff
@@ -189,9 +213,19 @@ public function current_user_can_read( $post_id, $context = 'view' ) {
...
 		if ( in_array( $post->post_status, get_post_stati( array( 'private' => true ) ) ) ) {
-			$post_type_object = get_post_type_object( $post->post_type );
+			if ( ! is_user_logged_in() ) {
+				return false;
+			}
+
 			$user = wp_get_current_user();
-			return is_user_logged_in() && ( current_user_can( $post_type_object->cap->read_private_posts ) || $user->ID == $post->post_author ); // Comparison must not be strict!
+
+			if ( $user->ID == $post->post_author ) { // Comparison must not be strict!
```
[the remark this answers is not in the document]
Screenfeed: Done in [19b73bf](https://github.com/polylang/polylang/pull/984/commits/19b73bfc175a4559d04798eb348343a61feadcea).

include/translated-object.php
```diff
@@ -127,44 +129,68 @@ public function update_language( $id, $lang ) {
...
+		$translations_term = get_term( $term_id, $this->tax_translations );
+
+		if ( ! $translations_term instanceof WP_Term || empty( $translations_term->description ) ) {
+			return array();
+		}
+
+		// Lang slugs as array keys, template IDs as array values.
+		$translations = maybe_unserialize( $translations_term->description );
+
+		return $this->validate_translations( $translations, $term_id, 'display' );
```
Chouby, line 193: I would rather pass 0 instead of `$term_id` which looks incorrect to me.
Coud you add tests to cover this new method?

Chouby (changes requested):

include/translated-object.php
Screenfeed: Fixed by [0a2d22d](https://github.com/polylang/polylang/pull/984/commits/0a2d22dcff92cf13990cd27e25c47b121c3225bb).
Test added by [6aaa8c2](https://github.com/polylang/polylang/pull/984/commits/6aaa8c2a75a6eac1b3cff6873776015f02f9f834).

PR #991, opened 2022-03-08
Introduce new class PLL_Db_Tools

Screenfeed:
This class regroups a small set of tools to work with the database.

2 methods:
- `prepare_value()` will wrap strings in quotes, or cast numeric values into integers.
- `prepare_values_list()` will change an array of values into a comma separated list of values, ready to be used in a `IN ()` clause. Ex: `'foo','bar','baz',4,7`.

Those 2 methods work only for strings and integers: we never work with float values for example, so I limited to those 2 types (for now) to keep the methods simple.

### DIFF
--- a/include/db-tools.php
+++ b/include/db-tools.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * @package Polylang
+ */
+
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👌

PR #993, opened 2022-03-08
Add previous lang data to the hook 'pll_update_language'

Screenfeed:
In `PLL_Admin_Model->update_language()`, this adds the previous value of the lang being edited as 2nd parameter of the hook `'pll_update_language'`. This eases comparison between old and new values.


Screenfeed:
This will be used in https://github.com/polylang/polylang-pro/pull/1236.

PR #1001, opened 2022-04-05
Fix parent page for parent unstranslatable post type

manooweb:
Fixes [HS#18717](https://secure.helpscout.net/conversation/1818755470/18717/)

In fact since https://github.com/polylang/polylang/pull/911 was merged we introduced an issue in the customer code.

Indeed before, with Polylang 2.9.x, the post parent translation assignment was done only in a bulk action context.
Since Polylang 3.1.x, to fix https://github.com/polylang/polylang-pro/issues/971 this assignment is done in all contexts.

However it's because we assume that the parent post type is the same as the child post type and thus translatable.

In the customer case, the parent post type is different than the child post type and also untranslatable.
This is what causes the issue.

## Changes
- add a condition to check if the parent post type is translatable before assigning the parent translation to the post.
- add a PHPUnit test to cover this case

## Expected behavior
- The customer validated the patch we provided and corresponding on this PR.

### DIFF
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -197,10 +197,12 @@ class PLL_CRUD_Posts {
 	 */
 	public function wp_insert_post_parent( $post_parent, $post_id ) {
 		$lang = $this->model->post->get_language( $post_id );
+		$parent_post_type = $post_parent > 0 ? get_post_type( $post_parent ) : null;
 		// Dont break the hierarchy in case the post has no language
-		if ( ! empty( $lang ) ) {
+		if ( ! empty( $lang ) && ! empty( $parent_post_type ) && $this->model->is_translated_post_type( $parent_post_type ) ) {
 			$post_parent = $this->model->post->get_translation( $post_parent, $lang );
 		}
+
 		return $post_parent;
 	}
 
--- a/tests/phpunit/tests/test-parent-page.php
+++ b/tests/phpunit/tests/test-parent-page.php
@@ -81,4 +81,18 @@ class Parent_Page_Test extends PLL_UnitTestCase {
 		$this->assertEquals( get_post( $child_page_id )->post_parent, 0 );
 
 	}
+
+	public function test_should_not_modify_parent_when_its_post_type_is_untranslatable() {
+		register_post_type( 'unstranslatable_cpt' );
+
+		$parent_id = $this->factory->post->create( array( 'post_title' => 'untranslated parent cpt', 'post_type' => 'unstranslatable_cpt' ) );
+
+		$child_page_id = $this->factory->post->create( array( 'post_title' => 'post with a untranslated parent', 'post_type' => 'page' ) );
+		self::$model->post->set_language( $child_page_id, 'en' );
+
+		wp_update_post( array( 'ID' => $child_page_id, 'post_parent' => $parent_id ) );
+
+		$child_page = get_post( $child_page_id );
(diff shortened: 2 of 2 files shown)


include/crud-posts.php
```diff
@@ -197,10 +197,12 @@ public function set_object_terms( $object_id, $terms, $tt_ids, $taxonomy ) {
 	 */
 	public function wp_insert_post_parent( $post_parent, $post_id ) {
 		$lang = $this->model->post->get_language( $post_id );
+		$parent_post_type = $post_parent > 0 ? get_post_type( $post_parent ) : null;
```
Screenfeed, line 200: This also prevents to run `get_translation()` when the post type is translated:

```suggestion
		if ( $parent_id <= 0 ) {
			return $parent_id;
		}

		$lang = $this->model->post->get_language( $post_id );
		$parent_post_type = get_post_type( $post_parent );
```

include/crud-posts.php
```diff
@@ -197,10 +197,12 @@ public function set_object_terms( $object_id, $terms, $tt_ids, $taxonomy ) {
...
 		// Dont break the hierarchy in case the post has no language
-		if ( ! empty( $lang ) ) {
+		if ( ! empty( $lang ) && ! empty( $parent_post_type ) && $this->model->is_translated_post_type( $parent_post_type ) ) {
 			$post_parent = $this->model->post->get_translation( $post_parent, $lang );
```
Screenfeed, line 203: `get_translation()` can return `false` if there is no translation:
```suggestion
			$post_parent = (int) $this->model->post->get_translation( $post_parent, $lang );
```

Screenfeed (changes requested):
First I considered performance and I thought about doing this:

```php

// Don't break the hierarchy in case the post has no language.
if ( ! empty( $post_parent ) && ! empty( $lang ) ) {
	$post_parent_tr = $this->model->post->get_translation( $post_parent, $lang );
	$post_parent    = ! empty( $post_parent_tr ) ? $post_parent_tr : $post_parent;
}
```

It ended up being a bad idea imho but I share it with my thoughts:
- If the post type is not translated then `get_translation()` would return `false`. Except that it may not: if the post type was translated at some point (but not anymore), `$post_parent` may still have translations in the database, then `get_translation()` will return a translation ID (yeah I often find funny edge cases) and I don't think this would be what is expected. (I didn't digged deep in the code so maybe there is something somewhere preventing this behavior.)
- If the post type is translated and `$post_parent` doesn't have a translation for this language, `get_translation()` will return `false`, then `$post_parent` doesn't change, then the post created by `wp_insert_post()` (which uses `wp_insert_post_parent()`) will have a parent post that is in another language (yay more fun). With the current implementation, the parent post ID will be `false`, which is better imho.

include/crud-posts.php
manooweb: You mean `post_parent` in the new condition you suggest?

include/crud-posts.php
manooweb: 🤔 I think we don't care as before because `false` converted to `0` in DB

include/crud-posts.php
Screenfeed: Forget it, if `$post_parent` is equal to `0`, then `$parent_post_type` will be `false`, then `get_translation()` won't run.

include/crud-posts.php
Screenfeed: Indeed, it doesn't break anything, but that doesn't mean we have to be lax about types ლ(ಠ益ಠლ

PR #1004, opened 2022-04-06
Prevent term_exists() to be filtered by language

Screenfeed:

WP 6.0 introduced a major change in `term_exists()`: it now uses `get_terms()` internally, which is filtered by language by Polylang.
This PR uses the new hook `term_exists_default_query_args` to prevent `term_exists()` to be filtered by language, by adding a `'lang' => '',` default argument.


Hug0-Drelon (approved):
OK 👍 
But one question remains for me. Why don't we add the filter (and others for that matter) only for the concerned WordPress version (and above) ? 
I understand that if the filter doesn't exist it'll have no impact, but we are still executing some code for nothing right ?

PR #1006, opened 2022-04-08
Delete all leftovers during language deletion

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1291

## Problem

Having two languages defined, when deleting one of these, `wp_term_relationships` rows remains in the database. 
The `term_taxanomy_id` corresponding are those from the deleted `post_translations` and `term_translations`.

## Changes

- If one `object_id` is left in the translation group, add its ID to the `wp_term_relationships.object_id` to delete. Note that the corresponding `wp_term_relationships.term_taxonomy_id` is already well set (see: https://github.com/polylang/polylang/blob/ef9fc783fa386debed895f92c6cf515865df3148/admin/admin-model.php#L462)
- Add test to cover language deletion with three languages then two languages defined. (The `PLL_Admin_Model::update_translations()` was not covered properly and the items in database was not checked)

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -456,16 +456,16 @@ class PLL_Admin_Model extends PLL_Model {
 	public function update_translations( $old_slug, $new_slug = '' ) {
 		global $wpdb;
 
-		$terms    = get_terms( array( 'post_translations', 'term_translations' ) );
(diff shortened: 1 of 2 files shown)


manooweb (approved):
OK 👌

PR #1007, opened 2022-04-08
Fix arrays comparisons in model's tests

Screenfeed:

Using `assertEquals()` to compare arrays [seems to fail with php 8.0 + WP 6.0](https://app.travis-ci.com/github/polylang/polylang/jobs/566142065). This fixes this issue by using `assertSameSets()`.


manooweb:
🤔 isn't the moment to do it everywhere? See https://github.com/polylang/polylang-pro/pull/952

PR #1010, opened 2022-04-08
Fixes failing tests for the inline edit form

Screenfeed:

In `Columns_Test::test_post_inline_edit()`:
- Use the constant `LIBXML_NOERROR` as 2nd argument for `DOMDocument->loadHTML()`. This prevents an error about duplicate HTML id tag that happens when the test is run separately.
- The HTML markup has changed in WP 6.0: there is a new `div` between the `td` and the `fieldset`. The solution is to use relative queries in `DOMXpath->query()`, and not specifying the full path.
- Prevent order issues with the `<option>` tags by comparing arrays and using `assertSameSets()`.

### DIFF
--- a/tests/phpunit/tests/test-columns.php
+++ b/tests/phpunit/tests/test-columns.php
@@ -256,8 +256,10 @@ class Columns_Test extends PLL_UnitTestCase {
 		ob_start();
 		$list_table->inline_edit();
 		$form = ob_get_clean();
-		$doc = new DomDocument();
-		$doc->loadHTML( $form, LIBXML_NOERROR );
+		// Temporarily fix issue https://core.trac.wordpress.org/ticket/55575.
+		$form = preg_replace( '@ (aria-labelledby|id)="inline-edit-legend"@', '', $form );
+		$doc  = new DomDocument();
+		$doc->loadHTML( $form );
 		$xpath = new DOMXpath( $doc );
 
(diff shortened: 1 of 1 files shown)


Hug0-Drelon (approved):
OK 👍

tests/phpunit/tests/test-columns.php
```diff
@@ -257,20 +257,34 @@ public function test_post_inline_edit() {
 		$list_table->inline_edit();
 		$form = ob_get_clean();
 		$doc = new DomDocument();
-		$doc->loadHTML( $form );
+		$doc->loadHTML( $form, LIBXML_NOERROR );
```
Chouby, line 260: Instead of ignoring the error, I suggest to open a ticket on wp.org to report the bug introduced by https://github.com/WordPress/WordPress/commit/337e4a8d9075e8376a3e39c8993ddc8e7d953d65

Screenfeed: Ticket opened [here](https://core.trac.wordpress.org/ticket/55575).

Chouby: Thanks. Would it be a good idea to remove the problematic id from `$form` with some `str_replace()` before passing it to `loadHTML()`? This would avoid to ignore all errors including those we could add ourselves.

PR #1011, opened 2022-04-12
Clean useless JS and centralize function

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/742

Remove useless part of JS code concerning the metabox. (the language selector and the translations table were modified while the page was being refreshed right after).

Centralization of the `initTranslations()` function to avoid duplication.


js/src/block-editor.js
```diff
@@ -106,25 +110,7 @@ jQuery(
...
-												init_translations();
-											break;
-											case 'flag': // Flag in front of the select dropdown
-												// Data is built and come from server side and is well escaped when necessary
-												$( '.pll-select-flag' ).html( this.data ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.html
-											break;
-										}
-									}
-								);
+							function() {
```
manooweb, line 113: It seems to me we can just call `blockEditorSavePostAndReloadPage()` without making the ajax call i.e. directly after clicking the confirmation button in the dialog box.
Indeed as we don't need to refresh the metabox and the code to do that was removed, it also not required to make the ajax call.

manooweb (changes requested):

manooweb:
Also look at PHPCS Error https://app.travis-ci.com/github/polylang/polylang/jobs/566822293, please

js/src/block-editor.js
```diff
@@ -107,13 +107,7 @@ jQuery(
...
 						}
 
-						$.post(
-							ajaxurl,
-							data,
-							function() {
-								blockEditorSavePostAndReloadPage();
-							}
-						);
+						blockEditorSavePostAndReloadPage();
```
manooweb, line 110: Sorry but `data` variable above used in the ajax call is also useless.

manooweb (changes requested):

manooweb (approved):
ok 👌

Chrystll:
That's Ok for me. I didn't detect any issue. Here are the done tests.

Test with Polylang Pro 3.3-dev and the Legacy metabox for block editor mode that I set in the wp-config.php define('PLL_USE_BLOCK_EDITOR_PLUGIN', false);
**_Block editor mode_**
Test 1: create a new post > don't fill a title and don't save it > change language > the page is reloaded
- check you get the Polylang notice: no but this the expected behavior when there isn't any title filled and the post is not saved
- check the categories, tags and Media Library are re-loaded in the right language : yes
- check autocomplete field, you should get the title in the right language : yes


Test 2: create a new post > fill a title and save it > change language > the page is reloaded
- check you get the Polylang notice: yes
- check the categories and tags and Media Library are re-loaded in the right language : yes
- check autocomplete field, you should get the title in the right language : yes

**_Classic editor mode_**
Test 3: create a new post > don't fill a title and don't save it > change language > the page is not reloaded
- check you get the Polylang notice : no but this the expected behavior when there isn't any title filled and the post is not saved
- check the categories, tags and Media Library are re-loaded in the right language :  yes
- check autocomplete field, you should get the title in the right language : yes

Test 4: create a new post > fill a title and save it > change language > the page is reloaded
- check you get the Polylang notice: yes
- check the categories, tags and Media Library are re-loaded in the right language : yes
- check autocomplete field, you should get the title in the right language : yes

PR #1014, opened 2022-04-14
[Container] Module Lingotek

Screenfeed:
This PR uses the new container to store and instanciate the Lingotek module.

This is quite simple since this module's constructor doesn't have any parameters.

What's new though, is that the instance can now be retrieved through the main `$polylang` object (with `$polylang->get( 'lingotek' )`), like all the other modules will be.

The test `Admin_Notices_Test->test_lingotek_notice()` has been slightly modified to reflect this change.

Follow-up #1012.

Note: based on #1012's branch, needs a rebase before reviewing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,8 @@
 	"homepage": "https://polylang.pro",
 	"type": "wordpress-plugin",
 	"require": {
-		"php": ">=5.6"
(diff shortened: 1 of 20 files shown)


Chouby:
We won't need it.

PR #1017, opened 2022-04-15
Avoid redirection in secondary language for terms request in permalinks

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1336

## Problem
- A redirection is made when a request is made on a term in a secondary language toward the default one.
- `PLL_Frontend_Filters_Links::get_queried_term_id()` seems to always return the term in default language when called in `check_canonical_url()`. See: https://github.com/polylang/polylang/blob/2abf0174d98ca97a668cc0e17e55be95ccc47a6e/frontend/frontend-filters-links.php#L400
- Later when retrieving the term link, it returns the wrong redirect link. See: https://github.com/polylang/polylang/blob/2abf0174d98ca97a668cc0e17e55be95ccc47a6e/frontend/frontend-filters-links.php#L404

## Changes
- In `PLL_Frontend_Filters_Links::get_queried_term_id()`, get the terms with the `lang` parameter set.
- If no term is returned, then the queried language might be the wrong one.
- In this case, get the terms with no language set and return it.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -395,7 +395,7 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 
 		elseif ( is_category() || is_tag() || is_tax() ) {
 			if ( $this->model->is_translated_taxonomy( $this->get_queried_taxonomy( $this->wp_query()->tax_query ) ) ) {
-				if ( $this->links_model->using_permalinks && ( ! empty( $this->wp_query()->query['cat'] ) || ! empty( $this->wp_query()->query['tag'] ) ) ) {
+				if ( $this->links_model->using_permalinks && ( ! empty( $this->wp_query()->query['cat'] ) || ! empty( $this->wp_query()->query['tag'] ) || ! empty( $this->wp_query()->query['category_name'] ) ) ) {
 					// When we receive a plain permalink with a cat or tag query var, we need to redirect to the pretty permalink.
 					$term_id = $this->get_queried_term_id( $this->wp_query()->tax_query );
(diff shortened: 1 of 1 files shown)


Screenfeed (approved):

frontend/frontend-filters-links.php
```diff
@@ -533,6 +534,15 @@ protected function get_queried_term_id( $tax_query ) {
 			'fields' => 'ids',
 		);
 		$terms = get_terms( $args );
+
+		$tr_term = $this->model->term->get_translation( reset( $terms ), $lang );
+
+		if ( $tr_term ) {
```
Chouby, line 540:
```suggestion
		if ( $tr_term && in_array( $tr_term, $terms ) ) {
```
You must check that the translated term has the same slug.

Chouby (changes requested):

frontend/frontend-filters-links.php
```diff
@@ -533,6 +534,24 @@ protected function get_queried_term_id( $tax_query ) {
 			'fields' => 'ids',
 		);
 		$terms = get_terms( $args );
+
+		$filtered_terms_by_lang = array_filter(
+			$terms,
+			function ( $term ) use ( $lang ) {
+				$term_lang = $this->model->term->get_language( $term );
+
+				return $term_lang->slug === $lang;
```
Screenfeed, line 543: `get_language()` may return `false`.

```suggestion
				return ! empty( $term_lang ) && $term_lang->slug === $lang;
```

Screenfeed (changes requested):

Screenfeed (approved):

PR #1018, opened 2022-04-15
Disable language dropdown for default categories in quick edit form

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1333

## Changes 
-  Both `PLL_Admin_Filters_Columns::term_column()` and `PLL_Admin_Default_Term::term_column()` are filtering `manage_category_custom_column`. But in `PLL_Admin_Filters_Columns` we do not use the passed `$out` variable and thus overriding  what's done in `PLL_Admin_Filters_Columns`. So we avoid that now by concatenating the strings. 
- Add a test for the `default_cat_{$term_id}` identifier.

## Test instructions
- Visit the quick edit in terms.
- Language dropdown should be disable for default terms.

### DIFF
--- a/tests/phpunit/tests/test-columns.php
+++ b/tests/phpunit/tests/test-columns.php
@@ -273,4 +273,27 @@ class Columns_Test extends PLL_UnitTestCase {
 		$this->assertEquals( 'en', $options->item( 1 )->getAttribute( 'value' ) );
 		$this->assertEquals( 'fr', $options->item( 2 )->getAttribute( 'value' ) );
 	}
+
+	public function test_custom_term_column_for_default_category() {
+		$GLOBALS['taxonomy']     = 'category';
+		$GLOBALS['post_type']    = 'post';
+		$out                     = '';
+		$column                  = 'language_en';
+		$default_cat_id          = get_option( 'default_category' );
+		self::$model->term->set_language( $default_cat_id, 'en' );
+		$admin_default_term = new PLL_Admin_Default_Term( $this->pll_admin );
+		$admin_default_term->add_hooks();
+		$column = apply_filters( 'manage_category_custom_column', $out, $column, $default_cat_id );
+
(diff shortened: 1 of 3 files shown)


Chrystll:
That's OK for me. I will test it again if changes are made in the code after the Marianne's review.

tests/phpunit/tests/test-columns.php
```diff
@@ -273,4 +275,25 @@ public function test_post_inline_edit() {
...
+
+		$doc = new DomDocument();
+		$doc->loadHTML( $column );
+		$xpath = new DOMXpath( $doc );
+
+		$def_cat = $xpath->query( "//div[@id=\"default_cat_{$default_cat_id}\"]" );
+		$this->assertSame( 1, $def_cat->length, 'Only one element with the default category ID should be rendered.' );
+		$def_cat_class_attr = $def_cat->item( 0 )->getAttribute( 'class' );
+		$this->assertSame( 'hidden', $def_cat_class_attr, 'The element of the default category should be hidden.' );
+	}
```
Marianne380, line 298: Shouldn't we pass this test in `Default_Term_Test` ( see https://github.com/polylang/polylang/blob/master/tests/phpunit/tests/test-default-term.php )? 
The tests for the default terms are here.
I don't know which is better or more consistent.

Hug0-Drelon: It seems to me that both locations are relevant. What do you think @Chouby ?

Chouby: I would follow Marianne's suggestion.

Hug0-Drelon: Fine, I've done it in 7f7d285
I leave the `PLL_Admin_Default_Term` instantiation in `Columns_Test` to be sur the column is also filtered for default category. Is it ok for you ?

PR #1024, opened 2022-04-22
Avoid to redirect if the queried object is defined not to use WordPress rewriting feature

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/1340

As explained in [the issue](https://github.com/polylang/polylang-pro/issues/1340) the infinite redirection loop occurs when:
- Polylang is set in directory
- WordPress permalink setting is set as pretty permalink and defined without trailing slash and with `/%category%/` placeholder
- A custom post type is defined not to use rewriting feature.

_Could be in conflicts with #1049_

## Changes

As the WP `redirect_canonical()` function remove the trailing slash in the URL, this one become something like that http://mysite.com/fr?elementor_library=my-template.

So this PR propose to correct the regular expression in `PLL_Links_Directory::remove_language_from_link()` to remove correctly the language from the URL in this case and then avoid to add twice the language by calling  `PLL_Links_Directory::add_language_to_link()`.

But this fix isn't sufficient because `PLL_Links_Directory::add_language_to_link()` always added a slash behind the language code. https://github.com/polylang/polylang/blob/3.2.2/include/links-directory.php#L71
So we loop again with the starting URL http://mysite.com/fr/?elementor_library=my-template and fall in a infinite loop too.

The PR also propose to use `user_trailingslashit()`, as WP does in `redirect_canonical()` https://github.com/WordPress/WordPress/blob/5.9.3/wp-includes/canonical.php#L665,  to remove the slash behind the language code when it is added to the URL and required and thus avoid this infinite loop.

## Expected behaviour
We don't have anymore the infinite redirection as in the customer case

### DIFF
--- a/include/links-directory.php
+++ b/include/links-directory.php
@@ -68,7 +68,7 @@ class PLL_Links_Directory extends PLL_Links_Permalinks {
 	public function add_language_to_link( $url, $lang ) {
 		if ( ! empty( $lang ) ) {
 			$base = $this->options['rewrite'] ? '' : 'language/';
-			$slug = $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : $base . $lang->slug . '/';
+			$slug = user_trailingslashit( $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : $base . $lang->slug );
 			$root = ( false === strpos( $url, '://' ) ) ? $this->home_relative . $this->root : preg_replace( '#^https?://#', '://', $this->home . '/' . $this->root );
 
 			if ( false === strpos( $url, $new = $root . $slug ) ) {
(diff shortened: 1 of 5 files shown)


Screenfeed (approved):

tests/phpunit/tests/test-canonical.php
```diff
@@ -50,6 +50,16 @@ function( $taxonomy ) {
 				}
 			}
 		);
+		$test = register_post_type(
```
Screenfeed, line 53:
```suggestion
		register_post_type(
```

Screenfeed (changes requested):

tests/phpunit/tests/test-canonical.php
manooweb: Forgotten after debugging and verify everything is ok with registering the post type 🙂

Done in 6b8770cdc399302902e14c7e24a74b29ffe243ac

include/links-directory.php
```diff
@@ -102,8 +107,12 @@ public function remove_language_from_link( $url ) {
...
 
 			$pattern = preg_quote( $root, '#' );
-			$pattern = '#' . $pattern . ( $this->options['rewrite'] ? '' : 'language/' ) . '(' . implode( '|', $languages ) . ')(/|$)#';
-			$url = preg_replace( $pattern, $root, $url );
+			$pattern = '#(?<root>' . $pattern . ')' . ( $this->options['rewrite'] ? '' : 'language/' ) . '(?:' . implode( '|', $languages ) . ')((?<query>\?)|/|$)#';
+			// $1 backreference corresponds on root part of URL.
+			// $2 backreference corresponds on the query string separator or slash or end of URL.
+			// $3 backreference corresponds on the query string separator only; returns nothing if there is something between the language code and the separator.
+			// Language part in the URL isn't captured.
+			$url = preg_replace( $pattern, '$1$3', $url );
```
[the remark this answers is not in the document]
manooweb: I understand the goal and some ideas are good but 🤔 I'm not sure it works. We have to discuss together.

PR #1025, opened 2022-04-22
Added a method to check for languages

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/1341

Added a method `has_languages()` to check if there are languages.

We call this method to check if there are languages before initializing `get_links_model()`.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -81,7 +81,7 @@ class PLL_Model {
 			return true;
 		}
 
-		$post_languages = get_terms( array( 'taxonomy' => 'language', 'hide_empty' => false ) );
+		$post_languages = get_terms( array( 'taxonomy' => 'language', 'hide_empty' => false, 'orderby' => 'term_group' ) );
 		if ( empty( $post_languages ) || is_wp_error( $post_languages ) ) {
 			return false;
 		} else {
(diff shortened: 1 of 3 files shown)


include/model.php
```diff
@@ -65,6 +65,30 @@ public function __construct( &$options ) {
...
+	public function has_languages() {
+		if ( false !== $this->cache->get( 'languages' ) ) {
+			return true;
+		}
+
+		if ( false !== get_transient( 'pll_languages_list' ) ) {
+			return true;
+		}
+
+		$post_languages = get_terms( array( 'taxonomy' => 'language', 'hide_empty' => false ) );
```
Screenfeed, line 84: I suggest to use the same arguments that the ones used in `get_languages_list()`:

```suggestion
		$post_languages = get_terms( array( 'taxonomy' => 'language', 'hide_empty' => false, 'orderby' => 'term_group' ) );
```

The reason:
`get_terms()` uses `WP_Term_Query->query( $args )`, which uses [a cache system based on the arguments](https://github.com/WordPress/WordPress/blob/5097c83eac3f1446bfa8c31c4ae3d7a858df76ba/wp-includes/class-wp-term-query.php#L761-L764). So if `get_languages_list()` is used after `has_languages()`, WP won't do another query to the DB. And this doesn't add any complexity to the query itself.

include/model.php
```diff
@@ -65,6 +65,30 @@ public function __construct( &$options ) {
...
+		if ( empty( $post_languages ) || is_wp_error( $post_languages ) ) {
+			return false;
+		} else {
+			return true;
+		}
```
Screenfeed, line 89: Simpler, imho:
```suggestion
		return ! empty( $post_languages ) && ! is_wp_error( $post_languages );
```

Screenfeed (changes requested):

Screenfeed (approved):

PR #1030, opened 2022-04-26
Re set the ajax call for post_lang_choice changes

Marianne380:
See the original issue : https://github.com/polylang/polylang-pro/issues/742

See previous PR that introduced the issue : https://github.com/polylang/polylang/pull/1011 

Update database as soon as possible with the ajax call, see : https://github.com/polylang/polylang/blob/master/admin/admin-classic-editor.php#L185


manooweb (changes requested):
OK👌

manooweb (approved):
ok 👌

PR #1038, opened 2022-05-03
Prevent random languages oreder with WP 6.0

Screenfeed:
With WP 6.0, the order of the languages seems inconsistent for languages having the same `term_group`.

This was fixed temporarilly in tests by https://github.com/polylang/polylang/pull/1020.

The first idea was to set the `term_group` each time a language is created, and add an upgrade routine to set the `term_group` on existing languages.

A simpler approach is to use the filter `'get_terms_orderby'` to set `orderby` to `term_group, term_id` instead of simply `term_group`.


Screenfeed:
Closing in favor of #1041 because GitHub Desktop messed up the rebase and it's faster to create a new branch instead of fixing this crap.

Issue #1027, opened 2022-04-25
Dropdown selector not correctly accessible


maecatnou:
Issue raised by client though Helscout issue => https://secure.helpscout.net/conversation/1857320988/19125?folderId=4673451

## Bug noticed
The dropdown selector isn't accessible as it should be. 
It appears to be missing the lang attributes and, according to the customer, should look more like what is referenced here => https://codyhouse.co/blog/post/accessible-language-picker#html-structure

## Side note
Greg noticed that there was some existing code that should do this but for the time being isn't likely to work due to how it is written.
Check => https://github.com/polylang/polylang/blob/master/include/walker-dropdown.php#L38=

PR #1039, opened 2022-05-03
Fix dropdown selector

Marianne380:
Fix https://github.com/polylang/polylang/issues/1027

The `element` object didn't have a `get_locale` method, so an empty string was returned.

Now we get the locale directly from the `element` object, and transform it to **w3c** format.

### DIFF
--- a/include/walker-dropdown.php
+++ b/include/walker-dropdown.php
@@ -35,7 +35,7 @@ class PLL_Walker_Dropdown extends Walker {
 		$output .= sprintf(
 			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
 			'url' === $value_type ? esc_url( $element->$value_type ) : esc_attr( $element->$value_type ),
-			! empty( $element->locale ) ? sprintf( ' lang="%s"', esc_attr( str_replace( '_', '-', $element->locale ) ) ) : '',
+			! empty( $element->locale ) ? sprintf( ' lang="%s"', esc_attr( $element->locale ) ) : '',
 			selected( isset( $args['selected'] ) && $args['selected'] === $element->$value_type, true, false ),
 			esc_html( $element->name )
 		);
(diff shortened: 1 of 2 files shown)


include/walker-dropdown.php
```diff
@@ -35,7 +35,7 @@ public function start_el( &$output, $element, $depth = 0, $args = array(), $curr
 		$output .= sprintf(
 			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
 			'url' === $value_type ? esc_url( $element->$value_type ) : esc_attr( $element->$value_type ),
-			method_exists( $element, 'get_locale' ) ? sprintf( ' lang="%s"', esc_attr( $element->get_locale( 'display' ) ) ) : '',
+			! empty( $element->locale ) ? sprintf( ' lang="%s"', esc_attr( str_replace( '_', '-', $element->locale ) ) ) : '',
```
Chouby, line 38: It seems to me that `$element->locale` is already what we are expectng: https://github.com/polylang/polylang/blob/3.2.2/include/switcher.php#L128

Chouby (changes requested):

include/walker-dropdown.php
manooweb: > It seems to me that `$element->locale` is already what we are expectng: https://github.com/polylang/polylang/blob/3.2.2/include/switcher.php#L128

Yes you're right and also already converted if required: https://github.com/polylang/polylang/blob/3.2.2/include/language.php#L217

tests/phpunit/tests/test-switcher.php
```diff
@@ -185,6 +185,9 @@ public function test_dropdown() {
 		$this->assertEquals( 'selected', $option->item( 0 )->getAttribute( 'selected' ) );
 		$this->assertNotEmpty( $xpath->query( '//select/option[.="Français"]' )->length );
 		$this->assertNotEmpty( $xpath->query( '//script' )->length );
+		$lang_attributes = $xpath->query( '//select/option/@lang' );
+		$this->assertEquals( self::$model->get_language( 'en' )->get_locale( 'display' ), $lang_attributes->item( 0 )->value );
+		$this->assertEquals( self::$model->get_language( 'fr' )->get_locale( 'display' ), $lang_attributes->item( 1 )->value );
```
Chouby, line 190: I would rather hardcode the expected values.

Chouby (changes requested):

Screenfeed (approved):

Chrystll:
That's OK for me. This PR add the 'lang' parameter for the Language switcher widget in dropdown mode. This parameter already exists for Language swicther inside menu dropdown mode.

PR #1040, opened 2022-05-04
Update PHPStan Baseline since esc_sql() parsing has now dynamic return type

Hug0-Drelon:
szepeviktor/phpstan-wordpress is now updated with this [PR](https://github.com/szepeviktor/phpstan-wordpress/pull/96) which allow a better dynamic return type for `esc_sql()`. Formerly it leads to several errors on implode function which accepts only array. Now PHPStan understand we give it an `array` and not `string|array` type.


manooweb (approved):
👌ok

PR #1041, opened 2022-05-04
Prevent random languages oreder with WP 6.0

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1345.

With WP 6.0, the order of the languages seems inconsistent for languages having the same `term_group`.

This was fixed temporarilly in tests by https://github.com/polylang/polylang/pull/1020.

The first idea was to set the `term_group` each time a language is created, and add an upgrade routine to set the `term_group` on existing languages.

A simpler approach is to use the filter `'get_terms_orderby'` to set `orderby` to `term_group, term_id` instead of simply `term_group`.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -60,6 +60,7 @@ class PLL_Model {
 		add_action( 'update_option_home', array( $this, 'clean_languages_cache' ) );
 
 		add_filter( 'get_terms_args', array( $this, 'get_terms_args' ) );
+		add_filter( 'get_terms_orderby', array( $this, 'filter_language_terms_orderby' ), 10, 3 );
 
 		// Just in case someone would like to display the language description ;).
 		add_filter( 'language_description', '__return_empty_string' );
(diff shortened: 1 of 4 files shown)


phpstan.neon.dist
```diff
@@ -33,6 +33,11 @@ parameters:
 			count: 1
 			path: admin/admin-base.php
 
+		-
```
manooweb, line 36: Could you please add a comment about this ignored error even if it's ununderstable ? 🤪

manooweb (changes requested):

phpstan.neon.dist
Screenfeed: Done in [8b381b2](https://github.com/polylang/polylang/pull/1041/commits/8b381b2611127599a88e49dc09dec6e8d2040dab) and it's perfect.

manooweb (approved):
ok 👌

include/model.php
```diff
@@ -60,6 +60,7 @@ public function __construct( &$options ) {
 		add_action( 'update_option_home', array( $this, 'clean_languages_cache' ) );
 
 		add_filter( 'get_terms_args', array( $this, 'get_terms_args' ) );
+		add_filter( 'get_terms_orderby', array( $this, 'filter_language_terms_orderby' ), 10, 3 );
```
Chouby, line 63: I suggest to add this filter only when we need it in `get_languages_list()` and `has_languages()` and to remove it once it has done its work, to avoid lots of calls to this filter.

Chouby (changes requested):

include/model.php
Screenfeed: Done in [0665014](https://github.com/polylang/polylang/pull/1041/commits/0665014f42bd585c4fcf28197f42087da9900567).
I created a new method that factorizes the common code from `get_languages_list()` and `has_languages()`. It's `protected` but maybe it could be useful as `public` someday.

PR #1042, opened 2022-05-05
Load strings translations accordingly to user locale when the password is retrieved 

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1346

## Issue
When the user locale is set on site default, the strings translations wasn't loaded during `retrieve_password()` call. Which means that the `blogname` was not translated neither in the mail subject nor in the body.

## Changes
- Hooks `PLL_Base::load_strings_translations` to `lostpassword_post` action. This way, strings are loaded even if there is no locale switching later in `retrieve_password()`. (e.g. if the user has site default language set)
- Add test for `blogname` translation in retrieve password mail for default language and for secondary language.

*Note:  `retrieve_password()` switch the locale before getting the `blogname` option, which allow us to translate this string correctly only for secondary language (when a true locale switch occurs). By loading the strings translations when  `lostpassword_post` action is done, we are sure to load at least the site default language anyway.*

### DIFF
--- a/include/base.php
+++ b/include/base.php
@@ -62,7 +62,7 @@ abstract class PLL_Base {
 		add_action( 'pll_language_defined', array( $this, 'load_strings_translations' ), 5 );
 		add_action( 'change_locale', array( $this, 'load_strings_translations' ) ); // Since WP 4.7
 		add_action( 'personal_options_update', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WP, for confirmation request when changing the user email.
-		add_action( 'lostpassword_post', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WordPress switch the locale in retrieve_password().
+		add_action( 'lostpassword_post', array( $this, 'load_strings_translations' ), 10, 0 ); // Password reset email.
 		// Switch_to_blog
 		add_action( 'switch_blog', array( $this, 'switch_blog' ), 10, 2 );
 	}
--- a/include/base.php
+++ b/include/base.php
@@ -62,6 +62,7 @@ abstract class PLL_Base {
 		add_action( 'pll_language_defined', array( $this, 'load_strings_translations' ), 5 );
 		add_action( 'change_locale', array( $this, 'load_strings_translations' ) ); // Since WP 4.7
 		add_action( 'personal_options_update', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WP, for confirmation request when changing the user email.
+		add_action( 'lostpassword_post', array( $this, 'load_strings_translations' ), 1, 0 );
 
 		// Switch_to_blog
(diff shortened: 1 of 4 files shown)


tests/phpunit/tests/test-strings.php
```diff
@@ -209,4 +210,60 @@ public function test_switch_to_locale() {
...
+			)
+		);
+
+		// Set Polylang environment.
+		$admin = new PLL_Admin( $this->links_model );
+		$admin->init();
+
+		// Let's send the mail.
+		$mailer = tests_retrieve_phpmailer_instance();
+		$result = retrieve_password( 'janeDoe' );
```
Chouby, line 235: To make it compatible with WP < 5.7, you can set the user with `$_POST['user_login']`.

include/base.php
```diff
@@ -62,6 +62,7 @@ public function __construct( &$links_model ) {
 		add_action( 'pll_language_defined', array( $this, 'load_strings_translations' ), 5 );
 		add_action( 'change_locale', array( $this, 'load_strings_translations' ) ); // Since WP 4.7
 		add_action( 'personal_options_update', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WP, for confirmation request when changing the user email.
+		add_action( 'lostpassword_post', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WP, for confirmation request when changing the user email.
```
Chouby, line 65: The comment is not correct.

Chouby (changes requested):
I suggest group all email tests together. The best is t create a new dedicated test class.

Chrystll (approved):

tests/phpunit/tests/test-strings.php
Hug0-Drelon: Thanks ! Done in f055b58

include/base.php
```diff
@@ -62,7 +62,7 @@ public function __construct( &$links_model ) {
 		add_action( 'pll_language_defined', array( $this, 'load_strings_translations' ), 5 );
 		add_action( 'change_locale', array( $this, 'load_strings_translations' ) ); // Since WP 4.7
 		add_action( 'personal_options_update', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WP, for confirmation request when changing the user email.
-
+		add_action( 'lostpassword_post', array( $this, 'load_strings_translations' ), 1, 0 ); // Before WordPress switch the locale in retrieve_password().
```
Chouby, line 65:
```suggestion
		add_action( 'lostpassword_post', array( $this, 'load_strings_translations' ), 10, 0 ); // Password reset email.
```
In this case, we don't need use a specific priority, so we can keep the default value. The second parameter set to 0 is important though.

Chouby (changes requested):

PR #1047, opened 2022-05-13
Fix language selector for product quick edit

Hug0-Drelon:
## Issue
In our Polylang plugin for WooCommerce, a regression occurred because of this [commit](https://github.com/polylang/polylang/commit/c010007ade943893f1d022859cf6135a61d9f707) .
We check only for post quick edit table on this line: https://github.com/polylang/polylang/blob/c010007ade943893f1d022859cf6135a61d9f707/js/src/post.js#L28

See: 
Capture d’écran 2022-05-13 à 17 52 54

## Changes
- ~Also test `inline-edit-product` class in the quick edit table.~
- Tests the inserted node contains the class `inline-editor` which seems to exists for all kind of post types in the quick edit form.

This PR is a quick fix, there should be another way to handle this allowing us to handle all post types. Maybe based on the `location.href` ?

### DIFF
--- a/js/src/post.js
+++ b/js/src/post.js
@@ -25,7 +25,7 @@ jQuery(
 		const handleQuickEditInsertion = ( mutationsList ) => {
 			for ( const mutation of mutationsList ) {
 				const form = mutation.addedNodes[0];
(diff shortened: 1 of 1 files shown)


Hug0-Drelon:
Note that the issue also occurs for pages.

PR #1049, opened 2022-05-16
keep query string in redirect url from the requested one

### LINKED ISSUE #1048: GET requests fail on archive (category) pages
Hi, I have notices the following issue:
When I try to do GET request my category page i.e. https://example.com/en/category/products/?sort=ASC it is redirected to https://example.com/en/category/products/ . When I disabled Polylang the redirect doesn't happen. and GET requests work fine. It happens only on archive URLs. On other pages I can send GET requests without any problems.

manooweb:
Fixes #1048 

Regression probably introduced #1017 and especially the condition on `category_name` https://github.com/polylang/polylang/blob/3.2.2/frontend/frontend-filters-links.php#L398

`get_term_link()` or `get_term_feed_link()`  functions generate the new URL from scratch without taking care of an existing query string in the URL.

## Changes
- Add a condition to check if a query string exist in the requested URL and add it to the redirect URL generated by the two functions mentioned above.
- Take care to remove query string parameters which can be used for rewriting.

## Expected behaviour
No redirection when the URL is correct and just has a query string in addition.
Same behaviour as before in Polylang 3.1.4

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -403,7 +403,7 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 					} else {
 						$redirect_url = $this->maybe_add_page_to_redirect_url( get_term_link( $term_id ) );
 					}
-					$redirect_url = $this->maybe_keep_query_string_in_redirect_url( $requested_url, $redirect_url);
+					$redirect_url = $this->maybe_keep_query_string_in_redirect_url( $requested_url, $redirect_url );
 					$language = $this->get_queried_term_language();
 				} else {
 					// We need to switch the language when there is no language provided in a pretty permalink.
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -403,10 +403,18 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 					} else {
 						$redirect_url = $this->maybe_add_page_to_redirect_url( get_term_link( $term_id ) );
 					}
-					// Keep query string from the requested URL if any.
-					$parsed_requested_url = wp_parse_url( $requested_url );
-					if ( $parsed_requested_url && isset( $parsed_requested_url['query'] ) && ! empty( $parsed_requested_url['query'] ) ) {
-						$redirect_url = $redirect_url . '?' . $parsed_requested_url['query'];
+					// Keeps query string from the requested URL if any.
+					$query_string = wp_parse_url( $requested_url, PHP_URL_QUERY );
+
+					if ( is_string( $query_string ) && ! empty( $query_string ) ) {
+						// Removes query string parameters which can be used for rewriting.
+						parse_str( $query_string, $parameters );
+						foreach ( $parameters as $parameter_name => $value ) {
+							if ( in_array( $parameter_name, array( 'cat', 'tag', 'category_name', 'feed', 'paged' ), true ) ) {
+								unset( $parameters[ $parameter_name ] );
+							}
+						}
+						$redirect_url = $redirect_url . ( empty( $parameters ) ? '' : '?' . http_build_query( $parameters ) );
 					}
 					$language = $this->get_queried_term_language();
 				} else {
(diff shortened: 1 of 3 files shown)


Chrystll (approved):

frontend/frontend-filters-links.php
```diff
@@ -403,6 +403,19 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
+					if ( is_string( $query_string ) && ! empty( $query_string ) ) {
+						// Removes query string parameters which can be used for rewriting.
+						parse_str( $query_string, $parameters );
+						foreach ( $parameters as $parameter_name => $value ) {
+							if ( in_array( $parameter_name, array( 'cat', 'tag', 'category_name', 'feed', 'paged' ), true ) ) {
+								unset( $parameters[ $parameter_name ] );
+							}
+						}
+						$redirect_url = $redirect_url . ( empty( $parameters ) ? '' : '?' . http_build_query( $parameters ) );
+					}
```
Hug0-Drelon, line 418: `check_canonical_url()` is going bigger and bigger. Maybe we could move this code inside a dedicated method ?

frontend/frontend-filters-links.php
```diff
@@ -403,6 +403,19 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
...
+
+					if ( is_string( $query_string ) && ! empty( $query_string ) ) {
+						// Removes query string parameters which can be used for rewriting.
+						parse_str( $query_string, $parameters );
+						foreach ( $parameters as $parameter_name => $value ) {
+							if ( in_array( $parameter_name, array( 'cat', 'tag', 'category_name', 'feed', 'paged' ), true ) ) {
+								unset( $parameters[ $parameter_name ] );
+							}
+						}
+						$redirect_url = $redirect_url . ( empty( $parameters ) ? '' : '?' . http_build_query( $parameters ) );
```
Hug0-Drelon, line 417: Could we bail out earlier if `$parameters` is empty (before the loop) ?

Hug0-Drelon (commented):
The issue occurs also for tags (and before 3.1.4 IMHO). Could we add some tests for tags too ? It seems more tests won't hurt `check_canonical_url()` as there is one issue coming after another 😅

frontend/frontend-filters-links.php
manooweb: I asked myself this question 😁

frontend/frontend-filters-links.php
manooweb: 🤔I'm not sure to understand what you thought.
If `$parameters` is empty we never come into the loop and we need to do the loop to remove WordPress the known query vars parameters used by the rewrinting feature.

manooweb:
> The issue occurs also for tags (and before 3.1.4 IMHO). Could we add some tests for tags too ? It seems more tests won't hurt `check_canonical_url()` as there is one issue coming after another 😅

Why not and for custom taxonomy too, I think.

frontend/frontend-filters-links.php
manooweb: Done in 3cbef55b6c8ba1b15c1fb52af29425839dc5bec1

frontend/frontend-filters-links.php
Hug0-Drelon: You're right, we would'n pass through the loop anyway. Sorry 😅

Hug0-Drelon (commented):
Some errors remain for custom taxonomy. It's weird that when `Canonical_Test` is called on its own, everything pass, but when the whole test suite is launched it's failing ... 🤔 
Also, when passing through `check_canonical_url()`, `is_tax()` is `true` when the class is called only but `false` with all the tests. It's hard to kwon what can change this behavior.

manooweb:
> Some errors remain for custom taxonomy. It's weird that when `Canonical_Test` is called on its own, everything pass, but when the whole test suite is launched it's failing ... 🤔 Also, when passing through `check_canonical_url()`, `is_tax()` is `true` when the class is called only but `false` with all the tests. It's hard to know what possibly change this behavior.

Solve by creating the custom taxonomy term in each test where it's required.

PR #1058, opened 2022-06-07
Force empty string translation to empty string

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1387

The issue was originally reported by https://secure.helpscout.net/conversation/1903439781/19600

Basically, when the strings translations include a translation for `'0'`, let say `'0'`, the empty string is translated to `'0'`.

https://github.com/polylang/polylang/blob/3.2.4/include/mo.php#L37 was probably intended to avoid this kind of issue, but this doesn't work. 

The root cause is explained in https://core.trac.wordpress.org/ticket/55941. However, we can work around this WP bug directly in Polylang, at least for string translations.

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -146,7 +146,11 @@ function pll_register_string( $name, $string, $context = 'Polylang', $multiline 
  * @return string The string translated in the current language.
  */
 function pll__( $string ) {
-	return is_scalar( $string ) ? __( $string, 'pll_string' ) : $string; // PHPCS:ignore WordPress.WP.I18n
+	if ( ! is_scalar( $string ) || '' === $string ) {
+		return $string;
+	}
+
+	return __( $string, 'pll_string' ); // PHPCS:ignore WordPress.WP.I18n
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Screenfeed (approved):

Hug0-Drelon (commented):
Should we add a test for `pll_translate_string()`?
Also, could we avoid registering empty string in the first place in `pll_register_string()`? Or is it done on purpose?

Screenfeed:
At first I was thinking this should be only a temporary fix that should be removed when fixed in core, because `''` is a "valid" translation string like `'0'`. But then I realized that such generic string would be used with a context (`_x()`), so not in PLL's translation strings.

Chouby:
>Also, could we avoid registering empty string in the first place in `pll_register_string()`? Or is it done on purpose?
I believe that it's already the case: https://github.com/polylang/polylang/blob/3.2.4/admin/admin-strings.php#L55

Chouby:
> But then I realized that such generic string would be used with a context (`_x()`), so not in PLL's translation strings.

They should indeed. But hopefully, if taken into account by WP core, then this should fix both cases as the issue's coming from https://github.com/WordPress/WordPress/blob/6.0/wp-includes/pomo/entry.php#L92-L94 which is excuted for both `__()` and `_x()`.

Hug0-Drelon (approved):
OK 👍

PR #1060, opened 2022-06-16
Avoid conflict with 3rd party

sebastienserre:
This PR fix https://github.com/polylang/polylang-pro/issues/1391


manooweb (approved):
`span.icon-default-lang:before` would be enough to fix the conflict however I also prefer this solution by targeting the parent HTML tag.
OK 👌

PR #1062, opened 2022-06-21
Make the Site Health report more human readable.

sebastienserre:
Some options in Polylang are stored with info which are  boolean or integer and are not easy to understand to non developer.
This PR will add the options text available in the Polylang settings with the old value.
Thanks to the old value, the developers will be able to debug their development easily.

### DIFF
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -179,6 +179,7 @@ class PLL_Admin_Site_Health {
 			case 'media_support':
 				if ( ! $value ) {
 					$value = esc_html__( 'The media are not translated (0)', 'polylang' );
+					break;
 				}
 				// translators: the placeholder is the option stored in database.
 				$value = esc_html__( 'The media are translated (%s)', 'polylang' );
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -118,7 +118,6 @@ class PLL_Admin_Site_Health {
 	 * @return mixed Option value.
 	 */
 	public function format_value( $key, $value ) {
-		$db_value = $value;
 		switch ( $key ) {
 			case 'browser':
 				if ( ! $value ) {
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -109,6 +109,96 @@ class PLL_Admin_Site_Health {
 	}
 
 	/**
+	 * Transform the option value to readable human sentence.
+	 *
+	 * @since 3.3
+	 *
+	 * @param string                $key option name.
+	 * @param string|array|bool|int $value option value.
+	 *
+	 * @return string|array $value option value.
+	 */
+	public function format_value( $key, $value ) {
+		$db_value = $value;
+		switch ( $key ) {
+			case 'browser':
+				if ( true === $value ) {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Browser prefered language activated (%s)', 'polylang' );
+				} else {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Browser prefered language deactivated (%s)', 'polylang' );
+				}
+				break;
+			case 'rewrite':
+				if ( 1 === $value ) {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Remove /language/ in pretty permalinks (%s)', 'polylang' );
+				} else {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Keep /language/ in pretty permalinks (%s)', 'polylang' );
+				}
+				break;
+			case 'hide_default':
+				if ( 1 === $value ) {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Hide URL language information for default language (%s)', 'polylang' );
+				} else {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Display URL language information for default language (%s)', 'polylang' );
+				}
+				break;
+			case 'force_lang':
+				switch ( $value ) {
+					case '0':
+						// translators: the placeholder is the option stored in database.
+						$value = __( 'The language is set from content (%s)', 'polylang' );
+						break;
+					case '1':
+						// translators: the placeholder is the option stored in database.
+						$value = __( 'The language is set from the directory name in pretty permalinks (%s)', 'polylang' );
+						break;
+					case '2':
+						// translators: the placeholder is the option stored in database.
+						$value = __( 'The language is set from the subdomain name in pretty permalinks (%s)', 'polylang' );
+						break;
+					case '3':
+						// translators: the placeholder is the option stored in database.
+						$value = __( 'The language is set from different domains (%s)', 'polylang' );
+						break;
+				}
+				break;
+			case 'redirect_lang':
+				if ( 1 === $value ) {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'The front page URL contains the language code instead of the page name or page id (%s)', 'polylang' );
+				} else {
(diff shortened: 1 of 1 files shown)


modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,96 @@ function ( &$value, $key ) {
...
+	 *
+	 * @return string|array $value option value.
+	 */
+	public function format_value( $key, $value ) {
+		$db_value = $value;
+		switch ( $key ) {
+			case 'browser':
+				if ( true === $value ) {
+					// translators: the placeholder is the option stored in database.
+					$value = __( 'Browser prefered language activated (%s)', 'polylang' );
```
Marianne380, line 127: Looks like using `%s` with boolean false values will give you this : 
**Browser prefered language deactivated ()**

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,96 @@ function ( &$value, $key ) {
 		return implode( ' | ', $array );
 	}
 
+	/**
+	 * Transform the option value to readable human sentence.
```
Marianne380, line 112:
```suggestion
	 * Transforms the option value to readable human sentence.
```

Marianne380 (changes requested):

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,96 @@ function ( &$value, $key ) {
...
+				break;
+
+			case 'sync':
+				if ( is_array( $value ) && empty( $value ) ) {
+					$value = __( 'Synchronization disabled (0)', 'polylang' );
+				}
+				break;
+		}
+		if ( ! is_array( $value ) ) {
+			$value = sprintf( $value, $db_value );
```
Marianne380, line 196: As PHPStan mentioned : 
```
  Line   modules/site-health/admin-site-health.php                 
 ------ ---------------------------------------------------------- 
  196    Parameter #1 $format of function sprintf expects string,  
         bool|int|string given.                                    
  196    Parameter #2 ...$values of function sprintf expects       
         bool|float|int|string|null, array|bool|int|string given
```
May be check the value's type before using `sprintf()` on it ?

sebastienserre: I really don't know how to manage this.
I've change to  `if ( is_string( $value ) ) {` but I always have the same error.
I wonder if the problem isn't in my PHPDoc declaration. I'm not sure of 
```
	 * @param string|array|bool|int $value option value.
	 * @return string|array $value option value.
```
it looks like a bag where  I put all possible type :-)

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,94 @@ function ( &$value, $key ) {
...
+					$value = esc_html__( 'The front page URL contains the language code instead of the page name or page id (%s)', 'polylang' );
+				} else {
+					// translators: the placeholder is the option stored in database.
+					$value = esc_html__( 'The front page URL contains the page name or page id instead of the language code (%s)', 'polylang' );
+				}
+				break;
+			case 'media_support':
+				if ( ! $value ) {
+					$value = esc_html__( 'The media are not translated (0)', 'polylang' );
+				}
```
Marianne380, line 182:
```suggestion
				}
				break;
```
Looks like you forget the `break`

Marianne380 (changes requested):

modules/site-health/admin-site-health.php
sebastienserre: It's in line 185

modules/site-health/admin-site-health.php
Marianne380: I don't really know the best way to handle this.

@Chouby What do you think ? 

Should we cast the expected type like this : 
```
     if ( ! is_array( $value ) ) {
         $value = (string) $value;
         if ( ! is_array( $db_value ) ) {
            $value = sprintf( $value, $db_value );
         }
      }
```

modules/site-health/admin-site-health.php
sebastienserre: Oh yes! Thank you @Marianne380

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,95 @@ function ( &$value, $key ) {
...
+		switch ( $key ) {
+			case 'browser':
+				if ( ! $value ) {
+					// translators: the placeholder is the option stored in database.
+					$value = esc_html__( 'Browser prefered language deactivated (0)', 'polylang' );
+					break;
+				}
+				// translators: the placeholder is the option stored in database.
+				$value = esc_html__( 'Browser prefered language activated (%s)', 'polylang' );
+				break;
```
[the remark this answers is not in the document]
sebastienserre: done

modules/site-health/admin-site-health.php
```diff
@@ -115,74 +115,62 @@ function ( &$value, $key ) {
 	 *
 	 * @param string                $key option name.
 	 * @param string|array|bool|int $value option value.
-	 * @return string|array $value option value.
+	 * @return array|bool|int|string $value option value.
```
Chouby, line 118:
```suggestion
	 * @return mixed Option value.
```

Chouby (changes requested):
I wonder if we should not put the option value in parenthesis outside the translated string because:
- this value is meaningful only to us, not to the user or the translators.
- this allows to reuse existing string from the setting interface.
What do you think?

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,80 @@ function ( &$value, $key ) {
 		return implode( ' | ', $array );
 	}
 
+	/**
+	 * Transforms the option value to readable human sentence.
+	 *
+	 * @since 3.3
+	 *
+	 * @param string                $key option name.
+	 * @param string|array|bool|int $value option value.
```
Chouby, line 117:
```suggestion
	 * @param string $key   Option name.
	 * @param mixed  $value Option value.
```

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,80 @@ function ( &$value, $key ) {
...
+	 * @return mixed Option value.
+	 */
+	public function format_value( $key, $value ) {
+		switch ( $key ) {
+			case 'browser':
+				if ( ! $value ) {
+					$value = '0: ' . esc_html__( 'Browser prefered language deactivated', 'polylang' );
+					break;
+				}
+				$value = '1: ' . esc_html__( 'Browser prefered language activated', 'polylang' );
```
Chouby, line 127:
```suggestion
					$value = '0: ' . esc_html__( 'Browser preferred language detection deactivated', 'polylang' );
					break;
				}
				$value = '1: ' . esc_html__( 'Browser preferred language detection activated', 'polylang' );
```

modules/site-health/admin-site-health.php
```diff
@@ -108,6 +108,80 @@ function ( &$value, $key ) {
...
+			case 'media_support':
+				if ( ! $value ) {
+					$value = '0: ' . esc_html__( 'The media are not translated', 'polylang' );
+					break;
+				}
+				$value = '1: ' . esc_html__( 'The media are translated', 'polylang' );
+				break;
+
+			case 'sync':
+				if ( is_array( $value ) && empty( $value ) ) {
```
Chouby, line 176:
```suggestion
				if ( empty( $value ) ) {
```

Chouby (changes requested):

modules/site-health/admin-site-health.php
sebastienserre: I wonder if it would not be better to reuse the same string as in the Settings ? `Detect browser language activated/deactivated`

Chouby: As you want

PR #445, opened 2020-01-27
Fix lang-item-first class in language switcher

Chouby:
The `lang-item-first` class is usually added to the first item of the menu language switcher. 
lang-item-first
This PR aims to fix a bug where this `lang-item-first` is sometimes missing.
**How to reproduce?**
1. Create 2 languages, English first, French second.
2. Creat a menu with a language switcher.
3. Select the option to hide the current language in the language switcher
4. Asign this menu to an English location
5. Visit an English page and examine the html code of the first item of the language switcher in the menu: the `lang-item-first` is missing.

PR #1065, opened 2022-06-24
Set current language just after saving a post type

manooweb:
By studying the [Polylang For WooCommerce #445 issue](https://github.com/polylang/polylang-wc/issues/445) it appears that attributes lookup table update process is run during the post save process and so when the edit post form is submitted.

Until now Polylang set the post current language only when the edit post page is requested by its URL `post.php?post=...`
See https://github.com/polylang/polylang/blob/3.2.4/admin/admin-base.php#L348

So the attributes lookup table update doesn't save the attribute values in the correct language because its `get_terms()` request 
isn't filtered by the correct language.
See https://github.com/woocommerce/woocommerce/blob/6.6.1/plugins/woocommerce/src/Internal/ProductAttributesLookup/LookupDataStore.php#L506-L520
In fact it is filtered by the Polylang admin filter language which could be different from the post ( product ) language.

This PR propose to also set the current language with the post language when it is saved.

This PR is only a part of the problem because it needs that the "Direct updates" is checked in the WooCommerce product advanced settings.

image

I will propose another Polylang For WooCommerce PR to solve the problem for scheduled attributes lookup table updates.

Even if there is no issue reported yet, this PR also fixes this kind of case which would probably cause issue on terms.

## Changes
- add a conditions to set the post  language as the current one when an edit post form is submitted
- remove an ignored PHPStan errors by adding a condition as in the new piece of code
- PHPunit test to cover this case
- form terms at the opposite of posts, the form input fields have the same names as the query string parameters. So I used `$_REQUEST` global variable to process both of `$_GET` and `$_POST` ones.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -345,7 +345,9 @@ abstract class PLL_Admin_Base extends PLL_Base {
 		// Edit Post
 		if ( isset( $_REQUEST['pll_post_id'] ) && $lang = $this->model->post->get_language( (int) $_REQUEST['pll_post_id'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 			$this->curlang = $lang;
-		} elseif ( 'post.php' === $GLOBALS['pagenow'] && isset( $_GET['post'] ) && $this->model->is_translated_post_type( get_post_type( (int) $_GET['post'] ) ) && $lang = $this->model->post->get_language( (int) $_GET['post'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		} elseif ( 'post.php' === $GLOBALS['pagenow'] && isset( $_GET['post'] ) && false !== get_post_type( (int) $_GET['post'] ) && $this->model->is_translated_post_type( get_post_type( (int) $_GET['post'] ) ) && $lang = $this->model->post->get_language( (int) $_GET['post'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$this->curlang = $lang;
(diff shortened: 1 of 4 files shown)


admin/admin-base.php
```diff
@@ -345,7 +345,9 @@ public function set_current_language() {
...
 		if ( isset( $_REQUEST['pll_post_id'] ) && $lang = $this->model->post->get_language( (int) $_REQUEST['pll_post_id'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 			$this->curlang = $lang;
-		} elseif ( 'post.php' === $GLOBALS['pagenow'] && isset( $_GET['post'] ) && $this->model->is_translated_post_type( get_post_type( (int) $_GET['post'] ) ) && $lang = $this->model->post->get_language( (int) $_GET['post'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		} elseif ( 'post.php' === $GLOBALS['pagenow'] && isset( $_GET['post'] ) && false !== get_post_type( (int) $_GET['post'] ) && $this->model->is_translated_post_type( get_post_type( (int) $_GET['post'] ) ) && $lang = $this->model->post->get_language( (int) $_GET['post'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$this->curlang = $lang;
+		} elseif ( 'post.php' === $GLOBALS['pagenow'] && isset( $_POST['post_ID'] ) && false !== get_post_type( (int) $_POST['post_ID'] ) && $this->model->is_translated_post_type( get_post_type( (int) $_POST['post_ID'] ) ) && $lang = $this->model->post->get_language( (int) $_POST['post_ID'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
Screenfeed, line 350: Why not doing 1 check with `$_REQUEST` instead of 2 with `$_GET` and `$_POST`? Like you did for the terms.

manooweb: Simply because in query string the parameter is named `post` instead of 'post_ID' when the form is submited

Screenfeed: Ah!

Screenfeed (approved):

PR #1067, opened 2022-07-01
Checks if current language is set in walker dropdown

FloLech:
In case languages are filtered it might occure, that the default language is not accessible. In this case $current might be empty. In this case I use the first avaliable language.

Context: I created a plugin to restrict editable languages for certain users.


Chouby:
Sorry for the late answer. We have worked a bit on this class recently so I'm coming back now on your PR.
However, I'm not sure to understand what you want to fix.
Should I understand that you are passing an intentionnally incomplete languages list as first argument and `$args['selected']` contains a language not present in this filtered languages list?
Could you explain more of your context?

PR #1068, opened 2022-07-04
Handle OR operator in WP_Tax_Query

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1389

## Issue
Given this query:
```
$args = array(
	'post_type' => 'post',
	'lang' => 'fr',
	'posts_per_page' => 10,
	'tax_query' => array(
		'relation' => 'OR',
		array(
			'taxonomy' => 'category',
			'field' => 'name',
			'terms' => array( '25' )
		),
		array(
			'taxonomy' => 'category',
			'field' => 'name',
			'terms' => array( 'FR' )
		)
	)
);
$loop = new WP_Query( $args );
```
WordPress will merge the `language` taxonomy with the queried ones [here](https://github.com/WordPress/wordpress-develop/blob/1c0a2efa5eac05dfd3b7d2b9cfe68e02da55b966/src/wp-includes/class-wp-query.php#L1131-L1170). Resulting in a taxonomy query like this:
```
'tax_query' => array(
		'relation' => 'OR',
		array(
			'taxonomy' => 'category',
			'field' => 'name',
			'terms' => array( '25' )
		),
		array(
			'taxonomy' => 'category',
			'field' => 'name',
			'terms' => array( 'FR' )
		),
		array(
			'taxonomy' => 'language',
			'field' => 'name',
			'terms' => array( 'fr' )
		)
	)
```
Unfortunately, `PLL_Query::set_language()` is not called when such a query is made since the language parameter is already existing, and thus the code handling the `OR` operator is not read. See: https://github.com/polylang/polylang/blob/3.2.5/include/query.php#L118-L123

## Changes
- Add a test to confirm the bug.
- In `PLL_Frontend::parse_tax_query()` add a `AND` operator between the language and the other queried terms if the `OR` operator is found in the first place. Seems to be a good place to do that, since this kind of query is made on front only, right?

### DIFF
--- a/include/query.php
+++ b/include/query.php
@@ -30,6 +30,7 @@ class PLL_Query {
 	public function __construct( &$query, &$model ) {
 		$this->query = &$query;
 		$this->model = &$model;
+		add_action( 'parse_tax_query', array( $this, 'parse_tax_query' ) );
 	}
 
 	/**
@@ -183,6 +184,51 @@ class PLL_Query {
 			if ( isset( $qvars['lang'] ) && 'all' === $qvars['lang'] ) {
 				unset( $qvars['lang'] );
 			}
+		}
+	}
+
+	/**
+	 * Filters the taxonomy query to add the language properly when the 'OR' operator is used.
+	 *
+	 * @since 3.3
+	 *
+	 * @param WP_Query $query The current query object to filter.
+	 * @return void
+	 */
+	public function parse_tax_query( $query ) {
+		// If the OR operator is used, the language is not the queried object.
+		/** @var WP_Tax_Query $tax_query */
+		$tax_query = $query->tax_query;
+		if ( ! empty( $tax_query->queried_terms ) && 'OR' === $tax_query->relation && array_key_exists( 'language', $tax_query->queried_terms ) ) {
+			// First get the queried language.
+			if ( ! isset( $tax_query->queried_terms['language']['terms'][0] ) ) {
+				// No language is queried.
+				return;
+			}
+			$lang = $this->model->get_language( $tax_query->queried_terms['language']['terms'][0] );
+
+			// Then remove the language from the queried terms.
+			$tax_query->queries = array_filter(
+				$tax_query->queries,
+				function( $query ) {
+					return ! isset( $query['taxonomy'] ) || 'language' !== $query['taxonomy'];
+				}
+			);
+
+			// Finally add the language to the queried terms with the AND operator between it and the orther terms.
+			$lang_query = array(
+				'taxonomy' => 'language',
+				'field'    => 'slug',
+				'terms'    => $lang,
+				'operator' => 'IN',
+			);
+			$tax_query = array(
+				$lang_query,
+				get_object_vars( $query->tax_query ),
+				'relation' => 'AND',
+			);
+			$tax_query        = new WP_Tax_Query( $tax_query );
+			$query->tax_query = $tax_query;
 		}
 	}
 }
--- a/include/query.php
+++ b/include/query.php
@@ -175,8 +175,8 @@ class PLL_Query {
 			}
 		} else {
 			// Set the language correctly in the query, since WordPress merges the language with the other query vars when the operator is OR.
-			if ( isset( $qvars['lang'] ) ) {
-				$lang = $this->model->get_language( $qvars['lang'] );
+			if ( isset( $this->query->tax_query->queried_terms['language'] ) && 1 === count( $this->query->tax_query->queried_terms['language']['terms'] ) ) {
+				$lang = $this->model->get_language( reset( $this->query->tax_query->queried_terms['language']['terms'] ) );
 				unset( $qvars['lang'] );
 				if ( $lang instanceof PLL_Language ) {
 					$this->set_language( $lang );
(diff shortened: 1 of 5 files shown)


tests/phpunit/tests/test-query.php
```diff
@@ -704,4 +704,211 @@ public function test_sticky_posts() {
...
+		$french_posts_cat_1 = $this->factory->post->create_many( 3, array( 'post_type' => 'post' ) );
+		foreach ( $french_posts_cat_1 as $index => $post ) {
+			self::$model->post->set_language( $post, 'fr' );
+			self::$model->post->save_translations(
+				$post,
+				array(
+					'en' => $eng_posts_cat_1[ $index ],
+					'fr' => $post,
+				)
+			);
```
Chouby, line 748: You should not need to assign translations for this test (here and everywhere else in your two tests).

include/query.php
```diff
@@ -174,6 +174,14 @@ public function filter_query( $lang ) {
 				}
 			}
 		} else {
+			// Set the language correctly in the query, since WordPress merges the language with the other query vars when the operator is OR.
+			if ( isset( $this->query->tax_query->queried_terms['language'] ) && 1 === count( $this->query->tax_query->queried_terms['language']['terms'] ) ) {
```
Chouby, line 178: We will have the same problem when assigning multiple language.
I propose to modify `set_language()` for it to accept `PLL_Language|PLL_Language[]`.

tests/phpunit/tests/test-query.php
```diff
@@ -704,4 +704,211 @@ public function test_sticky_posts() {
...
+		self::$model->term->set_language( $first_cat_fr, 'fr' );
+
+		$second_cat_en = $this->factory->term->create( array( 'taxonomy' => 'category', 'name' => 'test2' ) );
+		self::$model->term->set_language( $second_cat_en, 'en' );
+
+		$second_cat_fr = $this->factory->term->create( array( 'taxonomy' => 'category', 'name' => 'essai2' ) );
+		self::$model->term->set_language( $second_cat_fr, 'fr' );
+
+		// English Posts.
+		$eng_posts_cat_1 = $this->factory->post->create_many( 3, array( 'post_type' => 'post' ) );
```
Chouby, line 723: Do you really need many posts? If no, you can save test time by creating only one.

Chouby (changes requested):

tests/phpunit/tests/test-query.php
Hug0-Drelon: I have to check agains post with no taxonomy. And I need two taxonomies created to perform a `OR` query. So 3 posts are required IMHO.

include/query.php
```diff
@@ -100,16 +100,27 @@ public function get_queried_taxonomies() {
 	 * Optimized for (and requires) WP 3.5+.
 	 *
 	 * @since 2.2
+	 * @since 3.3 Accepts now an array of languages.
 	 *
-	 * @param PLL_Language $lang Language object.
+	 * @param PLL_Language|PLL_Language[] $langs Language object(s).
```
Chouby, line 105:
```suggestion
	 * @param PLL_Language|PLL_Language[] $languages Language object(s).
```

include/query.php
```diff
@@ -100,16 +100,27 @@ public function get_queried_taxonomies() {
...
 	 * @return void
 	 */
-	public function set_language( $lang ) {
+	public function set_language( $langs ) {
+		if ( $langs instanceof PLL_Language ) {
+			$langs = array( $langs );
+		}
```
Chouby, line 111: For me, it's more logical to check it like this:
```suggestion
	public function set_language( $languages ) {
		if ( ! is_array( $languages ) ) {
			$languages = array( $languages );
		}
```

include/query.php
```diff
@@ -100,16 +100,27 @@ public function get_queried_taxonomies() {
...
+		$tt_ids = array_map(
+			function( $lang ) {
+				return $lang->term_taxonomy_id;
+			},
+			$langs
+		);
```
Chouby, line 117: Think to this function:
```suggestion
		$tt_ids = wp_list_pluck( $languages, 'term_taxonomy_id' );
```

include/query.php
```diff
@@ -174,6 +185,20 @@ public function filter_query( $lang ) {
...
 		} else {
+			// Set the language correctly in the query, since WordPress merges the language with the other query vars when the operator is OR.
+			if ( isset( $this->query->tax_query->queried_terms['language'] ) && 'all' !== $this->query->tax_query->queried_terms['language']['terms'] ) {
+				$langs = $this->query->tax_query->queried_terms['language']['terms'];
+				if ( is_string( $langs ) ) {
+					$langs = explode( ',', $langs );
+				}
+				$langs = array_map( array( $this->model, 'get_language' ), $langs );
+				$langs = array_filter( $langs );
+				unset( $qvars['lang'] );
```
Chouby, line 196: Do we ne need to unset?

tests/phpunit/tests/test-query.php
Chouby: Reopening as this is not solved.

include/query.php
Hug0-Drelon: Thought it would be necessary, but you're right it's not.

tests/phpunit/tests/test-query.php
Hug0-Drelon: Forgot one test, my bad!

PR #1071, opened 2022-07-13
Fix - Generate links alternate hreflang for front page static only if is actived for that

MarcosFurquim:
## Problem
When you have a Front Page configured with an static page, all links  alternate:
like:
```<link rel="alternate" href="http://localhost:8080/pt/" hreflang="pt-PT" />```
 are printed on head HTML,  even it is not bound to all languages
 
<p align="center"> 
<br/><i>Columbia and Mexico Lang isnt bound to front page 1/2</i>
</p>

<br><br>

<p align="center"> 
<br/><i>Columbia and Mexico Lang isnt bound to front page 2/2</i>
</p>

<br><br>

Captura de Tela 2022-07-13 às 05 18 40

<p align="center">
<i>HTML <code>head</code> printed with all languagues actived on Polylang, ignoring the configs of front-page atribuation.</i></p>

<br><br>
# Solution

Checkk, when is a page is front page, if the language has `page_on_front`
```php
// Front page when it is the list of posts
elseif ( is_front_page() && $language->page_on_front) {
    $url = $this->get_home_url( $language );
}
```

### After this fix, printed only tags of langs with front page are configured:
Captura de Tela 2022-07-13 às 05 18 03

### DIFF
--- a/frontend/frontend-links.php
+++ b/frontend/frontend-links.php
@@ -63,7 +63,7 @@ class PLL_Frontend_Links extends PLL_Links {
 		 *
 		 * @since 1.8
 		 *
-		 * @param string       $url               Empty or the url of the translation of teh current page.
+		 * @param string       $url               Empty or the url of the translation of the current page.
 		 * @param PLL_Language $language          Language of the translation.
 		 * @param int          $queried_object_id Queried object id.
 		 */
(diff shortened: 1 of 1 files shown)


Chouby:
This is a specific use case, not applicable to all users. See also my answer: https://github.com/polylang/polylang/issues/1072#issuecomment-1191528337

PR #1054, opened 2022-05-30
Get back to 3.1.4 state in check_canonical_url()

Hug0-Drelon:
The following PRs opened the pandora box by trying to handle more redirection cases.
- https://github.com/polylang/polylang/pull/901
- https://github.com/polylang/polylang/pull/887
- https://github.com/polylang/polylang/pull/1017

Prior to Polylang 3.1.4, we didn't handle redirection for feed link. But since then we introduced several bugs...
To go back to a more stable state we decided to revert the modification made in `check_canonical_url()` and leave the new tests to improve them a bit. 

*Note: Some failing tests related to former fixed issues have been commented at the end of the file `tests/phpunit/includes/testcase-canonical.php`*

PR #1074, opened 2022-07-13
Fix canonical redirect

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/1380
Closes #1024
Closes #1049

This PR rewrites the canonical redirect mechanism. 

First, the mechanism is moved to a dedicated class `PLL_Canonical`. It now fully relies on the WP canonical redirect first. To achieve this: 
- For taxonomies, we trick WordPress by temporarily remove the language from the WP_Query in `PLL_Canonical::redirect_canonical()`.
- To let `redirect_canonical()` correctly handle multiple domains, this function is added to the white list of our `home_url` filter.

The PR focuses on:
 - redirect to the same url in the right language if the language in the initial url is incorrect.
 - handling of `www` vs non `www` urls when using multiple domains.

For everything else, it relies on WP.  Thus it doesn't attempt to fix missing redirects from WP. This includes the tickets that we recently opened:
- https://core.trac.wordpress.org/ticket/55880
- https://core.trac.wordpress.org/ticket/55993

The PR uncomments tests added by #1054 and reorganizes canoncal tests by type of content to ease future maintenance. It also moves `www` redirect tests to a dedicated class. Note that the test on homepage has been removed as it fires an error in WP `redirect_canonical` function (~~ticket to follow~~ I assume that it's a mistake to call this function with `http://example.org` instead of `http://example.org/`).


frontend/canonical.php
```diff
@@ -0,0 +1,261 @@
...
+		$this->options     = &$polylang->options;
+		$this->curlang     = &$polylang->curlang;
+	}
+
+	/**
+	 * If the language code is not in agreement with the language of the content,
+	 * redirects incoming links to the proper URL to avoid duplicate content.
+	 *
+	 * @since 0.9.6
+	 *
```
Hug0-Drelon, line 57:
```suggestion
	 * @global WP_Query $wp_query
```

frontend/canonical.php
```diff
@@ -0,0 +1,261 @@
...
+		unset( $queried_terms['language'] );
+
+		return (string) key( $queried_terms );
+	}
+
+	/**
+	 * Evaluates the canonical redirect url through the deidcated WP function.
+	 *
+	 * @since 3.3
+	 *
```
Hug0-Drelon, line 238:
```suggestion
	 * @global WP_Query $wp_query
```

frontend/canonical.php
```diff
@@ -0,0 +1,261 @@
...
+		$this->curlang = $language; // Hack to filter the `page_for_posts` option in the correct language.
+
+		$backup_wp_query = $wp_query;
+
+		if ( isset( $wp_query->tax_query ) ) {
+			unset( $wp_query->tax_query->queried_terms['language'] );
+			unset( $wp_query->query['lang'] );
+		}
+
+		$redirect_url = redirect_canonical( $url, false );
```
Hug0-Drelon, line 255: I wonder if unsetting the language from the query could lead to problem in `redirect_canonical()` for shared slugs?

frontend/canonical.php
```diff
@@ -0,0 +1,261 @@
...
+			return $requested_url;
+		}
+
+		if ( ! $do_redirect ) {
+			return $redirect_url;
+		}
+
+		// Protect against chained redirects.
+		if ( $redirect_url === $this->check_canonical_url( $redirect_url, false ) && wp_validate_redirect( $redirect_url ) ) {
+			wp_safe_redirect( $redirect_url, 301, POLYLANG );
```
Hug0-Drelon, line 154: Do you think we could perform the redirection in an other method, this way we could easily test `check_canonical_url()` with `$do_redirect` set to `true` and use a mock or something to avoid  the redirection in unit test. I remember we had some issue with this part of the code with infinite redirection.

Chouby: I took care to position the [problematic condition](https://github.com/polylang/polylang/pull/1074/files#diff-1f56dc45bed29b524bb07f3e0f6588c99486c88cf33178dd16ec12f984b18521R147) before the `$do_redirect` so we should not be trapped the same way.

Using a dedicated method for redirect was discussed in the past but (temporarily?) abandonned. See https://github.com/polylang/polylang-pro/pull/460. If we restart the idea, we should do it everywhere at once.

Hug0-Drelon (approved):
LGTM 👍 
I try to figure new test, but nothing came to my mind. Except maybe redirection to SSL, but I didn't find anything saying WordPress does that natively (no `w-redirect: WordPress` nor in the doc). Is it the case ?

PR #1076, opened 2022-07-18
Fix CPT and Custom taxonomy information

sebastienserre:
- Fixes https://github.com/polylang/polylang-pro/issues/1412
- Hides domains when the url option is not set to multiple domains.


sebastienserre:
I do not understand why Travis is reporting an error on file I didn't changed ...

PR #1080, opened 2022-07-29
Fix redirect when permalink structure has no trailing slash

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1340
Replaces and closes #1024

This PR fixes `PLL_Links_Directory::remove_language_from_link()` when the path ends with a language code without trailing slash. This could cause wrong redirects due to the [call to `switch_language_in_link()` in `check_canonical_url()`](https://github.com/polylang/polylang/blob/39ca7d4c09f69497a505f99976378a6e4784a855/frontend/canonical.php#L132). 

It adds several tests for `remove_language_from_link()` and `switch_language_in_link()` and also adds tests for the edge case which allowed us to detect the bug (redirect of a custom post with `rewrite` set to `false` and permalink structure without trailing slash).


manooweb (approved):
OK 👌

PR #1081, opened 2022-07-30
Filter search block

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/1305

The block search doesn't honor the `get_search_form` filter, preventing our usual filter to work. I just discovered the filter `render_block_{$block_name}` which can replace it.


manooweb (approved):
OK 👌

PR #1083, opened 2022-08-02
Fix admin render widget nav menu

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1420

## Problem
- `PLL_Frontend_Nav_Menu` is not instantiated during a REST request and thus the method `PLL_Frontend_Nav_Menu::wp_get_nav_menu_items` which render our custom menu items is never called.
- No current language is set in a REST request made from the Widgets page.

## Changes
- Instantiate `PLL_Frontend_Nav_Menu` instead of `PLL_Nav_Menu` in `PLL_REST_Request`.
- Hook on `rest_pre_dispatch` to set the current language to be able to catch a `lang` parameter passed in the body of the request.
- Enqueue `block-editor.js` in the widget and site editors pages.
- Add inline script to pass the default language to `block-editor.js`.
- Change the doc on `PLL_Frontend_Nav_Menu::$curlang` since it be `false` really.
- Add a new test class for widget nav menu.
- Adapt the test for rest request.

## 🆘 
~~I wrote the test in `test-rest-request.php` in the first place and it seems to be OK. But since I move the test to its own class, when the whole suit runs the test fails...~~ Fixed thanks to @Chouby 

## Note
~~We discussed another option to set the current language. Which is to filter the sent requests in `block-editor.js` and use a fallback (the default language) is the current post doesn't have one. Then inject the `block-editor.js` file on the Widget Editor and the Site Editor.
But some issues occurred with it. Indeed, the current language is defined [here](https://github.com/polylang/polylang/blob/3.2.5/include/rest-request.php#L62-L71) but since we add the language parameter in the body of the `POST` request [here](https://github.com/polylang/polylang/blob/3.2.5/js/src/block-editor.js#L24) the global `$_REQUEST` is not populated with the value... 
We could use the `rest_pre_dispatch` filter and get the parameter ourselves but this filter is applied a bit late ...~~
`rest_pre_dispatch` comes early enough for us eventually.


Chouby:
> I wrote the test in `test-rest-request.php` in the first place and it seems to be OK. But since I move the test to its own class, when the whole suit runs the test fails...

Under the hood, the REST request uses [global variables](https://github.com/WordPress/wordpress-develop/blob/6.0.1/src/wp-includes/rest-api/endpoints/class-wp-rest-widget-types-controller.php#L229).
Tests are executed in the alphabetical order:
- Before `test-rest-request.php`, no widget test is made, so you have the default globals set by WordPress.
- In `test-strings.php`, we test widgets and thus [clean up the globals related to widgets](https://github.com/polylang/polylang/blob/3.2.5/tests/phpunit/tests/test-strings.php#L24-L37).
- Then comes your nex `test-widget-nav-menu.php` which gets (as expected) empty globals.

So you need to setup the globals used by `WP_REST_Widget_Types_Controller` yourself.

Hug0-Drelon:
> So you need to setup the globals used by `WP_REST_Widget_Types_Controller` yourself.

Thanks ! It works just fine, I really didn't think about the globals cleaned up earlier in the test suit ...

manooweb (approved):
ok 👌

tests/phpunit/tests/test-rest-request.php
```diff
@@ -39,45 +44,148 @@ public function tear_down() {
...
+	 * }
+	 */
+	public function test_should_define_language_when_language_is_valid( $data ) {
 		self::$model->options['default_lang'] = 'en';
+		$this->pll_rest->init();
 
-		$_REQUEST['lang'] = 'fr';
+		$request = new WP_REST_Request( $data['method'], $data['route'] );
+		$request = $this->set_lang_param( $request, 'fr' );
+		$response = rest_do_request( $request );
```
Chouby, line 62: I suggest to use `Spy_REST_Server` and `$server->dispatch()` instead.
See our other [REST API tests](https://github.com/polylang/polylang-pro/blob/3.2.5/tests/phpunit/tests/pro/test-rest-api.php) or WP REST API tests for examples how to proceed.

tests/phpunit/tests/test-rest-request.php
```diff
@@ -39,45 +44,148 @@ public function tear_down() {
...
+			'comment'  => '/wp/v2/comments',
+			'taxonomy' => '/wp/v2/taxonomies',
+			'media'    => '/wp/v2/media',
+			'user'     => '/wp/v2/users',
+		);
+		$methods = array(
+			'GET',
+			'POST',
+			'PUT',
+			'PATCH',
```
Chouby, line 158: WP aliases them both to POST.

tests/phpunit/tests/test-rest-request.php
```diff
@@ -39,45 +44,148 @@ public function tear_down() {
...
+	 *    @type string $resource Resource name.
+	 * }
+	 */
+	public function test_should_define_language_when_language_is_valid( $data ) {
 		self::$model->options['default_lang'] = 'en';
+		$this->pll_rest->init();
 
-		$_REQUEST['lang'] = 'fr';
+		$request = new WP_REST_Request( $data['method'], $data['route'] );
+		$request = $this->set_lang_param( $request, 'fr' );
```
Chouby, line 61: Would `$request->set_param( 'lang', 'fr' );` not work without needing to create our own method?

tests/phpunit/tests/test-widget-nav-menu.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+class Widget_Nav_Menu_Test extends PLL_UnitTestCase {
```
Chouby, line 3: Could you please clean up the global scope as done in other widget tests?
See https://github.com/polylang/polylang/blob/3.2.5/tests/phpunit/tests/test-widgets-filter.php#L34-L45

tests/phpunit/tests/test-widget-nav-menu.php
```diff
@@ -0,0 +1,133 @@
...
+	 */
+	public function set_up() {
+		global $wp_widget_factory;
+
+		parent::set_up();
+
+		// We need to register the nav menu ourselves since widget globals are cleaned up in test-strings.php.
+		// @see https://github.com/polylang/polylang/blob/3.2.5/tests/phpunit/tests/test-strings.php#L24-L37.
+		register_widget( new WP_Nav_Menu_Widget() );
+		$wp_widget_factory->_register_widgets(); // This will populate the $wp_registered_widgets global used in WP_REST_Widget_Types_Controller::get_widgets().
```
Chouby, line 38: A simple call to `wp_widgets_init()` would do the job.

tests/phpunit/tests/test-widget-nav-menu.php
```diff
@@ -0,0 +1,133 @@
...
+		global $wp_version;
+
+		// Widgets routes are available since WordPress 5.8.
+		// @see https://github.com/WordPress/wordpress-develop/blob/6.0.1/src/wp-includes/rest-api/endpoints/class-wp-rest-widget-types-controller.php#L10-L16.
+		if ( version_compare( $wp_version, '5.8.0', '<' ) ) {
+			$this->markTestSkipped( 'This test requires WordPress 5.8 or higher.' );
+		}
+
+		// Let's create a menu.
+		register_widget( 'WP_Nav_Menu_Widget' );
```
Chouby, line 67: Already done in `set_up()`.

Chouby (changes requested):

tests/phpunit/tests/test-rest-request.php
Hug0-Drelon: Indeed, after some research I found `set_param()` acknowledge the request method and adds the parameter accordingly.

admin/admin-base.php
```diff
@@ -208,7 +212,21 @@ public function admin_enqueue_scripts() {
...
+	 * Tells whether or not the given screen is block editor kind.
+	 * e.g. widget, site or post editor.
+	 *
+	 * @since 3.3
+	 *
+	 * @param WP_Screen $screen Screen object.
+	 *
+	 * @return bool True if the screen is a block editor, false otherwise.
+	 */
+	private function is_block_editor( $screen ) {
```
Chouby, line 228:
```suggestion
	 * @param WP_Screen $screen Screen object.
	 * @return bool True if the screen is a block editor, false otherwise.
	 */
	protected function is_block_editor( $screen ) {
```

tests/phpunit/tests/test-widget-nav-menu.php
```diff
@@ -0,0 +1,144 @@
...
+		self::create_language( 'fr_FR' );
+	}
+
+	/**
+	 * @global $wp_widget_factory
+	 *
+	 * @return void
+	 */
+	public function set_up() {
+		// global $wp_widget_factory;
```
Chouby, line 31:
```suggestion
```

tests/phpunit/tests/test-widget-nav-menu.php
```diff
@@ -0,0 +1,144 @@
...
+
+		parent::set_up();
+
+		// We need to register the nav menu ourselves since widget globals are cleaned up in test-strings.php.
+		// @see https://github.com/polylang/polylang/blob/3.2.5/tests/phpunit/tests/test-strings.php#L24-L37.
```
Chouby, line 36: This comment may quickly become obsolete as things may change when adding new tests, I suggest something more generic if need a comment.
```suggestion
		// We need to register widgets after globals cleanup.
```

Chouby (changes requested):

admin/admin-base.php
```diff
@@ -187,11 +187,15 @@ public function admin_enqueue_scripts() {
 			}
 
 			// Block editor with legacy metabox in WP 5.0+.
-			if ( method_exists( $screen, 'is_block_editor' ) && $screen->is_block_editor() && ! pll_use_block_editor_plugin() ) {
+			if ( $this->is_block_editor( $screen ) ) {
 				$scripts['block-editor'] = array( array( 'post' ), array( 'jquery', 'wp-ajax-response', 'wp-api-fetch', 'jquery-ui-dialog', 'wp-i18n' ), 0, 1 );
 			}
```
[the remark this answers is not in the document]
Hug0-Drelon: Not sure since we check for translatable post type [here](https://github.com/polylang/polylang/blob/3.2.5/admin/admin-base.php#L181) before adding the script for the block editor.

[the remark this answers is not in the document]
Chouby: If I simplify your code, it looks like this:
```
		if ( ! empty( $screen->post_type ) && $this->model->is_translated_post_type( $screen->post_type ) ) {
			// Block editor with legacy metabox in WP 5.0+.
			if ( $this->is_block_editor( $screen ) ) {
				$scripts['block-editor'] = array( array( 'post' ), array( 'jquery', 'wp-ajax-response', 'wp-api-fetch', 'jquery-ui-dialog', 'wp-i18n' ), 0, 1 );
			}
		}

		if ( $this->is_block_editor( $screen ) ) {
			$scripts['block-editor'] = array( array( 'widgets', 'site-editor' ), array( 'jquery', 'wp-ajax-response', 'wp-api-fetch', 'jquery-ui-dialog', 'wp-i18n' ), 0, 1 );
		}
```
The the block editor script is added 2 times:
1. when a translated post type is edited
2. inconditionnally

When 2 is added, 1 is added too. So it looks useless to me.

Chouby (changes requested):

Chouby (approved):

PR #1087, opened 2022-08-05
Fix admin scripts tests

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1426

## Problem
- `Admin_Test::_test_scripts()` wasn't testing all passed scripts. For instance, the test `test_scripts_in_edit_post()` is passing the value `classic-editor` [here](https://github.com/polylang/polylang/blob/3.2.5/tests/phpunit/tests/test-admin.php#L113-L119) whereas the script really enqueued is `block-editor`.

## Changes
- `Admin_Test::_test_scripts()` accepts now any scripts to test if they're enqueued and use a baseline to asserts others aren't.
- `Admin_Test::_test_scripts()` argument is now an array with the position as key and the script (whether it is a name or source) as value.
- Add a test specific to the classic editor.
- Add a test for the widgets.

## Notes
- I wonder if we could make a part of this class a test case, so we can use it in Polylang Pro easily.
- Maybe we could add more tests that didn't come to my mind.


manooweb (approved):
Ok 👌

Hug0-Drelon:
@Chouby About the `polylang_admin` stylesheet, in the current state we have to check for `polylang_admin-css` name because we only check against JS files for sources. Do you want to make it even more generic and differentiate CSS from JS files?

tests/phpunit/includes/testcase-assets.php
```diff
@@ -0,0 +1,158 @@
...
+class PLL_Assets_UnitTestCase extends PLL_UnitTestCase {
+	protected static $editor;
+	protected static $stylesheet;
+	protected $polylang_assets = array(
+		'header' => array(
+			'user' => 'source',
+		),
+		'footer' => array(
+			'pll_ajax_backend'   => 'name',
+			'polylang_admin-css' => 'name',
```
Chouby, line 13: There is a problem if we expect the CSS in the footer.

Chouby (changes requested):

tests/phpunit/includes/testcase-assets.php
Hug0-Drelon: After looking at it it seems we have to call `do_action( 'admin_print_styles' );` in the header to render the stylesheets, otherwise they are rendered in the footer thanks to `_wp_footer_scripts()` which is hooked on `admin_print_footer_scripts` and prints remaining stylesheets (see: https://github.com/WordPress/wordpress-develop/blob/6.0.1/src/wp-includes/script-loader.php#L2118-L2126). 
This explains why no error was raised.

Hug0-Drelon: Fixed in c180077

Chouby:
Instead of a review I proposed my views in #1093

PR #1089, opened 2022-08-12
New tool to run PHPCS only on modified files.

Hug0-Drelon:
## Description
New bash script to run PHPCS only on modified files. The purpose is to make our life easier, nothing more. 😄 

## Note
If modified files are committed, then PHPCS will run as usual. That's why I changed the composer script `composer cs`.


Hug0-Drelon:
Obsolete since https://github.com/polylang/polylang/pull/1562.

PR #1098, opened 2022-08-24
Set language in tax query correctly when OR relation is used

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1389

## Changes:
- Add a new method that checks the current query contains a `OR` relation, if so set the language(s) correctly in the taxonomy query and unset the `lang` parameter (so WordPress doesn't merge it with the other queried terms once again...).
- `PLL_Query::set_language()` now accepts an array of languages.
- Remove an ignored error in the phpstan baseline.
- Fix a phpstan error in `choose-lang.php`.
- Add tests for query with OR relations on category, untranslated custom taxonomy, translated taxonomy and with several languages.

## Notes:
Once it's merged, we will have to take care of shared slugs. See this PR: https://github.com/polylang/polylang/pull/1092

### DIFF
--- a/include/query.php
+++ b/include/query.php
@@ -203,21 +203,28 @@ class PLL_Query {
 	 * @return void
 	 */
 	protected function maybe_set_language_for_or_relation() {
-		if ( $this->query->tax_query instanceof WP_Tax_Query
-			&& 'OR' === $this->query->tax_query->relation
-			&& isset( $this->query->tax_query->queried_terms['language'] )
-			&& 'all' !== $this->query->tax_query->queried_terms['language']['terms'] ) {
-			$langs = $this->query->tax_query->queried_terms['language']['terms'];
-			if ( is_string( $langs ) ) {
-				$langs = explode( ',', $langs );
-			}
-			$langs = array_map( array( $this->model, 'get_language' ), $langs );
-			$langs = array_filter( $langs );
+		if ( ! $this->query->tax_query instanceof WP_Tax_Query ) {
+			return;
+		}
 
-			if ( ! empty( $langs ) ) {
-				$this->set_language( $langs );
-				unset( $this->query->query_vars['lang'] ); // Unset the language query var otherwise WordPress would add the language query by slug in WP_Query::parse_tax_query().
-			}
+		if ( 'OR' !== $this->query->tax_query->relation ) {
+			return;
+		}
+
+		if ( ! isset( $this->query->tax_query->queried_terms['language'] ) ) {
+			return;
+		}
+
+		$langs = $this->query->tax_query->queried_terms['language']['terms'];
+		if ( is_string( $langs ) ) {
+			$langs = explode( ',', $langs );
+		}
+		$langs = array_map( array( $this->model, 'get_language' ), $langs );
+		$langs = array_filter( $langs );
+
+		if ( ! empty( $langs ) ) {
+			$this->set_language( $langs );
+			unset( $this->query->query_vars['lang'] ); // Unset the language query var otherwise WordPress would add the language query by slug in WP_Query::parse_tax_query().
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


include/query.php
```diff
@@ -185,4 +193,31 @@ public function filter_query( $lang ) {
...
+	 * since WordPress merges the language with the other query vars when the relation is OR.
+	 *
+	 * @since 3.3
+	 *
+	 * @return void
+	 */
+	protected function set_language_for_or_relation() {
+		if ( $this->query->tax_query instanceof WP_Tax_Query
+			&& 'OR' === $this->query->tax_query->relation
+			&& isset( $this->query->tax_query->queried_terms['language'] )
```
Chouby, line 208: Suggest splitting in conditions like:
```php
if ( ! $this->query->tax_query instanceof WP_Tax_Query ) {
return;
}
```
for better readability.

include/query.php
```diff
@@ -185,4 +193,31 @@ public function filter_query( $lang ) {
...
+			&& 'all' !== $this->query->tax_query->queried_terms['language']['terms'] ) {
```
Chouby, line 209: Is this condition useful? It looks to me that next lines will remove 'all' as anyway.

tests/phpunit/tests/test-query.php
```diff
@@ -704,4 +721,356 @@ public function test_sticky_posts() {
...
+		}
+		$french_posts_cat_2 = $this->factory->post->create_many( 3, array( 'post_type' => 'post' ) );
+		foreach ( $french_posts_cat_2 as $post ) {
+			self::$model->post->set_language( $post, 'fr' );
+			wp_set_post_terms( $post, array( $second_cat_fr ), 'category' );
+		}
+		$french_posts_no_cat = $this->factory->post->create_many( 3, array( 'post_type' => 'post' ) );
+		foreach ( $french_posts_no_cat as $post ) {
+			self::$model->post->set_language( $post, 'fr' );
+		}
```
Chouby, line 1036: I suggest that you write your tests fixtures in separate functions as most of your tests share more or less the same code. This could paramaterized by taxonomy to avoid repetitions as much as possible.

Chouby (changes requested):

tests/phpunit/tests/test-query.php
Hug0-Drelon: Good idea!

tests/phpunit/tests/test-query.php
```diff
@@ -64,6 +64,102 @@ public function tear_down() {
...
+	/**
+	 * Asserts language in taxonomy query by slug is removed.
+	 *
+	 * @param WP_Tax_Query $tax_query Taxonomy query.
+	 *
+	 * @throws AssertionFailedError Thrown when language slug query is not removed.
+	 *
+	 * @return void
+	 */
+	public function assert_no_lang_query_by_slug( $tax_query ) {
```
Chouby, line 76:
```suggestion
	protected function assert_no_lang_query_by_slug( $tax_query ) {
```

tests/phpunit/tests/test-query.php
```diff
@@ -64,6 +64,102 @@ public function tear_down() {
...
+	 * @return array {
+	 *     Arrays of post IDs, grouped by language.
+	 *
+	 *     @type int[] $eng_posts_with_tax English post IDs linked to the taxonomy.
+	 *     @type int[] $eng_posts_no_tax   English post IDs without the taxonomy.
+	 *     @type int[] $fr_posts_with_tax  French post IDs linked to the taxonomy.
+	 *     @type int[] $fr_posts_no_tax    French post IDs without the taxonomy.
+	 * }
+	 */
+	public function create_posts_with_tax( $taxonomy ) {
```
Chouby, line 98:
```suggestion
	protected function create_posts_with_tax( $taxonomy ) {
```

Chouby (approved):

PR #1100, opened 2022-08-29
Fix media translation setting has no effect

Chouby:
If the `attachement` post type is added in the database (in the `polylang` option) as a translated post type, then deactivating the media translation in the settings screen has no effect. The media are always translatable whatever the option value.

NB: The `attachement` post type may come from WPML imports. The bug was fixed in [WPML to Polylang 0.4](https://github.com/polylang/wpml-to-polylang/issues/7) but many sites may have been imported with this bug, so it's better to fix things in Polylang too.


manooweb (approved):
👌 ok

PR #1102, opened 2022-08-29
Add ajax on front tests

Chouby:
Closes https://github.com/polylang/polylang-pro/issues/48

For ajax on front, we used to test only the translation for logged in users.

This PR adds a set of tests with the following matrix:
- The user is logged in or anonymous.
- The language is default or secondary
- The language is defined by the `lang` parameter or the cookie.

Both the translation and the current language are checked.

### DIFF
--- a/tests/phpunit/tests/test-ajax-on-front.php
+++ b/tests/phpunit/tests/test-ajax-on-front.php
@@ -11,7 +11,7 @@ class Ajax_On_Front_Test extends PLL_Ajax_UnitTestCase {
 		self::create_language( 'en_US' );
 		self::create_language( 'fr_FR' );
 
-		// Copy language file
+		// Copy the language file.
 		@mkdir( DIR_TESTDATA );
 		@mkdir( WP_LANG_DIR );
 		copy( dirname( __FILE__ ) . '/../data/fr_FR.mo', WP_LANG_DIR . '/fr_FR.mo' );
@@ -20,6 +20,8 @@ class Ajax_On_Front_Test extends PLL_Ajax_UnitTestCase {
 	public function tear_down() {
 		parent::tear_down();
 
+		$_REQUEST = array();
+		$_COOKIE  = array();
 		unload_textdomain( 'default' );
 	}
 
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/test-ajax-on-front.php
```diff
@@ -28,26 +30,81 @@ public function _ajax_test_locale() {
 		wp_die( wp_json_encode( __( 'Dashboard' ) ) ); // phpcs:ignore WordPress.WP.I18n.MissingArgDomain
```
Hug0-Drelon, line 30:
```suggestion
		echo wp_json_encode( __( 'Dashboard' ) ); // phpcs:ignore WordPress.WP.I18n.MissingArgDomain
		wp_die();
```

According to the WordPress test suite documentation, `WPAjaxDieStopException` should be used for error and response without output. See: https://github.com/WordPress/wordpress-develop/blob/6.0.2/tests/phpunit/includes/exceptions.php#L12-L34
So maybe we should `echo` the response and call `wp_die()` right after. This way `$this->_handleAjax()` will throw a `WPAjaxDieContinueException`. Then we will be able to use `$this->_last_response` to do our tests.

tests/phpunit/tests/test-ajax-on-front.php
```diff
@@ -28,26 +30,81 @@ public function _ajax_test_locale() {
...
+		add_action( 'wp_ajax_test', array( $this, '_ajax_test_locale' ) );
+
+		$links_model = self::$model->get_links_model();
+		$frontend = new PLL_Frontend( $links_model );
+		$frontend->init();
 
 		try {
-			$this->_handleAjax( 'test_locale' );
+			$this->_handleAjax( 'test' );
 		} catch ( WPAjaxDieStopException $e ) {
```
Hug0-Drelon, line 62:
```suggestion
} catch ( WPAjaxDieContinueException $e ) {
			unset( $e );
		}
		$response = json_decode( $this->_last_response, true );
```
Follows my previous comment.

tests/phpunit/tests/test-ajax-on-front.php
Chouby: Interesting. I never paid attention to this.
In fact I believe there is even better: https://developer.wordpress.org/reference/functions/wp_send_json/

tests/phpunit/tests/test-ajax-on-front.php
Chouby: Done in a1aae1c67a9baec6fbbc1704a0843e3aa691abbf

tests/phpunit/tests/test-ajax-on-front.php
Hug0-Drelon: Indeed, it's even better!

Hug0-Drelon (approved):
OK 👍

PR #1103, opened 2022-08-30
Fix posts without languages displayed in Yoast SEO sitemap

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/279

We used to filter the posts sql query in only 2 cases:
1. When one language per subdomain or per domain (to have one sitemap per language).
2. When a language was inactive (to remove the posts in the inactive language from the sitemap).

We now always filter the query to make sure to remove the posts without language (they must be removed as they are 404).

### DIFF
--- a/integrations/wpseo/wpseo.php
+++ b/integrations/wpseo/wpseo.php
@@ -127,40 +127,43 @@ class PLL_WPSEO {
 	}
 
 	/**
-	 * Modifies the sql request for posts sitemaps
-	 * Only when using multiple domains or subdomains or if some languages are not active
+	 * Modifies the sql request for posts sitemaps.
 	 *
 	 * @since 1.6.4
 	 *
-	 * @param string $sql       JOIN clause
-	 * @param string $post_type
+	 * @param string $sql       JOIN clause.
+	 * @param string $post_type Post type.
 	 * @return string
 	 */
 	public function wpseo_posts_join( $sql, $post_type ) {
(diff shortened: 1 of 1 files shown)


integrations/wpseo/wpseo.php
```diff
@@ -127,40 +127,43 @@ protected function wpseo_get_active_languages() {
...
-			}
+		if ( ! pll_is_translated_post_type( $post_type ) ) {
+			return $sql;
+		}
 
-			if ( $languages = $this->wpseo_get_active_languages() ) {
-				return $sql . PLL()->model->post->where_clause( $languages );
-			}
+		if ( PLL()->options['force_lang'] > 1 ) {
+			return $sql . PLL()->model->post->where_clause( PLL()->curlang );
```
Hug0-Drelon, line 157: Shouldn't we test `PLL()->curlang` is an instance of `PLL_Language` and not `false` or `null` before calling `where_clause()`?

Screenfeed: +1, since we are editing this part...
PhpStan reports the issue:
```
Parameter #1 $lang of method PLL_Translated_Object::where_clause() expects array<string>|PLL_Language|string, PLL_Language|false|null given.
```

Screenfeed (changes requested):

integrations/wpseo/wpseo.php
Screenfeed: Done in [1beb0d1](https://github.com/polylang/polylang/pull/1103/commits/1beb0d1d684463d75c6297c0c46452024a72cd0e).

PR #1105, opened 2022-09-02
Fix parent of translated category removed when assigning an untranslated parent

### LINKED ISSUE #1440: Yoast SEO post type archive breadcrumb translation
## Prerequisites

- [x] This bug happens with only Polylang plugin active (and Yoast SEO)

1. Use Polylang Pro and translate the post type name for a CPT
2. Use Yoast SEO to display breadcrumbs including the post type archive
3. The CPT archive isn't always showing the correct language

## Solution

This topic has already been active on the Yoast SEO forums and WPML (which has the same issue) has scheduled a fix.
I've provided a solution here: https://github.com/Yoast/wordpress-seo/issues/19739#issuecomment-1971764269

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1440 and adds the corresponding PHPUnit test.

My first attempt to fix this issue re-introduced https://github.com/polylang/polylang-pro/issues/266. 
So we need to take care with this PR. Even if https://github.com/polylang/polylang-pro/issues/266 is protected by this [test](https://github.com/polylang/polylang/blob/3.2.5/tests/phpunit/tests/test-sync.php#L412-L444).

### DIFF
--- a/modules/sync/sync.php
+++ b/modules/sync/sync.php
@@ -212,7 +212,7 @@ class PLL_Sync {
 						$tr_parent = $this->model->term->get_translation( $term->parent, $lang );
 						$tr_term   = get_term( (int) $tr_id, $taxonomy );
 
-						if ( $tr_term instanceof WP_Term ) {
+						if ( $tr_term instanceof WP_Term && ! ( $term->parent && empty( $tr_parent ) ) ) {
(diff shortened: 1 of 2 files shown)


manooweb (approved):
ok  👌

PR #1109, opened 2022-09-12
Fix Behat tests

Chouby:
This PR is a consequence of  https://core.trac.wordpress.org/ticket/56514#comment:6

Our tests took profit of the magic method `__get()` to access to the factory object which is now protected.
In this PR, I now copy the `factory()` method and use it instead.


Hug0-Drelon (commented):
To avoid `factory()` copy, `BrowserPreferredLanguageContext` could extends `PLL_UnitTestCase`?

Chouby:
Done!

Hug0-Drelon (approved):
OK 👍

PR #1110, opened 2022-09-12
The factory property is now deprecated

Chouby:
See https://core.trac.wordpress.org/ticket/56514#comment:6


Hug0-Drelon (commented):
I found a few more `factory` non static calls. I made a little commit to replace it, see 906cf6b06b0. I hope you don't mind.

manooweb:
There are still several `$this->factory()` calls. Shouldn't we use `self::factory()` instead ?

manooweb (approved):
LGTM 👌

PR #1111, opened 2022-09-12
Fix update options when two PLL_Translate_Option are instanciated

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1074

This bug was detected with Yoast SEO and using it is convenient to reproduce. However, the bug itself doesn't involve Yoast SEO. It occurs when 2 instances of `PLL_Translate_Option` are created for the same option. SO, I expect that the issue with Yoast SEO was introduced by #796

This PR removes conflicts created by the instantiation of two `PLL_Translate_Option` objects for the same option.
1. In `pre_update_option`, we used to remove the filter `"option_{$name}"`. With 2 instances, removing the filter in the current instance is not sufficient as the second filter is still active when attempting to get the unfiltered option. This is issue is solved by removing the filter in a separated method hooked sooner.
2. Relying on the global `pll__()` to translate the option while updating it resulted in incorrect values in the second instance. So I ended to create local `$mo` objects to work on translations.

A few tests are added to covers potential sources of conflicts ( 2 instances for the same option, or the same string in 2 different options).

### DIFF
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -18,6 +18,13 @@ class PLL_Translate_Option {
 	 */
 	private $keys;
 
+  /**
+	 * Used to prevent filtering when retrieving the raw value of the option.
+	 *
+	 * @var bool
+	 */
+	private static $raw = false;
+
 	/**
 	 * Array of updated strings.
 	 *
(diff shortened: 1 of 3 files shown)


Chrystll:
Your PR fixes https://github.com/polylang/polylang-pro/issues/1074. But I need to make some deeper tests.

Screenfeed (changes requested):

Chrystll (approved):

PR #1112, opened 2022-09-13
CPT and Tax settings: display slug of CPT or Tax

nicomollet:
In the Polylang Settings, in order to easily understand where the CPT or Taxonomies come from, this PR adds to the label, the slug of the CPT or Taxonomy. See image below.

image

### DIFF
--- a/settings/settings-cpt.php
+++ b/settings/settings-cpt.php
@@ -105,11 +105,12 @@ class PLL_Settings_CPT extends PLL_Settings_Module {
 					if ( ! empty( $pt ) ) {
 						$disabled = in_array( $post_type, $this->disabled_post_types );
 						printf(
-							'<li><label><input name="post_types[%s]" type="checkbox" value="1" %s %s/> %s</label></li>',
+							'<li><label><input name="post_types[%s]" type="checkbox" value="1" %s %s/> %s %s</label></li>',
 							esc_attr( $post_type ),
 							checked( in_array( $post_type, $this->options['post_types'] ) || $disabled, true, false ),
 							disabled( $disabled, true, false ),
-							esc_html( $pt->labels->name )
+							esc_html( $pt->labels->name ),
+							'(' . esc_html( $pt->name ) . ')'
 						);
 					}
 				}


Screenfeed (changes requested):
Hello @nicomollet, thank you for your proposal. We would like to go ahead with it after a few adjustments mentioned in my other comment.
Also, it would be nice to also apply this for taxonomies, as this change only applies to post types with the current code 😅

Merci.

settings/settings-cpt.php
```diff
@@ -107,9 +107,16 @@ protected function form() {
 						printf(
 							'<li><label><input name="post_types[%s]" type="checkbox" value="1" %s %s/> %s</label></li>',
 							esc_attr( $post_type ),
-							checked( in_array( $post_type, $this->options['post_types'] ) || $disabled, true, false ),
+							checked( $disabled || in_array( $post_type, $this->options['post_types'] ), true, false ),
```
Screenfeed, line 110: The strict comparison was still missing.
```suggestion
							checked( $disabled || in_array( $post_type, $this->options['post_types'], true ), true, false ),
```

Screenfeed (changes requested):
Could you also reproduce this [for the taxonomies](https://github.com/polylang/polylang/pull/1112/files#diff-cd88aa798d74abe60d0f99e297af9c82cb3eb325eb6ade5d86b9285dc75816b3R138-R144) please?

nicomollet:
Thanks a lot @Screenfeed for the syntax tips.

My PR is done, I just have one suggestion for the UI.
Actually it is 
image

With
```css
.pll-inline-block-list li {
    display: inline-block;
    margin: 0;
    width: 250px;
    vertical-align: top;
}
```
It could look like this:
image

settings/settings-cpt.php
```diff
@@ -131,9 +138,16 @@ protected function form() {
 						printf(
 							'<li><label><input name="taxonomies[%s]" type="checkbox" value="1" %s %s/> %s</label></li>',
 							esc_attr( $taxonomy ),
-							checked( in_array( $taxonomy, $this->options['taxonomies'] ) || $disabled, true, false ),
+							checked( $disabled || in_array( $taxonomy, $this->options['taxonomies'] ), true, false ),
```
Screenfeed, line 141: We're missing the strict comparison here too:
```suggestion
							checked( $disabled || in_array( $taxonomy, $this->options['taxonomies'], true ), true, false ),
```

settings/settings-cpt.php
```diff
@@ -131,9 +138,16 @@ protected function form() {
...
 							disabled( $disabled, true, false ),
-							esc_html( $tax->labels->name )
+							esc_html(
+								sprintf(
+									/* translators: 1 is a post type or taxonomy label, 2 is a post type or taxonomy key. */
+									_x( '%1$s (%2$s)', 'taxonomy setting choice', 'polylang' ),
```
Screenfeed, line 146: @Chouby May think different upon his review, but I think both fields should use the same comment and context.

```suggestion
									_x( '%1$s (%2$s)', 'content type setting choice', 'polylang' ),
```

Screenfeed (changes requested):

Screenfeed:
Thanks @nicomollet.
`vertical-align` is indeed what I intended to use in another PR.
We're almost all good, just a few more tinkering.

Screenfeed (approved):
Thanks, perfect for me 👍

nicomollet:
Thanks for the help @Screenfeed ❤️ and keep up the good work

Chouby:
Thanks to you @nicomollet :)

PR #1124, opened 2022-09-27
Fix PHP warning when a filtered taxonomy has no query var

### LINKED ISSUE #1120: php Warning
array_flip(): Can only flip STRING and INTEGER values! in: https://github.com/polylang/polylang/blob/fefc34807c3187dd1749384c64b2eb4fd7641b3a/frontend/frontend-links.php#L154

The error appears on the author's posts page

Chouby:
Fixes #1120 

The warning occurs because `WP_Taxonomy::query_var` can be false. Thus `array_flip()` fires the warning is `false` is among the query vars returned by `PLL_Model::get_filtered_taxonomies_query_vars()`.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -406,13 +406,13 @@ class PLL_Model {
 	 *
 	 * @since 1.7
 	 *
-	 * @return array
+	 * @return string[]
 	 */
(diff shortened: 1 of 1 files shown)


manooweb (approved):
LGTM 👌

PR #1125, opened 2022-09-27
Fix term slug modification during update

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1468

This https://github.com/polylang/polylang/pull/1113 introduced a bug by changing the behavior. Indeed, before that we were using the updated sanitized term name as slug. See: https://github.com/polylang/polylang/blob/fefc34807c3187dd1749384c64b2eb4fd7641b3a/admin/admin-filters-term.php#L469

This PR propose to use the inserted/updated term name again and add a test to confirm the bug.


Screenfeed (changes requested):
I am not able to change the slug manually now.

Screenfeed (approved):

PR #1126, opened 2022-09-28
Don't report objects without languages when no language exist

### LINKED ISSUE #1121: SQL error on languages page
## Prerequisites

1. PHP 7.3, Mysql 5.7.39, Polylang 3.2.7
2. Access languages page on dashboard or Setup

When in the languages page or in the setup i can't add any language. Then i go back to the languages page and there is this error on top of page, it dissapears when i deactivate the polylang plugin:

[You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ') ) LIMIT 1' at line 4]
SELECT pdh_term_taxonomy.term_id FROM pdh_term_taxonomy WHERE taxonomy IN ('category','post_tag') AND pdh_term_taxonomy.term_id NOT IN ( SELECT object_id FROM pdh_term_relationships WHERE term_taxonomy_id IN () ) LIMIT 1

From Query Monitor plugin i got this traceback:

PLL_Model->get_terms_with_no_lang()
wp-content/plugins/polylang/include/model.php:704
PLL_Model->get_objects_with_no_lang()
wp-content/plugins/polylang/include/model.php:631
PLL_Settings->notice_objects_with_no_lang()
wp-content/plugins/polylang/settings/settings.php:347
do_action('admin_notices')
wp-includes/plugin.php:476

Chouby:
Fix #1121

When no languages exist, `get_objects_with_no_lang()` was not supposed to be called. This was protected [by the test of the existence of the default language](https://github.com/polylang/polylang/blob/3.2.7/settings/settings.php#L347). This protection proved not to be sufficient. This PR prevents an SQL error by checking that languages exist before injecting them in the SQL query. This check is also made for posts even if we don't use a direct SQL query in this case.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -679,6 +679,11 @@ class PLL_Model {
 	 * @return int[]
 	 */
 	public function get_posts_with_no_lang( $post_types, $limit ) {
+		$languages = $this->get_languages_list( array( 'fields' => 'term_id' ) );
+		if ( empty( $languages ) ) {
+			return array(); // Don't report if no languages have been defined yet.
+		}
+
 		return get_posts(
 			array(
 				'numberposts' => $limit,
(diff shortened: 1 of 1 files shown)


manooweb (approved):
LGTM 👌

PR #1127, opened 2022-09-28
Extract gutenberg-blocks tags from WPML config to be able to export and import strings to translate for blocks

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/1456

All is in the title.

The goal is to create rules for our xliff exporter/importer modules to be able to process strings inside blocks content or blocks attributes.

This PR deals only with simple cases documented in https://wpml.org/documentation/support/language-configuration-files/#gutenberg-blocks :
- `<xpath>` tag for blocks content
- simple `<key>` tag for blocks attributes

_Draft for the moment because we need to solve PHPStan issue with SimpleXMLElement_

## Going further

This PR doesn't deal with with more complex WPML configuration:
- attributes with sub-attributes
- attributes targeted by `search-method` attribute (`wildcards `or `regex`) or wildcard character `*`
- URL-Encoded JSON string attributes
- blocks targeted by namespace

### DIFF
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -329,26 +329,24 @@ class PLL_WPML_Config {
 			}
 			foreach ( $blocks as $block ) {
 				$attributes = $block->attributes();
-				if ( empty( $attributes ) ) {
+				if ( empty( $attributes ) || 1 !== (int) $attributes['translate'] ) {
 					continue;
 				}
-				if ( 1 === (int) $attributes['translate'] ) {
-					$block_name = (string) $attributes['type'];
-					foreach ( $block->children() as $child ) {
-						if ( $child_tag !== $child->getName() ) {
+				$block_name = (string) $attributes['type'];
+				foreach ( $block->children() as $child ) {
+					if ( $child_tag !== $child->getName() ) {
+						continue;
+					}
+					if ( $is_in_child_attribute ) {
+						$child_attributes = $child->attributes();
+						if ( empty( $child_attributes ) ) {
 							continue;
 						}
-						if ( $is_in_child_attribute ) {
-							$child_attributes = $child->attributes();
-							if ( empty( $child_attributes ) ) {
-								continue;
-							}
-							$rules = (string) $child_attributes[ $child_attribute_name ];
-						} else {
-							$rules = (string) $child;
-						}
-						$parsing_rules[ $block_name ][] = $rules;
+						$rules = (string) $child_attributes[ $child_attribute_name ];
+					} else {
+						$rules = (string) $child;
 					}
+					$parsing_rules[ $block_name ][] = $rules;
 				}
 			}
 		}
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -284,11 +284,11 @@ class PLL_WPML_Config {
 	 *
 	 * @since 3.3
 	 *
-	 * @param array[] $parsing_rules Rules as Xpath expressions to evaluate in the blocks content.
-	 * @return array[] Rules completed with ones from wpml-config file.
+	 * @param string[][] $parsing_rules Rules as Xpath expressions to evaluate in the blocks content.
+	 * @return string[][] Rules completed with ones from wpml-config file.
 	 *
 	 * @phpstan-param array<string,array<string>> $parsing_rules
-	 * @phpstan-return array<string,array<string>> $parsing_rules
+	 * @phpstan-return array<string,array<string>>
 	 */
 	public function translate_blocks( $parsing_rules ) {
 		return $this->extract_blocks_parsing_rules( $parsing_rules, 'xpath' );
@@ -299,8 +299,8 @@ class PLL_WPML_Config {
 	 *
 	 * @since 3.3
 	 *
-	 * @param array[] $parsing_rules Rules for block attributes to translate.
-	 * @return array[] Rules completed with ones from wpml-config file.
+	 * @param string[][] $parsing_rules Rules for block attributes to translate.
+	 * @return string[][] Rules completed with ones from wpml-config file.
 	 *
 	 * @phpstan-param array<string,array<string>> $parsing_rules
 	 * @phpstan-return array<string,array<string>>
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -353,23 +353,26 @@ class PLL_WPML_Config {
 				if ( empty( $attributes ) || 1 !== (int) $attributes['translate'] ) {
 					continue;
 				}
-				$block_name = (string) $attributes['type'];
+				$block_name = trim( (string) $attributes['type'] );
+				if ( '' === $block_name ) {
+					continue;
+				}
 				foreach ( $block->children() as $child ) {
 					$rule = '';
 					$child_tag = $child->getName();
 					switch ( $child_tag ) {
 						case 'xpath':
-							$rule = (string) $child;
+							$rule = trim( (string) $child );
 							break;
 						case 'key':
 							$child_attributes = $child->attributes();
 							if ( empty( $child_attributes ) ) {
 								break;
 							}
-							$rule = (string) $child_attributes['name'];
+							$rule = trim( (string) $child_attributes['name'] );
 							break;
 					}
-					if ( ! empty( $rule ) ) {
+					if ( '' !== $rule ) {
 						$parsing_rules[ $child_tag ][ $block_name ][] = $rule;
 					}
 				}
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -295,7 +295,7 @@ class PLL_WPML_Config {
 	}
 
 	/**
-	 * Translation management for blocks attributes.
+	 * Translation management for block attributes.
 	 *
 	 * @since 3.3
 	 *
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -302,11 +302,11 @@ class PLL_WPML_Config {
 	}
 
 	/**
-	 * Translation management for block attributes.
+	 * Translation management for blocks attributes.
 	 *
 	 * @since 3.3
 	 *
-	 * @param string[][] $parsing_rules Rules for block attributes to translate.
+	 * @param string[][] $parsing_rules Rules for blocks attributes to translate.
 	 * @return string[][] Rules completed with ones from wpml-config file.
 	 *
 	 * @phpstan-param array<string,array<string>> $parsing_rules
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -326,10 +326,10 @@ class PLL_WPML_Config {
 	 */
 	protected function get_blocks_parsing_rules( $rule_tag ) {
 
-		if ( null === $this->parsing_rules  ) {
+		if ( null === $this->parsing_rules ) {
 			$this->parsing_rules = $this->extract_blocks_parsing_rules();
 		}
-		
+
 		return isset( $this->parsing_rules[ $rule_tag ] ) ? $this->parsing_rules[ $rule_tag ] : array();
 	}
 
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -291,7 +291,7 @@ class PLL_WPML_Config {
 	 * @phpstan-return array<string,array<string>>
 	 */
 	public function translate_blocks( $parsing_rules ) {
-		return $this->extract_blocks_parsing_rules( $parsing_rules, 'xpath' );
+		return array_merge( $parsing_rules, $this->extract_blocks_parsing_rules( 'xpath' ) );
 	}
 
 	/**
@@ -306,7 +306,7 @@ class PLL_WPML_Config {
 	 * @phpstan-return array<string,array<string>>
 	 */
 	public function translate_blocks_attributes( $parsing_rules ) {
-		return $this->extract_blocks_parsing_rules( $parsing_rules, 'key', true );
+		return array_merge( $parsing_rules, $this->extract_blocks_parsing_rules( 'key', true ) );
 	}
 
 	/**
@@ -314,13 +314,14 @@ class PLL_WPML_Config {
 	 *
 	 * @since 3.3
 	 *
-	 * @param array[] $parsing_rules         Rules to complete with ones from wpml-config file..
 	 * @param string  $child_tag             Tag name to extract.
 	 * @param bool    $is_in_child_attribute Extract tag value in attribute or not. Default false.
 	 * @param string  $child_attribute_name  Attribute name where to extract the value. Default 'name'. Used if $is_in_child_attribute is set to true.
 	 * @return array[] Rules completed with ones from wpml-config file.
 	 */
-	protected function extract_blocks_parsing_rules( $parsing_rules, $child_tag, $is_in_child_attribute = false, $child_attribute_name = 'name' ) {
+	protected function extract_blocks_parsing_rules( $child_tag, $is_in_child_attribute = false, $child_attribute_name = 'name' ) {
+		$parsing_rules = array();
+
 		foreach ( $this->xmls as $xml ) {
 			$blocks = $xml->xpath( 'gutenberg-blocks/gutenberg-block' );
 			if ( ! is_array( $blocks ) ) {
@@ -346,9 +347,7 @@ class PLL_WPML_Config {
 						} else {
 							$rules = (string) $child;
 						}
-						if ( ! isset( $parsing_rules[ $block_name ] ) || ! in_array( $rules, $parsing_rules[ $block_name ] ) ) {
-							$parsing_rules[ $block_name ][] = $rules;
-						}
+						$parsing_rules[ $block_name ][] = $rules;
 					}
 				}
 			}
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -316,7 +316,7 @@ class PLL_WPML_Config {
 	 *
 	 * @param array[] $parsing_rules         Rules to complete with ones from wpml-config file..
 	 * @param string  $child_tag             Tag name to extract.
-	 * @param bool $is_in_child_attribute Extract tag value in attribute or not. Default false.
+	 * @param bool    $is_in_child_attribute Extract tag value in attribute or not. Default false.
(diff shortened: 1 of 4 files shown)


modules/wpml/wpml-config.php
```diff
@@ -225,7 +236,10 @@ public function translate_types( $types, $hide ) {
 			if ( is_array( $pts ) ) {
 				foreach ( $pts as $pt ) {
 					$attributes = $pt->attributes();
-					if ( 1 == $attributes['translate'] && ! $hide ) {
+					if ( empty( $attributes ) ) {
+						continue;
+					}
+					if ( '1' === (string) $attributes['translate'] && ! $hide ) {
```
Marianne380, line 242: Why do we cast it ? What type is this initially ?

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,81 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	public function translate_blocks( $parsing_rules ) {
+		return $this->extract_blocks_parsing_rules( $parsing_rules, 'xpath' );
+	}
+
+	/**
+	 * Translation management for blocks attributes.
+	 *
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules Rules for block attributes to translate.
```
Marianne380, line 302: We may need to choose between **blocks attributes** and  `block attributes` for consistency.

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,81 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	}
+
+	/**
+	 * Extract what kind of string to translate for blocks from WPML config file.
+	 *
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules         Rules to complete with ones from wpml-config file..
+	 * @param string  $child_tag             Tag name to extract.
+	 * @param boolean $is_in_child_attribute Extract tag value in attribute or not. Default false.
```
Marianne380, line 319:
```suggestion
	 * @param bool $is_in_child_attribute Extract tag value in attribute or not. Default false.
```

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,81 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	 * @param string  $child_attribute_name  Attribute name where to extract the value. Default 'name'. Used if $is_in_child_attribute is set to true.
```
Marianne380, line 320: Is this fourth parameter useful? We always use the default value here.

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,81 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	 * @return array[] Rules completed with ones from wpml-config file.
+	 */
+	protected function extract_blocks_parsing_rules( $parsing_rules, $child_tag, $is_in_child_attribute = false, $child_attribute_name = 'name' ) {
+		foreach ( $this->xmls as $xml ) {
+			$blocks = $xml->xpath( 'gutenberg-blocks/gutenberg-block' );
+			if ( is_array( $blocks ) ) {
```
Marianne380, line 326: To avoid too many indentation : 
```php
if ( ! is_array( $blocks ) {
    continue;
}
```

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,81 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+				foreach ( $blocks as $block ) {
+					$attributes = $block->attributes();
+					if ( empty( $attributes ) ) {
+						continue;
+					}
+					if ( '1' === (string) $attributes['translate'] ) {
+						$block_name = (string) $attributes['type'];
+						foreach ( $block->children() as $child ) {
+							if ( $child_tag === $child->getName() ) {
```
Marianne380, line 335: Same here for indentation.

Marianne380 (changes requested):

modules/wpml/wpml-config.php
manooweb: Done in 929a19365d65402d6b51bd655f0ba16f79aabfd0

modules/wpml/wpml-config.php
manooweb: `SimpleXMLElement` https://www.php.net/manual/fr/simplexmlelement.attributes.php 
So we need to cast for the [SimpleXMLElement::__toString()](https://www.php.net/manual/fr/simplexmlelement.tostring.php) would be aplied to read the attribute value.

For other PHPStan reason we have to pass the lazy '==' comparison to a strict comparison therefore casting this required.

BTW, I see other lazy comparison elsewhere in this class.

modules/wpml/wpml-config.php
manooweb: Done in 80d22f7eb519d52aeff34a180e0628ee6d0e3bd5

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+			$blocks = $xml->xpath( 'gutenberg-blocks/gutenberg-block' );
+			if ( ! is_array( $blocks ) ) {
+				continue;
+			}
+			foreach ( $blocks as $block ) {
+				$attributes = $block->attributes();
+				if ( empty( $attributes ) ) {
+					continue;
+				}
+				if ( '1' === (string) $attributes['translate'] ) {
```
Marianne380, line 334: Same here for intendation.

modules/wpml/wpml-config.php
Marianne380: If you see a potential need in the future, I guess we can leave it.

modules/wpml/wpml-config.php
Marianne380: This is not the same in the name of the method just below.

modules/wpml/wpml-config.php
Chouby: Does it work if we cas to `int`? It would look more logical to me.

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	/**
+	 * Translation management for strings in blocks content.
+	 *
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules Rules as Xpath expressions to evaluate in the blocks content.
+	 * @return array[] Rules completed with ones from wpml-config file.
+	 *
+	 * @phpstan-param array<string,array<string>> $parsing_rules
+	 * @phpstan-return array<string,array<string>> $parsing_rules
```
Chouby, line 291:
```suggestion
	 * @param string[][] $parsing_rules Rules as Xpath expressions to evaluate in the blocks content.
	 * @return string[][] Rules completed with ones from wpml-config file.
	 *
	 * @phpstan-param array<string,array<string>> $parsing_rules
	 * @phpstan-return array<string,array<string>>
```

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+		return $this->extract_blocks_parsing_rules( $parsing_rules, 'xpath' );
+	}
+
+	/**
+	 * Translation management for block attributes.
+	 *
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules Rules for block attributes to translate.
+	 * @return array[] Rules completed with ones from wpml-config file.
```
Chouby, line 303:
```suggestion
	 * @param string[][] $parsing_rules Rules for block attributes to translate.
	 * @return string[][] Rules completed with ones from wpml-config file.
```

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules Rules as Xpath expressions to evaluate in the blocks content.
+	 * @return array[] Rules completed with ones from wpml-config file.
+	 *
+	 * @phpstan-param array<string,array<string>> $parsing_rules
+	 * @phpstan-return array<string,array<string>> $parsing_rules
+	 */
+	public function translate_blocks( $parsing_rules ) {
+		return $this->extract_blocks_parsing_rules( $parsing_rules, 'xpath' );
```
Chouby, line 294: You are overwriting existing rules.

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules Rules for block attributes to translate.
+	 * @return array[] Rules completed with ones from wpml-config file.
+	 *
+	 * @phpstan-param array<string,array<string>> $parsing_rules
+	 * @phpstan-return array<string,array<string>>
+	 */
+	public function translate_blocks_attributes( $parsing_rules ) {
+		return $this->extract_blocks_parsing_rules( $parsing_rules, 'key', true );
```
Chouby, line 309: You are overwriting existing rules.

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	 *
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules         Rules to complete with ones from wpml-config file..
+	 * @param string  $child_tag             Tag name to extract.
+	 * @param bool    $is_in_child_attribute Extract tag value in attribute or not. Default false.
+	 * @param string  $child_attribute_name  Attribute name where to extract the value. Default 'name'. Used if $is_in_child_attribute is set to true.
+	 * @return array[] Rules completed with ones from wpml-config file.
+	 */
+	protected function extract_blocks_parsing_rules( $parsing_rules, $child_tag, $is_in_child_attribute = false, $child_attribute_name = 'name' ) {
```
Chouby, line 323: You are parsing the files with this function 2 times. I would parse only once and save the result for the second call.

modules/wpml/wpml-config.php
Chouby, line 334: Yo can avoid a pyramid step:
```php
if ( empty( $attributes ) || '1' !== (string) $attributes['translate'] ) {
	continue;
}
```

tests/phpunit/data/wpml-config.xml
```diff
@@ -45,12 +45,23 @@
...
+			<gutenberg-block type="core/image" translate="1">
+				<xpath>//figure/figcaption</xpath>
+				<xpath>//figure/img/@alt</xpath>
+			</gutenberg-block>
+		</gutenberg-blocks>
+		<gutenberg-blocks>
+			<gutenberg-block type="core/navigation-link" translate="1">
+				<key name="description" />
+			</gutenberg-block>
+		</gutenberg-blocks>
```
Chouby, line 64: Don't use core blocks.

tests/phpunit/tests/test-wpml-config.php
```diff
@@ -336,4 +336,47 @@ public function test_register_string_object() {
...
+		$expected_parsing_rules_for_attributes = array(
+			'core/navigation-link'      => array(
+				'description',
+			),
+		);
+
+		$parsing_rules                = apply_filters( 'pll_blocks_xpath_rules', array() );
+		$parsing_rules_for_attributes = apply_filters( 'pll_blocks_rules_for_attributes', array() );
+
+		$this->assertSameSets( $expected_parsing_rules, $parsing_rules, 'Rules should not be added to the existings ones' );
```
Chouby, line 358: >'Rules should not be added to the existings ones'

I don't understand.

modules/wpml/wpml-config.php
```diff
@@ -262,6 +279,83 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	public function translate_blocks_attributes( $parsing_rules ) {
+		return $this->extract_blocks_parsing_rules( $parsing_rules, 'key', true );
+	}
+
+	/**
+	 * Extract what kind of string to translate for blocks from WPML config file.
+	 *
+	 * @since 3.3
+	 *
+	 * @param array[] $parsing_rules         Rules to complete with ones from wpml-config file..
```
Chouby, line 317: You can be more specific with `string[][]` according to your PHPStan information.

Chouby (changes requested):

modules/wpml/wpml-config.php
manooweb: Done in e525cf49b3de1c932b6c11ab3e943ca929402f6c

modules/wpml/wpml-config.php
manooweb: I think it's always possible.

modules/wpml/wpml-config.php
manooweb: Done in 9b925b2c7475202314e5f60db3455800bcd521ae

modules/wpml/wpml-config.php
manooweb: Done in 9b925b2c7475202314e5f60db3455800bcd521ae

modules/wpml/wpml-config.php
manooweb: See https://github.com/polylang/polylang/pull/1127/files#r984341958
Done in 8aafa6ebf94ae4267a3f42c788b427bf83db6548

tests/phpunit/data/wpml-config.xml
manooweb: Done in 626270c464222444ed65a79ea783085253d34a17

tests/phpunit/tests/test-wpml-config.php
manooweb: Done in f6e1b5ab78efce4054fe26764e3f9c8e891b99c4

modules/wpml/wpml-config.php
manooweb: See https://github.com/polylang/polylang/pull/1127/files#r985464707

modules/wpml/wpml-config.php
manooweb: Done in 5c8c4da5fd920a3c525e8bbaa1d90f4327c95f17

modules/wpml/wpml-config.php
manooweb: Done in 9432b96776eb587bd337eb840adb5a36fdfb1326

modules/wpml/wpml-config.php
```diff
@@ -262,6 +286,98 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+							break;
+						case 'key':
+							$child_attributes = $child->attributes();
+							if ( empty( $child_attributes ) ) {
+								break;
+							}
+							$rule = (string) $child_attributes['name'];
+							break;
+					}
+					if ( ! empty( $rule ) ) {
```
Screenfeed, line 372: I think this can be reinforced a bit with `trim()`s (L356-359, L365, L372, L375):
```suggestion
				$block_name = trim( (string) $attributes['type'] );
				if ( '' === $block_name ) {
					continue;
				}
				foreach ( $block->children() as $child ) {
					$rule = '';
					$child_tag = $child->getName();
					switch ( $child_tag ) {
						case 'xpath':
							$rule = trim( (string) $child );
							break;
						case 'key':
							$child_attributes = $child->attributes();
							if ( empty( $child_attributes ) ) {
								break;
							}
							$rule = trim( (string) $child_attributes['name'] );
							break;
					}
					if ( '' !== $rule ) {
```
This should allow what I suggested in the tests to pass.

Screenfeed (approved):

Screenfeed (approved):

manooweb:
> By the way, I'm introducing a method [in my PR](https://github.com/polylang/polylang/pull/1133/files#diff-bd0009919dc4f061b1965ffc902fdbe9bd35285e0b5e99f825894f125ff2548fR458-R475) that could be used here.

I note and wait for the #1133 to be merged before using it

modules/wpml/wpml-config.php
```diff
@@ -262,6 +286,101 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+		return array_merge( $parsing_rules, $this->get_blocks_parsing_rules( 'key' ) );
+	}
+
+	/**
+	 * Returns rules to extract a kind of string for blocks.
+	 *
+	 * @since 3.3
+	 *
+	 * @param string $rule_tag Tag name to extract.
+	 * @return string[][] Rules completed with ones from wpml-config file.
```
Chouby, line 325:
```suggestion
	 * Returns rules to extract translatable strings from blocks.
	 *
	 * @since 3.3
	 *
	 * @param string $rule_tag Tag name to extract.
	 * @return string[][] The rules.
```

modules/wpml/wpml-config.php
```diff
@@ -262,6 +286,101 @@ public function translate_taxonomies( $taxonomies, $hide ) {
...
+	 */
+	protected function get_blocks_parsing_rules( $rule_tag ) {
+
+		if ( isset( $this->parsing_rules[ $rule_tag ] ) ) {
+			return $this->parsing_rules[ $rule_tag ];
+		}
+		$this->parsing_rules = $this->extract_blocks_parsing_rules();
+		return $this->parsing_rules[ $rule_tag ];
```
Chouby, line 333: I'm not sure `$this->parsing_rules[ $rule_tag ]` is always set when you return, so suggesting this:
```suggestion

		if ( null === $this->parsing_rules  ) {
			$this->parsing_rules = $this->extract_blocks_parsing_rules();
		}
		
		return isset( $this->parsing_rules[ $rule_tag ] ) ? $this->parsing_rules[ $rule_tag ] : array();
```

modules/wpml/wpml-config.php
```diff
@@ -32,6 +32,13 @@ class PLL_WPML_Config {
 	 */
 	protected $files;
 
+	/**
+	 * List of rules to extract strings to translate from blocks.
+	 *
+	 * @var string[][][]
+	 */
+	protected $parsing_rules;
```
Chouby, line 40:
```suggestion
	 * @var string[][][]|null
	 */
	protected $parsing_rules = null;
```

Chouby (changes requested):

modules/wpml/wpml-config.php
manooweb: `isset()` seems to return `false` when the variable has a `null` value
https://www.php.net/manual/en/function.isset.php

PR #1133, opened 2022-10-04
Support custom fields in wpml-config.xml

### LINKED ISSUE #210: Doesn't save language cookie with WP Fastest Cache
## Prerequisites

- [ ] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).

1. Install cache plugin "WP Fastest Cache"
2. Go directly site e.g. through new private tab in Chrome to see cached version
3. There is no new cookie saved with language code. Because of this issue there is a problem with e.g. data getting by wp rest api.

I don't know when but previously this bug wasn't exists. I check version from master with fix for such issues but this doesn't work for me.

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/210.
Related #535 (should be closed).

This PR allows to translate post and term metas listed in `wpml-config.xml` files.
This adds hooks to `'pll_post_metas_to_export'` and `'pll_term_metas_to_export'`. Post meta sub-keys are supported but not term meta sub-keys: WPML doesn't currently support this.
Wilcard (`*`) is not currently supported.
The `encoding` attribute is not supported.

Unit tests have been added but we should add integration tests to the xliff plugin.

### DIFF
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -95,8 +95,7 @@ class PLL_WPML_Config {
 				$keys = $xml->xpath( 'admin-texts/key' );
 				if ( is_array( $keys ) ) {
 					foreach ( $keys as $key ) {
-						$attributes = $key->attributes();
-						$name = (string) $attributes['name'];
+						$name = $this->get_field_attribute( $key, 'name' );
 
 						if ( false !== strpos( $name, '*' ) ) {
 							$pattern = '#^' . str_replace( '*', '(?:.+)', $name ) . '$#';
@@ -174,17 +173,22 @@ class PLL_WPML_Config {
 	public function copy_post_metas( $metas, $sync ) {
 		foreach ( $this->xmls as $xml ) {
 			$cfs = $xml->xpath( 'custom-fields/custom-field' );
-			if ( is_array( $cfs ) ) {
-				foreach ( $cfs as $cf ) {
-					$attributes = $cf->attributes();
-					if ( 'copy' == $attributes['action'] || ( ! $sync && in_array( $attributes['action'], array( 'translate', 'copy-once' ) ) ) ) {
-						$metas[] = (string) $cf;
-					} else {
-						$metas = array_diff( $metas, array( (string) $cf ) );
-					}
+
+			if ( ! is_array( $cfs ) ) {
+				continue;
+			}
+
+			foreach ( $cfs as $cf ) {
+				$action = $this->get_field_attribute( $cf, 'action' );
+
+				if ( 'copy' === $action || ( ! $sync && in_array( $action, array( 'translate', 'copy-once' ), true ) ) ) {
+					$metas[] = (string) $cf;
+				} else {
+					$metas = array_diff( $metas, array( (string) $cf ) );
 				}
 			}
 		}
+
 		return $metas;
 	}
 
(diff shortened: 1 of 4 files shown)


modules/wpml/wpml-config.php
```diff
@@ -281,23 +403,60 @@ protected function register_or_translate_option( $context, $name, $key ) {
...
+	/**
+	 * Get the value of an attribute.
+	 *
+	 * @since 3.3
+	 *
+	 * @param  SimpleXMLElement $field          A XML node.
+	 * @param  string           $attribute_name Node of the attribute.
+	 * @return string
+	 */
+	private function get_field_attribute( SimpleXMLElement $field, $attribute_name ) {
```
Hug0-Drelon, line 453: Could this method be used in `copy_post_metas()` and `copy_term_metas()`?

modules/wpml/wpml-config.php
```diff
@@ -281,23 +403,60 @@ protected function register_or_translate_option( $context, $name, $key ) {
...
 	 *
-	 * @param SimpleXMLElement $key XML node.
-	 * @param array            $arr Array of option keys to translate.
+	 * @param SimpleXMLElement $key        XML node.
+	 * @param array            $arr        Array of option keys to translate.
+	 * @param mixed            $fill_value Value to use when filling entries. Default is true.
 	 * @return array
 	 */
-	protected function xml_to_array( $key, &$arr = array() ) {
+	protected function xml_to_array( SimpleXMLElement $key, array &$arr = array(), $fill_value = true ) {
```
Hug0-Drelon, line 414: `$arr` is passed by reference even if `xml_to_array()` returns the modified array. Is it on purpose?
Wouldn't it be better to avoid passing by ref?

modules/wpml/wpml-config.php
```diff
@@ -281,23 +403,60 @@ protected function register_or_translate_option( $context, $name, $key ) {
...
+			return $arr;
+		}
+
 		$children = $key->children();
 
 		if ( count( $children ) ) {
 			foreach ( $children as $child ) {
-				$arr[ $name ] = $this->xml_to_array( $child, $arr[ $name ] );
+				if ( ! isset( $arr[ $name ] ) || ! is_array( $arr[ $name ] ) ) {
+					$arr[ $name ] = array();
```
Hug0-Drelon, line 432: I don't understand why you didn't `continue` and add an empty array instead? 😅

modules/wpml/wpml-config.php
```diff
@@ -210,6 +214,124 @@ public function copy_term_metas( $metas, $sync ) {
...
+	 *        'sub_key_to_translate_1' => 1,
+	 *        'sub_key_to_translate_2' => array(
+	 *             'sub_sub_key_to_translate_1' => 1,
+	 *         ),
+	 *      ),
+	 *    )
+	 * }
+	 * @return array
+	 *
+	 * @phpstan-param array<string, int|array> $keys
```
Hug0-Drelon, line 238:
```suggestion
	 * @phpstan-param array<string, int|string|array> $keys
```
I think @Chouby told me array value doesn't matter (except for sub key, which imply an array), and thus users can fill it with whatever the want.

Hug0-Drelon (approved):

modules/wpml/wpml-config.php
```diff
@@ -211,23 +224,143 @@ public function copy_post_metas( $metas, $sync ) {
...
+				$keys[ (string) $field ] = 1;
+			}
+		}
+
+		// Deal with sub-field translations.
+		foreach ( $this->xmls as $xml ) {
+			$fields = $xml->xpath( 'custom-fields-texts/key' );
+
+			if ( ! is_array( $fields ) ) {
+				// Wrong configuration: no `<key>` nodes.
```
Chouby, line 296: For me, there's nothing wrong, as we don't always need 'custom-fields-texts' (when meta are single strings). 
```suggestion
				// No 'custom-fields-texts` nodes.
```

PR #1134, opened 2022-10-04
Replace first parameter of WordPress get_terms() function passed as a string with an array

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/1474

First step to fix the issue by replacing the first parameter of get_terms() WordPress function with an array instead of a string.

Accordingly : 
- Clean up the baseline 
- Update some existing ignored PHPStan errors because the count decreases.
- Update some existing ignored PHPStan errors because only the error messages are different due to this fix.

This error is ignored in config file:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  Line   frontend\frontend-static-pages.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------    
  233    Parameter #1 $args of function get_posts expects array{numberposts?: int, category?: int|string, include?: array<int>, exclude?: array<int>, suppress_filters?: bool, attachment_id?: int, author?: int|string, author_name?: string, ...}|null, array{posts_per_page: 99, post_type: 'page', post__in: array<int, int|null>, lang: ''} given.     
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------    

```
- due to the Polylang `lang` parameter unknown in the WordPress stubs documentation of `get_posts()`
- and also probably due to `post__in` parameter expected as `int[]` but not always returned as this by `PLL_Model::get_languages_list()` method.

This PR also removes deprecated usage of `get_terms()` in PHPUnit tests.

### DIFF
--- a/frontend/frontend-auto-translate.php
+++ b/frontend/frontend-auto-translate.php
@@ -277,8 +277,11 @@ class PLL_Frontend_Auto_Translate {
 		} else {
 			$terms = get_terms( array( 'taxonomy' => $taxonomy, $field => $term, 'lang' => '' ) );
 
-			if ( ! empty( $terms ) && ! is_wp_error( $terms ) ) {
+			if ( ! empty( $terms ) && is_array( $terms ) ) {
 				$t = reset( $terms );
+				if ( ! $t instanceof WP_Term ) {
+					return $term;
+				}
 				$tr_id = $this->get_term( $t->term_id );
 
 				if ( ! is_wp_error( $tr = get_term( $tr_id, $taxonomy ) ) ) {
--- a/frontend/frontend-auto-translate.php
+++ b/frontend/frontend-auto-translate.php
@@ -275,7 +275,7 @@ class PLL_Frontend_Auto_Translate {
 				return $tr_id;
 			}
 		} else {
-			$terms = get_terms( $taxonomy, array( $field => $term, 'lang' => '' ) );
+			$terms = get_terms( array( 'taxonomy' => $taxonomy ), array( $field => $term, 'lang' => '' ) );
 
 			if ( ! empty( $terms ) && ! is_wp_error( $terms ) ) {
 				$t = reset( $terms );
--- a/frontend/frontend-auto-translate.php
+++ b/frontend/frontend-auto-translate.php
@@ -275,7 +275,7 @@ class PLL_Frontend_Auto_Translate {
 				return $tr_id;
 			}
 		} else {
-			$terms = get_terms( array( 'taxonomy' => $taxonomy ), array( $field => $term, 'lang' => '' ) );
+			$terms = get_terms( array( 'taxonomy' => $taxonomy, $field => $term, 'lang' => '' ) );
 
 			if ( ! empty( $terms ) && ! is_wp_error( $terms ) ) {
 				$t = reset( $terms );
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -38,8 +38,14 @@ parameters:
 			count: 1
 			path: frontend/choose-lang.php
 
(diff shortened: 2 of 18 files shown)


frontend/frontend-auto-translate.php
```diff
@@ -275,7 +275,7 @@ protected function get_translated_term_by( $field, $term, $taxonomy ) {
 				return $tr_id;
 			}
 		} else {
-			$terms = get_terms( $taxonomy, array( $field => $term, 'lang' => '' ) );
+			$terms = get_terms( array( 'taxonomy' => $taxonomy, $field => $term, 'lang' => '' ) );
 
 			if ( ! empty( $terms ) && ! is_wp_error( $terms ) ) {
 				$t = reset( $terms );
```
Screenfeed, line 281:
```
 ------ -------------------------------------------------------------
  Line   frontend/frontend-auto-translate.php
 ------ -------------------------------------------------------------
  281    Parameter #1 $array of function reset expects array|object,
         array<int, int|string|WP_Term>|string given.
 ------ -------------------------------------------------------------
```

I think this can be fixed by replacing `&& ! is_wp_error( $terms )` by `&& is_array( $terms )` at L280:
```suggestion
			$terms = get_terms( array( 'taxonomy' => $taxonomy, $field => $term, 'lang' => '' ) );

			if ( ! empty( $terms ) && is_array( $terms ) ) {
				$t = reset( $terms );
```

Screenfeed (changes requested):

frontend/frontend-auto-translate.php
manooweb: I'll look at this. I thought about something like that. Thank you for your confirmation / suggestion.

manooweb: It seems not to be sufficient. I complete with another condition on the `reset()` result to be sure we can do `$t->term_id`

Screenfeed (approved):

frontend/frontend-auto-translate.php
Screenfeed: Ok, because `get_terms()` can return `string[]`.

phpstan.neon.dist
```diff
@@ -37,3 +37,15 @@ parameters:
...
+		-
+			message: "#^Parameter \\#1 \\$args of function get_posts expects array(.+)\\|null, array\\{posts_per_page: 99, post_type: 'page', post__in: array\\<int, int\\|null\\>, lang: ''\\} given\\.$#"
+			count: 1
+			path: frontend/frontend-static-pages.php
+
+		# Ignored because the WordPress stubs doesn't know a dynamic key in the associative array passed to the get_terms() parameter
+		-
+			message: "#^Parameter \\#1 \\$args of function get_terms expects array(.+), non\\-empty\\-array\\<string, int\\|string\\> given\\.$#"
+			count: 1
+			path: frontend/frontend-auto-translate.php
```
Chouby, line 51: Can we use generic ignored errors in case we introduced new such cases in the future?

Chouby (changes requested):

phpstan.neon.dist
manooweb: Done in 8b2c369db85ce9a55537b554d5236101e23fc241

Screenfeed (approved):

Chouby:
It's not clear in the commit description, however 7c2412f65eaf323cedcdb400f44b32fe8d3202bf removes dead code.

PR #1140, opened 2022-10-13
Add support for wpml-config.xml to MU plugins

### LINKED ISSUE #1088: [FIX] `wpml-config.xml` doesn't work for `mu-plugins`
## Prerequisites

1. Use an `mu-plugin` and have a `wpml-config.xml` inside it
2. `wpml-config.xml` does nothing

I expected an `mu-plugin` `wpml-config.xml` to be used but it's not.

Screenfeed:
Closes #1088.

This PR brings support for `wpml-config.xml` to Must-Use plugins.
The whole method `get_files()` has been reworked to improve code readability: it has been split into 4 other private methods that handle different `wpml-config.xml` files (1 for MU plugins, 1 for plugins, 1 for themes, 1 for custom PLL file).

### DIFF
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -629,16 +629,22 @@ class PLL_WPML_Config {
 	 */
 	private function get_plugin_files() {
 		$files   = array();
-		$plugins = get_option( 'active_plugins', array() );
-		$plugins = is_array( $plugins ) ? $plugins : array();
+		$plugins = array();
 
 		if ( is_multisite() ) {
 			// Don't forget sitewide active plugins thanks to Reactorshop http://wordpress.org/support/topic/polylang-and-yoast-seo-plugin/page/2?replies=38#post-4801829.
 			$sitewide_plugins = get_site_option( 'active_sitewide_plugins', array() );
 
 			if ( ! empty( $sitewide_plugins ) && is_array( $sitewide_plugins ) ) {
-				$plugins = array_merge( $plugins, array_keys( $sitewide_plugins ) );
+				$plugins = array_keys( $sitewide_plugins );
 			}
+		}
+
+		// By-site plugins.
+		$active_plugins = get_option( 'active_plugins', array() );
+
+		if ( ! empty( $active_plugins ) && is_array( $active_plugins ) ) {
+			$plugins = array_merge( $plugins, $active_plugins );
 		}
 
 		$plugin_path = trailingslashit( WP_PLUGIN_DIR ) . '%s/wpml-config.xml';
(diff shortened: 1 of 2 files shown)


modules/wpml/wpml-config.php
```diff
@@ -594,4 +578,140 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
...
+
+		// Search for top level wpml-config.xml file.
+		$file_path = WPMU_PLUGIN_DIR . '/wpml-config.xml';
+
+		if ( file_exists( $file_path ) ) {
+			$files['mu-plugins'] = $file_path;
+		}
+
+		// Search in proxy loaded MU plugins.
+		try {
```
Hug0-Drelon, line 606: `try` and `catch` are here because of `DirectoryIterator::__construc()`?
If so it shouldn't raise any exception or error since `WPMU_PLUGIN_DIR` is checked earlier, should it?

Screenfeed: You're right, I added this before the `is_dir()` check, it has no value now.
Removed in [47eca19](https://github.com/polylang/polylang/pull/1140/commits/47eca19d2fb23ce3fb795784611725dcb1199ad1).

modules/wpml/wpml-config.php
```diff
@@ -594,4 +578,136 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
...
+
+		if ( file_exists( $file_path ) ) {
+			$files[ 'themes/' . get_template() ] = $file_path;
+		}
+
+		// Child theme.
+		$stylesheet_path = get_stylesheet_directory();
+		$file_path       = "{$stylesheet_path}/wpml-config.xml";
+
+		if ( $stylesheet_path !== $template_path && file_exists( $file_path ) ) {
```
Hug0-Drelon, line 686: This is probably not the topic of this PR, but I wonder what happens if `wpm-config.php` files from parent and child themes have conflicting data. Is there a hierarchy (like in WordPress) in the way `wpm-config.php` files are loaded?

Hug0-Drelon (approved):

modules/wpml/wpml-config.php
Screenfeed: This is done in [28f7c2c](https://github.com/polylang/polylang/pull/1140/commits/28f7c2ccf1c4f16f829c0540a2e77c3b193708a4) and [537bc71](https://github.com/polylang/polylang/pull/1140/commits/537bc715e4f04231b388689b09387daac9bf5022).

Screenfeed:
The final order is:
- Network-activated plugins,
- Site-activated plugins,
- Parent theme,
- Child theme,
- Must-Use plugins,
- Polylang.

Chouby (approved):

PR #1141, opened 2022-10-17
Use WP format for arrays doc

Chouby:
This PR replaces usage of `array<..>` in doc blocks.

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -171,8 +171,8 @@ abstract class PLL_Translated_Object {
 	 *
 	 * @since 3.2
 	 *
-	 * @param  int $term_id Term ID.
-	 * @return array<int>   An associative array of translations with language code as key and translation id as value.
+	 * @param int $term_id Term ID.
+	 * @return int[] An associative array of translations with language code as key and translation id as value.
 	 */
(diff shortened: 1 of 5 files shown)


include/translated-object.php
```diff
@@ -496,8 +496,9 @@ public function current_user_can_synchronize( $id ) {
 	 *
 	 * @since 3.2
 	 *
-	 * @param  mixed $id  A supposedly numeric ID.
-	 * @return int<0,max> A positive integer. `0` for non numeric values and negative integers.
+	 * @param mixed $id A supposedly numeric ID.
+	 * @return int A positive integer. `0` for non numeric values and negative integers.
+	 * @phpstan-return int<0,max>
```
Screenfeed, line 501: Like in https://github.com/polylang/polylang-pro/pull/1498#discussion_r997893752, we tend to add a blank line before the `@phpstan-*` tags apparently.
```suggestion
	 * @return int A positive integer. `0` for non numeric values and negative integers.
	 *
	 * @phpstan-return int<0,max>
```

Screenfeed (changes requested):

Screenfeed (approved):

PR #1144, opened 2022-10-20
Add a test for untranslated taxonomy slug

Hug0-Drelon:
Add a test following https://github.com/polylang/polylang/pull/1142, it does not show any bug but prevent modifying untranslated term slugs.

### DIFF
--- a/tests/phpunit/tests/test-slugs.php
+++ b/tests/phpunit/tests/test-slugs.php
@@ -121,4 +121,53 @@ class Slugs_Test extends PLL_UnitTestCase {
 		$this->assertSame( 'New Test', $en_new->name );
 		$this->assertSame( 'test-en', $en_new->slug );
 	}
+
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/test-slugs.php
```diff
@@ -121,4 +121,53 @@ public function test_update_existing_term_slugs_with_parent() {
...
 	}
+
+	public function test_untranslatable_taxonomy() {
+		register_taxonomy(
+			'test-tax',
+			'post',
+			array(
+				'public' => false, // Not translatable by default.
+			)
+		);
```
Chouby, line 132: In fact all taxonomies and post types are not translatable by default.
```suggestion
		register_taxonomy( 'test-tax', 'post' ); // Not translatable by default.
```

Chouby (changes requested):

PR #1148, opened 2022-11-03
Fix PHPStan errors.

Hug0-Drelon:
Since PHPStan 1.9.0, the following errors are raised:
```
------ ------------------------------------------------------------------------------------------ 
  Line   frontend/frontend-filters-search.php                                                      
 ------ ------------------------------------------------------------------------------------------ 
  :72    Parameter #3 $subject of function preg_replace expects array|string, string|false given.  
  :73    Parameter #1 $search of function str_replace expects array|string, string|false given.    
  :73    Parameter #2 $replace of function str_replace expects array|string, string|null given.    
 ------ ------------------------------------------------------------------------------------------ 

 ------ ----------------------------------------------------------------------------------------- 
  Line   include/model.php                                                                        
 ------ ----------------------------------------------------------------------------------------- 
  :553   Binary operation "." between '' and array<int, string>|string|null results in an error.  
 ------ -----------------------------------------------------------------------------------------
 ```

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -550,7 +550,10 @@ class PLL_Model {
 			$groupby = ' GROUP BY pll_tr.term_taxonomy_id';
 
 			if ( ! empty( $q['m'] ) ) {
-				$q['m'] = '' . preg_replace( '|[^0-9]|', '', $q['m'] );
+				$date_query = preg_replace( '|[^0-9]|', '', $q['m'] );
+				if ( is_string( $date_query ) ) {
+					$q['m'] = '' . $date_query;
+				}
 				$where .= $wpdb->prepare( " AND YEAR( {$wpdb->posts}.post_date ) = %d", substr( $q['m'], 0, 4 ) );
 				if ( strlen( $q['m'] ) > 5 ) {
 					$where .= $wpdb->prepare( " AND MONTH( {$wpdb->posts}.post_date ) = %d", substr( $q['m'], 4, 2 ) );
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: include/model.php [hunk 1]
```diff
@@ -550,7 +550,10 @@ public function count_posts( $lang, $q = array() ) {
 			$groupby = ' GROUP BY pll_tr.term_taxonomy_id';
 
 			if ( ! empty( $q['m'] ) ) {
```
Screenfeed, line 552: While at it, I think we should make sure that `$q['m']` is a string.
```suggestion
			if ( ! empty( $q['m'] ) && is_string( $q['m'] ) ) {
```

Screenfeed:
Oh no.

Screenfeed (approved):

include/model.php
```diff
@@ -550,7 +550,10 @@ public function count_posts( $lang, $q = array() ) {
...
-				$q['m'] = '' . preg_replace( '|[^0-9]|', '', $q['m'] );
+				$date_query = preg_replace( '|[^0-9]|', '', $q['m'] );
+				if ( is_string( $date_query ) ) {
+					$q['m'] = $date_query;
+				}
```
Chouby, line 556: I don't reproduce any error here on my local install. Do you?

Hug0-Drelon: After updating my PHPStan dependency locally, I didn't reproduce the error...

Hug0-Drelon: Revert in bf268a9

PR #1149, opened 2022-11-04
Refactor PLL_Translated_Object

Screenfeed:
`PLL_Translated_Object` is now split into two classes.

- `PLL_Object_With_Language` deals with object types that support at least one language. It implements the new interface `PLL_Object_With_Language_Interface`.
- `PLL_Translated_Object` deals with translation groups and extends `PLL_Object_With_Language`. It implements the new interface `PLL_Translated_Object_Interface`.
- `PLL_Translated_Post` and `PLL_Translated_Term` still extend `PLL_Translated_Object`.

A few important details:
- New methods `PLL_Object_With_Language::get_tax_language()` and `PLL_Translated_Object::get_tax_translations()`. This is something that was missing imho.
- The 2nd parameter of `PLL_Object_With_Language::update_language()` is now type-hinted. This triggered an error in a test.
- `PLL_Translated_Post::get_language()` and `PLL_Translated_Term::get_language()` are now identical, to respect the interface `PLL_Object_With_Language_Interface`. Which means `PLL_Translated_Term::get_language()` cannot be used with term slug + taxonomy anymore: it still works but it is deprecated. `PLL_Translated_Term::get_language_by_term_slug()` can be used instead (it is used only once, in `PLL_Choose_Lang_Content`).
- Hooks are now launched in a separate method `init()`. For backward compatibility, `init()` is called in `PLL_Translated_Post`'s and `PLL_Translated_Term`'s constructor.
- Parameters for post/term ID are renamed into `$id` everywhere, even in child classes, for consistency and compatibility with named parameters in php 8.
- An array is passed as 2nd argument of `register_taxonomy()`.
- `pll_get_post()` and `pll_get_term()` now return `0` instead of `false`.
- Several improvements in DocBlocks for PHPStan.

A last word: maybe what is done about cache in `PLL_Translated_Term` should be done for all object types that are not `post` 🤔.


include/object-with-language-interface.php
```diff
@@ -0,0 +1,98 @@
...
+	 * @phpstan-return non-empty-string
+	 */
+	public function get_tax_language();
+
+	/**
+	 * Adds hooks.
+	 *
+	 * @since 3.3
+	 *
+	 * @return void
```
Hug0-Drelon, line 29: Wouldn't it be better if `init()` return `self`?

include/object-with-language-interface.php
```diff
@@ -0,0 +1,98 @@
...
+	public function init();
+
+	/**
+	 * Stores the object's language in the database.
+	 *
+	 * @since 3.3
+	 *
+	 * @param int                     $id   Object ID.
+	 * @param int|string|PLL_Language $lang Language (term_id, slug, or object).
+	 * @return void
```
Hug0-Drelon, line 40: Maybe we could return a boolean to validate language assignation?

include/object-with-language-interface.php
```diff
@@ -0,0 +1,98 @@
...
+	public function get_language( $id );
+
+	/**
+	 * Assigns a new language to an object.
+	 *
+	 * @since 3.3
+	 *
+	 * @param int          $id   Object ID.
+	 * @param PLL_Language $lang New language to assign to the object.
+	 * @return void
```
Hug0-Drelon, line 61: Same as above.

include/translated-object-interface.php
```diff
@@ -0,0 +1,116 @@
...
+	 * @phpstan-return non-empty-string
+	 */
+	public function get_tax_translations();
+
+	/**
+	 * Adds hooks.
+	 *
+	 * @since 3.3
+	 *
+	 * @return void
```
Hug0-Drelon, line 29: Same question as above.

include/translated-object-interface.php
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * Among the object and its translations, returns the ID of the object which is in `$lang`.
+	 *
+	 * @since 3.3
+	 *
+	 * @param int                     $id   Object ID.
+	 * @param int|string|PLL_Language $lang Language (term_id or slug or object).
+	 * @return int|false The translation object ID if exists, otherwise the passed ID, false if the passed object has no language.
+	 *
+	 * @phpstan-return positive-int|false
```
Hug0-Drelon, line 103: Would it be a breaking change if it returns `0` instead of false?
I saw you cast the result of this method into int in `static-pages.php` for instance.

include/object-with-language-interface.php
Screenfeed: Not useful currently but why not, for consistency. Done in [9600eb0](https://github.com/polylang/polylang/pull/1149/commits/9600eb061ecbd769b1d1e4292408f3b74b747fb9).

include/object-with-language-interface.php
Screenfeed: Done in [6d0a7b4](https://github.com/polylang/polylang/pull/1149/commits/6d0a7b4c742ce9d952a103ea07c747c31984a413).

include/object-with-language-interface.php
Screenfeed: [6d0a7b4](https://github.com/polylang/polylang/pull/1149/commits/6d0a7b4c742ce9d952a103ea07c747c31984a413)

include/translated-object-interface.php
Screenfeed: [9600eb0](https://github.com/polylang/polylang/pull/1149/commits/9600eb061ecbd769b1d1e4292408f3b74b747fb9)

include/translated-object-interface.php
Screenfeed: I wondered the same. It does change the API though (which could be reverted).
Done in [26713d5](https://github.com/polylang/polylang/pull/1149/commits/26713d5cfcca5eb2efac1d6137c19dcfaa1b02a3).

Hug0-Drelon (approved):

Screenfeed:
Considering removing the `$id` parameter from `PLL_Translated_Object::save_translations()`, this is what I was thinking about:
```php
	/**
	 * Saves the object's translations.
	 *
	 * @since 0.5
	 * @since 3.3 Removed the parameter $id.
	 *
	 * @param int[] $translations An associative array of translations with language code as key and translation ID as value.
	 * @return int[] An associative array with language codes as key and object IDs as values.
	 *
	 * @phpstan-return array<non-empty-string, positive-int>
	 */
	public function save_translations( $translations ) {
		if ( func_num_args() > 1 ) {
			// Backward compatibility.
			_deprecated_argument(
				__METHOD__ . '()',
				'3.3',
				esc_html(
					sprintf(
						/* translators: %s is a function name. */
						__( 'Please use %s instead.', 'polylang' ),
						get_class( $this ) . '::save_translations( $translations )'
					)
				)
			);
			$id           = $translations;
			$translations = func_get_arg( 1 );
		}

		$id = $this->sanitize_int_id( $id );

		// ...
	}
```

include/translated-term.php
```diff
@@ -213,4 +300,26 @@ public function _prime_terms_cache( $terms, $taxonomies ) {
 	public function clean_term_cache( $ids ) {
 		clean_object_term_cache( $this->sanitize_int_ids_list( $ids ), 'term' );
 	}
+
+	/**
+	 * Tells whether a translation term must updated.
```
Chouby, line 305:
```suggestion
	 * Tells whether a translation term must be updated..
```

modules/sync/sync-tax.php
```diff
@@ -105,7 +105,7 @@ protected function maybe_translate_terms( $object_id, $terms, $taxonomy, $lang )
 				 * @param int    $term    Source term id
 				 * @param string $lang    Language slug
 				 */
-				if ( $term_id = apply_filters( 'pll_maybe_translate_term', $this->model->term->get_translation( $term, $lang ), $term, $lang ) ) {
+				if ( $term_id = apply_filters( 'pll_maybe_translate_term', (int) $this->model->term->get_translation( $term, $lang ), $term, $lang ) ) {
```
Chouby, line 108: What's the reason for this change? Did you check that we never explicitely test `false` when using the filter?

admin/admin-filters-term.php
```diff
@@ -588,8 +588,12 @@ public function split_shared_term( $term_id, $new_term_id, $term_taxonomy_id, $t
...
 			}
 			else {
-				$tr_term = get_term( $tr_id, $taxonomy );
-				$translations[ $key ] = _split_shared_term( $tr_id, $tr_term->term_taxonomy_id );
+				$tr_term   = get_term( $tr_id, $taxonomy );
+				$new_tr_id = _split_shared_term( $tr_id, $tr_term->term_taxonomy_id );
+
+				if ( is_int( $new_tr_id ) ) {
+					$translations[ $key ] = $new_tr_id;
+				}
```
Chouby, line 596: Should we go further when we don't get an `int`?

include/api.php
```diff
@@ -73,31 +73,51 @@ function pll_default_language( $field = 'slug' ) {
 }
 
 /**
- * Among the post and its translations, returns the id of the post which is in the language represented by $lang.
+ * Among the post and its translations, returns the ID of the post which is in the language represented by $lang.
  *
  * @api
  * @since 0.5
+ * @since 3.4 Returns 0 instead of false.
```
Chouby, line 80: This is an API change. We should discuss that.

include/api.php
```diff
@@ -73,31 +73,51 @@ function pll_default_language( $field = 'slug' ) {
...
+ * @since 3.4 $lang accepts PLL_Language, string, or int.
```
Chouby, line 81: OK for `PLL_Language` but I don't agree for `int`.

include/api.php
```diff
@@ -73,31 +73,51 @@ function pll_default_language( $field = 'slug' ) {
...
 }
 
 /**
- * Among the term and its translations, returns the id of the term which is in the language represented by $lang.
+ * Among the term and its translations, returns the ID of the term which is in the language represented by $lang.
  *
  * @api
  * @since 0.5
+ * @since 3.4 Returns 0 instead of false.
+ * @since 3.4 $lang accepts PLL_Language, string, or int.
```
Chouby, line 105: Same as above.

include/api.php
```diff
@@ -303,9 +323,10 @@ function pll_languages_list( $args = array() ) {
  *
  * @api
  * @since 1.5
+ * @since 3.4 $lang accepts PLL_Language, string, or int.
```
Chouby, line 326: Same as above.

include/api.php
```diff
@@ -317,9 +338,10 @@ function pll_set_post_language( $id, $lang ) {
...
```
Chouby, line 341: Same as above.

tests/phpunit/tests/test-translated-post.php
```diff
@@ -161,7 +161,7 @@ public function test_update_language( $original_group, $to, $expected_new_group,
 		$post_id = array_shift( $translations );
 		self::$model->post->save_translations( $post_id, $translations );
 
-		self::$model->post->update_language( $post_id, $to, 'post' );
+		self::$model->post->update_language( $post_id, self::$model->get_language( $to ) );
```
Chouby, line 164: WTF? Why the test did not break?

Chouby (changes requested):

include/object-with-language.php
```diff
@@ -0,0 +1,337 @@
...
+	/**
+	 * Add hooks.
+	 *
+	 * @since 3.4
+	 *
+	 * @return self
+	 */
+	public function init() {
+		return $this;
+	}
```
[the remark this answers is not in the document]
Chouby: Forget.

include/object-with-language-interface.php
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @package Polylang
+ */
+
+/**
+ * Interface to use for object types that support at least one language.
+ *
+ * @since 3.4
+ */
+interface PLL_Object_With_Language_Interface {
```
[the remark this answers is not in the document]
Screenfeed: Removed in [37734ad](https://github.com/polylang/polylang/pull/1149/commits/37734ad31a94f38c767fcf1fa48bc402d712d1b3).

include/translated-object-interface.php
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * @package Polylang
+ */
+
+/**
+ * Interface to use for object types that support translations.
+ *
+ * @since 3.4
+ */
+interface PLL_Translated_Object_Interface {
```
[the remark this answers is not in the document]
Screenfeed: Removed in [37734ad](https://github.com/polylang/polylang/pull/1149/commits/37734ad31a94f38c767fcf1fa48bc402d712d1b3).

include/object-with-language.php
```diff
@@ -0,0 +1,337 @@
...
+
+		if ( empty( $id ) ) {
+			return false;
+		}
+
+		$old_lang = $this->get_language( $id );
+		$old_lang = $old_lang ? $old_lang->tl_term_id : 0;
+
+		$lang = $this->model->get_language( $lang );
+		$lang = $lang ? $lang->tl_term_id : 0;
```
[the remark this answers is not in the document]
Screenfeed: Right. Fixed in [0a21288](https://github.com/polylang/polylang/pull/1149/commits/0a212885048ea1fc10ef4f52fda0e5ed45ae7f7c).

include/translated-object.php
```diff
@@ -560,6 +433,7 @@ protected function validate_translations( $translations, $id = 0, $context = 'sa
 		);
 
 		// Make sure values are clean before working with them.
+		/** @phpstan-var array<non-empty-string, positive-int> $translations */
```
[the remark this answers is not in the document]
Screenfeed: The result. Since this variable appears twice on the line, PHPStan seems to consider the first one.

include/object-with-language.php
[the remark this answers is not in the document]
Screenfeed: For the record:
- Hooks are removed from the constructor to this method, so an instance can be created without launching the hooks.
- This method is created in this abstract class, so classes that extend this class don't have to create this method if they don't need it.
- For backward compatibility and preventing us to edit a lot of files (especially in tests), this method is called in `PLL_Translated_Post`'s and `PLL_Translated_Term`'s constructor. This may happen in the future though.
- New classes that extend this one, like the one coming to Polylang for WooCommerce, can call this method outside of the constructor.

include/translated-term.php
```diff
@@ -3,112 +3,223 @@
...
+	 *
+	 * @since 1.8
+	 *
+	 * @param PLL_Model $model Instance of `PLL_Model`.
+	 */
+	public function __construct( PLL_Model &$model ) {
 		parent::__construct( $model );
 
+		$this->init();
+	}
```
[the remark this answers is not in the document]
Screenfeed: https://github.com/polylang/polylang/pull/1149#discussion_r1023865370

include/translated-term.php
```diff
@@ -3,112 +3,223 @@
...
-		wp_delete_object_term_relationships( $this->sanitize_int_id( $term_id ), $this->tax_language );
+	public function delete_language( $id ) {
+		$id = $this->sanitize_int_id( $id );
+
+		if ( empty( $id ) ) {
+			return;
+		}
+
+		wp_delete_object_term_relationships( $id, $this->tax_language );
 	}
```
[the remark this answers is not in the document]
Screenfeed: Done in [552f9ee](https://github.com/polylang/polylang/pull/1149/commits/552f9ee07c0ccb489c053b034a80943624fceee1).

include/translated-term.php
```diff
@@ -3,112 +3,223 @@
...
-			return;
+			return true;
 		}
 
-		wp_set_object_terms( $term_id, $lang, $this->tax_language );
+		$assigned = is_array( wp_set_object_terms( $id, $lang, $this->tax_language ) );
+
+		if ( ! $assigned ) {
+			return false;
+		}
```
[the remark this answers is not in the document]
Screenfeed: Done in [06de35d](https://github.com/polylang/polylang/pull/1149/commits/06de35d624eacee92a8aa0c6a0076787f072fc46).

include/translated-object.php
```diff
@@ -3,176 +3,120 @@
...
+	 * @return string
 	 *
-	 * @param int $id Object id.
-	 * @return PLL_Language|false PLL_Language object, false if no language is associated to that object.
+	 * @phpstan-return non-empty-string
 	 */
-	abstract public function get_language( $id );
+	public function get_tax_translations() {
+		return $this->tax_translations;
+	}
```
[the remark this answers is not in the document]
Screenfeed: For the record:
- This allows to get the taxonomy name instead of hard-coding it everywhere: this could come handy for future and third-party developments.
- There are similar methods in Polylang for WooCommerce.

include/translated-post.php
```diff
@@ -3,92 +3,131 @@
...
-	 * @param int                     $post_id Post id.
-	 * @param int|string|PLL_Language $lang    Language (term_id or slug or object).
-	 * @return void
+	 * @param int                     $id   Post ID.
+	 * @param PLL_Language|string|int $lang Language (object, slug, or term ID).
+	 * @return bool True on success (or if the given language is already assigned to the object). False otherwise.
 	 */
-	public function set_language( $post_id, $lang ) {
-		$post_id = $this->sanitize_int_id( $post_id );
+	public function set_language( $id, $lang ) {
```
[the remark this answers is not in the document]
Screenfeed: Done in [6376f6a](https://github.com/polylang/polylang/pull/1149/commits/6376f6ae32a7020a81a42f64c404386e6bdaba98).

include/translated-object.php
```diff
@@ -3,176 +3,120 @@
...
-		$this->model = &$model;
+	public function __construct( PLL_Model &$model ) {
+		parent::__construct( $model );
+
+		$this->tax_to_cache[] = $this->tax_translations;
 
 		/*
-		 * Register our taxonomies as soon as possible.
+		 * Register our taxonomy as soon as possible.
 		 * This is early registration, not ready for rewrite rules as $wp_rewrite will be setup later.
```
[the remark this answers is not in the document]
Screenfeed: Done in [2f5c32e](https://github.com/polylang/polylang/pull/1149/commits/2f5c32e9a94888da1ac86011e604404375c6a0a1).

include/translated-object.php
```diff
@@ -3,176 +3,120 @@
...
-		// Stores it the way WP expects it. Set an empty cache if no term was found in the taxonomy.
-		foreach ( $taxonomies as $tax ) {
-			wp_cache_add( $object_id, empty( $terms[ $tax ] ) ? array() : array( $terms[ $tax ] ), $tax . '_relationships' );
-		}
+		// Remove the object's former language from the new translations group.
+		$translations = array_diff( $translations, array( $id ) );
+		$this->save_translations( $id, $translations );
 
-		return $term;
+		return true;
```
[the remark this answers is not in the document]
Screenfeed: Done in [06de35d](https://github.com/polylang/polylang/pull/1149/commits/06de35d624eacee92a8aa0c6a0076787f072fc46).

include/api.php
Screenfeed: Done in [10ec660](https://github.com/polylang/polylang/pull/1149/commits/10ec660bc67a078dde283dd1c86b5566cee60343).

include/api.php
Screenfeed: Done in [10ec660](https://github.com/polylang/polylang/pull/1149/commits/10ec660bc67a078dde283dd1c86b5566cee60343).

include/api.php
Screenfeed: Done in [10ec660](https://github.com/polylang/polylang/pull/1149/commits/10ec660bc67a078dde283dd1c86b5566cee60343).

include/api.php
Screenfeed: Done in [10ec660](https://github.com/polylang/polylang/pull/1149/commits/10ec660bc67a078dde283dd1c86b5566cee60343).

include/translated-term.php
```diff
@@ -3,112 +3,223 @@
...
+		// get_term_by() still not cached in WP 3.5.1 but internally, the function is always called by term_id.
+		$term = get_term_by( 'slug', $slug, $taxonomy );
+
+		if ( ! $term instanceof WP_Term ) {
 			return false;
 		}
 
 		// Get the language and make sure it is a PLL_Language object.
-		$lang = $this->get_object_term( $term_id, $this->tax_language );
+		$lang = $this->get_object_term( $term->term_id, $this->tax_language );
```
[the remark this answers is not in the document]
Screenfeed: Good idea. Done in [d6c1f24](https://github.com/polylang/polylang/pull/1149/commits/d6c1f247129566cfd09636a8f6cd7c54f2dfa428).

include/translated-object.php
```diff
@@ -365,92 +299,45 @@ public function get_translation( $id, $lang ) {
 	}
 
 	/**
-	 * Among the object and its translations, returns the id of the object which is in $lang
+	 * Among the object and its translations, returns the ID of the object which is in `$lang`.
 	 *
 	 * @since 0.1
+	 * @since 3.4 Returns 0 instead of false.
```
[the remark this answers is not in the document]
Screenfeed: I found only [one reference](https://github.com/polylang/polylang-pro/blob/3567e9416947690bdb6d882df0a02d6b118a3c8b/modules/sync/sync-content.php#L452) in PLL pro, nothing in PLL nor PLLWC.

modules/sync/sync-tax.php
Screenfeed: I found only `PLL_Sync_Content::duplicate_term()` that uses this filter and [it uses `empty()`](https://github.com/polylang/polylang-pro/blob/3567e9416947690bdb6d882df0a02d6b118a3c8b/modules/sync/sync-content.php#L187).

include/api.php
Screenfeed: I found no references.

Screenfeed:
I wonder if `PLL_Translated_Term::_prime_terms_cache()`, `PLL_Translated_Term::clean_term_cache()`, and `PLL_Translated_Term::should_update_translation_group()` should be moved in a parent class 🤔.

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+		}
+	}
+
+	/**
+	 * Returns a language term property value (like term ID, term taxonomy ID, or count).
+	 *
+	 * @since 3.4
+	 *
+	 * @param string $taxonomy_name Name of the taxonomy.
+	 * @param string $prop_name     Name of the property: 'term_taxonomy_id', 'term_id', 'count'.
```
Hug0-Drelon, line 347:
```suggestion
	 * @param string $prop_name     Name of the property (e.g. 'term_taxonomy_id', 'term_id', 'count').
```
Right now, I understand it accepts only 'term_taxonomy_id', 'term_id', 'count'.

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+		return isset( $this->term_props[ $taxonomy_name ][ $prop_name ] ) ? $this->term_props[ $taxonomy_name ][ $prop_name ] : 0;
+	}
+
+	/**
+	 * Stores a language term property value (like term ID, term taxonomy ID, or count).
+	 *
+	 * @since 3.4
+	 *
+	 * @param string $taxonomy_name Name of the taxonomy.
+	 * @param string $prop_name     Name of the property: 'term_taxonomy_id', 'term_id', 'count'.
```
Hug0-Drelon, line 364: Same as above :)

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+	 * @param int    $prop_value    Property value.
+	 * @return int
```
Hug0-Drelon, line 366: Why returning `int` instead of `bool`?

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+	 * @phpstan-param non-empty-string $prop_name
+	 * @phpstan-return int<0, max>
 	 */
-	public function __construct( $language, $term_language = null ) {
-		if ( empty( $term_language ) ) {
-			// Build the object from all properties stored as an array.
-			foreach ( $language as $prop => $value ) {
-				$this->$prop = $value;
+	public function set_tax_prop( $taxonomy_name, $prop_name, $prop_value ) {
+		if ( ! is_numeric( $prop_value ) ) {
```
Hug0-Drelon, line 373: You expect numeric string?
Otherwise `is_int()` would do the trick?

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+			return 0;
+		}
+
+		$prop_value = $prop_value >= 1 ? abs( (int) $prop_value ) : 0;
```
Hug0-Drelon, line 377: Why use `abs()`? I don't understand 😅

include/language.php
```diff
@@ -432,8 +620,18 @@ public function get_display_flag_url() {
...
+		}
+
+		/**
+		 * Fires when updating language term counts.
+		 *
+		 * @since 3.4
+		 *
+		 * @param PLL_Language $language The language object.
+		 */
+		do_action( 'pll_updated_language_term_counts', $this );
```
Hug0-Drelon, line 634: Why this action? Do you plan to use it elsewhere? Or is it for our users?

include/language.php
```diff
@@ -490,4 +693,288 @@ public function set_url_scheme() {
...
+		}
+
+		/**
+		 * @phpstan-var array{
+		 *     term_id: positive-int,
+		 *     term_taxonomy_id: positive-int,
+		 *     count: int<0, max>
+		 * }
+		 */
+		return array_intersect_key(
```
Hug0-Drelon, line 971: What happens if some new language term requires a new prop?
Should we add a filter to allow to sanitize it?

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
-	 * Html markup of the custom flag if it exists.
+	 * HTML markup of the custom flag if it exists.
 	 *
 	 * @var string|null
 	 */
 	public $custom_flag;
 
 	/**
-	 * Constructor: builds a language object given its two corresponding terms in 'language' and 'term_language' taxonomies.
+	 * Stores language term properties (like term IDs and counts) for each language taxonomy (`language`,
```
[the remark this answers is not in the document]
Screenfeed: Nope.
>for all language taxonomies

Or:
>for each language taxonomy

include/language.php
Screenfeed: Indeed, it only accepts those.

include/language.php
Screenfeed: To return the sanitized value.

include/language.php
Screenfeed: Yes, numeric, because we often have to deal with integers in string, that's one of the purposes of this method.

include/language.php
Screenfeed: So `$prop_value` is considered as a 0+ integer by PHPStan 😉.

include/language.php
Screenfeed: We should keep it hard-coded so far imho: if we allow other properties, we would also need to do something about sanitization and validation.

include/api.php
```diff
@@ -303,9 +323,10 @@ function pll_languages_list( $args = array() ) {
...
  * @since 1.5
+ * @since 3.4 $lang accepts PLL_Language or string.
  *
- * @param int    $id   Post id.
- * @param string $lang Language code.
+ * @param int                 $id   Post id.
+ * @param PLL_Language|string $lang Language (object or slug).
  * @return void
  */
 function pll_set_post_language( $id, $lang ) {
```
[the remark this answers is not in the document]
Screenfeed: Yeah, why not, this shouldn't break anything and since we have the data at hand...
Done in [fda227d](https://github.com/polylang/polylang/pull/1149/commits/fda227d7895656589e0c94c73d20db74e2395790).

include/api.php
```diff
@@ -317,9 +338,10 @@ function pll_set_post_language( $id, $lang ) {
...
  * @since 1.5
+ * @since 3.4 $lang accepts PLL_Language or string.
  *
- * @param int    $id   Term id.
- * @param string $lang Language code.
+ * @param int                 $id   Term id.
+ * @param PLL_Language|string $lang Language (object or slug).
  * @return void
  */
 function pll_set_term_language( $id, $lang ) {
```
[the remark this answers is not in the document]
Screenfeed: Same answer then 😊.

include/language.php
```diff
@@ -39,106 +36,141 @@ class PLL_Language {
...
 	public $term_group;
 
+	/**
+	 * ID of the term in 'language' taxonomy.
+	 *
+	 * @var int
+	 *
+	 * @phpstan-var int<0, max>
+	 */
+	public $term_id;
```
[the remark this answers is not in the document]
Screenfeed: Yes, good idea.
Done in [0f26b2f](https://github.com/polylang/polylang/pull/1149/commits/0f26b2fb66a69b88679f1721bc12739066cc5744).

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
 
 	/**
-	 * Constructor: builds a language object given its two corresponding terms in 'language' and 'term_language' taxonomies.
+	 * Stores language term properties (like term IDs and counts) for each language taxonomy (`language`,
+	 * `term_language`, etc).
+	 * This stores the values of the properties `$term_id`, `$term_taxonomy_id`, `$count`, `$tl_term_id`,
+	 * `$tl_term_taxonomy_id`, and `$tl_count`.
+	 *
+	 * @var array[] Array keys are property names.
+	 *
```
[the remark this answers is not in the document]
Screenfeed: Good idea.
Done (and fixed) in [21f6b81](https://github.com/polylang/polylang/pull/1149/commits/21f6b81cd0ae39bc50e74a3b1a2a1faf15bcea34).

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+				foreach ( $prop_values as $prop_name => $prop_value ) {
+					$this->set_tax_prop( $taxonomy_name, $prop_name, $prop_value );
+				}
+			}
+		}
+
+		unset( $language_data['term_props'] );
+
+		// Add all the other values.
+		foreach ( $language_data as $prop => $value ) {
```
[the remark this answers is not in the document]
Screenfeed: I thought about it too, but so far I did nothing about it. @Chouby will give us his opinion in his review.

include/language.php
Screenfeed: Same answer 😁.

include/language.php
Screenfeed: It's a hook I added at first, but we don't need it anymore.
I removed it in [47b6a57](https://github.com/polylang/polylang/pull/1149/commits/47b6a578803e5c072fb6f30eec7d9a6dfe19a188).

include/language.php
```diff
@@ -148,78 +180,234 @@ class PLL_Language {
...
+	 *     @type string  $flag            Optional. HTML markup of the flag.
+	 *     @type string  $custom_flag_url Optional. URL of the custom flag if it exists.
+	 *     @type string  $custom_flag     Optional. HTML markup of the custom flag if it exists.
+	 *     @type int     $page_on_front   Optional. ID of the page on front in this language.
+	 *     @type int     $page_for_posts  Optional. ID of the page for posts in this language.
+	 * }
 	 *
-	 * @param WP_Term|array $language      Term in 'language' taxonomy or language object properties stored as an array.
-	 * @param WP_Term       $term_language Corresponding 'term_language' term.
+	 * @phpstan-param array{
```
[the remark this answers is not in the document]
Screenfeed: `term_props` is mandatory since [2946a79](https://github.com/polylang/polylang/pull/1149/commits/2946a79f7e4892a88bd251b557e9e52c3e977eba) (not in the code thought, only in the doc), but I kept the old properties for PHPStan as optional, for backward compatibility. Note that they have been removed from the doc for humans.
Maybe we should remove them for PHPStan too indeed.

include/language.php
Hug0-Drelon: Fine, I understand your point.

include/language.php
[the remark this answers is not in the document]
Hug0-Drelon: oops ^^

include/language.php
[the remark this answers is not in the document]
Hug0-Drelon: I'll let you chose if we should remove them from PHPStan 😉

include/language.php
Hug0-Drelon: 🙈

include/language.php
Hug0-Drelon: Ok, I see.

include/language.php
Hug0-Drelon: 👍

include/language.php
Hug0-Drelon: Ok! Nice trick

include/translated-object.php
```diff
@@ -3,176 +3,113 @@
...
 
-		$this->set_language( $id, $lang );
+		$id = $this->sanitize_int_id( $id );
 
 		$translations = $this->get_translations( $id );
 
-		if ( $translations ) {
-			// Remove the post's former language from the new translations group.
+		// Don't create translation groups with only 1 value.
+		if ( ! empty( $translations ) ) {
```
Hug0-Drelon, line 95: If I'm not wring, term translations are stored even if they contains one value for the moment.
This would avoid that? Could it be a problem @Chouby?

Hug0-Drelon:
Also, `PLL_Model` is getting harder and harder to read 😅 
It would be cool to split it out some day!

I have a remaining question about `PLL_Translatable_Objects_Registry`. Could we offer the opportunity to remove an translatable object?

include/mo.php
```diff
@@ -78,14 +78,30 @@ public function import_from_db( $lang ) {
...
+	 * Returns the post ID of the post storing the strings translations.
+	 *
+	 * @since 3.4
+	 *
+	 * @param int $term_id The language term ID.
+	 * @return int|null
+	 *
+	 * @phpstan-return positive-int|null
+	 */
+	public static function get_id_from_term_id( $term_id ) {
```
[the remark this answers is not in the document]
Screenfeed: So we can use it in `PLL_Model::prepare_data_for_language()` without creating a `PLL_Language` object.

include/model.php
```diff
@@ -50,8 +59,9 @@ public function __construct( &$options ) {
 		$this->options = &$options;
 
 		$this->cache = new PLL_Cache();
-		$this->post = new PLL_Translated_Post( $this ); // Translated post sub model.
-		$this->term = new PLL_Translated_Term( $this ); // Translated term sub model.
+		$this->translatable_objects = new PLL_Translatable_Objects_Registry();
+		$this->post = $this->translatable_objects->register( new PLL_Translated_Post( $this ) ); // Translated post sub model.
```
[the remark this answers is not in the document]
Screenfeed: I realized that these objects are duplicated now but I haven't give it more thinking so far.

include/mo.php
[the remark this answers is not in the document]
Hug0-Drelon: Ok, thanks!

include/model.php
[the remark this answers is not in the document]
Hug0-Drelon: Fine, let's see what @Chouby thinks about it then 🙂

Chouby (changes requested):

include/object-with-language.php
```diff
@@ -0,0 +1,416 @@
...
+	 * @since 3.1
+	 *
+	 * @param int          $id   Object ID.
+	 * @param PLL_Language $lang New language to assign to the object.
+	 * @return bool True when successfully assigned. False otherwise (or if the given language is already assigned to
+	 *              the object).
+	 */
+	public function update_language( $id, PLL_Language $lang ) {
+		return $this->set_language( $id, $lang );
+	}
```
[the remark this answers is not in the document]
Screenfeed: I believe so, because it is overwritten in `PLL_Translated_Object`: in that class, `update_language()` takes care of the translations while `update_language()` doesn't.

include/language.php
```diff
@@ -460,21 +729,18 @@ public function set_home_url( $search_url, $home_url ) {
...
-		if ( ! empty( $this->flag_url ) ) {
-			$this->flag_url = set_url_scheme( $this->flag_url );
-		}
+			$url = set_url_scheme( $this->$prop );
 
-		if ( ! empty( $this->custom_flag_url ) ) {
-			$this->custom_flag_url = set_url_scheme( $this->custom_flag_url );
+			if ( ! empty( $url ) ) {
+				$this->$prop = $url;
+			}
```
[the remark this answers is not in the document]
Screenfeed: Because `set_url_scheme()` returns `string` (not `non-empty-string`), and PHPStan doesn't like it because those properties are tagged with `non-empty-string`. The other solution is to do this:
```php
$url = set_url_scheme( $this->$prop );
/** @var non-empty-string $url */
$this->$prop = $url;
```

include/object-with-language.php
```diff
@@ -0,0 +1,416 @@
...
+	 * @param PLL_Language|PLL_Language[]|string|string[] $lang A `PLL_Language` object, or a comma separated list of language slugs, or an array of language slugs or objects.
+	 * @return string The WHERE clause.
+	 */
+	public function where_clause( $lang ) {
+		/*
+		 * $lang is an object.
+		 * This is generally the case if the query is coming from Polylang.
+		 */
+		if ( $lang instanceof PLL_Language ) {
+			return ' AND pll_tr.term_taxonomy_id = ' . $lang->get_tax_prop( $this->tax_language, 'term_taxonomy_id' );
```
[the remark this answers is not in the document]
Screenfeed: With the internal sanitization, it shouldn't be possible to have something else than an integer >= 0.
An other solution would be:
```php
return sprintf( ' AND pll_tr.term_taxonomy_id = %d', $lang->get_tax_prop( $this->tax_language, 'term_taxonomy_id' ) );
```

include/translatable-objects-registry.php
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	public function get_from_tax( $tax_name ) {
+		foreach ( $this->objects as $object ) {
+			if ( $object->get_tax_language() === $tax_name ) {
+				return $object;
+			}
+		}
+
+		return null;
+	}
```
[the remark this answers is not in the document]
Screenfeed: I think I planned to use them, but maybe I found another way after creating them. I'll take a look.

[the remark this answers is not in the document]
Screenfeed: They don't seem to be useful anymore, they can be deleted.

include/model.php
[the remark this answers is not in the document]
Screenfeed: Note: when the container is here, each object will be stored in one place.

include/translated-object.php
```diff
@@ -3,188 +3,113 @@
...
 	/**
-	 * PLL_Language property name for the term_taxonomy id.
+	 * Object type to use when checking capabilities.
 	 *
 	 * @var string
+	 *
+	 * @phpstan-var non-empty-string
 	 */
-	protected $tax_tt;
+	protected $cap_type;
```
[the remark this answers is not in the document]
Screenfeed: Done in [0bd442c](https://github.com/polylang/polylang/pull/1159/commits/0bd442c9e05624126f0ee840200460354032184c).

include/translated-object.php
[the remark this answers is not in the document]
Screenfeed: Fixed by https://github.com/polylang/polylang-pro/pull/1550.

include/translatable-objects-registry.php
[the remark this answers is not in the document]
manooweb: Removed

include/translatable-objects-registry.php
```diff
@@ -0,0 +1,108 @@
...
+	 * Tells if a translatable object is registered, given an object type.
+	 *
+	 * @since 3.4
+	 *
+	 * @param string $object_type The object type.
+	 * @return bool
+	 */
+	public function has( $object_type ) {
+		return isset( $this->objects[ $object_type ] );
+	}
```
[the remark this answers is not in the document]
manooweb: Method remove and its calls replaced with this suggested in PLL_Model and PLL_Admin_Model class

admin/admin-model.php
```diff
@@ -456,11 +480,19 @@ public function set_translation_in_mass( $type, $translations ) {
...
+		$term_ids   = array();
+		$dr         = array();
+		$dt         = array();
+		$ut         = array();
+
+		foreach ( $this->translatable_objects->get_all() as $object ) {
+			if ( $object instanceof PLL_Translated_Object ) {
+				$taxonomies[] = $object->get_tax_translations();
+			}
+		}
```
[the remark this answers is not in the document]
manooweb: Add `PLL_Model::get_taxonomy_names()` and replace loops with it.

Screenfeed:
Closed in favor of https://github.com/polylang/polylang/pull/1159, https://github.com/polylang/polylang/pull/1160, https://github.com/polylang/polylang/pull/1158, and https://github.com/polylang/polylang/pull/1161.

PR #1151, opened 2022-11-14
Fix PHPStan errors since szepeviktor/phpstan-wordpress new release

Hug0-Drelon:
New szepeviktor/phpstan-wordpress releases have shown new errors in PHPStan.

Some errors are trivial and concerns number of required parameters not matching callbacks.

But others concern mixed between actions and filters. See these pieces of documentation:
- [`pre_option_{$option}`](https://developer.wordpress.org/reference/hooks/pre_option_option/)
- [`request`](https://developer.wordpress.org/reference/hooks/request/)

Special case for `template_redirect` action, where I ignored the error since `PLL_Canonical::check_canonical_url()` is used as an action callback (which shouldn't return anything)  but also inside itself here:
https://github.com/polylang/polylang/blob/3.3-beta1/frontend/canonical.php#L156

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -210,11 +210,12 @@ class PLL_Filters {
 			// Take care that 'exclude' argument accepts integer or strings too.
 			$args['exclude'] = array_merge( wp_parse_id_list( $args['exclude'] ), $this->get_related_page_ids( $language, 'NOT IN', $args ) ); // phpcs:ignore WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude
 			$pages = get_pages( $args );
+			$pages = ! $pages ? array() : $pages;
 		}
 
 		$ids = wp_list_pluck( $pages, 'ID' );
 
-		if ( ! $once ) {
+		if ( ! $once && is_array( $pages ) ) {
 			// Filters the queried list of pages by language.
 			$ids = array_intersect( $ids, $this->get_related_page_ids( $language, 'IN', $args ) );
 
--- a/include/filters.php
+++ b/include/filters.php
@@ -209,13 +209,13 @@ class PLL_Filters {
 
 			// Take care that 'exclude' argument accepts integer or strings too.
 			$args['exclude'] = array_merge( wp_parse_id_list( $args['exclude'] ), $this->get_related_page_ids( $language, 'NOT IN', $args ) ); // phpcs:ignore WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude
-			$pages = get_pages( $args );
-			$pages = ! $pages ? array() : $pages;
+			$numbered_pages  = get_pages( $args );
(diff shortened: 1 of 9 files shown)


manooweb (approved):
👌 LGTM

include/filters.php
```diff
@@ -210,11 +210,12 @@ public function get_pages( $pages, $args ) {
 			// Take care that 'exclude' argument accepts integer or strings too.
 			$args['exclude'] = array_merge( wp_parse_id_list( $args['exclude'] ), $this->get_related_page_ids( $language, 'NOT IN', $args ) ); // phpcs:ignore WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude
 			$pages = get_pages( $args );
+			$pages = ! $pages ? array() : $pages;
 		}
 
 		$ids = wp_list_pluck( $pages, 'ID' );
 
-		if ( ! $once ) {
+		if ( ! $once && is_array( $pages ) ) {
```
Chouby, line 218: - These 2 modifications look redundant. Am I wrong?
- If you assume that `get_pages()` returns false then something's broken (because we start from a non empty `$pages` and transform it to `false`. So, it would be better to take the opportunity to use an intermediate variable to fix this.

include/translate-option.php
```diff
@@ -194,7 +194,7 @@ public function pre_update_option( $value, $old_value, $name ) {
 		// Stores the unfiltered old option value before it is updated in DB.
 		remove_filter( 'option_' . $name, array( $this, 'translate' ) );
 		$unfiltered_old_value = get_option( $name );
-		add_filter( 'option_' . $name, array( $this, 'translate' ), 20, 2 );
+		add_filter( 'option_' . $name, array( $this, 'translate' ), 20 );
```
Chouby, line 197: Please temporarily ignore the error. This line will be removed by #1111

modules/wpml/wpml-api.php
```diff
@@ -28,7 +28,7 @@ public function __construct() {
 		// Site Wide Language informations
 
 		add_filter( 'wpml_active_languages', array( $this, 'wpml_active_languages' ), 10, 2 );
-		add_filter( 'wpml_display_language_names', array( $this, 'wpml_display_language_names' ), 10, 5 );
+		add_filter( 'wpml_display_language_names', array( $this, 'wpml_display_language_names' ), 10, 2 );
```
Chouby, line 31:
```suggestion
		add_filter( 'wpml_display_language_names', array( $this, 'wpml_display_language_names' ), 10, 2 ); // Because we don't translate language names, 3rd to 5th parameters are not supported.
```

Chouby (changes requested):

include/filters.php
Hug0-Drelon: - Indeed, it's redundant...
- Good idea to use an intermediate variable, didn't think about overriding `$pages` ...
Fixed in f034b02.

PR #1154, opened 2022-11-20
Only store term_ids in taxonomy relationships cache

ocean90:
In WordPress 6.0 the cache value of the `{$taxonomy}_relationships` has been simplified to only store term IDs instead of full `WP_Term` objects, see https://core.trac.wordpress.org/changeset/52836 and https://make.wordpress.org/core/2022/04/28/taxonomy-performance-improvements-in-wordpress-6-0/.

This PR adjusts `PLL_Translated_Object::get_object_term()` to only store term IDs in 6.0+ to keep the cache small and to avoid potential fatal errors in the future. This was noticed while working on https://github.com/WordPress/wordpress-develop/pull/3647.
So far this is not an issue due to the back-compat handling in [get_object_term_cache()](https://github.com/WordPress/wordpress-develop/blob/cfd09a412765e072b842e3216605c577f667d462/src/wp-includes/taxonomy.php#L3677).

While #1149 also touches the function it's not changing the caching.

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -129,6 +129,8 @@ abstract class PLL_Translated_Object {
 	 * @return WP_Term|false The term associated to the object in the requested taxonomy if it exists, false otherwise.
 	 */
 	public function get_object_term( $object_id, $taxonomy ) {
+		global $wp_version;
+
 		$object_id = $this->sanitize_int_id( $object_id );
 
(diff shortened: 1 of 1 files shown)


Screenfeed (approved):

Screenfeed:
@ocean90 Thank you for the heads up and this PR.

PR #1156, opened 2022-11-24
Check rewrite rules before processing them.

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1540

## Description
- Check `$rules` is an array containing string keys and values.
- ~~Add a test passing bad rules and checking there is no errors on our side (⚠️ weird test using `expectNotToPerformAssertions()`).~~ `expectNotToPerformAssertions()` isn't available for PHPUnit < 7.2 so I used `$this->assertTrue( true );`...


Chouby:
6fbd525 replaces this PR (force push) after a git mistake.

PR #1165, opened 2022-12-14
Remove useless PLL_Translatable_Object::update_language()

Marianne380:
See https://github.com/polylang/polylang-wc/issues/502

Removal of `update_language()`. 
`PLL_Translated_Object::set_language()` handles translation groups.

It looks like that the handling of terms that must always have a translation group is already handled [here](https://github.com/polylang/polylang/blob/3.4-dev/include/translated-term.php#L106-L113).

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -78,6 +78,8 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 			return false;
 		}
 
+		$id = $this->sanitize_int_id( $id );
+
 		$translations = $this->get_translations( $id );
 
 		// Don't create translation groups with only 1 value.
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -68,8 +68,8 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 	 *
 	 * @since 3.4
 	 *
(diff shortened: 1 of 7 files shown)


Screenfeed:
For reference, [the PR that introduced `update_language()`](https://github.com/polylang/polylang/pull/807/files#diff-b622973a908ca73dc41a63a7722f68a1659c6cfa9c9d26da4cb416730b9c2aa1).

include/translated-object.php
```diff
@@ -64,17 +64,17 @@ public function get_tax_translations() {
...
 	/**
-	 * Assigns a new language to an object, taking care of the translations group.
+	 * Assigns a language to an object, taking care of the translations group.
 	 *
-	 * @since 3.1
+	 * @since 3.4
 	 *
 	 * @param int          $id   Object ID.
-	 * @param PLL_Language $lang New language to assign to the object.
+	 * @param PLL_Language $lang Language to assign to the object.
```
Screenfeed, line 72: `$lang`'s type should be `PLL_Language|string|int`, like the parent method.

Screenfeed (changes requested):

Screenfeed (approved):

Chouby (approved):

PR #1167, opened 2022-12-15
Get rid of deprecated PLL_Language properties.

Marianne380:
See https://github.com/polylang/polylang-wc/issues/502

Remove direct call to deprecated properties and use `PLL_Language::get_tax_prop()` instead.

### DIFF
--- a/integrations/wp-sweep/wp-sweep.php
+++ b/integrations/wp-sweep/wp-sweep.php
@@ -49,11 +49,19 @@ class PLL_WP_Sweep {
 
 		$excluded_term_ids = array_merge( $excluded_term_ids, $_term_ids );
 
+		$language_taxonomies = PLL()->model->translatable_objects->getIterator();
+		$languages           = PLL()->model->get_languages_list();
+		$lt_term_ids         = array();
+		foreach ( $language_taxonomies as $language_taxonomy ) {
+			foreach ( $languages as $language ) {
+				$lt_term_ids[] = $language->get_tax_prop( $language_taxonomy->get_tax_language(), 'term_id' );
+			}
+		}
+
 		// Add the terms of our languages.
 		$excluded_term_ids = array_merge(
 			$excluded_term_ids,
-			pll_languages_list( array( 'fields' => 'term_id' ) ),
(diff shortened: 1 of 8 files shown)


Screenfeed (changes requested):
Some are missing:
- https://github.com/polylang/polylang/blob/11a25e1862455212108f93aadbfef6950a08ed54/include/query.php#L112
- https://github.com/polylang/polylang/blob/11a25e1862455212108f93aadbfef6950a08ed54/settings/table-languages.php#L71-L72
- https://github.com/polylang/polylang/blob/11a25e1862455212108f93aadbfef6950a08ed54/integrations/wp-sweep/wp-sweep.php#L56 This one will require to loop through the language taxonomies to add the `term_id` of `language`, `term_language`, `wc_order_language`, etc.
- https://github.com/polylang/polylang/blob/11a25e1862455212108f93aadbfef6950a08ed54/frontend/frontend-filters.php#L82
- https://github.com/polylang/polylang/blob/11a25e1862455212108f93aadbfef6950a08ed54/include/model.php#L148

Screenfeed (approved):

integrations/wp-sweep/wp-sweep.php
```diff
@@ -49,11 +49,19 @@ public function wp_sweep_excluded_termids( $excluded_term_ids ) {
...
+			}
+		}
+
 		// Add the terms of our languages.
 		$excluded_term_ids = array_merge(
 			$excluded_term_ids,
-			pll_languages_list( array( 'fields' => 'term_id' ) ),
-			pll_languages_list( array( 'fields' => 'tl_term_id' ) )
+			$lt_term_ids
 		);
```
Chouby, line 65: You can do much simpler by:
1. Avoiding a direct call to `getIterator()`. It's for PHP, you probably never it.
2. Avoiding an intermediate array ( again ;) )
3. Use `get_tax_prop**s**()` which definitely kills the bullet 1.

```php
		// Add the terms of our languages.
			$excluded_term_ids = array_merge(
				$excluded_term_ids,
				$language->get_tax_props( 'term_id' );
			);
		}
```

Chouby (changes requested):

PR #1170, opened 2022-12-16
Fix PHPunit errors since language objects refactor.

Hug0-Drelon:
After all the work done to refactor language objects, several PHPUnit errors are raised (see: https://app.travis-ci.com/github/polylang/polylang/jobs/591309741).
This PR only takes care of the errors. I suggest to handle failures in another PR.

There is a remaining error in `Settings_Test::test_edit_language`, this will be handled in another PR to make review easier.

PR #1161, opened 2022-12-06
API and types improvement

Marianne380:
See https://github.com/polylang/polylang-wc/issues/502

PR #1173, opened 2022-12-16
Fix PHPunit failures since language objects refactor.

Hug0-Drelon:
After all the work done to refactor language objects, several PHPUnit failures are raised (see: https://app.travis-ci.com/github/polylang/polylang/jobs/591410166).
Only one failure remains, the one in `Settings_Test::test_notice_for_objects_with_no_lang`. Indeed some work has to be done regarding `get_object_with_no_lang()` (c.f. https://github.com/polylang/polylang-pro/issues/1554)

For convenience, we might want to merge PR #1170 before, since this branch is based on the one from the mentioned PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,18 +23,18 @@ jobs:
       script:
         - "vendor/bin/phpunit --verbose"
 
-    - name: "WP 6.0 - PHP 5.6"
+    - name: "WP 6.1 - PHP 5.6"
       dist: "xenial"
       php: "5.6"
-      env: "WP_VERSION=6.0 WP_MULTISITE=0"
+      env: "WP_VERSION=6.1 WP_MULTISITE=0"
 
-    - name: "WP 6.0 - PHP 8.0"
+    - name: "WP 6.1 - PHP 8.0"
       php: "8.0"
-      env: "WP_VERSION=6.0 WP_MULTISITE=0"
+      env: "WP_VERSION=6.1 WP_MULTISITE=0"
 
-    - name: "WP 6.0 multisite - PHP 7.2"
+    - name: "WP 6.1 multisite - PHP 7.2"
       php: "7.2"
-      env: "WP_VERSION=6.0 WP_MULTISITE=1"
+      env: "WP_VERSION=6.1 WP_MULTISITE=1"
 
     - name: "WP nightly - PHP 5.6"
       dist: "xenial"
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,18 +23,18 @@ jobs:
       script:
         - "vendor/bin/phpunit --verbose"
 
-    - name: "WP 6.1 - PHP 5.6"
+    - name: "WP 6.0 - PHP 5.6"
       dist: "xenial"
       php: "5.6"
-      env: "WP_VERSION=6.1 WP_MULTISITE=0"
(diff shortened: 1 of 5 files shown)


tests/phpunit/tests/test-media.php
```diff
@@ -80,38 +79,6 @@ public function test_attachment_fields_to_edit() {
 		$this->assertFalse( isset( $fields['language'] ) );
 	}
 
-	/**
-	 * @since 3.1 Since the language and translations are updated through a previous AJAX call, we'd rather not perform an unnecessary update now.
-	 */
-	public function test_attachment_fields_to_save() {
```
Marianne380, line 86: Why don't we need this test anymore?

Hug0-Drelon: This test is checking that updating a media (but not its language) does not update the langue nor the translations. But `PLL_Translated_Object::set_language` and `PLL_Translatable_Object::set_language` are taking care of this right now. See: https://github.com/polylang/polylang/blob/29854c72ddaf357772bc12b1e7627236bf839f8d/include/translated-object.php#L86-L90 and https://github.com/polylang/polylang/blob/3.4-dev/include/translatable-object.php#L167-L169
And since the test is checking that `set_language` and `save_translations` are not called through a mock, this is becoming obsolete. Indeed, these function are now called anyway.

Marianne380 (approved):

.travis.yml
```diff
@@ -23,18 +23,18 @@ jobs:
       script:
         - "vendor/bin/phpunit --verbose"
 
-    - name: "WP 6.0 - PHP 5.6"
+    - name: "WP 6.1 - PHP 5.6"
```
Chouby, line 26: Please don't modify the Travis config to avoid merge conflicts. This is already done on master branch.

tests/phpunit/tests/test-create-delete-languages.php
```diff
@@ -146,7 +146,6 @@ public function test_language_properties_in_transient() {
 			'name',
 			'slug',
 			'term_group',
-			'term_taxonomy_id',
```
Chouby, line 149: Maybe can we had more assertions to check the content of `$term_props`. What do you think?

tests/phpunit/tests/test-media.php
Chouby: This was indeed the purpose of this test. Can we add alternative test(s) that will adapt this check to our new way to update languages?

Chouby (changes requested):

tests/phpunit/tests/test-media.php
Hug0-Drelon: We started a list of tests to implement, and it's added into it. Should I create the test in this PR?

.travis.yml
Hug0-Drelon: Yup, done in 811fe4d

tests/phpunit/tests/test-create-delete-languages.php
Hug0-Drelon: It has already been done in https://github.com/polylang/polylang/pull/1161
See: https://github.com/polylang/polylang/blob/3.4-dev/tests/phpunit/tests/test-create-delete-languages.php#L172-L180

Chouby: :fish:

PR #1160, opened 2022-12-02
Add a registry for PLL_Translatable_Object

manooweb:
Related #1149, https://github.com/polylang/polylang-wc/issues/502.

Requires #1159 

`get_terms()` function obviously accept an array of taxonomy names.
See https://developer.wordpress.org/reference/functions/get_terms/#parameters

PR #1175, opened 2022-12-19
Simplify languages list creation

manooweb:
After merging #1160 it remained to simplify how the languages list is created.

It is the goal of this PR

## Changes
- Modify the terms request in `PLL_Model::get_language_terms()` to request all taxonomies in one time. Also introduce `'orderby' => 'term_group'` to take in account the language display order.
- Remove `PLL_Model::get_terms_by_type` method become useless
- Simplify `PLL_Model::get_languages_from_taxonomies()` to prepare data to send to `PLL_Language_Factory::get_from_terms()` method
- Remove reference to content type in the parameter of `PLL_Language_Factory::get_from_terms()` method and use language taxonomies instead.

### DIFF
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -41,7 +41,7 @@ class PLL_Language_Factory {
 	 *
 	 * @since 3.4
 	 *
-	 * @param WP_Term[] $terms List of language terms, with the language taxonomies language as array keys.
+	 * @param WP_Term[] $terms List of language terms, with the language taxonomy names as array keys.
 	 *                         `language` and `term_language` are mandatory keys.
 	 * @return PLL_Language
 	 *
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -45,7 +45,7 @@ class PLL_Language_Factory {
 	 *                         `language` and `term_language` are mandatory keys.
 	 * @return PLL_Language
 	 *
-	 * @phpstan-param array{post:WP_Term, term:WP_Term}&array<string, WP_Term> $terms
+	 * @phpstan-param array{language:WP_Term, term_language:WP_Term}&array<string, WP_Term> $terms
 	 */
 	public static function get_from_terms( array $terms ) {
 		$languages = self::get_languages();
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -41,8 +41,8 @@ class PLL_Language_Factory {
 	 *
 	 * @since 3.4
 	 *
-	 * @param WP_Term[] $terms List of language terms, with the type as array keys.
-	 *                         `post` and `term` are mandatory keys.
+	 * @param WP_Term[] $terms List of language terms, with the language taxonomies language as array keys.
+	 *                         `language` and `term_language` are mandatory keys.
 	 * @return PLL_Language
 	 *
 	 * @phpstan-param array{post:WP_Term, term:WP_Term}&array<string, WP_Term> $terms
@@ -50,9 +50,9 @@ class PLL_Language_Factory {
 	public static function get_from_terms( array $terms ) {
 		$languages = self::get_languages();
 		$data      = array(
-			'name'       => $terms['post']->name,
-			'slug'       => $terms['post']->slug,
-			'term_group' => $terms['post']->term_group,
+			'name'       => $terms['language']->name,
+			'slug'       => $terms['language']->slug,
+			'term_group' => $terms['language']->term_group,
 			'term_props' => array(),
 			'mo_id'      => PLL_MO::get_id_from_term_id( $terms['post']->term_id ),
 		);
@@ -66,7 +66,7 @@ class PLL_Language_Factory {
 		}
 
 		// The description field can contain any property.
-		$description = maybe_unserialize( $terms['post']->description );
+		$description = maybe_unserialize( $terms['language']->description );
 
 		if ( is_array( $description ) ) {
 			$description = array_intersect_key(
--- a/include/model.php
+++ b/include/model.php
@@ -703,12 +703,6 @@ class PLL_Model {
 	 */
 	protected function get_languages_from_taxonomies() {
 
-		$taxonomies = array();
-
-		foreach ( $this->translatable_objects as $type => $sub_model ) {
-			$taxonomies[ $sub_model->get_tax_language() ] = $type;
-		}
-
 		// To get language taxonomies first.
 		$reversed_terms = array_reverse( $this->get_language_terms() );
 
@@ -717,7 +711,7 @@ class PLL_Model {
 		foreach ( $reversed_terms as $term ) {
 			if ( $term instanceof WP_Term ) {
 				$key = 'language' === $term->taxonomy ? $term->slug : substr( $term->slug, 4 );
-				$terms_by_slug[ $key ][ $taxonomies[ $term->taxonomy ] ] = $term;
+				$terms_by_slug[ $key ][ $term->taxonomy ] = $term;
 			}
 		}
 
--- a/include/model.php
+++ b/include/model.php
@@ -710,6 +710,7 @@ class PLL_Model {
 
 		foreach ( $reversed_terms as $term ) {
 			if ( $term instanceof WP_Term ) {
+				// Except for language taxonomy term slugs, remove 'pll_' prefix from the other language taxonomy term slugs.
 				$key = 'language' === $term->taxonomy ? $term->slug : substr( $term->slug, 4 );
 				$terms_by_slug[ $key ][ $term->taxonomy ] = $term;
 			}
--- a/include/model.php
+++ b/include/model.php
@@ -703,7 +703,11 @@ class PLL_Model {
 	 */
 	protected function get_languages_from_taxonomies() {
 
-		// To get language taxonomies first.
+		/*
+		 * Only terms of the taxonomy 'language' include a 'term_group' for the order.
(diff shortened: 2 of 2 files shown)


include/language-factory.php
```diff
@@ -41,18 +41,18 @@ public static function get( $language_data ) {
 	 *
 	 * @since 3.4
 	 *
-	 * @param WP_Term[] $terms List of language terms, with the type as array keys.
-	 *                         `post` and `term` are mandatory keys.
+	 * @param WP_Term[] $terms List of language terms, with the language taxonomies language as array keys.
```
Hug0-Drelon, line 44:
```suggestion
	 * @param WP_Term[] $terms List of language terms, with language taxonomy names as array keys.
```
Just suggest a more descriptive sentence.

include/language-factory.php
```diff
@@ -41,18 +41,18 @@ public static function get( $language_data ) {
...
+	 *                         `language` and `term_language` are mandatory keys.
 	 * @return PLL_Language
 	 *
 	 * @phpstan-param array{post:WP_Term, term:WP_Term}&array<string, WP_Term> $terms
```
Hug0-Drelon, line 48:
```suggestion
	 * @phpstan-param array{language:WP_Term, term_language:WP_Term}&array<string, WP_Term> $terms
```

include/model.php
```diff
@@ -702,34 +702,25 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
-		$languages = array();
+		$terms_by_slug = array();
 
-		foreach ( $terms_by_type['post'] as $term_slug => $lang_term ) {
-			if ( ! isset( $terms_by_type['term'][ "pll_{$term_slug}" ] ) ) {
-				// A corresponding `term` language is mandatory.
-				continue;
+		foreach ( $reversed_terms as $term ) {
+			if ( $term instanceof WP_Term ) {
+				$key = 'language' === $term->taxonomy ? $term->slug : substr( $term->slug, 4 );
```
Hug0-Drelon, line 713: Today I see why using `substr()`, but not sure I'll still in the future. Could you add a comment on why?

include/model.php
```diff
@@ -702,34 +702,25 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
 	 */
 	protected function get_languages_from_taxonomies() {
-		$terms_by_type = $this->get_language_terms_by_type();
 
-		if ( empty( $terms_by_type['post'] ) || empty( $terms_by_type['term'] ) ) {
-			// `post` and `term` languages are mandatory.
-			return array();
-		}
+		// To get language taxonomies first.
+		$reversed_terms = array_reverse( $this->get_language_terms() );
```
Hug0-Drelon, line 707: Is it really necessary? Since terms are ordered with keys, not sure we have to do it.

Hug0-Drelon: Also, they are ordered by `term_group` which already order language right?

include/language-factory.php
manooweb: Done in cbe5d977df3438ec872accd81e82eb4604e390cd

include/language-factory.php
manooweb: Done in 89fa58a0d18c2fd1480a57bdfb7d3ab477932abb

include/model.php
manooweb: All the language taxonomy slugs except the main one (language) are prefixed with `pll_`

See in 3.3 release https://github.com/polylang/polylang/blob/3.3.x/admin/admin-model.php#L47
and now in the future 3.4 release
https://github.com/polylang/polylang/blob/3.4-dev/admin/admin-model.php#L242

So to have the same key (language slug) between the language taxonomy and the other language taxonomies we need to remove the `pll_` prefix from their slug.

manooweb: Done in 2653e10cf16d0541ae4c6a66880ed433198155c9

include/model.php
manooweb: In most case, it's right because all `term_group` are set to `0`.
However it exists one case where it's wrong: when you order the languages with the `term_group` (`order` column  in the languages list UI). `term_language` taxonomy terms have always `term_group` set to `0`

Let's take an example with 2 languages (en and fr):

`language` taxonomy term `en` with `term_group` set to `2`
`language` taxonomy term `fr` with `term_group`  set to `1`
`term_language taxonomy term `pll_en` with `term_group` set to `0`
`term_language taxonomy term `pll_fr` with `term_group` set to `0`

So the terms request ordered by `term_group` returns:

pll_en (term_group = 0)
pll_fr (term_group = 0)
fr (term_group = 1)
en (term_group = 2)

At the end we can't ensure the languages order because we start the process with the English before the French as we expect.

`array_reverse` is a concise manner to put the language taxonomy terms at the beginning of the results.
As you put these terms in the reverse order compared to the `term_group` order, we need to also reverse the output array. Thus the language taxonmy terms are in the right order.

include/model.php
Hug0-Drelon: My future self thanks you 😁

include/model.php
Hug0-Drelon: OK I see now, thanks for the explanation.

Hug0-Drelon (approved):

include/model.php
```diff
@@ -702,34 +702,26 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+				$key = 'language' === $term->taxonomy ? $term->slug : substr( $term->slug, 4 );
+				$terms_by_slug[ $key ][ $term->taxonomy ] = $term;
 			}
+		}
 
-			$lang_terms = array();
-
-			foreach ( $terms_by_type as $type => $terms ) {
-				$key = 'post' === $type ? $term_slug : "pll_{$term_slug}";
+		// To put the language in the right order.
```
Chouby, line 719:
```suggestion
		// Restore the right order after the first `array_reverse()`.
```

include/model.php
```diff
@@ -702,34 +702,26 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
-				if ( isset( $terms[ $key ] ) ) {
-					$lang_terms[ $type ] = $terms[ $key ];
-				}
-			}
+		$languages = array();
 
-			if ( ! isset( $lang_terms['post'], $lang_terms['term'] ) ) {
-				continue;
-			}
+		foreach ( $terms_by_slug as $lang_terms ) {
```
Chouby, line 724: I suggest to simply the loop because the factory always returns an object.
```php
foreach ( $terms_by_slug as $lang_terms ) {
	$languages[] = PLL_Language_Factory::get_from_terms( $lang_terms );
}
```
```

File: include/model.php [hunk 7]
```diff
@@ -827,7 +775,13 @@ protected function get_language_terms_by_type() {
 	 */
 	protected function get_language_terms() {
 		add_filter( 'get_terms_orderby', array( $this, 'filter_language_terms_orderby' ), 10, 3 );
-		$post_languages = get_terms( array( 'taxonomy' => 'language', 'hide_empty' => false, 'orderby' => 'term_group' ) );
+		$post_languages = get_terms(
```
Chouby, line 778: I suggest ro `$post_languages` maybe to `$terms` as these are no more only terms associated to posts.

include/model.php
```diff
@@ -827,7 +775,13 @@ protected function get_language_terms_by_type() {
...
+			array(
+				'taxonomy'   => $this->translatable_objects->get_taxonomy_names( array( 'language' ) ),
+				'orderby' => 'term_group',
+				'hide_empty' => false,
```
Chouby, line 782: Alignment:
```suggestion
				'taxonomy'   => $this->translatable_objects->get_taxonomy_names( array( 'language' ) ),
				'orderby'    => 'term_group',
				'hide_empty' => false,
```

Chouby (changes requested):

include/model.php
```diff
@@ -702,34 +702,26 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
 	 * @phpstan-return list<PLL_Language>
 	 */
 	protected function get_languages_from_taxonomies() {
-		$terms_by_type = $this->get_language_terms_by_type();
 
-		if ( empty( $terms_by_type['post'] ) || empty( $terms_by_type['term'] ) ) {
-			// `post` and `term` languages are mandatory.
-			return array();
-		}
+		// To get language taxonomies first.
```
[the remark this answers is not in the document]
Hug0-Drelon: Yup! Much more understandable.

include/model.php
manooweb: Done in 6f3885e36e5950089a072d17901cf88f15275d96

include/model.php
manooweb: Done in 6f3885e36e5950089a072d17901cf88f15275d96

include/model.php
manooweb: Done in 6f3885e36e5950089a072d17901cf88f15275d96

include/model.php
manooweb: Done in 6f3885e36e5950089a072d17901cf88f15275d96

include/model.php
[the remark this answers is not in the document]
manooweb: Done in 6f3885e36e5950089a072d17901cf88f15275d96

PR #1176, opened 2022-12-19
Tools to add secondary language terms

Screenfeed:
This adds several methods to ease the creation of secondary language terms.
This is required for https://github.com/polylang/polylang-wc/pull/526.


manooweb (approved):

admin/admin-model.php
```diff
@@ -461,4 +466,94 @@ public function update_default_lang( $slug ) {
...
+	 *     slug: non-empty-string,
+	 *     name: non-empty-string,
+	 *     term_id?: int<0, max>
+	 * } $args
+	 * @phpstan-return array{
+	 *     term_id: positive-int,
+	 *     term_taxonomy_id: positive-int
+	 * }|null
+	 */
+	protected function update_secondary_language_term( array $args ) {
```
Chouby, line 537: Is there any reason why you don't include the `foreach ( $this->translatable_objects->get_secondary_translatable_objects() as $object )` in this method? 
As I see:
- The call is always made in such a loop
- You need to pass `$object` related things as parameters.
So I would rather include the loop in the method and get the taxonomy and term_id from here.

admin/admin-model.php
```diff
@@ -461,4 +466,94 @@ public function update_default_lang( $slug ) {
...
+			return null;
+		}
+
+		/**
+		 * @phpstan-var array{
+		 *     term_id: positive-int,
+		 *     term_taxonomy_id: positive-int
+		 * }
+		 */
+		return array_map( 'absint', $result );
```
Chouby, line 557: Do we need to return anything? I don't see it's used.

admin/admin-model.php
```diff
@@ -461,4 +466,94 @@ public function update_default_lang( $slug ) {
...
+				if ( ! empty( $result ) ) {
+					$updated = true;
+				}
+			}
+		}
+
+		if ( $updated ) {
+			// Clear the cache, so the new `term_id` and `term_taxonomy_id` appear in the languages list.
+			$this->clean_languages_cache();
+		}
```
Chouby, line 508: I wouldn't bother to check if seomthing's has been updated and always clean the cache. Just as we do in `add_language()` and `update_language()`.

include/translatable-objects.php
```diff
@@ -98,13 +98,14 @@ public function get_secondary_translatable_objects() {
...
-	 * @phpstan-param array<'language'|'translations'> $filter
+	 * @phpstan-param array<'language'|'translations'|'secondary'> $filter
 	 * @phpstan-return list<non-empty-string>
 	 */
 	public function get_taxonomy_names( $filter = array( 'language', 'translations' ) ) {
 		$taxonomies = array();
+		$objects    = in_array( 'secondary', $filter, true ) ? $this->get_secondary_translatable_objects() : $this->objects;
 
-		foreach ( $this->objects as $object ) {
+		foreach ( $objects as $object ) {
```
Chouby, line 108: Not sure we should keep this change. Apart that you don't use it anymore, the `secondary` filter is not used as the othesr.
- `array( 'language', 'translations' )` returns the union.
- `array( 'language', 'secondary' )` returns the intersection.

Chouby (changes requested):

include/translatable-objects.php
Screenfeed: Reverted in [0f1fa8e](https://github.com/polylang/polylang/pull/1176/commits/0f1fa8e99b47bc3a80609878fc32b054988b7728).

admin/admin-model.php
Screenfeed: Done in [a610c9d](https://github.com/polylang/polylang/pull/1176/commits/a610c9da49eded79f94df96cb94bc40e4b7177d7).

admin/admin-model.php
Screenfeed: It was needed but not anymore. Removed in [a610c9d](https://github.com/polylang/polylang/pull/1176/commits/a610c9da49eded79f94df96cb94bc40e4b7177d7).

admin/admin-model.php
Screenfeed: Done in [a610c9d](https://github.com/polylang/polylang/pull/1176/commits/a610c9da49eded79f94df96cb94bc40e4b7177d7).

include/model.php
```diff
@@ -695,6 +695,54 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+
+		foreach ( $this->translatable_objects->get_secondary_translatable_objects() as $object ) {
+			if ( ! empty( $language ) ) {
+				$term_id = $language->get_tax_prop( $object->get_tax_language(), 'term_id' );
+			} else {
+				$term_id = 0;
+			}
+
+			if ( ! empty( $term_id ) ) {
+				wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
```
Chouby, line 738: I thought that WP was comparing old and new terms before the update, but it seems that's not the case so maybe we should do it ourselves to avoid useless code execution (this part of the code may be called more frequently now than before). 
```suggestion
				if ( $language->slug !== $slug || $language->name !== $name ) {
					wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
				}
```

admin/admin-model.php
```diff
@@ -43,11 +43,9 @@ public function add_language( $args ) {
 		wp_update_term( (int) $r['term_id'], 'language', array( 'term_group' => (int) $args['term_group'] ) ); // can't set the term group directly in wp_insert_term
 
 		// The other language taxonomies.
-		foreach ( $this->translatable_objects->get_secondary_translatable_objects() as $object ) {
-			wp_insert_term( $args['name'], $object->get_tax_language(), array( 'slug' => 'pll_' . $args['slug'] ) );
-		}
+		$this->update_secondary_language_terms( $args['slug'], $args['name'] );
 
-		$this->clean_languages_cache(); // Update the languages list now !
+		$this->clean_languages_cache(); // Update the languages list now!
```
Chouby, line 48: I thought we had already removed this line. But it looks like it's not the case.

It should not be needed anymaore. It was previously useful to access incomplete `PLL_Language` object in `PLL_MO` and in action `pll_add_language` but we want to prevent this now.

Chouby (changes requested):

include/model.php
```diff
@@ -695,6 +695,54 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+	 */
+	protected function update_secondary_language_terms( $slug, $name, PLL_Language $language = null ) {
+		$slug = 0 === strpos( $slug, 'pll_' ) ? $slug : "pll_$slug";
+
+		foreach ( $this->translatable_objects->get_secondary_translatable_objects() as $object ) {
+			if ( ! empty( $language ) ) {
+				$term_id = $language->get_tax_prop( $object->get_tax_language(), 'term_id' );
+			} else {
+				$term_id = 0;
+			}
```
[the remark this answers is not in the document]
Screenfeed: It is needed, otherwise we could get the `$term_id` of the previous `foreach` lap (we're in a `foreach` loop).

admin/admin-model.php
Screenfeed: Removed in [b055000](https://github.com/polylang/polylang/pull/1176/commits/b055000b93861cacd840af1ef29c28fe0bcf46e2).

include/model.php
Screenfeed: Done in [e21f962](https://github.com/polylang/polylang/pull/1176/commits/e21f962a5b5fd4ef37e2208242f65f8356da728c).
I tweaked a bit your suggestion to prevent adding another floor to the pyramid.

PR #1178, opened 2022-12-19
Improve flag properties setting

Hug0-Drelon:
In order to remove `PLL_Language::set_flag()`, we chose to set flag props from within the `PLL_Language_Factory`.

With no extra code, there is still an issue. Indeed, an infinite loop can occur because when creating the languages objects from terms, `PLL_Language::get_flag_informations()` calls `plugin_url()` which is filtered by `PLL_Links_Abstract_Domain::site_url()`. This very last method is also trying to create the languages list, leading to an new call to `PLL_Language::get_flag_informations()` then `plugin_url()` then ...

To prevent it, we propose to add a new property to `PLL_Model` telling whether or not languages objects are properly created. This can be used in `PLL_Links_Abstract_Domain::site_url()`, if objects are not created yet, we don't filter the URL. With that in mind, we have to clarify that `PLL_Language::$flag_url` is storing the flag URL only on the main domain now.

Another infinite loop occurs in tests due to `PLL_Links_Subdomain::get_hosts()` which requires the languages list (leading to the same issue than before...). This was because a language was added before running tests without generating the languages list again... To fix it we a calling `PLL_Model::get_languages_list()` after a language is added.


include/language-factory.php
```diff
@@ -144,4 +147,106 @@ private static function get_languages() {
...
+
+		foreach ( $directories as $dir ) {
+			if ( file_exists( $file = "{$dir}/{$locale}.png" ) || file_exists( $file = "{$dir}/{$locale}.jpg" ) || file_exists( $file = "{$dir}/{$locale}.svg" ) ) {
+				$flags['custom_flag']['url'] = content_url( '/' . str_replace( WP_CONTENT_DIR, '', $file ) );
+				break;
+			}
+		}
+
+		/**
+		 * Filters the custom flag informations.
```
Marianne380, line 194:
```suggestion
		 * Filters the custom flag information.
```

include/model.php
```diff
@@ -141,8 +150,10 @@ public function get_languages_list( $args = array() ) {
 			 */
 			$languages = apply_filters( 'pll_after_languages_cache', $languages );
 			$this->cache->set( 'languages', $languages );
+			$this->are_languages_created = true;
 		}
 
+
```
Marianne380, line 156: Empty line
```suggestion
```

include/language.php
```diff
@@ -628,7 +546,17 @@ public function get_display_flag() {
...
+
+		/**
+		 * Filters flag URL. The custom one is passed if exists.
+		 *
+		 * @since 3.4
+		 *
+		 * @param string       $flag_url Flag URL.
+		 * @param PLL_Language $language Current language object.
+		 */
+		return apply_filters( 'pll_get_display_flag_url', $flag_url, $this );
```
Marianne380, line 559: Why do we need the language object in this filter ?

Hug0-Drelon: Today `pll_get_display_flag_url` is only filtered by `site_url()`. But I added `$this` for our future self and for our users. Do you think it'd be better without it?

include/language-factory.php
Hug0-Drelon: "information" is uncountable IMHO.

Marianne380: Isn't that why we shouldn't add an "s"?

Hug0-Drelon: Yup, for instance, we have to use "a piece of information" to about a specific one.

Hug0-Drelon: Sorry, didn't read correctly...

Marianne380 (approved):

include/links-abstract-domain.php
```diff
@@ -67,8 +68,15 @@ protected function set_home_url( $language ) {
 	 * @return string The modified url.
 	 */
 	public function site_url( $url ) {
+		if ( ! $this->model->are_languages_created() ) {
```
Chouby, line 71: In order to simplify, could we use `PLL_Model::has_languages()` instead of introducing this new method?

admin/admin-model.php
```diff
@@ -70,6 +70,8 @@ public function add_language( $args ) {
 
 		$this->clean_languages_cache(); // Again to set add mo_id in the cached languages list
 		flush_rewrite_rules(); // Refresh rewrite rules
+		$this->get_languages_list(); // Refresh languages.
```
Chouby, line 73:
```suggestion
		// Refresh languages.
		$this->clean_languages_cache();
		$this->get_languages_list();
		
		flush_rewrite_rules(); // Refresh rewrite rules.
```
- I propose to change the order of the calls as it makes more sense to deal withthe languages all at conce.
- Removed the `Again...| comment. It seems to me that now the first call to `clean_languages_cache()` higher in the method could be removed.
- Maybe we should do the same for `update_language()` for consistency.

include/language.php
Chouby, line 559: Maybe a way to avoid a new filter :
```php
$url = site_url( set_url_scheme( $url, 'relative' ) );
```
with a comment to explain what we are doing.

include/language-factory.php
```diff
@@ -144,4 +147,106 @@ private static function get_languages() {
 
 		return self::$languages;
 	}
+
+
+	/**
+	 * Creates flag_url and flag language properties. Also takes care of custom flag.
+	 *
+	 * @since 3.4
```
Chouby, line 155: I suggest to keep `@since 1.2` with a changelog for `3.4`.

Chouby (changes requested):

admin/admin-model.php
Hug0-Drelon: Done in b173e86.
But I didn't remove the previous call to `clean_languages_cache()` in `add_language()` because otherwise the later `$this->get_language( $args['slug'] )` returns `false`.

include/language.php
Hug0-Drelon: Good idea! Done in 1ea4829.

include/links-abstract-domain.php
Hug0-Drelon: It appears that `has_languages()` doesn't do the job. Indeed it check also for languages term in the database. Whereas `are_languages_created()` tells us when language objects are created and that's what we want to know before filtering URLs.
I propose a better naming in ff7c22e.

include/language.php
```diff
@@ -628,7 +546,12 @@ public function get_display_flag() {
 	 * @return string
 	 */
 	public function get_display_flag_url() {
-		return empty( $this->custom_flag_url ) ? $this->flag_url : $this->custom_flag_url;
+		$flag_url = empty( $this->custom_flag_url ) ? $this->flag_url : $this->custom_flag_url;
+
+		/**
+		 * Let's use `site_url()` so the returned URL will be filtered properly according to permalink structure.
```
Chouby, line 552:
```suggestion
		 * Let's use `site_url()` so the returned URL will be filtered properly according to the current domain.
```

include/links-abstract-domain.php
```diff
@@ -25,6 +25,7 @@ public function __construct( &$model ) {
 		add_filter( 'theme_root_uri', array( $this, 'site_url' ) ); // The above filter is not sufficient with WPMU Domain Mapping.
 		add_filter( 'plugins_url', array( $this, 'site_url' ) );
 		add_filter( 'rest_url', array( $this, 'site_url' ) );
+		add_filter( 'pll_get_display_flag_url', array( $this, 'site_url' ) );
```
Chouby, line 28: Not existing anymore.

include/links-abstract-domain.php
```diff
@@ -67,8 +68,15 @@ protected function set_home_url( $language ) {
 	 * @return string The modified url.
 	 */
 	public function site_url( $url ) {
+		if ( ! $this->model->are_language_objects_created() ) {
```
Chouby, line 71: Another attempt ;)
Would `if ( false === $this->cache->get( 'languages' ) )` work?

Chouby (changes requested):

include/links-abstract-domain.php
Hug0-Drelon: As we saw, sometimes the cache is cleaned before passing through it...

include/model.php
```diff
@@ -46,6 +46,13 @@ class PLL_Model {
 	 */
 	public $term;
 
+	/**
+	 * Whether or not the languages objects have been created.
+	 *
+	 * @var bool
+	 */
+	private $are_language_objects_created = false;
+
```
Chouby, line 55:
```suggestion
	/**
	 * Flag set to true during the language objects creation.
	 *
	 * @var bool
	 */
	private $is_creating_language_objects = false;

```

include/model.php
```diff
@@ -111,6 +118,8 @@ public function get_languages_list( $args = array() ) {
 		$languages = $this->cache->get( 'languages' );
 
 		if ( ! is_array( $languages ) ) {
+			$this->are_language_objects_created = false;
```
Chouby, line 121:
```suggestion
			$this->is_creating_language_objects = true;
```

include/model.php
```diff
@@ -141,6 +150,7 @@ public function get_languages_list( $args = array() ) {
 			 */
 			$languages = apply_filters( 'pll_after_languages_cache', $languages );
 			$this->cache->set( 'languages', $languages );
+			$this->are_language_objects_created = true;
```
Chouby, line 153:
```suggestion
			$this->is_creating_language_objects = false;
```

include/model.php
```diff
@@ -786,4 +795,15 @@ protected function get_language_terms() {
...
+	/**
+	 * Tells if language objects are currently created.
+	 *
+	 * @since 3.4
+	 *
+	 * @return bool Whether language objects are created or not.
+	 */
+	public function is_creating_language_objects() {
+		return ! $this->are_language_objects_created;
+	}
```
Chouby, line 808:
```suggestion
	/**
	 * Tells if language objects are currently being created.
	 *
	 * @since 3.4
	 *
	 * @return bool
	 */
	public function is_creating_language_objects() {
		return $this->is_creating_language_objects;
	}
```

PR #1182, opened 2022-12-23
Adds an object type parameter to get_objects_with_no_lang()

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/1554

- Adds an object type parameter to `PLL_Translatable_Object::get_objects_with_no_lang()`.

- Adds this same parameter to `PLL_Translatable_Object::get_objects_with_no_lang_sql()`. This one won't be used, see the PHPStan exclusion.

- Adds this same parameter to `PLL_Translatable_Object_With_Types_Trait::get_objects_with_no_lang_sql()`.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -614,7 +614,7 @@ class PLL_Model {
 		foreach ( $this->translatable_objects as $type => $object ) {
 			// The trailing 's' in the array key is for backward compatibility.
 			if ( $object instanceof PLL_Translatable_Object_With_Types_Interface ) {
-				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit, array( 'type' => $object->get_translated_object_types() ) );
+				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit, array( 'types' => $object->get_translated_object_types() ) );
 			} else {
 				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit );
 			}
@@ -646,7 +646,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit, array( 'type' => $post_types ) );
+		return $object->get_objects_with_no_lang( $limit, array( 'types' => $post_types ) );
 	}
 
 	/**
@@ -665,7 +665,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit, array( 'type' => $taxonomies ) );
+		return $object->get_objects_with_no_lang( $limit, array( 'types' => $taxonomies ) );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -614,7 +614,7 @@ class PLL_Model {
 		foreach ( $this->translatable_objects as $type => $object ) {
 			// The trailing 's' in the array key is for backward compatibility.
 			if ( $object instanceof PLL_Translatable_Object_With_Types_Interface ) {
-				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit, array( 'types' => $object->get_translated_object_types() ) );
+				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit, $object->get_translated_object_types() );
 			} else {
 				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit );
 			}
@@ -646,7 +646,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit, array( 'types' => $post_types ) );
+		return $object->get_objects_with_no_lang( $limit, is_array( $post_types ) ? $post_types : array( $post_types ) );
 	}
 
 	/**
@@ -665,7 +665,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit, array( 'types' => $taxonomies ) );
+		return $object->get_objects_with_no_lang( $limit, is_array( $taxonomies ) ? $taxonomies : array( $taxonomies ) );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -642,7 +642,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $object->get_translated_object_types(), $limit );
+		return $object->get_objects_with_no_lang( $post_types, $limit );
 	}
 
 	/**
@@ -661,7 +661,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $object->get_translated_object_types(), $limit );
+		return $object->get_objects_with_no_lang( $taxonomies, $limit );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -665,7 +665,7 @@ class PLL_Model {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit, $taxonomies );
+		return $object->get_objects_with_no_lang( $limit, array( 'type' => $taxonomies ) );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -613,7 +613,9 @@ class PLL_Model {
 
 		foreach ( $this->translatable_objects as $type => $object ) {
 			// The trailing 's' in the array key is for backward compatibility.
-			$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $object->get_translated_object_types(), $limit );
+			if ( $object instanceof PLL_Translatable_Object_With_Types_Interface ) {
+				$objects[ "{$type}s" ] = $object->get_objects_with_no_lang( $limit, array( 'type' => $object->get_translated_object_types() ) );
+			}
 		}
 
 		/**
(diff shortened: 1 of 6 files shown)


Screenfeed (approved):

include/model.php
```diff
@@ -642,7 +644,7 @@ public function get_posts_with_no_lang( $post_types, $limit ) {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit );
+		return $object->get_objects_with_no_lang( $limit, $post_types );
```
Screenfeed, line 647: Woopsy 🤭
```suggestion
		return $object->get_objects_with_no_lang( $limit, array( 'type' => $post_types ) );
```

include/model.php
```diff
@@ -661,7 +663,7 @@ public function get_terms_with_no_lang( $taxonomies, $limit ) {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit );
+		return $object->get_objects_with_no_lang( $limit, $taxonomies );
```
Screenfeed, line 666: Same as above.
```suggestion
		return $object->get_objects_with_no_lang( $limit, array( 'type' => $taxonomies ) );
```

include/translatable-object-with-types-trait.php
```diff
@@ -34,15 +34,19 @@ trait PLL_Translatable_Object_With_Types_Trait {
 	 *
 	 * @param int[] $language_ids List of language `term_taxonomy_id`.
 	 * @param int   $limit        Max number of objects to return. `-1` to return all of them.
+	 * @param array $args {
+	 * 		The object args.
+	 *
+	 * 		@type string|string[] $object_types A translated object type or an array of translated object types.
```
Screenfeed, line 40: Typo I think.
```suggestion
	 * 		@type string|string[] $type A translated object type or an array of translated object types.
```

include/translatable-object-with-types-trait.php
```diff
@@ -34,15 +34,19 @@ trait PLL_Translatable_Object_With_Types_Trait {
...
+	 * }
 	 * @return string
 	 *
 	 * @phpstan-param array<positive-int> $language_ids
 	 * @phpstan-param -1|positive-int $limit
+	 * @phpstan-param array<array<string, string>> $args
```
Screenfeed, line 46: We can be more precise, to force the `type` key:
```suggestion
	 * @phpstan-param array{
	 *     type?: string|array<string>
	 * } $args
```
I'd like to use `non-empty-string` here instead of `string` but it would require to improve the documentation of `PLL_Model::get_posts_with_no_lang()`, `PLL_Model::get_terms_with_no_lang()`, and `PLL_Model::get_translated_post_types()`. So, maybe another time.

include/translatable-object-with-types-trait.php
```diff
@@ -34,15 +34,19 @@ trait PLL_Translatable_Object_With_Types_Trait {
...
 	 */
-	protected function get_objects_with_no_lang_sql( $language_ids, $limit ) {
-		$object_types = $this->get_translated_object_types();
-
-		if ( empty( $object_types ) ) {
+	protected function get_objects_with_no_lang_sql( $language_ids, $limit, $args = array() ) {
+		if ( empty( $args ) || ! isset( $args['type'] ) ) {
```
Screenfeed, line 49: A bit shorter:
```suggestion
		if ( empty( $args['type'] ) ) {
```

include/translatable-object.php
```diff
@@ -340,17 +340,24 @@ public function where_clause( $lang ) {
 	 *
 	 * @since 3.4
 	 *
-	 * @param int $limit Max number of objects to return. `-1` to return all of them.
+	 * @param int   $limit  Max number of objects to return. `-1` to return all of them.
+	 * @param array $args {
+	 * 		The object args.
+	 *
+	 * 		@type string|string[] $object_types A translated object type or an array of translated object types.
```
Screenfeed, line 347: Same typo here.
```suggestion
	 * 		@type string|string[] $type A translated object type or an array of translated object types.
```

include/translatable-object.php
```diff
@@ -340,17 +340,24 @@ public function where_clause( $lang ) {
...
+	 * }
 	 * @return int[] Array of object IDs.
 	 *
 	 * @phpstan-param -1|positive-int $limit
+	 * @phpstan-param array<array<string, string>> $args
```
Screenfeed, line 352:
```suggestion
	 * @phpstan-param array{
	 *     type?: string|array<string>
	 * } $args
```

include/translatable-object-with-types-trait.php
```diff
@@ -34,15 +34,19 @@ trait PLL_Translatable_Object_With_Types_Trait {
...
 	 *
 	 * @phpstan-param array<positive-int> $language_ids
 	 * @phpstan-param -1|positive-int $limit
+	 * @phpstan-param array<array<string, string>> $args
 	 */
-	protected function get_objects_with_no_lang_sql( $language_ids, $limit ) {
-		$object_types = $this->get_translated_object_types();
-
-		if ( empty( $object_types ) ) {
+	protected function get_objects_with_no_lang_sql( $language_ids, $limit, $args = array() ) {
```
Screenfeed, line 48: Let's add some type hints:
```suggestion
	protected function get_objects_with_no_lang_sql( array $language_ids, $limit, array $args = array() ) {
```

include/translatable-object.php
```diff
@@ -340,17 +340,24 @@ public function where_clause( $lang ) {
...
 	 * @phpstan-return list<positive-int>
 	 */
-	public function get_objects_with_no_lang( $limit ) {
-		$language_ids = $this->model->get_languages_list();
+	public function get_objects_with_no_lang( $limit, $args = array() ) {
```
Screenfeed, line 355: And here too:
```suggestion
	public function get_objects_with_no_lang( $limit, array $args = array() ) {
```

include/translatable-object.php
```diff
@@ -340,17 +340,24 @@ public function where_clause( $lang ) {
...
+		$languages = $this->model->get_languages_list();
 
-		foreach ( $language_ids as $i => $language ) {
-			$language_ids[ $i ] = $language->get_tax_prop( $this->get_tax_language(), 'term_taxonomy_id' );
+		$language_ids = array();
+		foreach ( $languages as  $language ) {
```
Screenfeed, line 359: If I can see it, PHPCS will 😅
```suggestion
		foreach ( $languages as $language ) {
```

include/translatable-object.php
```diff
@@ -422,12 +429,14 @@ public function sanitize_int_ids_list( $ids ) {
 	 *
 	 * @param int[] $language_ids List of language `term_taxonomy_id`.
 	 * @param int   $limit        Max number of objects to return. `-1` to return all of them.
+	 * @param array $args         The object args.
 	 * @return string
 	 *
 	 * @phpstan-param array<positive-int> $language_ids
 	 * @phpstan-param -1|positive-int $limit
+	 * @phpstan-param array<empty> $args
```
Screenfeed, line 437: Good catch, it will help us to spot code that adds arguments that shouldn't be there 👍

Screenfeed (changes requested):

Screenfeed:
By the way, what do you think about using `types` instead of `type` for the key in `$args`? We already use the plural form in `PLL_Translatable_Object_With_Types_Interface` and `get_translated_object_types()` for example.

include/translatable-object-with-types-trait.php
Screenfeed: We still need this one. I didn't meant to say we should remove the tag, just to forget `non-empty-string` for now.

Screenfeed (changes requested):

Screenfeed (approved):

Screenfeed (approved):

include/model.php
```diff
@@ -642,7 +646,7 @@ public function get_posts_with_no_lang( $post_types, $limit ) {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit );
+		return $object->get_objects_with_no_lang( $limit, is_array( $post_types ) ? $post_types : array( $post_types ) );
```
Screenfeed, line 649: Simpler:
```suggestion
		return $object->get_objects_with_no_lang( $limit, (array) $post_types );
```

include/model.php
```diff
@@ -661,7 +665,7 @@ public function get_terms_with_no_lang( $taxonomies, $limit ) {
 			return array();
 		}
 
-		return $object->get_objects_with_no_lang( $limit );
+		return $object->get_objects_with_no_lang( $limit, is_array( $taxonomies ) ? $taxonomies : array( $taxonomies ) );
```
Screenfeed, line 668: Simpler:
```suggestion
		return $object->get_objects_with_no_lang( $limit, (array) $taxonomies );
```

Screenfeed (changes requested):

Screenfeed (approved):

Chouby (approved):

Chouby:
I'll merge this PR as is. However there is an issue (not only linked to this PR, so I propose fix it in a separate one.
We use `"{$this->type}s"` for the cache key. This currently works by luck and we can't be sure that it will in other cases that we currently know. I suggest to change this. One idea could be to use the cache key "posts" and "tems" as `$type` if you believe that it can work. Otherwise we will have to create separate property.

PR #1183, opened 2022-12-23
Create language before test run

manooweb:
To correctly initialize polylang default language option which is missing during the test run.

After more investigations with @Hug0-Drelon, we found that the languages list could be an empty array in the Polylang cache at the opposite of the 3.3 version.
See https://github.com/polylang/polylang/blob/3.4-dev/include/model.php#L143

So as there is also a bug in `PLL_Model::has_languages()` method as reported by @Marianne380 in #1185, when Polylang is initialized with no languages, it is considered as a `PLL_Frontend` object instead of null. It's why the `YARPP_Test` was in error.

So, I reverted this PR commit and cherry-pick https://github.com/polylang/polylang/commit/5e2831bdbe73734df445f23547f554e42b06fe40 instead

## Going further
It's interesting to look at replacing `PLL_Model::get_languages_list()` calls in all modules loaders with `PLL_Model::has_languages()`

for example https://github.com/polylang/polylang/blob/3.3.x/modules/share-slug/load.php#L12

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -182,7 +182,7 @@ class PLL_Admin_Classic_Editor {
 			wp_die( 'You are not allowed to edit this post.' );
 		}
 
-		$this->model->post->update_language( $post_ID, $lang );
+		$this->model->post->set_language( $post_ID, $lang );
 
 		ob_start();
 		if ( 'attachment' === $post_type ) {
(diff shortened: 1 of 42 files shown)


Screenfeed (approved):

Screenfeed (approved):

Chouby:
Closing as obsolete.

PR #1185, opened 2022-12-29
Fix has_languages return

Marianne380:
`has_languages` could return `true` in cases without languages.

Check `! empty()` instead of `!== false` for `$this->cache->get( 'languages' )` since the return can be `mixed`.

I guess the same thing should be done for the transient.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -77,7 +77,7 @@ class PLL_Model {
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


Chouby:
There should be no issue with the transient since it's set only if at least one pair of language terms have been found: https://github.com/polylang/polylang/blob/3.3/include/model.php#L114-L141. But I expect your change to do no harm.

Chouby (approved):

PR #1187, opened 2022-12-30
Use custom query to search for untranslated posts.

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1551

## Issue
In some case, the desired posts doesn't show up in search. Especially when there is a lot of already translated posts (more than the limit queried, which is set to 20 by default) and the desired posts is returned at the end of the list.

## Description
To be sure to return the correct post, I propose to use a custom query. It's partially taken from the WordPress search query, with a join on `post_translations` taxonomy to look into the description (i.e. the translation table) and reject translated posts.

### DIFF
--- a/include/translated-post.php
+++ b/include/translated-post.php
@@ -333,53 +333,63 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 		 */
 		$limit = apply_filters( 'pll_ajax_posts_not_translated_limit', 20 );
 
-		$not_translated_query = $wpdb->prepare(
-			"LEFT JOIN {$wpdb->term_relationships} AS tr1 ON ({$wpdb->posts}.ID = tr1.object_id)
-			INNER JOIN {$wpdb->term_taxonomy} AS tt ON (tt.term_taxonomy_id = tr1.term_taxonomy_id)
-				AND (tt.taxonomy = 'post_translations')
-				AND tt.description NOT LIKE '%%%s%%'",
-			$untranslated_in->slug
-		);
+		$search_like = '%' . $wpdb->esc_like( $search ) . '%';
 
 		$search_query = $wpdb->prepare(
-			"INNER JOIN {$wpdb->term_relationships} AS tr2 ON ({$wpdb->posts}.ID = tr2.object_id)
-			AND tr2.term_taxonomy_id IN (%d)
-			AND (({$wpdb->posts}.post_title LIKE '%%%s%%')
-				OR ({$wpdb->posts}.post_excerpt LIKE '%%%s%%')
-				OR ({$wpdb->posts}.post_content LIKE '%%%s%%'))
+			"INNER JOIN {$wpdb->term_relationships} AS tr1 ON ({$wpdb->posts}.ID = tr1.object_id)
+			AND tr1.term_taxonomy_id IN (%d)
+			AND (({$wpdb->posts}.post_title LIKE %s)
+				OR ({$wpdb->posts}.post_excerpt LIKE %s)
+				OR ({$wpdb->posts}.post_content LIKE %s))
 			AND {$wpdb->posts}.post_type = %s
 			AND (({$wpdb->posts}.post_status <> 'trash'
 			AND {$wpdb->posts}.post_status <> 'auto-draft'))",
 			$lang->term_taxonomy_id,
-			$search,
-			$search,
-			$search,
+			$search_like,
+			$search_like,
+			$search_like,
 			$type
 		);
 
+		$untranslated_like = '%' . $wpdb->esc_like( $untranslated_in->slug ) . '%';
+
+		$not_translated_query = $wpdb->prepare(
+			"WHERE {$wpdb->posts}.ID NOT IN (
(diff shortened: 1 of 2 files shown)


include/translated-post.php
```diff
@@ -259,45 +259,76 @@ public function current_user_can_read( $post_id, $context = 'view' ) {
...
+		 * @since 3.4
 		 *
-		 * @since 1.7
-		 *
-		 * @param array $args WP_Query arguments
+		 * @param int $limit Limit to searched posts, default 20.
 		 */
-		$args  = apply_filters( 'pll_ajax_posts_not_translated_args', $args );
-		$posts = get_posts( $args );
+		$limit = apply_filters( 'pll_ajax_posts_not_translated_limit', 20 );
```
Chouby, line 272: I'd like to keep the current filter. We could allow to filter only 'number_posts' for now but would be able to add new parameters if asked by users.

include/translated-post.php
```diff
@@ -259,45 +259,76 @@ public function current_user_can_read( $post_id, $context = 'view' ) {
...
+				{$search_query}
+                {$not_translated_query}
+                GROUP BY {$wpdb->posts}.ID
+                ORDER BY {$wpdb->posts}.post_date DESC
+                {$limit_query}";
+
+		/**
+		 * Already prepared above.
+		 */
+		$posts = $wpdb->get_results( $sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
```
Chouby, line 320: I wonder if it would be absolutely unreadable to write all in only one operation to avoid this  `phpcs:ignore` as I don't see any special operation which could not gor through `prepare()`.

Chouby (changes requested):

Screenfeed (changes requested):
My review was pending for 2 months 😭

include/translated-post.php
Hug0-Drelon: I set it back in 4181a052.

include/translated-post.php
Hug0-Drelon: Done in 21f6e6f2, LGTM. What do you think?

include/translated-post.php
```diff
@@ -321,44 +321,79 @@ public function current_user_can_read( $id, $context = 'view' ) {
...
 			'post_type'        => $type,
 			'tax_query'        => array(
 				array(
 					'taxonomy' => $this->tax_language,
-					'field'    => 'term_taxonomy_id', // WP 3.5+.
+					'field'    => 'term_taxonomy_id', // WP 3.5+
 					'terms'    => $lang->get_tax_prop( $this->tax_language, 'term_taxonomy_id' ),
 				),
 			),
 		);
```
Chouby, line 341: You can remove all useless args and keep only `number_posts`.

include/translated-post.php
```diff
@@ -321,44 +321,79 @@ public function current_user_can_read( $id, $context = 'view' ) {
...
-
 		/**
-		 * Filter the query args when auto suggesting untranslated posts in the Languages metabox.
-		 * This should help plugins to fix some edge cases.
+		 * Filter the query args when auto suggesting untranslated posts in the Languages metabox
+		 * This should help plugins to fix some edge cases
 		 *
 		 * @see https://wordpress.org/support/topic/find-translated-post-when-10-is-not-enough
 		 *
```
Chouby, line 347: We can remove comments which are not relevant anymore.
```suggestion
		 * Filters the query args when auto suggesting untranslated posts in the Languages metabox.
		 *
```

include/translated-post.php
```diff
@@ -321,44 +321,79 @@ public function current_user_can_read( $id, $context = 'view' ) {
...
 		 * @since 1.7
+		 * @since 3.4 Handled arguments restricted to `numberposts` to limit queried posts.
+		 * @since 3.4 No `WP_Query` is made anymore, a custom one is used instead.
```
Chouby, line 350:
```suggestion
		 * @since 3.4 Handled arguments restricted to `numberposts` to limit queried posts.
		 *            No `WP_Query` is made anymore, a custom one is used instead.
```

Chouby (changes requested):

include/translated-post.php
```diff
@@ -321,44 +321,76 @@ public function current_user_can_read( $id, $context = 'view' ) {
...
+			'suppress_filters' => 0, // To make the post_fields filter work
+			'lang'             => 0, // Avoid admin language filter
+			'numberposts'      => 20, // Limit to 20 posts
 			'post_status'      => 'any',
 			'post_type'        => $type,
 			'tax_query'        => array(
 				array(
 					'taxonomy' => $this->tax_language,
-					'field'    => 'term_taxonomy_id', // WP 3.5+.
+					'field'    => 'term_taxonomy_id', // WP 3.5+
```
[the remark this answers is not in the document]
Hug0-Drelon: Oops, multiline editing drama 😅 
Fixed in 7e1b9393

Screenfeed (approved):

include/translated-post.php
Hug0-Drelon: Useless args removed in 5f1b6cf8.

PR #1191, opened 2023-01-05
Add "active" language property.

Hug0-Drelon:
In order to ease the "Activated Language" feature in Polylang Pro, we chose to set this property in `PLL_Language`.
`PLL_Language::$active` default value is `true`.

### DIFF
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -122,6 +122,8 @@ class PLL_Language_Factory {
 			$data[ $field ] = ! empty( $data[ $field ] ) ? absint( $data[ $field ] ) : 0;
 		}
 
+		$data['active'] = isset( $data['active'] ) ? $data['active'] : true;
+
 		/**
 		 * @var LanguageData
 		 */
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -71,7 +71,7 @@ class PLL_Language_Factory {
 		if ( is_array( $description ) ) {
 			$description = array_intersect_key(
 				$description,
-				array( 'locale' => null, 'rtl' => null, 'flag_code' => null )
+				array( 'locale' => null, 'rtl' => null, 'flag_code' => null, 'active' => null )
 			);
 
 			foreach ( $description as $prop => $value ) {
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -122,7 +122,7 @@ class PLL_Language_Factory {
 			$data[ $field ] = ! empty( $data[ $field ] ) ? absint( $data[ $field ] ) : 0;
 		}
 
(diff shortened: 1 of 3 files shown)


include/language.php
```diff
@@ -40,7 +40,8 @@
  *     custom_flag_url?: string,
  *     custom_flag?: string,
  *     page_on_front:positive-int,
- *     page_for_posts:positive-int
+ *     page_for_posts:positive-int,
+ *     active?: bool
```
manooweb, line 44: Should we consider active is optional if it is intialized to true?

include/language.php
```diff
@@ -272,6 +280,7 @@ class PLL_Language {
 	 *     @type string  $custom_flag     Optional. HTML markup of the custom flag if it exists.
 	 *     @type int     $page_on_front   ID of the page on front in this language.
 	 *     @type int     $page_for_posts  ID of the page for posts in this language.
+	 *     @type bool    $active          Whether or not the language is active. Default `true`.
```
Screenfeed, line 283:
```suggestion
	 *     @type bool    $active          Optional. Whether or not the language is active. Default `true`.
```

Screenfeed (changes requested):

include/language.php
Chouby: I would disagree. Since [you always set it in the factory](https://github.com/polylang/polylang/pull/1191/files/c9576921acff74f6cfd78018a87d06029b6e8dec#diff-b79b7331e30aa849724ec6eb266c88b82d91510c53ca648d39262207aa88a9b3R125), I would require it.

Chouby (changes requested):

include/language.php
Screenfeed: But [it has a default value](https://github.com/polylang/polylang/pull/1191/files/c9576921acff74f6cfd78018a87d06029b6e8dec#diff-15448452fcf2eda2e4d5c61d73ff632a908cfa35d6fb0ce7b0162654dd0c658eR213), so there is no reason to make it required.

Hug0-Drelon: 🍿

Hug0-Drelon: Made it mandatory in 143b26a

include/language-factory.php
```diff
@@ -122,6 +122,8 @@ private static function sanitize_data( array $data ) {
 			$data[ $field ] = ! empty( $data[ $field ] ) ? absint( $data[ $field ] ) : 0;
 		}
 
+		$data['active'] = isset( $data['active'] ) ? $data['active'] : true;
```
Chouby, line 125:
```suggestion
		$data['active'] = isset( $data['active'] ) ? (bool) $data['active'] : true;
```

Chouby (changes requested):

PR #1192, opened 2023-01-05
Prepare for locale fallbacks

Screenfeed:
This PR makes the `PLL_Language::$fallbacks` property official by declaring it (without saying its purpose). This allows Polylang Pro to use it for locale fallbacks. However, the value of this property is fetch from the `language` term description, if available.
Also, the `#[AllowDynamicProperties]` attribute has been removed.

Related: https://github.com/polylang/polylang-pro/pull/1573.

### DIFF
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -204,65 +204,49 @@ class PLL_Upgrade {
 	 * @return void
 	 */
 	protected function upgrade_3_4() {
-		global $wpdb;
-
 		delete_transient( 'pll_languages_list' );
 
-		if ( ! defined( 'POLYLANG_PRO' ) || ! POLYLANG_PRO ) {
-			return;
-		}
-
 		// Migrate locale fallbacks from term metas to language term description.
-		$metas = $wpdb->get_results(
-			"
-			SELECT m.meta_id, m.term_id, m.meta_value, tt.description
-			FROM {$wpdb->termmeta} AS m
-			LEFT JOIN {$wpdb->term_taxonomy} AS tt
-				ON tt.term_id = m.term_id
-			WHERE
-				tt.taxonomy = 'language'
-				AND m.meta_key = 'fallback'
-			"
-		);
-
-		if ( empty( $metas ) ) {
-			// Nothing to migrate.
-			return;
-		}
-
-		$meta_ids = array();
-		$migrated = array();
-
-		foreach ( $metas as $meta ) {
-			$meta_ids[] = (int) $meta->meta_id;
-
-			if ( isset( $migrated[ $meta->term_id ] ) ) {
-				// Duplicate? Should not happen.
-				continue;
-			}
-
-			$fallbacks = maybe_unserialize( $meta->meta_value );
-
-			if ( empty( $fallbacks ) || ! is_array( $fallbacks ) ) {
-				// No fallbacks to migrate. Should not happen.
-				continue;
-			}
-
-			$description = maybe_unserialize( $meta->description );
-			$description = is_array( $description ) ? $description : array();
-
-			$description['fallbacks']   = $fallbacks;
-			$migrated[ $meta->term_id ] = 1;
-
-			wp_update_term( (int) $meta->term_id, 'language', array( 'description' => maybe_serialize( $description ) ) );
-		}
-
-		// Delete term metas.
-		$wpdb->query(
-			sprintf(
-				"DELETE FROM {$wpdb->termmeta} WHERE meta_id IN (%s)",
-				PLL_Db_Tools::prepare_values_list( $meta_ids ) // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+		$terms = get_terms(
+			array(
+				'taxonomy'   => 'language',
+				'hide_empty' => false,
+				'meta_query' => array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+					array(
+						'key'         => 'fallback',
+						'compare_key' => 'EXISTS',
+					),
+				),
 			)
 		);
+
+		if ( ! is_array( $terms ) ) {
+			return;
+		}
+
+		foreach ( $terms as $term ) {
+			$fallbacks = get_term_meta( $term->term_id, 'fallback', true );
+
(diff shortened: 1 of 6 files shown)


manooweb (approved):
LGTM 👌

install/upgrade.php
```diff
@@ -204,6 +204,65 @@ protected function upgrade_2_7() {
 	 * @return void
 	 */
 	protected function upgrade_3_4() {
+		global $wpdb;
+
 		delete_transient( 'pll_languages_list' );
+
+		if ( ! defined( 'POLYLANG_PRO' ) || ! POLYLANG_PRO ) {
+			return;
+		}
```
Chouby, line 213: We don't even need this as we just won't get any meta. Also removing it would account for the (improbable) case that some one does Polylang Pro -> Polylang -> Polylang Pro and updates while using Polylang.

install/upgrade.php
```diff
@@ -204,6 +204,65 @@ protected function upgrade_2_7() {
...
+			"
+			SELECT m.meta_id, m.term_id, m.meta_value, tt.description
+			FROM {$wpdb->termmeta} AS m
+			LEFT JOIN {$wpdb->term_taxonomy} AS tt
+				ON tt.term_id = m.term_id
+			WHERE
+				tt.taxonomy = 'language'
+				AND m.meta_key = 'fallback'
+			"
+		);
```
Chouby, line 226: WHen possible, I prefer to avoid direct SQL queries. This is interesting for critical performance cases but is less readable. Even with 20 languages, I don't thinks the spee would be critical here.

What do you think of:
```php
$args = array(
	'taxonomy' => 'language',
	'hide_empty' => 'false',
);
	
$terms = get_terms( $args );

foreach ( $terms as $term ) {
	$meta = get_term_meta( $term->term_id, 'fallback', true );

	if ( empty( $meta ) ) {
		continue;
	}

	$description = maybe_unserialize( $term->description );
	$description['fallbacks'] = $meta; 
	wp_update_term( (int) $meta->term_id, 'language', array( 'description' => maybe_serialize( $description ) ) );

	delete_term_meta( $term->term_id, 'fallback' );
}
```

phpstan.neon.dist
```diff
@@ -61,3 +61,9 @@ parameters:
 			message: "#^Parameter \\#2 \\.\\.\\.\\$args of function array_merge expects array, array\\|false given\\.$#"
 			count: 1
 			path: include/translated-term.php
+
+		# Ignored because maybe_serialize() cannot return anything else than a string when feeded with an array.
+		-
+			message: "#^Parameter \\#3 \\$args of function wp_update_term expects array\\{alias_of\\?\\: string, description\\?\\: string, parent\\?\\: int, slug\\?\\: string\\}, array\\{description\\: mixed\\} given\\.$#"
+			count: 1
+			path: install/upgrade.php
```
Chouby, line 69: Maybe we can avodi ignoring this with something like:
```php
	// @var string
	$description = maybe_serialize( $description );
```

Chouby (changes requested):

install/upgrade.php
Screenfeed: Removed in [7eb63dd](https://github.com/polylang/polylang/pull/1192/commits/7eb63dddb12251bec3b3174a05286563bf774f77).

install/upgrade.php
Screenfeed: Done in [7eb63dd](https://github.com/polylang/polylang/pull/1192/commits/7eb63dddb12251bec3b3174a05286563bf774f77).
On second thought, `get_term_meta()` shouldn't add queries since the values are cached by `get_terms()` (arg `update_term_meta_cache`).
It still adds a lot of queries (29 for 3 languages, only 1 with fallbacks) but I don't know why, maybe something else happened at the same time.

phpstan.neon.dist
Screenfeed: Good idea. Done in [7eb63dd](https://github.com/polylang/polylang/pull/1192/commits/7eb63dddb12251bec3b3174a05286563bf774f77).

install/upgrade.php
Screenfeed: Ok, the bunch of queries comes from the things hooked to `wp_update_term()` and `delete_term_meta()`. They don't seem to slow things down by a lot though.

install/upgrade.php
```diff
@@ -205,5 +205,48 @@ protected function upgrade_2_7() {
...
+
+			$description = maybe_unserialize( $term->description );
+			$description = is_array( $description ) ? $description : array();
+
+			$description['fallbacks'] = $fallbacks;
+			/** @var string */
+			$description = maybe_serialize( $description );
+
+			wp_update_term( $term->term_id, 'language', array( 'description' => $description ) );
+			delete_term_meta( $term->term_id, 'fallback' );
```
Chouby, line 249: Unconditionnally seleting the meta first allows to simplify the code:

```suggestion
			delete_term_meta( $term->term_id, 'fallback' );
			
			if ( empty( $fallbacks ) || ! is_array( $fallbacks ) ) {
				// Empty or invalid value, should not happen.
				continue;
			}

			$description = maybe_unserialize( $term->description );
			$description = is_array( $description ) ? $description : array();

			$description['fallbacks'] = $fallbacks;
			/** @var string */
			$description = maybe_serialize( $description );

			wp_update_term( $term->term_id, 'language', array( 'description' => $description ) );
```

install/upgrade.php
```diff
@@ -205,5 +205,48 @@ protected function upgrade_2_7() {
...
+		// Migrate locale fallbacks from term metas to language term description.
+		$terms = get_terms(
+			array(
+				'taxonomy'   => 'language',
+				'hide_empty' => false,
+				'meta_query' => array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+					array(
+						'key'         => 'fallback',
+						'compare_key' => 'EXISTS',
+					),
```
Chouby, line 218: I thought to this too. I did not include it in my proposal because it doesn't simplify the code afterwards. But why not.

install/upgrade.php
Chouby: For sure I knew that your initial proposal was more performant.

PR #1193, opened 2023-01-05
Don't override 'active' language property on update

manooweb:
fixes https://github.com/polylang/polylang-pro/issues/1569

And as we decided to manage active language property in Polylang in the PR #1191

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -248,7 +248,7 @@ class PLL_Admin_Model extends PLL_Model {
 			}
 		}
 
-		$description = maybe_serialize( array( 'locale' => $args['locale'], 'rtl' => (int) $args['rtl'], 'flag_code' => empty( $args['flag'] ) ? '' : $args['flag'] ) );
+		$description = $this->update_language_description( $args );
 		wp_update_term( $lang->get_tax_prop( 'language', 'term_id' ), 'language', array( 'slug' => $slug, 'name' => $args['name'], 'description' => $description, 'term_group' => (int) $args['term_group'] ) );
 
 		/**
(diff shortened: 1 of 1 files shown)


admin/admin-model.php
```diff
@@ -228,7 +228,12 @@ public function update_language( $args ) {
 		update_option( 'polylang', $this->options );
 
 		// And finally update the language itself
-		$description = maybe_serialize( array( 'locale' => $args['locale'], 'rtl' => (int) $args['rtl'], 'flag_code' => empty( $args['flag'] ) ? '' : $args['flag'] ) );
+		$description = maybe_serialize(
+			array_merge(
+				array( 'active' => $lang->active ? true : $lang->active ),
+				array( 'locale' => $args['locale'], 'rtl' => (int) $args['rtl'], 'flag_code' => empty( $args['flag'] ) ? '' : $args['flag'] )
```
Hug0-Drelon, line 234: Since `locale` will became `locales` [here](https://github.com/polylang/polylang/pull/1192). Maybe we should make the change here right away.

Hug0-Drelon: Another way to avoid future modification (if one day another feature is added in description) is to retrieve the term and its description, merge the modification and only then update the term?

admin/admin-model.php
```diff
@@ -227,8 +227,15 @@ public function update_language( $args ) {
 
 		update_option( 'polylang', $this->options );
 
+		$old_description = maybe_unserialize( get_term( (int) $args['lang_id'] )->description );
```
Hug0-Drelon, line 230: Maybe we should check `get_term()` is returning a `WP_Term` object before processing it?
Otherwise PHPStan won't be happy.

Hug0-Drelon (approved):

Chouby (changes requested):
I suggest to move the term description evaluation in a separate method.
Alos please make this PR on the 3.4-dev branch to avoid merge conflicts.

manooweb:
> I suggest to move the term description evaluation in a separate method. Also please make this PR on the 3.4-dev branch to avoid merge conflicts.

The branch is overriden and the suggestion is applied in d85dc70f8e87f36799fb7f8a6c9bcba38445b10e

admin/admin-model.php
```diff
@@ -282,6 +282,40 @@ public function update_language( $args ) {
...
+		$language_term = get_term( (int) $args['lang_id'] );
+
+		$old_description = array();
+
+		if ( $language_term instanceof WP_Term ) {
+			$description = maybe_unserialize( $language_term->description );
+			if ( is_array( $description ) ) {
+				$old_description = $description;
+			}
+		}
```
Chouby, line 309: Proposing a simplified way to avoid an intermediate variable:
```suggestion
		if ( $language_term instanceof WP_Term ) {
			$old_description = maybe_unserialize( $language_term->description );
		}

		if ( empty( $old_description ) || ! is_array( $old_description ) ) {
				$old_description = array();
		}
```

manooweb: Done in 7095e98d027f964cd25d2007152f0b6589234335

Chouby (approved):

PR #1195, opened 2023-01-09
Remove language home and search urls setters.

Hug0-Drelon:
The purpose is to not set `PLL_language::$home_url` and `PLL_language::$search_url` once language object is created anymore.
To do this I propose to add a new filter when creating a language from terms called `pll_language_urls` which will allow `PLL_Links_Model::set_language_urls()` to add the properties before the object is created.
Also, I added getters for this props to ensure that when they are not cached they are still rendered dynamically.
Now `add_language_to_link()` and `home_url()` accepts a language slug as parameter since that's only what they need in the end. I had to do it because `PLL_Language` object isn't available when called.


include/links-model.php
```diff
@@ -147,64 +148,37 @@ public function get_hosts() {
...
-	 * @since 1.8
-	 *
-	 * @param PLL_Language $language The language object.
-	 * @return void
-	 */
-	protected function set_home_url( $language ) {
-		$search_url = $this->home_url( $language );
-		$home_url = empty( $language->page_on_front ) || $this->options['redirect_lang'] ? $search_url : $this->front_page_url( $language );
-		$language->set_home_url( $search_url, $home_url );
+		return $this->options['hide_default'] && $language == $this->options['default_lang'] ? $url : $this->add_language_to_link( $url, $language );
```
Marianne380, line 163: Opportunity to change `==` ?

include/links-permalinks.php
```diff
@@ -108,29 +108,39 @@ public function add_paged_to_link( $url, $page ) {
...
 	 * @return string The static front page url.
 	 */
-	public function front_page_url( $lang ) {
-		if ( $this->options['hide_default'] && $lang->slug == $this->options['default_lang'] ) {
+	public function front_page_url( $language ) {
+		if ( $language instanceof PLL_Language ) {
+			$language = $language->get_object_vars();
+		}
+
+		if ( $this->options['hide_default'] && $language['slug'] == $this->options['default_lang'] ) {
```
Marianne380, line 138: Same here ?

include/links-subdomain.php
```diff
@@ -34,14 +34,19 @@ public function __construct( &$model ) {
...
-	public function add_language_to_link( $url, $lang ) {
-		if ( ! empty( $lang ) && false === strpos( $url, '://' . $lang->slug . '.' ) ) {
-			$url = $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? $url : str_replace( $this->www, '://' . $lang->slug . '.', $url );
+	public function add_language_to_link( $url, $language ) {
+		if ( $language instanceof PLL_Language ) {
+			$language = $language->slug;
+		}
+
+		if ( ! empty( $language ) && false === strpos( $url, '://' . $language . '.' ) ) {
+			$url = $this->options['default_lang'] == $language && $this->options['hide_default'] ? $url : str_replace( $this->www, '://' . $language . '.', $url );
```
Marianne380, line 49: Same here ?

include/links-default.php
```diff
@@ -21,13 +21,18 @@ class PLL_Links_Default extends PLL_Links_Model {
...
 	 * @return string The modified url.
 	 */
-	public function add_language_to_link( $url, $lang ) {
-		return empty( $lang ) || ( $this->options['hide_default'] && $this->options['default_lang'] == $lang->slug ) ? $url : add_query_arg( 'lang', $lang->slug, $url );
+	public function add_language_to_link( $url, $language ) {
+		if ( $language instanceof PLL_Language ) {
+			$language = $language->slug;
+		}
+
+		return empty( $language ) || ( $this->options['hide_default'] && $this->options['default_lang'] == $language ) ? $url : add_query_arg( 'lang', $language, $url );
```
Marianne380, line 35: Same here ?

include/links-directory.php
```diff
@@ -132,13 +137,18 @@ public function get_language_from_url( $url = '' ) {
...
 	 */
-	public function home_url( $lang ) {
+	public function home_url( $language ) {
+		if ( $language instanceof PLL_Language ) {
+			$language = $language->slug;
+		}
+
 		$base = $this->options['rewrite'] ? '' : 'language/';
-		$slug = $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : '/' . $this->root . $base . $lang->slug;
+		$slug = $this->options['default_lang'] == $language && $this->options['hide_default'] ? '' : '/' . $this->root . $base . $language;
```
Marianne380, line 151: Here too ?

include/links-model.php
Hug0-Drelon: Indeed! Let's do it 👍

Hug0-Drelon: Done in a806cc9

include/links-permalinks.php
Hug0-Drelon: Done in a806cc9

include/links-default.php
Hug0-Drelon: Done in a806cc9

include/links-directory.php
Hug0-Drelon: Done in a806cc9

include/links-subdomain.php
Hug0-Drelon: Done in b80c68a

Marianne380 (approved):

include/language.php
```diff
@@ -345,6 +345,24 @@ public function __get( $property ) {
...
+						esc_html( get_class( $this ) ),
+						esc_html( $property ),
+					),
+					E_USER_DEPRECATED
+				);
+			}
+
+			$property_name = '_' . $property;
+
+			return $this->$property_name;
```
Screenfeed, line 363: Maybe we should use the getter functions instead of the properties directly.

include/links-model.php
```diff
@@ -147,64 +148,37 @@ public function get_hosts() {
...
-		return $languages;
-	}
-
-	/**
-	 * Sets the home urls when not cached.
-	 * Sets the home urls scheme.
-	 *
-	 * @since 1.8
+	 * @param array $urls Array of language home and search URLs.
+	 * @param array $language Language data.
```
Screenfeed, line 172:
```suggestion
	 * @param array $urls     Array of language home and search URLs.
	 * @param array $language Language data.
```

Screenfeed (changes requested):
There is one detail that we'll need to discuss, maybe with @Chouby, and I haven't really gave it a thought.
It's `PLL_Language::get_object_vars()` and filtered values like `$home_url` and `$search_url`. This method is used twice:
- To store the language's data into the transient,
- In the REST API in PLL Pro.

I'm not sure but maybe this method should return filtered values instead of raw values.

include/language-factory.php
```diff
@@ -98,6 +98,29 @@ public static function get_from_terms( array $terms ) {
...
+		 *
+		 *     @type string $_home_url   Language home URL.
+		 *     @type string $_search_url Language search URL.
+		 * }
+		 * @param array $data Language data.
+		 *
+		 * @phpstan-param array{_home_url: string, _search_url: string} $urls
+		 * @phpstan-param non-empty-array<string, mixed> $data
+		 */
+		$data = array_merge( $data, apply_filters( 'pll_language_urls', $urls, $data ) );
```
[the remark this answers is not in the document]
Hug0-Drelon: You're right!

include/language.php
```diff
@@ -123,7 +123,7 @@ class PLL_Language {
 	 *
 	 * @phpstan-var non-empty-string
 	 */
-	public $home_url;
+	private $_home_url; // phpcs:ignore PSR2.Classes.PropertyDeclaration.Underscore
```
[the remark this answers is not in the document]
Hug0-Drelon: I thought `__get()` wouldn't be called for a private property ... Again I should have read the doc 😅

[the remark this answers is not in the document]
Hug0-Drelon: Done in fc6bf6e.

include/language.php
Hug0-Drelon: Indeed, done in 50f0a4f.

include/language.php
```diff
@@ -639,14 +619,25 @@ public function get_locale( $filter = 'raw' ) {
...
-	 * @return array
+	 * @param bool $raw Whether or not properties should be raw. Default to `true`.
+	 *
+	 * @return array Array of language object properties.
+	 *
+	 * @phpstan-return LanguageData
 	 */
-	public function get_object_vars() {
-		return get_object_vars( $this );
+	public function get_object_vars( $raw = true ) {
```
Screenfeed, line 632: I suppose the raw values should be used for the transient and the filtered values everywhere else.
So I think we should set `$raw` to `false` by default.

Screenfeed (changes requested):

Screenfeed:
Maybe we should add `$home_url` and `$search_url` to `__isset()`.

include/language.php
Hug0-Drelon: Ouch! Thanks.

include/language.php
```diff
@@ -358,9 +358,7 @@ public function __get( $property ) {
 				);
 			}
 
-			$property_name = '_' . $property;
-
-			return $this->$property_name;
+			return $this->{'get_' . $property};
```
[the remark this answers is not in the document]
Hug0-Drelon: Ouch! Thanks 👍

include/language.php
Hug0-Drelon: Done in dfa485d.

include/language-factory.php
```diff
@@ -98,6 +98,30 @@ public static function get_from_terms( array $terms ) {
 		$flag_props = self::get_flag( $data['flag_code'], $data['name'], $data['slug'], $data['locale'] );
 		$data       = array_merge( $data, $flag_props );
 
+		$default = array(
+			'home_url'   => '',
+			'search_url' => '',
+		);
```
Chouby, line 104: Do we really need this or can we just pass `array()` to the filter?

include/language-factory.php
```diff
@@ -98,6 +98,30 @@ public static function get_from_terms( array $terms ) {
...
+		 *
+		 *     @type string $home_url   Language home URL.
+		 *     @type string $search_url Language search URL.
+		 * }
+		 * @param array $data Language data.
+		 *
+		 * @phpstan-param array{home_url: string, search_url: string} $default
+		 * @phpstan-param non-empty-array<string, mixed> $data
+		 */
+		$_default = apply_filters( 'pll_filterable_language_data', $default, $data );
```
Chouby, line 122: You can keep the same variable for the filtered parameter and returned value.
Suggesting `pll_language_data` or `pll_additional_language_data` as filter name.

include/language.php
```diff
@@ -345,6 +345,22 @@ public function __get( $property ) {
...
+						"Class property %1\$s::\$%2\$s is deprecated, use %1\$s::get_%2\$s() instead.\nError handler",
+						esc_html( get_class( $this ) ),
+						esc_html( $property ),
+					),
+					E_USER_DEPRECATED
+				);
+			}
+
+			return $this->{"get_{$property}"}();
+		}
```
Chouby, line 362: Could we avoid some duplicate code in the method?

include/links-model.php
```diff
@@ -147,64 +148,35 @@ public function get_hosts() {
...
+	 * @return array Language data with URLs addeddefault.
 	 */
-	public function pll_after_languages_cache( $languages ) {
-		foreach ( $languages as $language ) {
-			// Get the home urls when not cached.
-			if ( ( defined( 'PLL_CACHE_LANGUAGES' ) && ! PLL_CACHE_LANGUAGES ) || ( defined( 'PLL_CACHE_HOME_URL' ) && ! PLL_CACHE_HOME_URL ) ) {
-				$this->set_home_url( $language );
-			}
+	public function set_language_urls( $default, $language ) {
```
Chouby, line 175: I'd like to keep `home` in the method name. I suggest `set_home_urls()` or `set_language_home_urls()`.

Chouby (changes requested):

include/language-factory.php
Hug0-Drelon: This is used to restrict passed data with `array_intersect_key()` bellow.

include/links-model.php
Hug0-Drelon: Done in 42de22a.

include/language-factory.php
Hug0-Drelon: Not sure I can use the same variable name with `array_intersect_key()` 🤔.
Filter renamed in 42de22a.

include/language.php
Hug0-Drelon: Indeed, done in f5db54d.

Chouby (changes requested):

admin/admin-model.php
```diff
@@ -51,6 +51,8 @@ public function add_language( $args ) {
 			update_option( 'polylang', $this->options );
 		}
 
+		$this->clean_languages_cache(); // Update the languages list now !
+
```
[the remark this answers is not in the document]
Hug0-Drelon: Yes and it lead to new errors unfortunately. This is required because right after this line we try to access to the language object newly created.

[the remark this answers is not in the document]
Hug0-Drelon: That was removed [here](https://github.com/polylang/polylang/pull/1176/files#diff-fa9d68e2c9e1ef4d6d50166392a16a9280dc9f8e1ce77523ce09577ca59e1b66L50)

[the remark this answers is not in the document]
Hug0-Drelon: Do you want me to do it in this PR?

Hug0-Drelon:
> I would have done more and created a new method to trigger the error.
Done in 27283bf.

Hug0-Drelon:
> OK maybe we can merge this `array_intersect_key()` with the other and since this is some way the purpose of this method, do this at the end of `sanitize_data()`.  

@Chouby Do you want to move the filter inside `sanitize_data()` as well? Because if I only move `array_intersect_key()` into it, everyone could still a or change any of the property values.
Or we could use `array_intersect_key()` with a "default" array of *all* language property values.

admin/admin-model.php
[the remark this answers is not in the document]
Chouby: In a separate PR.

include/language-factory.php
Chouby: OK. Forget my last comment. I'd prefer to pass `array()` to the filter and use the array only to filter the output of `apply_filters()`.

PR #1202, opened 2023-01-13
Remove PLL_Language dynamic property in PLL_Walker_Dropdown

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1577

In order to get rid of the remaining dynamic property in PLL_Walker_Dropdown, I propose to create a new method to convert into `stdClass` any value before being passed to parent method `Walker::display_element()`.

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -669,6 +669,17 @@ class PLL_Language {
 	}
 
 	/**
+	 * Converts current `PLL_language` into a stdClass object. Mostly used to allow dynamic properties.
+	 *
+	 * @since 3.4
+	 *
+	 * @return stdClass Converted `PLL_Language` object.
+	 */
+	public function to_standard_class() {
+		return (object) $this->get_object_vars();
+	}
+
+	/**
 	 * Returns a predefined HTML flag.
 	 *
 	 * @since 3.4
--- a/include/language.php
+++ b/include/language.php
@@ -656,7 +656,7 @@ class PLL_Language {
 	 *
 	 * @phpstan-return LanguageData
 	 */
-	public function get_object_vars( $context = 'display' ) {
+	public function to_array( $context = 'display' ) {
 		$language = get_object_vars( $this );
 
 		if ( 'db' !== $context ) {
@@ -669,14 +669,14 @@ class PLL_Language {
 	}
 
 	/**
-	 * Converts current `PLL_language` into a stdClass object. Mostly used to allow dynamic properties.
+	 * Converts current `PLL_language` into a `stdClass` object. Mostly used to allow dynamic properties.
 	 *
 	 * @since 3.4
 	 *
 	 * @return stdClass Converted `PLL_Language` object.
 	 */
-	public function to_standard_class() {
-		return (object) $this->get_object_vars();
+	public function to_std_class() {
+		return (object) $this->to_array();
 	}
 
 	/**
--- a/include/switcher.php
+++ b/include/switcher.php
@@ -241,6 +241,11 @@ class PLL_Switcher {
 			return $elements;
 		}
 
+		// Cast each elements to stdClass.
+		foreach ( $elements as $i => $element ) {
+			$elements[ $i ] = (object) $element;
+		}
+
 		if ( $args['dropdown'] ) {
 			$args['name'] = 'lang_choice_' . $args['dropdown'];
 			$args['class'] = 'pll-switcher-select';
--- a/include/switcher.php
+++ b/include/switcher.php
@@ -252,12 +252,9 @@ class PLL_Switcher {
 		}
 
 		// Cast each elements to stdClass because $walker::walk() expects an array of objects.
-		$elements = array_map(
-			function( $element ) {
-				return (object) $element;
-			},
-			$elements
-		);
+		foreach ( $elements as $i => $element ) {
+			$elements[ $i ] = (object) $element;
+		}
 
 		/**
 		 * Filter the whole html markup returned by the 'pll_the_languages' template tag
--- a/include/switcher.php
+++ b/include/switcher.php
@@ -251,7 +251,7 @@ class PLL_Switcher {
 			$walker = new PLL_Walker_List();
 		}
 
-		// Cast each elements to stdClass because $walker::walk() expects an array of objects.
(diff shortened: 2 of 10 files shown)


include/walker-dropdown.php
```diff
@@ -19,19 +19,20 @@ class PLL_Walker_Dropdown extends Walker {
...
+	 * @param string $output            Passed by reference. Used to append additional content.
+	 * @param object $element           The data object.
+	 * @param int    $depth             Depth of the item.
+	 * @param array  $args              An array of additional arguments.
+	 * @param int    $current_object_id ID of the current item.
 	 * @return void
 	 */
 	public function start_el( &$output, $element, $depth = 0, $args = array(), $current_object_id = 0 ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
 		$value_type = $args['value'];
+		$element = $this->convert_to_standard_class( $element );
```
Chouby, line 35: Could this work?
```php
		$element = (object) (array) $element;
```
Alos could it be sufficient to make teh conversion in `walk()`?

Chouby (changes requested):
Could th

include/walker-dropdown.php
Hug0-Drelon: Not according to this discussion: https://github.com/polylang/polylang/pull/1195#pullrequestreview-1245609057
If we want to use `PLL_Language::get_object_vars()` we can't do that IMHO.

include/language.php
```diff
@@ -649,6 +649,17 @@ public function get_object_vars() {
...
 	}
 
+	/**
+	 * Converts current `PLL_language` into a stdClass object. Mostly used to allow dynamic properties.
+	 *
+	 * @since 3.4
+	 *
+	 * @return stdClass Converted `PLL_Language` object.
+	 */
+	public function to_standard_class() {
```
Chouby, line 659: Suggesting `to_std_class()`.
Also, for consistency, we could rename `get_object_vars()` to `to_array()`. What do you think ? In this case, this would require to rebase the branch before doing so.

include/switcher.php
```diff
@@ -251,6 +251,14 @@ public function the_languages( $links, $args = array() ) {
 			$walker = new PLL_Walker_List();
 		}
 
+		// Cast each elements to stdClass because $walker::walk() expects an array of objects.
+		$elements = array_map(
+			function( $element ) {
+				return (object) $element;
+			},
+			$elements
+		);
```
Chouby, line 260: Same with less code:
```suggestion
		foreach( $elements as &$element ) {
			$element = (object) $element;
		}
```

include/walker-dropdown.php
```diff
@@ -19,19 +19,24 @@ class PLL_Walker_Dropdown extends Walker {
...
+	 * @param int                   $current_object_id ID of the current item.
 	 * @return void
 	 */
 	public function start_el( &$output, $element, $depth = 0, $args = array(), $current_object_id = 0 ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
 		$value_type = $args['value'];
+
+		if ( $element instanceof PLL_Language ) {
+			$element = $element->to_standard_class();
+		}
+
```
Chouby, line 39: In fact, is the conversion useful here? I would think it's useful for `display_element()` only.

If it's useful, could we avoid doing this operation 2 times by moving the conversion to `walk()`?
```php
foreach ( $elements as &$element ) {
	if ( $element instanceof PLL_Language ) {
		$element = $element->to_standard_class();
	}
}
```

Chouby (changes requested):

include/switcher.php
Hug0-Drelon: Done in 8660249
Don't you mind I used the array index instead of passing by reference? I find it less confusing and more readable 😅

include/walker-dropdown.php
Hug0-Drelon: It's not useful as is, but in my mind, since these are public method I wanted to handle as many cases as possible. But it may be overkill.
As you said, I leave the conversion in `display_element()` in 1064110.

include/language.php
Hug0-Drelon: Good idea! Done in bc81d22.

Chouby (changes requested):

include/language.php
```diff
@@ -656,7 +656,7 @@ public function get_locale( $filter = 'raw' ) {
 	 *
 	 * @phpstan-return LanguageData
 	 */
-	public function get_object_vars( $context = 'display' ) {
+	public function to_array( $context = 'display' ) {
```
Screenfeed, line 659: We'll need to do the change in [PLL Pro](https://github.com/polylang/polylang-pro/blob/0bfb34007694cbb989484698bed4e2fe4aa152fb/modules/rest/rest-api.php#L221) too.

include/walker.php
```diff
@@ -0,0 +1,73 @@
...
+	 * and triggers an error in case of misuse of them.
+	 *
+	 * @since 3.4
+	 *
+	 * @param array|int $max_depth The maximum hierarchical depth. Passed by reference.
+	 * @param array     $args      Additional arguments. Passed by reference.
+	 * @return void
+	 */
+	protected function maybe_fix_walk_args( &$max_depth, &$args ) {
+		if ( ! is_array( $max_depth ) ) {
```
Screenfeed, line 59: I think we should compare with the expected type here. What do you think?
```suggestion
		if ( is_int( $max_depth ) ) {
```

include/switcher.php
```diff
@@ -251,6 +251,11 @@ public function the_languages( $links, $args = array() ) {
 			$walker = new PLL_Walker_List();
 		}
 
+		// Cast each elements to stdClass because $walker::walk() expects an array of objects.
```
Screenfeed, line 254:
```suggestion
		// Cast each element to stdClass because $walker::walk() expects an array of objects.
```

Screenfeed (changes requested):

include/language.php
Hug0-Drelon: Yup, I wanted to be sure to merge this change before doing a PR.

Screenfeed (approved):

PR #1203, opened 2023-01-17
Add a new property to use for cache key

Marianne380:
See https://github.com/polylang/polylang/pull/1182#issuecomment-1379997871

### DIFF
--- a/include/translatable-object.php
+++ b/include/translatable-object.php
@@ -140,21 +140,6 @@ abstract class PLL_Translatable_Object {
 	}
 
 	/**
-	 * Returns the cache key.
-	 *
-	 * @since 3.4
-	 *
-	 * @param string $key          The md5 hash of the SQL request.
-	 * @param string $last_changed The last changed date for the specified cache group.
-	 * @return string
-	 *
-	 * @phpstan-return non-empty-string
-	 */
-	public function get_cache_key( $key, $last_changed ) {
-		return "{$this->cache_type}_no_lang:{$key}:{$last_changed}";
-	}
-
-	/**
 	 * Adds hooks.
 	 *
 	 * @since 3.4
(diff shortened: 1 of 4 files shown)


include/translatable-object.php
```diff
@@ -170,7 +170,9 @@ public function set_language( $id, $lang ) {
 
 		$term_taxonomy_ids = wp_set_object_terms( $id, $lang, $this->tax_language );
 
-		wp_cache_set( 'last_changed', microtime(), "{$this->type}s" );
+		if ( $this instanceof PLL_Translated_Object ) {
```
Hug0-Drelon, line 173: Not sure it's a good idea `PLL_Translatable_Object` be aware of its child class `PLL_Translated_Object`. Could the property `cache_type` be declared in `PLL_Translatable_Object`?

include/translatable-object.php
```diff
@@ -371,15 +373,18 @@ public function get_objects_with_no_lang( $limit, array $args = array() ) {
...
+			return array();
+		}
+
 		$key          = md5( $sql );
-		$cache_type   = "{$this->type}s";
-		$last_changed = wp_cache_get_last_changed( $cache_type );
-		$cache_key    = "{$cache_type}_no_lang:{$key}:{$last_changed}";
-		$object_ids   = wp_cache_get( $cache_key, $cache_type );
+		$last_changed = wp_cache_get_last_changed( $this->cache_type );
+		$cache_key    = "{$this->cache_type}_no_lang:{$key}:{$last_changed}";
```
Hug0-Drelon, line 382: I suggest to create a dedicated method to retrieve the cache key, (like it's done for WooCommerce if I say correctly). This way anyone will be able to get the key easily and invalidate the cache anytime.

Hug0-Drelon (approved):

include/translatable-object.php
```diff
@@ -372,14 +396,13 @@ public function get_objects_with_no_lang( $limit, array $args = array() ) {
 		}
 
 		$key          = md5( $sql );
-		$cache_type   = "{$this->type}s";
-		$last_changed = wp_cache_get_last_changed( $cache_type );
-		$cache_key    = "{$cache_type}_no_lang:{$key}:{$last_changed}";
-		$object_ids   = wp_cache_get( $cache_key, $cache_type );
+		$last_changed = wp_cache_get_last_changed( $this->cache_type );
+		$cache_key    = $this->get_cache_key( $key, $last_changed );
```
Chouby, line 400:
```suggestion
		$cache_key    =  "{$this->cache_type}_no_lang:{$key}:{$last_changed}";
```
I suggest to revert the introduction of the function.
1. The code is short and cache key is used only once so introducing a new function is not useful.
2. The function name is too generic (the cache key is for a specific SQL query).

Chouby (changes requested):

PR #1205, opened 2023-01-17
Improvement following #1195

Hug0-Drelon:
After the #1195 review, there is still some improvement to do:
- `PLL_LAnguage::get_object_vars()` should accept a string and not a boolean to improve readability when called.
- Create a new filter `deprecated_property_trigger_error`, it doesn't exist in WordPress core yet but could be in a near future.
- Fix a syntax error in `PLL_Language::__get()` on a dynamic call to method.

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -401,8 +401,10 @@ class PLL_Language {
 	 * @return void
 	 */
 	private function deprecated_property( $property, $replacement ) {
-		/** This filter is documented in wordpress/wp-includes/functions.php */
-		if ( WP_DEBUG && apply_filters( 'deprecated_function_trigger_error', true ) ) {
+		/**
+		 * This filter doesn't exist in WordPress core yet, but let's use a similar system as `deprecated_function_trigger_error`.
+		 */
+		if ( WP_DEBUG && apply_filters( 'deprecated_property_trigger_error', true ) ) {
 			trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
 				sprintf(
 					"Class property %1\$s::\$%2\$s is deprecated, use %1\$s::%3\$s instead.\nError handler",
--- a/include/language.php
+++ b/include/language.php
@@ -630,7 +630,7 @@ class PLL_Language {
 	 *
 	 * @since 3.4
 	 *
-	 * @param string $context Whether or not properties should be filtered. Accepts `raw` or `display`.
+	 * @param string $context Whether or not properties should be filtered. Accepts `db` or `display`.
 	 *                        Default to `display` which filters some properties.
 	 *
 	 * @return array Array of language object properties.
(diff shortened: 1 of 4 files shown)


include/language.php
```diff
@@ -628,16 +630,17 @@ public function get_locale( $filter = 'raw' ) {
 	 *
 	 * @since 3.4
 	 *
-	 * @param bool $raw Whether or not properties should be raw. Default to `false`.
+	 * @param string $context Whether or not properties should be filtered. Accepts `raw` or `display`.
+	 *                        Default to `display` which filters some properties.
```
Screenfeed, line 634: When I read this, I'm not sure that `'display'` is the best. Maybe `'filtered'`?

Screenfeed: Idea: `'not-raw'`! (kidding)

include/language.php
```diff
@@ -628,16 +630,17 @@ public function get_locale( $filter = 'raw' ) {
...
 	 * @return array Array of language object properties.
 	 *
 	 * @phpstan-return LanguageData
 	 */
-	public function get_object_vars( $raw = false ) {
+	public function get_object_vars( $context = 'display' ) {
 		$language = get_object_vars( $this );
 
-		if ( ! $raw ) {
+		if ( 'display' === $context ) {
```
Chouby, line 643: I would invert the logic. Indeed, in most cases, we want to use the methods and not send the internal properties. So unknown values should not return the internal properties.
```suggestion
		if ( 'raw' !== $context ) {
```

Chouby (changes requested):

include/language.php
Chouby: I suggest `'db'` instead of ``raw'` as this is really in this context that we want this method to return the unfiltered url.

include/language.php
Hug0-Drelon: Done in 2b4c5d4, as well as the change for `db`.

Screenfeed (approved):

include/language.php
```diff
@@ -401,8 +401,10 @@ public function __isset( $property ) {
 	 * @return void
 	 */
 	private function deprecated_property( $property, $replacement ) {
-		/** This filter is documented in wordpress/wp-includes/functions.php */
-		if ( WP_DEBUG && apply_filters( 'deprecated_function_trigger_error', true ) ) {
+		/**
+		 * This filter doesn't exist in WordPress core yet, but let's use a similar system as `deprecated_function_trigger_error`.
+		 */
```
Chouby, line 406:
```suggestion
		/**
		 * Filters whether to trigger an error for deprecated properties.
		 *
		 * The filter name is intentionnaly not prefixed to use the same as WordPress
		 * in case it is added in the future. 
		 *
		 * @since 3.4
		 *
		 * @param bool $trigger Whether to trigger the error for deprecated properties. Default true.
		 */
```

Chouby (changes requested):

PR #1209, opened 2023-02-02
Removes mo_id from PLL_Language

Marianne380:
Closes https://github.com/polylang/polylang-pro/issues/1584
See also (https://github.com/polylang/polylang/pull/1195#discussion_r1069686911

Delete the `polylang_mo` post type and its associated meta `_pll_strings_translations` to replace it with a term meta linked to the language term.

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -58,6 +58,12 @@ class PLL_Admin_Model extends PLL_Model {
 			update_option( 'polylang', $this->options );
 		}
 
+		// Refresh languages.
+		$this->clean_languages_cache();
+		$this->get_languages_list();
+
+		flush_rewrite_rules(); // Refresh rewrite rules.
+
 		/**
 		 * Fires when a language is added.
 		 *
@@ -66,12 +72,6 @@ class PLL_Admin_Model extends PLL_Model {
 		 * @param array $args Arguments used to create the language. @see PLL_Admin_Model::add_language().
 		 */
 		do_action( 'pll_add_language', $args );
-
-		// Refresh languages.
-		$this->clean_languages_cache();
-		$this->get_languages_list();
-
-		flush_rewrite_rules(); // Refresh rewrite rules.
 
 		return true;
 	}
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -58,10 +58,6 @@ class PLL_Admin_Model extends PLL_Model {
 			update_option( 'polylang', $this->options );
 		}
 
-		// Init a PLL_MO for this language
-		$mo = new PLL_MO();
-		$mo->export_to_db_from_id( $r['term_id'] );
-
 		/**
 		 * Fires when a language is added.
 		 *
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -58,7 +58,7 @@ class PLL_Admin_Model extends PLL_Model {
 			update_option( 'polylang', $this->options );
 		}
 
-		// Init a mo_id for this language
+		// Init a PLL_MO for this language
 		$mo = new PLL_MO();
 		$mo->export_to_db_from_id( $r['term_id'] );
 
@@ -139,12 +139,6 @@ class PLL_Admin_Model extends PLL_Model {
 		foreach ( get_users( array( 'fields' => 'ID' ) ) as $user_id ) {
 			delete_user_meta( $user_id, 'pll_filter_content', $lang->slug );
 			delete_user_meta( $user_id, 'description_' . $lang->slug );
-		}
-
-		// Delete the string translations
-		$mo_id = PLL_MO::get_id( $lang );
-		if ( ! empty( $mo_id ) ) {
-			wp_delete_post( $mo_id );
 		}
 
 		// Delete domain
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -58,11 +58,9 @@ class PLL_Admin_Model extends PLL_Model {
 			update_option( 'polylang', $this->options );
 		}
 
-		$this->clean_languages_cache(); // Update the languages list now !
-
 		// Init a mo_id for this language
 		$mo = new PLL_MO();
-		$mo->export_to_db( $this->get_language( $args['slug'] ) );
+		$mo->export_to_db_from_id( $r['term_id'] );
 
 		/**
 		 * Fires when a language is added.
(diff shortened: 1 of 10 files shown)


include/mo.php
```diff
@@ -35,7 +22,7 @@ public function __construct() {
 	 */
 	public function export_to_db( $lang ) {
 		/*
-		 * It would be convenient to store the whole object but it would take a huge space in DB.
+		 * It would be convenient to store the whole object, but it would take a huge space in DB.
 		 * So let's keep only the strings in an array.
 		 * The strings are slashed to avoid breaking slashed strings in update_post_meta.
```
Hug0-Drelon, line 27:
```suggestion
		 * The strings are slashed to avoid breaking slashed strings in update_term_meta.
```

install/upgrade.php
```diff
@@ -191,7 +191,41 @@ protected function upgrade_2_7() {
...
+		}
+
+		foreach ( $posts as $post ) {
+			$meta = get_post_meta( $post->ID, '_pll_strings_translations', true );
+			if ( empty( $meta ) || ! is_array( $meta ) ) {
+				continue;
+			}
+
+			$term_id = (int) substr( $post->post_title, 12 );
+			update_term_meta( $term_id, '_pll_strings_translations', $meta );
```
Hug0-Drelon, line 223: Shouldn't we use `wp_slash()` on `$meta`?

install/upgrade.php
```diff
@@ -191,7 +191,41 @@ protected function upgrade_2_7() {
...
+			wp_delete_post( $post->ID );
```
Hug0-Drelon, line 224: Do you think we have to delete **all** post, even if returned post meta is empty or not an array?
Because of the previous if, the post may not deleted at all.

uninstall.php
```diff
@@ -112,6 +97,7 @@ public function uninstall() {
 			$term_ids = array_unique( $term_ids );
 			$wpdb->query( "DELETE FROM {$wpdb->terms} WHERE term_id IN ( " . implode( ',', $term_ids ) . ' )' ); // PHPCS:ignore WordPress.DB.PreparedSQL.NotPrepared
 			$wpdb->query( "DELETE FROM {$wpdb->term_taxonomy} WHERE term_id IN ( " . implode( ',', $term_ids ) . ' )' ); // PHPCS:ignore WordPress.DB.PreparedSQL.NotPrepared
+			$wpdb->query( "DELETE FROM {$wpdb->termmeta} WHERE term_id IN ( " . implode( ',', $term_ids ) . ' )' ); // PHPCS:ignore WordPress.DB.PreparedSQL.NotPrepared
```
Hug0-Drelon, line 100: Here we are deleting **all** language term meta, even the ones that aren't `_pll_strings_translations` (for instance third party added metas). Is that what we want?

install/upgrade.php
```diff
@@ -203,7 +237,7 @@ protected function upgrade_2_7() {
 	 *
 	 * @return void
 	 */
-	protected function upgrade_3_4() {
+	protected function migrate_locale_fallback_to_language_description() {
```
Hug0-Drelon, line 240: Not sure the method documentation is still accurate.
Also, do you think that creating a separated method for the term meta migration (as you did with this new method) is a good idea?

admin/admin-model.php
```diff
@@ -58,11 +58,11 @@ public function add_language( $args ) {
...
 
-		$this->clean_languages_cache(); // Update the languages list now !
+		// Refresh languages.
+		$this->clean_languages_cache();
+		$this->get_languages_list();
 
-		// Init a mo_id for this language
-		$mo = new PLL_MO();
-		$mo->export_to_db( $this->get_language( $args['slug'] ) );
+		flush_rewrite_rules(); // Refresh rewrite rules.
```
Chouby, line 65: OK to move the action at the end, but for consistency, you should move `pll_update_language` too.

install/upgrade.php
Chouby: I agree that you should move the strings migration to a separate method too.
But you must keep the `delete_transient()` and its related comment in the min method as it's not fully related to the fallback migration.

Chouby (changes requested):

include/language.php
```diff
@@ -331,6 +331,21 @@ public function __get( $property ) {
 			return $this->{$url_getter}();
 		}
 
+		if ( 'mo_id' === $property ) {
+			trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
```
Hug0-Drelon, line 335: There's a dedicated method to trigger errors for deprecated props, see `$this->deprecated_property()`.

include/mo.php
```diff
@@ -8,25 +8,12 @@
  *
  * @since 1.2
  * @since 2.1 Stores the strings in a post meta instead of post content to avoid unserialize issues (See #63)
+ * @since 3.4 Stores the strings in a term meta instead of post meta.
```
Hug0-Drelon, line 11:
```suggestion
 * @since 3.4 Stores the strings into language taxonomy term meta instead of a post meta.
```
Should we be more precise?

tests/phpunit/tests/test-install.php
```diff
@@ -47,6 +47,7 @@ public function test_uninstall_removing_data() {
 		$english = self::$model->get_language( 'en' );
 
 		self::create_language( 'fr_FR' );
+		$french = self::$model->get_language( 'fr' );
```
Hug0-Drelon, line 50: Why is it necessary? 
If it's to refresh the languages list, it should has already been done at the end of `add_language()` right?

Marianne380: Only to use it here: https://github.com/polylang/polylang/blob/87e631ce277ef4aa93d556a07f2e8e110750f2fa/tests/phpunit/tests/test-install.php#L131

Hug0-Drelon: Oops 😅

include/language.php
Hug0-Drelon: I see, don't know if it worth making `deprecated_property()` less specific.

Hug0-Drelon (approved):

include/language.php
Chouby: Is it worth? Why someone would like to access to this property? To access strings translations, we pass the complete language to `PLL_MO` methods..

PR #1210, opened 2023-02-03
Deprecate language list filters

Screenfeed:
Since the languages list has been refactored to contain immutable objects, the filters `pll_languages_list` and `pll_after_languages_cache` are deprecated.

- `pll_languages_list` is replaced by `pll_additional_language_data`.
- `pll_after_languages_cache` has no replacements: since its purpose was to be able to filter the URLs after the cache, we can't say that `pll_additional_language_data` is a fitted replacement (which is triggered before the `PLL_Language` objects are even created, so, long before the cache. However, a comment has been added to advise to hook `'site_url'` instead (not added in the deprecation notice because the hook could be used for something else).
- A bit unrelated to this PR, but while discussing the previous point's cache issue, we discovered that `get_display_flag_url()` didn't use the constants `PLL_CACHE_LANGUAGES` and `PLL_CACHE_HOME_URL` like `get_home_url()` and `get_search_url()` do, so they have been added. Props @Hug0-Drelon for the idea.

This PR should not be merged before https://github.com/polylang/polylang-wc/pull/548, as Polylang for WooCommerce still uses those filters.

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -597,7 +597,7 @@ class PLL_Language {
 	public function get_display_flag_url() {
 		$flag_url = empty( $this->custom_flag_url ) ? $this->flag_url : $this->custom_flag_url;
 
-		if ( defined( 'PLL_CACHE_LANGUAGES' ) && ! PLL_CACHE_LANGUAGES ) {
+		if ( ( defined( 'PLL_CACHE_LANGUAGES' ) && ! PLL_CACHE_LANGUAGES ) || ( defined( 'PLL_CACHE_HOME_URL' ) && ! PLL_CACHE_HOME_URL ) ) {
 			/**
 			 * Let's use `site_url()` so the returned URL will be filtered properly according to the current domain.
 			 */
--- a/include/language.php
+++ b/include/language.php
@@ -405,7 +405,7 @@ class PLL_Language {
 		 * Filters whether to trigger an error for deprecated properties.
 		 *
 		 * The filter name is intentionnaly not prefixed to use the same as WordPress
-		 * in case it is added in the future.
+		 * in case it is added in the future. 
 		 *
 		 * @since 3.4
 		 *
@@ -597,14 +597,10 @@ class PLL_Language {
 	public function get_display_flag_url() {
 		$flag_url = empty( $this->custom_flag_url ) ? $this->flag_url : $this->custom_flag_url;
 
-		if ( defined( 'PLL_CACHE_LANGUAGES' ) && ! PLL_CACHE_LANGUAGES ) {
-			/**
-			 * Let's use `site_url()` so the returned URL will be filtered properly according to the current domain.
-			 */
-			return site_url( set_url_scheme( $flag_url, 'relative' ) );
-		}
-
-		return $flag_url;
+		/**
+		 * Let's use `site_url()` so the returned URL will be filtered properly according to the current domain.
+		 */
+		return site_url( set_url_scheme( $flag_url, 'relative' ) );
 	}
 
 	/**
--- a/include/language.php
+++ b/include/language.php
@@ -597,7 +597,7 @@ class PLL_Language {
 	public function get_display_flag_url() {
 		$flag_url = empty( $this->custom_flag_url ) ? $this->flag_url : $this->custom_flag_url;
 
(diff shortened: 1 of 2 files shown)


include/model.php
```diff
@@ -150,10 +150,12 @@ public function get_languages_list( $args = array() ) {
 			 * /!\ This filter is fired *before* the $polylang object is available.
 			 *
 			 * @since 1.8
+			 * @since 3.4 Deprecated.
+			 * @deprecated
 			 *
 			 * @param PLL_Language[] $languages The list of language objects.
 			 */
-			$languages = apply_filters( 'pll_after_languages_cache', $languages );
+			$languages = apply_filters_deprecated( 'pll_after_languages_cache', array( $languages ), '3.4', 'pll_additional_language_data' );
```
Hug0-Drelon, line 158: Could we add information with the `$message` parameter?
If I understood correctly, this filter was created to avoid getting cached values (for URLs at least).
I'm not sure `pll_additional_language_data` is an ideal replacement here. Maybe we could steer users to use the proper getters?

Screenfeed: But which hook would be more suited? (it's the one you use in https://github.com/polylang/polylang-wc/pull/548 😏😜).

Hug0-Drelon: I know 😅 I'm a bit confused too. Let's say I know what I'm doing (really?!?). But third party might want to use `pll_additional_language_data` instead of `pll_after_languages_cache` thinking it would allow them to bypass cached properties.
Also, I purely removed `pll_after_languages_cache` in PLLWC in favor of the getters.

Screenfeed: Currently there is no way to bypass the cache, it would require filters in `get_home_url()` and `get_search_url()`, wouldn't it?

Hug0-Drelon: I think that's already the case [here](https://github.com/polylang/polylang/blob/3.4-dev/include/language.php#L695-L702), right?

Screenfeed: Well, I meant a dedicated PLL filter but filtering `site_url()` is enough indeed.
So you think we should add a message about the filter `'site_url'`?

Screenfeed: After discussing we decided to remove the replacement for the hook `'pll_after_languages_cache'` and add a comment about using `'site_url'` instead.
Done in [9d7554b](https://github.com/polylang/polylang/pull/1210/commits/9d7554bc6e961d14b5a8b92c1556bd3ed8053e7c).

Hug0-Drelon (approved):

include/language.php
```diff
@@ -597,10 +597,14 @@ public function get_display_flag() {
...
-		 */
-		return site_url( set_url_scheme( $flag_url, 'relative' ) );
+		if ( ( defined( 'PLL_CACHE_LANGUAGES' ) && ! PLL_CACHE_LANGUAGES ) || ( defined( 'PLL_CACHE_HOME_URL' ) && ! PLL_CACHE_HOME_URL ) ) {
+			/**
+			 * Let's use `site_url()` so the returned URL will be filtered properly according to the current domain.
+			 */
+			return site_url( set_url_scheme( $flag_url, 'relative' ) );
+		}
+
+		return $flag_url;
```
Chouby, line 607: It seems that you misunderstood the purpose of the function. We can not cache the flag url. At least not in the context proposed by this test. The reason is that the url of the flags will change depending on the domain we are visiting.

Ex: 2 languages, `mysite.com` (default), `monsite.fr`:
`$flag_url` will be `mysite.com/wp-content/plugins/polylang/flags/us.png` for English and `mysite.com/wp-content/plugins/polylang/flags/fr.png` for French.
These 2 urls are correct when we display the language switcher on `mysite.com`.
But they must be switched to `monsite.fr/wp-content/plugins/polylang/flags/us.png` and `monsite.fr/wp-content/plugins/polylang/flags/fr.png when we display the language switcher on `monsite.fr`. 

So the urls are dependent on the domain displayed. This is true whether the languages are cached or not. I propose to revert this change.

Chouby (changes requested):

include/language.php
Screenfeed: Indeed, that's not exactly the same thing.
Reverted.

Hug0-Drelon: Oops I didn't thought about it that way ...

PR #1211, opened 2023-02-03
Don't cache incomplete languages list

Screenfeed:
This PR prevents caching the languages list until the languages are complete, which happens after the "static page" instance is created.
This also triggers a "doing it wrong" warning if the list is retrieved before it is complete.

I found 4 ways to achieve that:

1. `doing_action( 'pll_pre_init' ) || did_action( 'pll_pre_init' )`:
    - no need to modify anything else in the plugin,
    - but `'pll_pre_init'` is never fired during the tests, so the cache would never been enabled.
2. Custom property (boolean) in `PLL_Model`, set to `true` in the class `Polylang`, after instantiating the `$polylang` object:
    - slight performance improvement, compared to `doing_action()` + `did_action()`,
    - minor change outside `PLL_Model`,
    - but requires to set the property in all of our tests.
3. Custom property (boolean) in `PLL_Model`, set to `true` at the end of each "context" class' constructor (like `PLL_Base`):
    - slight performance improvement, compared to `doing_action()` + `did_action()`,
    - minor changes outside `PLL_Model`,
    - but not all tests instanciate a "context" class.
4. Custom property (boolean) set by a callback hooked on `'pll_pre_init'` with priority `PHP_INT_MIN`:
    - no need to modify anything else in the plugin,
    - but `'pll_pre_init'` is never fired during the tests, so the cache would never been enabled.

I chose the 3rd option because of the tiny performance improvement and because only 10% of the tests wouldn't trigger the cache.
Note: the warning is suppressed in tests for now, maybe we'll find a way to improve our setup in tests later.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -855,7 +855,7 @@ class PLL_Model {
 		 */
 		$languages_data = array_map(
 			function ( $language ) {
-			return $language->to_array( 'db' );
+				return $language->to_array( 'db' );
 			},
 			$languages
 		);
--- a/include/model.php
+++ b/include/model.php
@@ -841,22 +841,26 @@ class PLL_Model {
 		 */
 		$languages = apply_filters( 'pll_languages_list', $languages, $this );
 
-		if ( $this->are_languages_ready() ) {
-			/**
-			 * Don't store directly objects as it badly break with some hosts ( GoDaddy ) due to race conditions when using object cache.
-			 * Thanks to captin411 for catching this!
-			 *
-			 * @see https://wordpress.org/support/topic/fatal-error-pll_model_languages_list?replies=8#post-6782255
-			 */
-			$languages_data = array_map(
-				function ( $language ) {
-				return $language->to_array( 'db' );
-				},
-				$languages
-			);
-
-			set_transient( 'pll_languages_list', $languages_data );
+		if ( ! $this->are_languages_ready() ) {
+			// Do not cache an incomplete list.
+			/** @var list<PLL_Language> $languages */
+			return $languages;
 		}
+
+		/**
+		 * Don't store directly objects as it badly break with some hosts ( GoDaddy ) due to race conditions when using object cache.
+		 * Thanks to captin411 for catching this!
+		 *
+		 * @see https://wordpress.org/support/topic/fatal-error-pll_model_languages_list?replies=8#post-6782255
+		 */
+		$languages_data = array_map(
+			function ( $language ) {
+			return $language->to_array( 'db' );
+			},
+			$languages
+		);
+
+		set_transient( 'pll_languages_list', $languages_data );
 
 		/** @var list<PLL_Language> $languages */
 		return $languages;
--- a/include/model.php
+++ b/include/model.php
@@ -125,7 +125,7 @@ class PLL_Model {
 		if ( ! $this->are_languages_ready() ) {
 			_doing_it_wrong(
 				__METHOD__ . '()',
-				"This function must not be called before the hook 'pll_pre_init'.",
+				"It must not be called before the hook 'pll_pre_init'.",
 				'3.4'
 			);
 		}
--- a/include/model.php
+++ b/include/model.php
@@ -123,11 +123,9 @@ class PLL_Model {
 	 */
 	public function get_languages_list( $args = array() ) {
 		if ( ! $this->are_languages_ready() ) {
-			/* translators: %s is the name of a hook. */
-			$message = function_exists( '__' ) ? __( 'This function must not be called before the hook %s.', 'polylang' ) : 'This function must not be called before the hook %s.';
 			_doing_it_wrong(
 				__METHOD__ . '()',
-				sprintf( esc_html( $message ), "'pll_pre_init'" ),
+				"This function must not be called before the hook 'pll_pre_init'.",
 				'3.4'
 			);
 		}
--- a/include/model.php
+++ b/include/model.php
@@ -54,6 +54,13 @@ class PLL_Model {
 	private $is_creating_language_objects = false;
 
 	/**
+	 * Tells if {@see PLL_Model::get_languages_list()} can be used.
+	 *
+	 * @var bool
+	 */
+	private $languages_ready = false;
+
+	/**
 	 * Constructor.
 	 * Setups translated objects sub models.
 	 * Setups filters and actions.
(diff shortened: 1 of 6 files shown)


Screenfeed:
I'm going to investigate an alternative suggested by @Hug0-Drelon: instead of returning incomplete objects before they are ready, return an empty array.

1. This will inevitably trigger errors in our tests, but maybe we can fix them if they are not too many.
2. We need that the work on `PLL_MO` is complete.

Screenfeed:
As expected...
```
..EF................EEEE.......................................  63 / 657 (  9%)
............................................................... 126 / 657 ( 19%)
.....S...FFFFFFFF.FFFFFFFF.FFF.FF.FF.F..FFF.F.FFFFF.FFFF..FF.FF 189 / 657 ( 28%)
FFFF.......F......................................S.EE......... 252 / 657 ( 38%)
.........EEEE...........................EEEE...............F..F 315 / 657 ( 47%)
..FF..FFFFFFF.F............FF.F.FF..FFFEEE.FFFFE............... 378 / 657 ( 57%)
.............F................................................. 441 / 657 ( 67%)
............................................................... 504 / 657 ( 76%)
..........................FFFFFFFFF.EEFF...FFFF................ 567 / 657 ( 86%)
...........................................EFFF.FFF......F..... 630 / 657 ( 95%)
...........................                                     657 / 657 (100%)
```

Hug0-Drelon (approved):

include/model.php
```diff
@@ -154,7 +161,11 @@ public function get_languages_list( $args = array() ) {
 			 * @param PLL_Language[] $languages The list of language objects.
 			 */
 			$languages = apply_filters( 'pll_after_languages_cache', $languages );
-			$this->cache->set( 'languages', $languages );
+
+			if ( $this->are_languages_ready() ) {
+				$this->cache->set( 'languages', $languages );
+			}
```
Chouby, line 167: Wouldn't it be appropriate to fire a `_doing_it_wrong()` in the other case?

Screenfeed: Done in [cc9572f](https://github.com/polylang/polylang/pull/1211/commits/cc9572f3ce0b03a544da6b41721a622a0edbfebb).

include/model.php
```diff
@@ -122,6 +122,16 @@ public function has_languages() {
 	 * @return array List of PLL_Language objects or PLL_Language object properties.
 	 */
 	public function get_languages_list( $args = array() ) {
+		if ( ! $this->are_languages_ready() ) {
+			/* translators: %s is the name of a hook. */
+			$message = function_exists( '__' ) ? __( 'This function must not be called before the hook %s.', 'polylang' ) : 'This function must not be called before the hook %s.';
```
Hug0-Drelon, line 127: I think we chose to not translate error message, am I wrong @Chouby?

Chouby: Exactly. Unlike WP, we don't translate messages displayed to developers only.

Screenfeed: Done in [7ddbfa8](https://github.com/polylang/polylang/pull/1211/commits/7ddbfa870ea94a405822d3f75a637c20cd5150fb).

Hug0-Drelon (approved):

tests/phpunit/includes/testcase-trait.php
```diff
@@ -63,6 +63,20 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) { // p
...
+		 * @see PLL_UnitTestCase_Trait::doing_it_wrong_run()
+		 */
+		add_filter(
+			'doing_it_wrong_trigger_error',
+			function ( $trigger_error, $function_name ) {
+				return $trigger_error && 'PLL_Model::get_languages_list()' !== $function_name;
+			},
+			10,
+			2
+		);
```
Chouby, line 79: I did not test but [this](https://github.com/WordPress/wordpress-develop/blob/6.1.1/tests/phpunit/includes/abstract-testcase.php#L557) makes me think that this is useless. If I'm wrong, then I would prefer not to use an anonymous function.

tests/phpunit/includes/testcase-trait.php
```diff
@@ -166,4 +180,25 @@ protected function require_wp_menus( $trigger_hooks = true ) {
 
 		return static::$submenu;
 	}
+
+	/**
+	 * Don't trigger an error if `PLL_Model::get_languages_list()` is called too early.
+	 * Note: the paramaters `$message` and `$version` are available since WP 6.1.
```
Chouby, line 186:
```suggestion
	 * Note: the parameters `$message` and `$version` are available since WP 6.1.
```

Chouby (changes requested):

tests/phpunit/includes/testcase-trait.php
Screenfeed: It is required because `expectDeprecated()` is called in `set_up()`, which is too late: we need this filter in `wpSetUpBeforeClass()`, before creating the languages, because `PLL_UnitTestCase::create_language` calls `PLL_Admin_Model::add_language()`, which calls `PLL_Admin_Model::validate_lang()`, then `PLL_Model::get_languages_list()`.

include/model.php
```diff
@@ -115,6 +122,14 @@ public function has_languages() {
 	 * @return array List of PLL_Language objects or PLL_Language object properties.
 	 */
 	public function get_languages_list( $args = array() ) {
+		if ( ! $this->are_languages_ready() ) {
+			_doing_it_wrong(
+				__METHOD__ . '()',
+				"This function must not be called before the hook 'pll_pre_init'.",
```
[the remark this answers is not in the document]
Screenfeed: Wording changed in [0a6937c](https://github.com/polylang/polylang/pull/1211/commits/0a6937cf7326ffa53ea5e199c141c95081932c3b).

include/model.php
```diff
@@ -154,7 +169,11 @@ public function get_languages_list( $args = array() ) {
 			 * @param PLL_Language[] $languages The list of language objects.
 			 */
 			$languages = apply_filters( 'pll_after_languages_cache', $languages );
-			$this->cache->set( 'languages', $languages );
+
+			if ( $this->are_languages_ready() ) {
+				$this->cache->set( 'languages', $languages );
+			}
```
Chouby, line 175:
```suggestion
			if ( $this->are_languages_ready() ) {
				$this->cache->set( 'languages', $languages );
			} else {
				_doing_it_wrong(
					__METHOD__ . '()',
					"It must not be called before the hook 'pll_pre_init'.",
					'3.4'
				);
			}
```

Screenfeed: As we discussed, we keep it here for a better readability.

tests/phpunit/includes/testcase-trait.php
Screenfeed: Simplified and added a comment with [8e4000c](https://github.com/polylang/polylang/pull/1211/commits/8e4000c2f3317386ac67244e10d4a192dcf05643).

PR #1215, opened 2023-02-08
Fix E_WARNING notice

buutqn:
Sometimes this line triggers E_WARNING with the following message: Undefined array key "default_lang".

This change prevents that from happening;

### DIFF
--- a/include/links-directory.php
+++ b/include/links-directory.php
@@ -66,7 +66,7 @@ class PLL_Links_Directory extends PLL_Links_Permalinks {
 	public function add_language_to_link( $url, $lang ) {
 		if ( ! empty( $lang ) ) {
 			$base = $this->options['rewrite'] ? '' : 'language/';
-			$slug = $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : $base . $lang->slug . '/';
+			$slug = isset( $this->options['default_lang'] ) && $this->options['default_lang'] == $lang->slug && $this->options['hide_default'] ? '' : $base . $lang->slug . '/';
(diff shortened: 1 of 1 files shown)


Chouby:
Thank you for your contribution. 
I would appreciate if you could prvide more context to this error.
As far as I understand, your site has at least one language but no default language. How did you reach this state?
1. [The default language should be automatically assigned to the first language created](https://github.com/polylang/polylang/blob/3.3.1/admin/admin-model.php#L51-L55).
2. [If the default language is deleted, it is automatically assigned to another existing language](https://github.com/polylang/polylang/blob/3.3.1/admin/admin-model.php#L90-L101). As you'll notice we unset the default language only when deleting the last language.

buutqn:
> Thank you for your contribution. I would appreciate if you could prvide more context to this error. As far as I understand, your site has at least one language but no default language. How did you reach this state?
> 
> 1. [The default language should be automatically assigned to the first language created](https://github.com/polylang/polylang/blob/3.3.1/admin/admin-model.php#L51-L55).
> 2. [If the default language is deleted, it is automatically assigned to another existing language](https://github.com/polylang/polylang/blob/3.3.1/admin/admin-model.php#L90-L101). As you'll notice we unset the default language only when deleting the last language.

The site is running with 2 language and default is defined, it works just fine, but some routine of a third party plugin before the display page time is calling this function, probably before fully loading polylang, not entirely sure.

If you want, I can verify further on this to check which plugin and hook is calling this method.

Chouby:
Yes thank you, that would be interesting to have a stack trace for this warning and also if you could check if the section "Polylang options" in Tools > Site Health > info displays a line "default_lang" (this section should display the content of the serialized array stored in the option "polylang").

PR #1207, opened 2023-01-27
Fix all new issues reported by PHPStan

Screenfeed:

PR #1216, opened 2023-02-08
Prevent warning

buutqn:
am using an error notify plugin, and i keep receiving this error:
Attempt to read property "home_url" on bool
As well, wordpress admin dashboard loading with php-error class;

This simple change prevent the warning being triggered.


Chouby:
Thank you for your contribution. However we have already fixed it in #1207 (not merged to master yet but planned for v3.4)

PR #1219, opened 2023-02-13
Fix message for deprecated property

Screenfeed:
Don't add the `()` after the replacement in the error message: the replacement itself must include them.

- 1st case: `get_tax_prop( '{$term_prop_type}', '{$term_prop}' )`,
- 2nd case: `get_{$property}()`.

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -335,7 +335,7 @@ class PLL_Language {
 		}
 
 		if ( 'search_url' === $property || 'home_url' === $property ) {
-			$url_getter = "get_{$property}";
+			$url_getter = "get_{$property}()";
 
 			$this->deprecated_property( $property, $url_getter );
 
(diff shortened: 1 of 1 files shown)


include/language.php
```diff
@@ -335,7 +335,7 @@ public function __get( $property ) {
 		}
 
 		if ( 'search_url' === $property || 'home_url' === $property ) {
-			$url_getter = "get_{$property}";
+			$url_getter = "get_{$property}()";
```
Hug0-Drelon, line 338: Won't this call `get_home_url()()` in line 342?

Screenfeed: Fixed in [843004c](https://github.com/polylang/polylang/pull/1219/commits/843004c3c3dc2998c3ebfa86274a20935ce9ff50).

Hug0-Drelon (approved):

PR #1224, opened 2023-02-27
Don't set update plugins transient to null

Chouby:
To force WordPress to check for the plugins updates we use to set the transient `update_plugins` to `null`.

This works as expected but it may be a source of conflict as WordPress usually deletes the transient to obtain the same result. Indeed, in our case `get_site_transient()` returns `null` instead of `false`. If a 3rd party plugin explicitely tests the retrun value against `false` instead the expected object, then it will result in a fatal error.

### DIFF
--- a/include/license.php
+++ b/include/license.php
@@ -149,8 +149,8 @@ class PLL_License {
 		$this->license_key = $license_key;
 		$this->api_request( 'activate_license' );
 
-		// Tell WordPress to look for updates
-		set_site_transient( 'update_plugins', null );
+		// Tell WordPress to look for updates.
+		delete_site_transient( 'update_plugins' );
 		return $this;
 	}
 


manooweb (approved):
👌 OK

Chouby:
I could examine the code of Jet Engine which fires a fatal error.
The faulty method is hooked to `pre_set_site_transient_update_plugins` and expects an object with a response property as value passed to the hook, while the hook doc accepts any value (mixed). No check is made before attempting to access to this object. 
There was no mistake (according to the doc) on our side but we did something that WP never does passing `null` to this transient. Now  with `delete_site_transient()`, we won't fire the hook anymore. The potential fatal error still exist in Jet Engine but Polylang will not trigger it.

PR #1225, opened 2023-03-03
WordPress managers do not detect Polylang Pro update

sebastienserre:
WordPress manger like ManageWP & UmbrellaWP are bypassing some WordPress hooks (admin_init is one of them)
Tested and works with UmbrellaWP

Reported on https://secure.helpscout.net/conversation/2170319527/22959?folderId=3655443

### DIFF
--- a/include/license.php
+++ b/include/license.php
@@ -91,7 +91,7 @@ class PLL_License {
 		}
 
 		// Updater
-		add_action( 'plugins_loaded', array( $this, 'auto_updater' ), 10 );
+		$this->auto_updater();
 
 		// Register settings
 		add_filter( 'pll_settings_licenses', array( $this, 'settings' ) );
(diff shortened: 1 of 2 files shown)


Chouby:
Since v2.10 EDD uses a [different way](https://github.com/awesomemotive/easy-digital-downloads/pull/8170/commits/481f80d6536a38924bb23fa12e4e588935f30a3d).

include/license.php
```diff
@@ -91,8 +91,7 @@ public function __construct( $file, $item_name, $version, $author, $api_url = nu
 		}
 
 		// Updater
-		add_action( 'admin_init', array( $this, 'auto_updater' ), 0 );
-		add_action( 'cli_init', array( $this, 'auto_updater' ), 0 ); // For WP CLI.
+		add_action( 'plugins_loaded', array( $this, 'auto_updater' ), 10 );
```
Chouby, line 94: Do we really need to use an action here? As far as I know, at this point we are already hooked to `plugins_loaded` (with priority 1).
```suggestion
		$this->auto_updater();
```

Hug0-Drelon: You're right, let's simplify!

PR #1228, opened 2023-03-07
New `$is_default` property in `PLL_Language`.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/1600
Add a new property `$is_default` to `PLL_Language`.

## Why?
To ease default language tests as well as to easily pass this piece of data to JS scripts (through REST for instance).

## How?
- Introduce new prop `$is_default` in `PLL_Language` class and make it mandatory.
- Inject Polylang's options in `PLL_Language_Factory::get_from_terms()` to pass `$is_default` correctly to `PLL_Language` constructor. This is not useful for the method `PLL_Language_Factory::get()` since it's used to create language with the transient (which should already have the property set).
- Use this prop everywhere we test if a given language is the default one.

### DIFF
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -41,13 +41,14 @@ class PLL_Language_Factory {
 	 *
 	 * @since 3.4
 	 *
-	 * @param WP_Term[] $terms List of language terms, with the language taxonomy names as array keys.
-	 *                         `language` and `term_language` are mandatory keys.
+	 * @param WP_Term[] $terms   List of language terms, with the language taxonomy names as array keys.
+	 *                           `language` and `term_language` are mandatory keys.
+	 * @param array     $options Array of Polylang's options.
 	 * @return PLL_Language
 	 *
 	 * @phpstan-param array{language:WP_Term, term_language:WP_Term}&array<string, WP_Term> $terms
 	 */
-	public static function get_from_terms( array $terms ) {
+	public static function get_from_terms( array $terms, array $options ) {
 		$languages = self::get_languages();
 		$data      = array(
 			'name'       => $terms['language']->name,
@@ -93,6 +94,8 @@ class PLL_Language_Factory {
 				$data['facebook'] = $languages[ $data['locale'] ]['facebook'];
 			}
 		}
+
+		$data['is_default'] = $options['default_lang'] === $data['slug'];
 
 		$flag_props = self::get_flag( $data['flag_code'], $data['name'], $data['slug'], $data['locale'] );
 		$data       = array_merge( $data, $flag_props );
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -18,7 +18,7 @@ class PLL_Language_Factory {
 	 *
 	 * @phpstan-var array<string, array<string, string>>
 	 */
-	private $languages;
+	private static $languages;
 
 	/**
 	 * Polylang's options.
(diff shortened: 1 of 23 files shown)


include/language-factory.php
```diff
@@ -94,6 +95,8 @@ public static function get_from_terms( array $terms ) {
 			}
 		}
 
+		$data['is_default'] = $options['default_lang'] === $data['slug'];
```
manooweb, line 98: For you, it's clearer to set the property than when we initialize the `$data` array at the beginning of the method?

manooweb (changes requested):
Just a question. If it seems clearer for you, we can keep like that ;)

manooweb (approved):
LGTM 👌

include/language-factory.php
```diff
@@ -41,19 +41,21 @@ public static function get( $language_data ) {
...
-	 *                         `language` and `term_language` are mandatory keys.
+	 * @param WP_Term[] $terms   List of language terms, with the language taxonomy names as array keys.
+	 *                           `language` and `term_language` are mandatory keys.
+	 * @param array     $options Array of Polylang's options.
 	 * @return PLL_Language
 	 *
 	 * @phpstan-param array{language:WP_Term, term_language:WP_Term}&array<string, WP_Term> $terms
 	 */
-	public static function get_from_terms( array $terms ) {
+	public static function get_from_terms( array $terms, array $options ) {
```
Chouby, line 51: I would have passed `$options`  as dependence of the class in a constructor (mainly because this is how we use options in the other classes), but why not like this? @Screenfeed your opinion?

Chouby (changes requested):

integrations/jetpack/jetpack.php
```diff
@@ -76,7 +76,8 @@ public function jetpack_widget_get_top_posts( $posts ) {
 	 */
 	public function grunion_contact_form_field_html_filter( $r, $field_label ) {
 		if ( function_exists( 'icl_translate' ) ) {
-			if ( pll_current_language() !== pll_default_language() ) {
+			$curlang = pll_current_language();
+			if ( empty( $curlang ) || ! $curlang->is_default ) {
```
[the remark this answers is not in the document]
Hug0-Drelon: Is it ok for you if we get an object instead? See 8fe0c86a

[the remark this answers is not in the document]
Chouby: Your proposal is probably ok but I would prefer to keep the code as close from the original as possible. See https://github.com/Automattic/jetpack/blob/9.3.2/3rd-party/wpml.php#L50-L60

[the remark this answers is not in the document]
Hug0-Drelon: Got it! Reverted in 640e513e

include/language-factory.php
Screenfeed: It makes sense. That gives us an excuse to make all methods non-static, for the small cost of it being a little bit more verbose.

include/language-factory.php
```diff
@@ -18,7 +18,26 @@ class PLL_Language_Factory {
 	 *
 	 * @phpstan-var array<string, array<string, string>>
 	 */
-	private static $languages;
+	private $languages;
```
Chouby, line 21: Maybe we could keep this property static, thus avoiding to read the languages file as many times as we create instances.

include/model.php
```diff
@@ -819,20 +823,20 @@ protected function get_languages_from_taxonomies() {
...
+		/**
+		 * @var (
+		 *     array{
+		 *         string: array{
+		 *             language: WP_Term,
+		 *             term_language: WP_Term
+		 *         }&array<non-empty-string, WP_Term>
+		 *     }
+		 * ) $terms_by_slug
+		 */
```
Chouby, line 835: I know it's not related to this PR but I'd like to disuss this with you and @Screenfeed

Chouby (changes requested):

include/language-factory.php
Hug0-Drelon: Good idea, done in c37ec018.

PR #1229, opened 2023-03-07
Add a new method query_objects_with_no_lang().

Hug0-Drelon:
## What?
Fixes partially https://github.com/polylang/polylang-wc/issues/502 (see last "Polylang" task).

## Why?
To allow third party to override the way objects with no language are queried (especially when the cache is hit).
For instance, if a third party doesn't use  `wp_cache_set_last_changed()`, the way we are using the cache won't work anymore.

## How?
Add a new protected method `PLL_Translatable_Object::query_objects_with_no_lang()`.

### DIFF
--- a/include/translatable-object.php
+++ b/include/translatable-object.php
@@ -379,17 +379,35 @@ abstract class PLL_Translatable_Object {
 			return array();
 		}
 
+		$object_ids = $this->query_objects_with_no_lang( $sql );
+
+		return array_values( $this->sanitize_int_ids_list( $object_ids ) );
+	}
+
+	/**
+	 * Returns object IDs without language given a specific SQL query.
+	 * Can be overriden by child classes in case queried object doesn't use
+	 * `wp_cache_set_last_changed()` or another cache system.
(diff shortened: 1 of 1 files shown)


include/translatable-object.php
```diff
@@ -379,17 +379,35 @@ public function get_objects_with_no_lang( $limit, array $args = array() ) {
...
+
+	/**
+	 * Returns object IDs without language given a specific SQL query.
+	 * Can be overriden by child classes in case queried object doesn't use
+	 * `wp_cache_set_last_changed()` or another cache system.
+	 *
+	 * @since 3.4
+	 *
+	 * @param string $sql A prepared SQL query for object IDs with no language.
+	 * @return mixed An array of numeric object IDs.
```
Screenfeed, line 395: Since `wpdb::get_col()` returns an array, I think it's safe to use a more specific return type:
```suggestion
	 * @return string[] An array of numeric object IDs.
```

Screenfeed (approved):

PR #1230, opened 2023-03-08
Tests post language

Marianne380:
Partially fix https://github.com/polylang/polylang-pro/issues/1561

Re set the media test for the `save_translations` method.
Add a test for testing that a post language is not updated if it is already set.

### DIFF
--- a/tests/phpunit/tests/test-translated-post.php
+++ b/tests/phpunit/tests/test-translated-post.php
@@ -202,4 +202,11 @@ class Translated_Post_Test extends PLL_Translated_Object_UnitTestCase {
 
 		$this->dont_save_translations_with_incorrect_language( $model->post );
 	}
+
+	public function test_post_language_not_updated_if_already_set() {
+		$post_id = self::factory()->post->create();
+		self::$model->post->set_language( $post_id, 'fr' );
+
+		$this->assertFalse( self::$model->post->set_language( $post_id, 'fr' ) );
+	}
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-translated-post.php
```diff
@@ -202,4 +202,14 @@ public function test_dont_save_translations_with_incorrect_language() {
...
 	}
+
+	public function test_post_language_not_updated_if_already_set() {
+		$post_id = self::factory()->post->create();
+
+		$this->assertNotEmpty( $post_id );
+
+		self::$model->post->set_language( $post_id, 'fr' );
+
+		$this->assertFalse( self::$model->post->set_language( $post_id, 'fr' ) );
```
Hug0-Drelon, line 213:
```suggestion
		$this->assertFalse( self::$model->post->set_language( $post_id, 'fr' ) );
```

Marianne380: We have to set it once first.

Hug0-Drelon: 🙈

Hug0-Drelon (approved):

Chouby (changes requested):
I suggest to move the test on top just after `test_post_language()` as both tests are related. I also suggest to assert the return value in this old test. And you could also test the update to a different language.

PR #1232, opened 2023-03-09
Simplify code to get objects with no language

Chouby:
This PR changes how we handle `get_objects_with_no_lang()` for objects with types.

### DIFF
--- a/include/translatable-object.php
+++ b/include/translatable-object.php
@@ -373,12 +373,7 @@ abstract class PLL_Translatable_Object {
 			return array();
 		}
 
-		$sql = $this->get_objects_with_no_lang_sql( $language_ids, $limit, $args );
-
-		if ( empty( $sql ) ) {
-			return array();
-		}
-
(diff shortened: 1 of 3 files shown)


include/translatable-object.php
```diff
@@ -373,12 +373,7 @@ public function get_objects_with_no_lang( $limit, array $args = array() ) {
 			return array();
 		}
 
-		$sql = $this->get_objects_with_no_lang_sql( $language_ids, $limit, $args );
-
-		if ( empty( $sql ) ) {
-			return array();
-		}
```
Marianne380, line 380: Why remove this?

Chouby: This is exactly what motivated this PR as I was surprised by the fact that `$sql` could be empty. So the purpose of this PR is in fact to make sure that `get_objects_with_no_lang_sql()` always returns a non empty SQL query.

Marianne380 (approved):

PR #1234, opened 2023-03-14
New parameter `hide_default` in `PLL_Model::get_languages_list()`.

Hug0-Drelon:
## What?
Filter out default language from languages list if required.

## Why?
Since the default language is "manually" removed in some places, refactor the logic to clean the code.

## How?
Add a new parameter `hide_default` to `PLL_Model::get_languages_list()` and use it when necessary.

**Note: Unit tests will break until #1228 is merged since it requires `PLL_Language::$is_default` property to be available.**

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -185,9 +185,10 @@ class PLL_Model {
 		$languages = array_filter(
 			$languages,
 			function( $lang ) use ( $args ) {
-				$empty_to_hide   = isset( $args['hide_empty'] ) && $args['hide_empty'] && empty( $lang->get_tax_prop( 'language', 'count' ) );
-				$default_to_hide = isset( $args['hide_default'] ) && $args['hide_default'] && $lang->is_default;
-				return ! ( $empty_to_hide || $default_to_hide );
+				$keep_empty   = empty( $args['hide_empty'] ) || $lang->get_tax_prop( 'language', 'count' );
+				$keep_default = empty( $args['hide_default'] ) || ! $lang->is_default;
+				return $keep_empty && $keep_default;
+
 			}
 		);
 
--- a/include/model.php
+++ b/include/model.php
@@ -116,8 +116,9 @@ class PLL_Model {
 	 * @since 0.1
 	 *
 	 * @param array $args {
-	 *   @type bool  $hide_empty Hides languages with no posts if set to true ( defaults to false ).
-	 *   @type string $fields    Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 *   @type bool  $hide_empty   Hides languages with no posts if set to true (defaults to `false`).
+	 *   @type bool  $hide_default Hides default language from the list (default to `false`).
+	 *   @type string $fields      Returns only that field if set; {@see PLL_Language} for a list of fields.
 	 * }
 	 * @return array List of PLL_Language objects or PLL_Language object properties.
 	 */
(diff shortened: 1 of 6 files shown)


include/model.php
```diff
@@ -186,6 +187,11 @@ public function get_languages_list( $args = array() ) {
 			}
 		}
 
+		// Remove default language if requested.
+		if ( ! empty( $args['hide_default'] ) ) {
+			$languages = wp_list_filter( $languages, array( 'is_default', true ) );
+		}
```
Chouby, line 193: Just a suggestion to merge both filters:
```php
		$languages = array_filter(
			$languages,
			function( $lang ) use ( $args ) {
				return ! ( ( $args['hide_empty'] && $lang->get_tax_prop( 'language', 'count' ) ) || ( $args['hide_default'] && $lang->is_default ) );
			}
		);

File: integrations/wpseo/wpseo.php [hunk 2]
```diff
@@ -265,9 +268,6 @@ public function add_post_type_archive( $str ) {
 			$slug = ( true === $post_type_obj->has_archive ) ? $post_type_obj->rewrite['slug'] : $post_type_obj->has_archive;
 
 			if ( ! wpcom_vip_get_page_by_path( $slug ) ) {
-				// The post type archive in the current language is already added by WPSEO.
-				$languages = wp_list_filter( $languages, array( 'slug' => pll_current_language() ), 'NOT' );
-
```
Screenfeed, line 270: Must not be removed: we want to keep the languages that are not the CURRENT language (!== default).

Screenfeed (changes requested):

include/model.php
Hug0-Drelon: I like it, done in 0ed8f3bc. I've just used variables to make it clearer and checked array keys exist.

include/model.php
```diff
@@ -186,6 +187,11 @@ public function get_languages_list( $args = array() ) {
 			}
 		}
 
+		// Remove default language if requested.
+		if ( ! empty( $args['hide_default'] ) ) {
+			$languages = wp_list_filter( $languages, array( 'is_default', true ) );
```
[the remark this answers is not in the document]
Hug0-Drelon: Oops, RTFM. Thought it was `NOT` by default... Still, I applied @chouby's suggestion.

integrations/wpseo/wpseo.php
Hug0-Drelon: Maybe I missed something, but is it done in line 252?

include/model.php
Hug0-Drelon: Instead of using `isset()`, should we use `wp_parse_args()`?

integrations/wpseo/wpseo.php
Screenfeed: As we discussed, everything has been reverted for this file because we're dealing with the default language on one place and with the current language on the other place.

Screenfeed (approved):

Chouby (changes requested):
I also would like that you add more tests to https://github.com/polylang/polylang/blob/3.3.2/tests/phpunit/tests/test-model.php#L17-L28

Hug0-Drelon:
> I also would like that you add more tests to https://github.com/polylang/polylang/blob/3.3.2/tests/phpunit/tests/test-model.php#L17-L28

Done in 166a992b, let me know if you want to add more.

PR #1242, opened 2023-03-24
Ease creation of 3rd party language taxonomies

Screenfeed:
This PR makes easier for 3rd party developers to translate custom DB tables.

## Register a language taxonomy

This introduces the hook `pll_model_init`: it can be used to register a 3rd party language taxonomy (by using `PLL_Translatable_Objects::register()`).

## Create the language taxonomy term

This also introduces the method `PLL_Model::maybe_add_missing_3rd_party_language_terms()`: it is called right after the hook `pll_model_init`, and creates the terms for 3rd party language taxonomies. Internally it uses a new entry in Polylang's option that contains the list of the 3rd party language taxonomies: if the list of the registered taxonomies (`PLL_Translatable_Objects`) has items that are not in the list from the option, the process begins.

Note that `language` and `term_language` are excluded from this automatic term creation.

## Bug fix

This PR also includes a bug fix in `PLL_Model::update_secondary_language_terms()`: `wp_update_term()` was always called, even when there were no changes in the term. This was because of term slug comparison (both slugs must have the `pll_` prefix.

## Other
The method `PLL_Model::get_default_language()` has been moved up in the class, to be right after `get_language()`. This keeps `maybe_create_language_terms()` and `update_secondary_language_terms()` together.

### DIFF
--- a/include/class-polylang.php
+++ b/include/class-polylang.php
@@ -239,7 +239,7 @@ class Polylang {
 			 */
 			do_action( 'pll_model_init', $model );
 
-			$model->maybe_add_missing_3rd_party_language_terms();
+			$model->maybe_create_language_terms();
 
 			/**
 			 * Fires after the $polylang object is created and before the API is loaded
--- a/include/class-polylang.php
+++ b/include/class-polylang.php
@@ -189,6 +189,7 @@ class Polylang {
 		 * @param string $class either PLL_Model or PLL_Admin_Model
 		 */
 		$class = apply_filters( 'pll_model', PLL_SETTINGS || self::is_wizard() ? 'PLL_Admin_Model' : 'PLL_Model' );
+		/** @var PLL_Model $model */
 		$model = new $class( $options );
 
 		if ( ! $model->has_languages() ) {
@@ -225,6 +226,20 @@ class Polylang {
 		if ( ! empty( $class ) ) {
 			$links_model = $model->get_links_model();
 			$polylang    = new $class( $links_model );
+
+			/**
+			 * Fires after Polylang's model init.
+			 * This is the best place to register a custom table (see `PLL_Model`'s constructor).
+			 * /!\ This hook is fired *before* the $polylang object is available.
+			 * /!\ The languages are also not available yet.
+			 *
+			 * @since 3.4
+			 *
+			 * @param PLL_Model $model Polylang model.
+			 */
+			do_action( 'pll_model_init', $model );
+
+			$model->maybe_add_missing_3rd_party_language_terms();
 
 			/**
 			 * Fires after the $polylang object is created and before the API is loaded
--- a/include/model.php
+++ b/include/model.php
@@ -820,31 +820,16 @@ class PLL_Model {
 		}
 
 		// We have at least one unknown 3rd party language taxonomy.
-		$this->add_missing_secondary_language_terms( $new_taxonomies );
-
-		// Keep the previous values, so this is triggered only once per taxonomy.
-		$this->options['language_taxonomies'] = array_merge( $known_taxonomies, $new_taxonomies );
-		update_option( 'polylang', $this->options );
-	}
-
-	/**
-	 * Adds the missing language terms for the secondary language taxonomies.
-	 *
-	 * @since 3.4
-	 *
-	 * @param string[]|null $taxonomies Optional. List of language taxonomies to deal with. An empty value means all of
-	 *                                  them. Defauls to all taxonomies.
-	 * @return void
-	 *
-	 * @phpstan-param array<non-empty-string>|null $taxonomies
-	 */
-	public function add_missing_secondary_language_terms( array $taxonomies = null ) {
 		foreach ( $this->get_languages_list() as $language ) {
-			$this->update_secondary_language_terms( $language->slug, $language->name, $language, $taxonomies );
+			$this->update_secondary_language_terms( $language->slug, $language->name, $language, $new_taxonomies );
 		}
 
 		// Clear the cache, so the new `term_id` and `term_taxonomy_id` appear in the languages list.
 		$this->clean_languages_cache();
+
+		// Keep the previous values, so this is triggered only once per taxonomy.
+		$this->options['language_taxonomies'] = array_merge( $known_taxonomies, $new_taxonomies );
+		update_option( 'polylang', $this->options );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -734,6 +734,35 @@ class PLL_Model {
 	}
 
 	/**
+	 * Assigns the default language to objects in mass.
+	 *
+	 * @since 1.2
+	 * @since 3.4 Moved from PLL_Admin_Model class.
+	 * @since 3.4 Change method signature.
+	 *
+	 * @param int $limit Max number of posts or terms to return. Defaults to -1 (no limit).
+	 * @return void
+	 *
+	 * @phpstan-param -1|positive-int $limit
+	 */
+	public function set_language_in_mass( $limit = -1 ) {
+		$nolang = $this->get_objects_with_no_lang( $limit );
+
+		if ( empty( $nolang ) ) {
+			return;
+		}
+
+		/** @var PLL_Language $lang */
+		$lang = $this->get_default_language();
+
+		foreach ( $this->translatable_objects as $type => $object ) {
+			if ( ! empty( $nolang[ "{$type}s" ] ) ) {
+				$object->set_language_in_mass( $nolang[ "{$type}s" ], $lang );
+			}
+		}
+	}
+
+	/**
 	 * Filters the ORDERBY clause of the languages query.
 	 * This allows to order languages by `term_group` and `term_id`.
 	 *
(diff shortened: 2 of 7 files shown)


include/class-polylang.php
```diff
@@ -189,6 +189,7 @@ public function init() {
 		 * @param string $class either PLL_Model or PLL_Admin_Model
 		 */
 		$class = apply_filters( 'pll_model', PLL_SETTINGS || self::is_wizard() ? 'PLL_Admin_Model' : 'PLL_Model' );
+		/** @var PLL_Model $model */
```
Hug0-Drelon, line 192: Shouldn't we check the type of the object returned instead? Since it comes from a filter ...

include/model.php
Hug0-Drelon, line 1: I wonder if we should make `$translatable_objects` private and add a wrapper around `PLL_Translatable_Objects::register()`? 
This way we could sanitize the object registered and avoid third party to play with the `PLL_Model::$translatable_objects` prop.

Hug0-Drelon (commented):
I'm not a big fan of calling "secondary" language terms or taxonomies those about anything but posts. This is a bit confusing with secondary language objects.
What do you think of something like "subsidiary" language taxonomies for instance?

Hug0-Drelon:
Also, since we are creating the terms, should we register the taxonomies as well?

include/model.php
Screenfeed: @Hug0-Drelon you mean something like this?

```php
class PLL_Model {
	public function register_translatable_object( PLL_Translatable_Object $object ) {
		$this->translatable_objects->register( $object );
	}
}
```

Why not. I'm not sure to see what could be sanitized though 🤔

Screenfeed:
> Also, since we are creating the terms, should we register the taxonomies as well?

Like this?

```php
abstract class PLL_Translatable_Object {
	public function __construct( PLL_Model &$model ) {
		// ...
		$model->translatable_objects->register( $this );
	}
}
```

That would make sense indeed.

include/class-polylang.php
Hug0-Drelon: As we discussed, it's not useful.

include/model.php
Screenfeed: As we discussed, it seems that it doesn't bring much value since:
- `PLL_Translatable_Objects::register()` is already type-hinted,
- the target audience is developers,
- we will provide an example implementation.

Hug0-Drelon:
> > Also, since we are creating the terms, should we register the taxonomies as well?
> 
> Like this?
> 
> ```
> abstract class PLL_Translatable_Object {
> 	public function __construct( PLL_Model &$model ) {
> 		// ...
> 		$model->translatable_objects->register( $this );
> 	}
> }
> ```
> 
> That would make sense indeed. This would also remove the need of a wrapper for `register()`. On the other hand this will forbid us to make `$translatable_objects` private, unless we create a wrapper of course, but I don't see the point in doing so.

I meant to call `register_taxonomy()` but it's already done.

Hug0-Drelon (approved):

include/model.php
```diff
@@ -748,53 +763,86 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+	 * @param string[]|null $taxonomies Optional. List of language taxonomies to deal with. An empty value means all of
+	 *                                  them. Defauls to all taxonomies.
+	 * @return void
+	 *
+	 * @phpstan-param array<non-empty-string>|null $taxonomies
 	 */
-	public function get_default_language() {
-		if ( empty( $this->options['default_lang'] ) ) {
-			return false;
+	public function add_missing_secondary_language_terms( array $taxonomies = null ) {
```
Chouby, line 812: I propose to rename to `create_language_terms()`. Is it useful for this method to be public? Is even this method useful now ? Would it be appropriate merge it to the calling method?

include/model.php
```diff
@@ -748,53 +763,86 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+	 * @param string[]|null     $taxonomies Optional. List of language taxonomies to deal with. An empty value means
+	 *                                      all of them. Defauls to all taxonomies.
 	 * @return void
 	 *
 	 * @phpstan-param non-empty-string $slug
 	 * @phpstan-param non-empty-string $name
+	 * @phpstan-param array<non-empty-string>|null $taxonomies
 	 */
-	protected function update_secondary_language_terms( $slug, $name, PLL_Language $language = null ) {
+	protected function update_secondary_language_terms( $slug, $name, PLL_Language $language = null, array $taxonomies = null ) {
```
Chouby, line 837: It looks strange to me to strict type to `array` and then passing `null`.
```suggestion
	protected function update_secondary_language_terms( $slug, $name, PLL_Language $language = null, array $taxonomies = array() ) {
```

include/model.php
```diff
@@ -808,7 +856,7 @@ protected function update_secondary_language_terms( $slug, $name, PLL_Language $
 			}
 
 			/** @var PLL_Language $language */
-			if ( $language->slug !== $slug || $language->name !== $name ) {
+			if ( "pll_{$language->slug}" !== $slug || $language->name !== $name ) {
```
Chouby, line 859: You are fixing a bug. Could you add a corresponding test? Or fix the tests already covering this line?

Chouby (changes requested):

include/model.php
```diff
@@ -284,6 +284,21 @@ public function get_language( $value ) {
...
+	 * @return PLL_Language|false Default language object, `false` if no language found.
+	 */
+	public function get_default_language() {
+		if ( empty( $this->options['default_lang'] ) ) {
+			return false;
+		}
+
+		return $this->get_language( $this->options['default_lang'] );
+	}
+
```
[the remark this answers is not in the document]
Screenfeed: I does, I only moved it in the class.
>The method `PLL_Model::get_default_language()` has been moved up in the class, to be right after `get_language()`. This keeps `maybe_add_missing_3rd_party_language_terms()`, `add_missing_secondary_language_terms()`, and `update_secondary_language_terms()` together.

include/model.php
```diff
@@ -748,53 +763,86 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+		if ( empty( $registered_taxonomies ) ) {
+			// No 3rd party language taxonomies.
+			return;
 		}
 
-		// Clear the cache, so the new `term_id` and `term_taxonomy_id` appear in the languages list.
-		$this->clean_languages_cache();
+		// We have at least one 3rd party language taxonomy.
+		$known_taxonomies = ! empty( $this->options['3rd_party_taxonomies'] ) && is_array( $this->options['3rd_party_taxonomies'] ) ? $this->options['3rd_party_taxonomies'] : array();
```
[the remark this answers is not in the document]
Screenfeed: I need it because it may not be set (until `PLL_Options`).

include/model.php
```diff
@@ -748,53 +763,86 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+	 * Maybe adds the missing language terms for 3rd party language taxonomies.
 	 *
 	 * @since 3.4
 	 *
 	 * @return void
 	 */
-	public function add_missing_secondary_language_terms() {
-		foreach ( $this->get_languages_list() as $language ) {
-			$this->update_secondary_language_terms( $language->slug, $language->name, $language );
+	public function maybe_add_missing_3rd_party_language_terms() {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e9cbdcd](https://github.com/polylang/polylang/pull/1242/commits/e9cbdcd2e92fdcacfe79d7caa4ef7d55787f83ab).

include/model.php
```diff
@@ -748,53 +763,86 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
...
+		if ( empty( $new_taxonomies ) ) {
+			// No new 3rd party language taxonomies.
+			return;
+		}
+
+		// We have at least one unknown 3rd party language taxonomy.
+		$this->add_missing_secondary_language_terms( $new_taxonomies );
+
+		// Keep the previous values, so this is triggered only once per taxonomy.
+		$this->options['3rd_party_taxonomies'] = array_merge( $known_taxonomies, $new_taxonomies );
```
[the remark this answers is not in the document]
Screenfeed: Renamed in [e9cbdcd](https://github.com/polylang/polylang/pull/1242/commits/e9cbdcd2e92fdcacfe79d7caa4ef7d55787f83ab).

include/model.php
Screenfeed: Not needed anymore indeed.
Removed in [b8df0da](https://github.com/polylang/polylang/pull/1242/commits/b8df0da2d593306e720cac41ad7fb608a971c1c0).

include/model.php
Screenfeed: Right. I think I used `isset( $taxonomies )` at some point, but thought it was not a good idea and didn't changed it for `array()`.
Done in [9a118a7](https://github.com/polylang/polylang/pull/1242/commits/9a118a77bdf7ff32ca551fae5ebcab1832f12d48).

include/model.php
Screenfeed: I'm not sure this worth adding a test.
The bug was only calling `wp_update_term()` with new data === old data, so it was adding unnecessary process but was doing nothing. I can still try to add one that makes sure certain actions are not triggered I guess.

Chouby: forget it.

PR #1243, opened 2023-03-30
Remove PLL_Block_Editor_Filter_Preload_Paths.

Hug0-Drelon:
## What?
Remove backward compatibility for `block_editor_preload_paths` filter.

## Why?
Polylang doesn't support WordPress 5.8 anymore.

## How?
Remove `PLL_Block_Editor_Filter_Preload_Paths` class and related test file and use `block_editor_rest_api_preload_paths` to hook callback directly.


admin/admin-block-editor.php
```diff
@@ -32,7 +32,7 @@ public function __construct( &$polylang ) {
 		$this->model     = &$polylang->model;
 		$this->pref_lang = &$polylang->pref_lang;
 
-		new PLL_Block_Editor_Filter_Preload_Paths( array( $this, 'preload_paths' ), 10, 2 );
+		add_filter( 'block_editor_rest_api_preload_paths', array( $this, 'preload_paths' ), 10, 2 );
```
Chouby, line 35: This won't work as is as the second parameter is no more a `WP_Post` but a `WP_Block_Editor_Context`.

Chouby (changes requested):

admin/admin-block-editor.php
Hug0-Drelon: Indeed ... 🙈 
Fixed in 540178ab

Hug0-Drelon:
@Chouby Also, should I bump the WordPress version in Travis?

PR #1244, opened 2023-03-30
Reorder language terms by slug with language taxonomy first

manooweb:
fixes https://github.com/polylang/polylang-pro/issues/1626

## Prerequisities

As it's explained in the #1626 is required to cause the issue:
- change the order (>0) of the language in the language settings page. It makes the `language` key first in the associative array passed to create the PLL_Language object. See https://github.com/polylang/polylang/blob/master/include/language-factory.php#L69. Otherwise without any order `term_language` key is first.
- have a content type (post or page) which this language is assigned. When we delete terms for the language taxonomies (`language` and `term_language`) See https://github.com/polylang/polylang/blob/master/admin/admin-model.php#L145 `wp_delete_term` WordPress function also removes taxonomies aasigned to content See https://github.com/WordPress/WordPress/blob/6.2/wp-includes/taxonomy.php#L2046-L2049
- In addition Polylang is hooked to `pre_delete_term` action tiggered by `wp_delete_term` WordPress function
See https://github.com/WordPress/WordPress/blob/6.2/wp-includes/taxonomy.php#L2022
and https://github.com/polylang/polylang/blob/3.3.2/include/crud-terms.php#L252-L255 
- Polylang is also hooked to `edited_term_taxonomy` action triggered in `_update_post_term_count` or `_update_generic_term_count` functions See https://github.com/WordPress/WordPress/blob/6.2/wp-includes/taxonomy.php#L4045-L4099 or https://github.com/WordPress/WordPress/blob/6.2/wp-includes/taxonomy.php#L4113-L4126. Polylang cleans the languages cache See https://github.com/polylang/polylang/blob/master/include/model.php#L235-L240

## What happens?

- The `language` term is deleted first by calling `wp_delete_term`  WordPress function.
- `wp_delete_term`  WordPress function then calls `wp_remove_object_terms` to remove language assignmnet to contents.
- `wp_remove_object_terms` calls `wp_update_term_count` which calls `_update_post_term_count` or `_update_generic_term_count` and then triggers the `edited_term_taxonomy` action.
- Polylang then cleans the language cache.
- The 'term_language' term is then deleted. However `wp_delete_term` tiggers `pre_delete_term` action which calls `PLL_CRUD_Terms::delete_term()` then `PLL_Translated_Term::delete_translation()` then `PLL_Translated_Term::get_translation()` then `PLL_Translated_Term::validate_translation()` where `PLL_Model::get_languages_list()` is called.
See https://github.com/polylang/polylang/blob/23b1a797bd3f71690a1ac68d87920340bbf08778/include/translated-object.php#L447
- At this point the languages cache is empty (no more cache and transient exist). So `PLL_Model::get_languages_from_taxonomies()` is called to build again the languages list. However the language term doesn't exist any more. Thus, when `PLL_Language_Factory::get_from_terms()` is called only the `term_language` key exists in the expected array parameter. The notice reported is thus triggered because the `language` key is referenced in the `PLL_Language_Factory::get_from_terms()` code unlike the `term_language` key. See https://github.com/polylang/polylang/blob/master/include/language-factory.php#L72-L74

## Proposal

This PR propose two kind of fixes.
- One to reverse the order of keys in the associative array passed to `PLL_Language_Factory::get_from_terms()`.
It makes `term_language` to be deleted first because it is created in the `PLL_Language` object properties first.
See https://github.com/polylang/polylang/blob/master/include/language-factory.php#L79-L85
- The other is to avoid to call  `PLL_Translated_Term::delete_translation()` on `pre_delete_term` action because a language isn't translatable. In fact it's also the case for all kind of terms. We need to check if a term is translatable before calling `PLL_Translated_Term::delete_translation()`.

### DIFF
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -140,13 +140,13 @@ class PLL_Admin_Model extends PLL_Model {
 		// Delete domain
 		unset( $this->options['domains'][ $lang->slug ] );
 
-		/**
+		/*
 		 * Delete the language itself.
 		 *
 		 * Reverses the language taxonomies order is required to make sure 'language' is deleted in last.
 		 *
 		 * The initial order with the 'language' taxonomy at the beginning of 'PLL_Language::term_props' property
-		 * is done by PLL_Model::filter_language_terms_orderby()
+		 * is done by {@see PLL_Model::filter_language_terms_orderby()}
 		 */
 		foreach ( array_reverse( $lang->get_tax_props( 'term_id' ) ) as $taxonomy_name => $term_id ) {
 			wp_delete_term( $term_id, $taxonomy_name );
--- a/admin/admin-model.php
+++ b/admin/admin-model.php
@@ -140,8 +140,15 @@ class PLL_Admin_Model extends PLL_Model {
 		// Delete domain
 		unset( $this->options['domains'][ $lang->slug ] );
 
-		// Delete the language itself.
-		foreach ( $lang->get_tax_props( 'term_id' ) as $taxonomy_name => $term_id ) {
+		/**
+		 * Delete the language itself.
+		 *
+		 * Reverses the language taxonomies order is required to make sure 'language' is deleted in last.
+		 *
+		 * The initial order with the 'language' taxonomy at the beginning of 'PLL_Language::term_props' property
+		 * is done by PLL_Model::filter_language_terms_orderby()
+		 */
+		foreach ( array_reverse( $lang->get_tax_props( 'term_id' ) ) as $taxonomy_name => $term_id ) {
 			wp_delete_term( $term_id, $taxonomy_name );
 		}
 
--- a/include/crud-terms.php
+++ b/include/crud-terms.php
@@ -78,7 +78,7 @@ class PLL_CRUD_Terms {
 		add_action( 'posts_selection', array( $this, 'unset_tax_query_lang' ), 0 );
 
 		// Deleting terms
-		add_action( 'pre_delete_term', array( $this, 'delete_term' ) );
+		add_action( 'pre_delete_term', array( $this, 'delete_term' ), 10, 2 );
 	}
 
 	/**
@@ -246,17 +246,12 @@ class PLL_CRUD_Terms {
 	 *
 	 * @since 0.1
 	 *
-	 * @param int $term_id
+	 * @param int    $term_id  Id of the term to delete.
+	 * @param string $taxonomy Name of the taxonomy.
 	 * @return void
 	 */
-	public function delete_term( $term_id ) {
-		$term = get_term( $term_id );
-
-		if ( ! $term instanceof WP_Term ) {
-			return;
-		}
-
-		if ( ! $this->model->is_translated_taxonomy( $term->taxonomy ) ) {
+	public function delete_term( $term_id, $taxonomy ) {
+		if ( ! $this->model->is_translated_taxonomy( $taxonomy ) ) {
 			return;
 		}
 
--- a/include/crud-terms.php
+++ b/include/crud-terms.php
@@ -250,6 +250,17 @@ class PLL_CRUD_Terms {
 	 * @return void
 	 */
 	public function delete_term( $term_id ) {
+		$term = get_term( $term_id );
+
+		if ( ! $term instanceof WP_Term ) {
+			return;
+		}
+
+		if ( ! $this->model->is_translated_taxonomy( $term->taxonomy ) ) {
+			return;
+		}
+
+		// Delete translation and relationships only if the term is translatable.
 		$this->model->term->delete_translation( $term_id );
 		$this->model->term->delete_language( $term_id );
 	}
--- a/include/model.php
+++ b/include/model.php
@@ -775,6 +775,9 @@ class PLL_Model {
 	 * Filters the ORDERBY clause of the languages query.
 	 * This allows to order languages by `term_group` and `term_id`.
 	 *
+	 * Makes the all terms in 'language' taxonomy first in the returned result by respecting the 'term_group' order.
+	 * The goal is not to mix terms between all language taxomonomies.
+	 * 
 	 * @since 3.2.3
 	 *
 	 * @param  string   $orderby    `ORDERBY` clause of the terms query.
@@ -797,10 +800,6 @@ class PLL_Model {
 			return $orderby;
 		}
 
-		/**
-		 * Makes the all terms in 'language' taxonomy first in the returned result by respecting the 'term_group' order.
-		 * The goal is not to mix terms between all language taxomonomies.
-		 */
 		return sprintf( 'tt.taxonomy = "language" DESC, %1$s.term_group, %1$s.term_id', $matches['alias'] );
 	}
 
--- a/include/model.php
+++ b/include/model.php
@@ -913,13 +913,10 @@ class PLL_Model {
 		}
 
 		// Restore the right order after the first `array_reverse()`.
-		// Also keep the terms order with the 'language' taxonomy first.
+		// Also keep the terms order with the 'term_language' taxonomy first.
 		$terms_by_slug = array_reverse(
 			array_map(
 				function( $term_by_slug ) {
-					if ( 0 === $term_by_slug['language']->term_group ) {
-						return $term_by_slug;
-					}
 					return array_reverse( $term_by_slug );
 				},
 				$terms_by_slug
--- a/include/model.php
+++ b/include/model.php
@@ -797,7 +797,7 @@ class PLL_Model {
 			return $orderby;
 		}
 
-		return sprintf( '%1$s.term_group, %1$s.term_id', $matches['alias'] );
+		return sprintf( 'tt.taxonomy = "language", %1$s.term_group, %1$s.term_id', $matches['alias'] );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -914,14 +914,7 @@ class PLL_Model {
 
 		// Restore the right order after the first `array_reverse()`.
 		// Also keep the terms order with the 'term_language' taxonomy first.
-		$terms_by_slug = array_reverse(
-			array_map(
-				function( $term_by_slug ) {
-					return array_reverse( $term_by_slug );
-				},
-				$terms_by_slug
-			)
-		);
+		$terms_by_slug = array_reverse( array_map( 'array_reverse', $terms_by_slug ) );
 
 		/**
 		 * @var (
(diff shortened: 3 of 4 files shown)


include/crud-terms.php
```diff
@@ -250,6 +250,17 @@ public function unset_tax_query_lang() {
 	 * @return void
 	 */
 	public function delete_term( $term_id ) {
+		$term = get_term( $term_id );
+
+		if ( ! $term instanceof WP_Term ) {
+			return;
+		}
```
Chouby, line 257: You don't need this. The taxonomy is accessible as second param of the hook.

include/model.php
```diff
@@ -842,7 +842,18 @@ protected function get_languages_from_taxonomies() {
...
+			array_map(
+				function( $term_by_slug ) {
+					if ( 0 === $term_by_slug['language']->term_group ) {
+						return $term_by_slug;
+					}
+					return array_reverse( $term_by_slug );
+				},
+				$terms_by_slug
+			)
+		);
```
Chouby, line 856: These `array_reverse()` things are getting too complex. I propose to remove all of them and correctly sort terms directly in the database query by replacing [the orderby](https://github.com/polylang/polylang/blob/ca8c61187157d227b43c916c8b15bcb9c49d0340/include/model.php#L747) with something like:
```php
return sprintf( 'tt.taxonomy = "language", %1$s.term_group, %1$s.term_id', $matches['alias'] );
```

Chouby (changes requested):

Chouby (changes requested):
Also please add a test reproducing the bug.

include/crud-terms.php
manooweb: Done in 6b2153a998a661fedd33ddc022ad59e45cccd6f8

include/model.php
Screenfeed: Ha, I see what you're trying to do. That may simplify things indeed.
However, I'm not sure if you want `language` terms first or last here, but they will be last with this.
To get them first, I did that:
```sql
ORDER BY tt.taxonomy = 'language' DESC, t.term_group ASC, t.term_id ASC
```

manooweb: 🤔 it can' work as we expected. 

Some tests as `Model_Test::test_languages_list_order()` confirm we break the languages list order instead of the solution I proposed at the beginning.

I will explain with more detailed explanations.

⚠we can't put `ASC` or `DESC` at the end of the orderby clause we modify because `get_terms()` add it with `ASC` value by default. If we want to change it, we need to use the `order` parameter passed to `get_terms()`
It is also useless to use ASC because it's the MySQL default.
See 
- https://github.com/WordPress/WordPress/blob/6.2/wp-includes/class-wp-term-query.php#L454 
- https://github.com/WordPress/WordPress/blob/6.2/wp-includes/class-wp-term-query.php#L738
Default values are set by the `parse_query()` call See https://github.com/WordPress/WordPress/blob/6.2/wp-includes/class-wp-term-query.php#L351

manooweb: So I propose a mixed solution.
A part of the code I proposed but simplified because the `get_terms()` is also order by `tt.taxonomy = "language"` which puts all the `language` terms at the beginning of the result. Thus we always have to reverse the order of `language` and `term_language` keys.

Screenfeed: > ⚠we can't put `ASC` or `DESC` at the end of the orderby clause we modify because `get_terms()` add it with `ASC` value by default.

This doesn't work?
```php
return sprintf( 'tt.taxonomy = "language" DESC, %1$s.term_group, %1$s.term_id', $matches['alias'] );
```

manooweb:
> Also please add a test reproducing the bug.

Done in dfdac0e5371db6a5e082387602fd99e94a10bd2d

🤔 strange, it triggers an error at the end of the class test execution when all languages are deleted because the latest language terms seems not to exists while it should.

include/model.php
manooweb: > > ⚠we can't put `ASC` or `DESC` at the end of the orderby clause we modify because `get_terms()` add it with `ASC` value by default.
> 
> This doesn't work?
> 
> ```
> return sprintf( 'tt.taxonomy = "language" DESC, %1$s.term_group, %1$s.term_id', $matches['alias'] );
> ```

No! It breaks the languages list order (frontend) and breaks the test I quoted.

include/model.php
```diff
@@ -842,7 +842,15 @@ protected function get_languages_from_taxonomies() {
...
-		$terms_by_slug = array_reverse( $terms_by_slug );
+		// Also keep the terms order with the 'term_language' taxonomy first.
+		$terms_by_slug = array_reverse(
+			array_map(
+				function( $term_by_slug ) {
+					return array_reverse( $term_by_slug );
+				},
+				$terms_by_slug
+			)
+		);
```
Screenfeed, line 853: I think it can be simplified:
```suggestion
		$terms_by_slug = array_reverse(
			array_map( 'array_reverse', $terms_by_slug )
		);
```

Screenfeed (changes requested):

Screenfeed (approved):

Chouby (changes requested):

include/model.php
```diff
@@ -797,7 +797,7 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
 			return $orderby;
 		}
 
-		return sprintf( '%1$s.term_group, %1$s.term_id', $matches['alias'] );
+		return sprintf( 'tt.taxonomy = "language", %1$s.term_group, %1$s.term_id', $matches['alias'] );
```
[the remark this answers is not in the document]
manooweb: Done in 47209bb0a61db4d44cbada9d22d3725591168d99

include/model.php
```diff
@@ -913,7 +913,8 @@ protected function get_languages_from_taxonomies() {
 		}
 
 		// Restore the right order after the first `array_reverse()`.
-		$terms_by_slug = array_reverse( $terms_by_slug );
+		// Also keep the terms order with the 'term_language' taxonomy first.
+		$terms_by_slug = array_reverse( array_map( 'array_reverse', $terms_by_slug ) );
```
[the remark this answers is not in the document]
manooweb: Done in 47209bb0a61db4d44cbada9d22d3725591168d99

tests/phpunit/tests/test-create-delete-languages.php
```diff
@@ -267,4 +267,26 @@ public function test_create_language_object_without_term_language_tax() {
...
+
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR', array( 'term_group' => 1 ) );
+
+		$fr = self::factory()->post->create();
+		self::$model->post->set_language( $fr, 'fr' );
+
+		$lang = self::$model->get_language( 'fr' );
+		self::$model->delete_language( $lang->term_id );
+		$this->assertCount( 1, self::$model->get_languages_list() );
```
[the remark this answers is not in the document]
manooweb: Solve by adding a clean up in `tear_down()` method. Done in 1a66c2d139964b5864517dc5a56992a064d52834

admin/admin-model.php
```diff
@@ -140,8 +140,15 @@ public function delete_language( $lang_id ) {
...
-		// Delete the language itself.
-		foreach ( $lang->get_tax_props( 'term_id' ) as $taxonomy_name => $term_id ) {
+		/**
+		 * Delete the language itself.
+		 *
+		 * Reverses the language taxonomies order is required to make sure 'language' is deleted in last.
+		 *
+		 * The initial order with the 'language' taxonomy at the beginning of 'PLL_Language::term_props' property
+		 * is done by PLL_Model::filter_language_terms_orderby()
+		 */
```
Chouby, line 150: `/**` should be used for doc blocks only.
```suggestion
		/*
		 * Delete the language itself.
		 *
		 * Reverses the language taxonomies order is required to make sure 'language' is deleted in last.
		 *
		 * The initial order with the 'language' taxonomy at the beginning of 'PLL_Language::term_props' property
		 * is done by {@see PLL_Model::filter_language_terms_orderby()}.
		 */
```

include/model.php
```diff
@@ -797,7 +797,11 @@ public function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
 			return $orderby;
 		}
 
-		return sprintf( '%1$s.term_group, %1$s.term_id', $matches['alias'] );
+		/**
+		 * Makes the all terms in 'language' taxonomy first in the returned result by respecting the 'term_group' order.
+		 * The goal is not to mix terms between all language taxomonomies.
+		 */
```
Chouby, line 803: `/**` should be used for doc blocks only. I suggest to merg this comment with th method description.

tests/phpunit/tests/test-create-delete-languages.php
```diff
@@ -263,8 +324,114 @@ public function test_create_language_object_without_term_language_tax() {
...
+		$this->assertTrue( $this->pll_env->model->add_language( $args ) );
+		$this->assertEquals( 'en', $this->pll_env->options['default_lang'] );
+
+		$fr = self::factory()->post->create();
+		$this->pll_env->model->post->set_language( $fr, 'fr' );
+
+		$lang = $this->pll_env->model->get_language( 'fr' );
+		$this->pll_env->model->delete_language( $lang->term_id );
+		$this->assertCount( 3, $this->pll_env->model->get_languages_list() );
+		remove_all_actions( 'pre_delete_term' );
```
Chouby, line 435: This is useless. Your hook is remove in `tear_down()` by WP.
```suggestion
```

tests/phpunit/tests/test-create-delete-languages.php
```diff
@@ -229,14 +246,49 @@ function ( $term_id, $tt_id, $taxonomy ) {
...
+
+		$args = array(
+			'name'       => 'Español',
+			'slug'       => 'es',
+			'locale'     => 'es_ES',
+			'rtl'        => 0,
+			'flag'       => 'es',
+			'term_group' => 3,
+		);
+		$this->assertTrue( $this->pll_env->model->add_language( $args ) );
```
Chouby, line 291: Could you factorize? You are using the same code in several tests.

Chouby (changes requested):

tests/phpunit/tests/test-create-delete-languages.php
manooweb: No!
In fact, action declared in this test is run when we delete the all languages in the `tear_down()` method which triggers a notice on `options['default_lang']` when we delete the last language because this option doesn't exist anymore.

Move clean up of this action in the `tear_down()` method and so replace the closure with a new method.

Done in 11b4be70ba5afc07d35339a42df5fd51c83e2d9a

admin/admin-model.php
manooweb: Done in 3dcac07b883dfbb9ab76ddf0e720012a007340b8

PR #1250, opened 2023-04-03
Allow to get a language by `term_taxonomy_id`

Screenfeed:
This PR allows to use `PLL_Model::get_language()` by using a `term_taxonomy_id` without conflicting with `term_id`.
This is useful to prevent us from looping the languages list again each time we want to fetch a language by its `term_taxonomy_id`.

To do so, prefix the ID with `tt:`. Ex: `"tt:{$tt_id}"`.

Note: this works only for `language`, not for `term_language` nor any secondary language taxonomy: this doesn't seem useful to do the same for the other taxonomies.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -260,8 +260,11 @@ class PLL_Model {
 	 * Returns the language by its term_id, tl_term_id, slug or locale.
 	 *
 	 * @since 0.1
+	 * @since 3.4 Allow to get a language by `term_taxonomy_id`.
 	 *
-	 * @param mixed $value term_id, tl_term_id, slug or locale of the queried language.
+	 * @param mixed $value `term_id`, `tl_term_id`, `term_taxonomy_id`, `slug`, `locale`, or `w3c` of the queried language.
+	 *                     /!\ For the `term_taxonomy_id`, prefix the ID by `tt:` (ex: `"tt:{$tt_id}"`),
+	 *                     this is to prevent confusion between `term_id` and `term_taxonomy_id`.
 	 * @return PLL_Language|false Language object, false if no language found.
 	 */
 	public function get_language( $value ) {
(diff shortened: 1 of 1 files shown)


include/model.php
```diff
@@ -274,6 +277,7 @@ public function get_language( $value ) {
 				foreach ( $lang->get_tax_props( 'term_id' ) as $term_id ) {
 					$this->cache->set( 'language:' . $term_id, $lang );
 				}
+				$this->cache->set( 'language:tt:' . $lang->get_tax_prop( 'language', 'term_taxonomy_id' ), $lang );
```
Chouby, line 280: Maybe we could move this inside the loop above.

Chouby (changes requested):

include/model.php
Screenfeed: I didn't think it was useful but if you think it is, let's move it.
On 2nd thought, either it is useful or not, I think we should do that for consistency.
Done in [107e031](https://github.com/polylang/polylang/pull/1250/commits/107e031398ab2e49225e88fb61acd7afad8d6201).

include/model.php
```diff
@@ -260,8 +260,11 @@ public function get_terms_args( $args ) {
 	 * Returns the language by its term_id, tl_term_id, slug or locale.
 	 *
 	 * @since 0.1
+	 * @since 3.4 Allow to get a language by `term_taxonomy_id`.
 	 *
-	 * @param mixed $value term_id, tl_term_id, slug or locale of the queried language.
+	 * @param mixed $value `term_id`, `tl_term_id`, `term_taxonomy_id`, `slug`, `locale`, or `w3c` of the queried language.
```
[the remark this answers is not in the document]
Screenfeed: Done in [107e031](https://github.com/polylang/polylang/pull/1250/commits/107e031398ab2e49225e88fb61acd7afad8d6201).
I also cleanup a bit.

Issue #1241, opened 2023-03-23
Array to string conversion in class-wp-debug-data.php


Screenfeed:
When printing the values of the properties of each language in the site health page, the property `PLL_Language::$term_props` triggers a PHP notice because it is an array because the site health supports arrays that are only 1 level deep.

Maybe we could display something like this:
```
term_props
    language/term_id: xxx
    language/term_taxonomy_id: xxx
    language/count: xx
    term_language/term_id: xxx
    term_language/term_taxonomy_id: xxx
    term_language/count: xx
```

Chouby:
We can even reduce and avoid to display "term_props" itself.

Screenfeed:
I think it can be useful, for example to know if all language taxonomies are well registered and have their terms.

Chouby:
I meant output like this :
```
language/term_id: xxx
language/term_taxonomy_id: xxx
language/count: xx
term_language/term_id: xxx
term_language/term_taxonomy_id: xxx
term_language/count: xx
```

Screenfeed:
OK, I see.

PR #1253, opened 2023-04-04
Fix the display of PLL_Language::$term_props in the health site.

Marianne380:
Fix https://github.com/polylang/polylang/issues/1241

Format the values of `term_props` to have the following display : 
```
language/term_id: xxx
language/term_taxonomy_id: xxx
language/count: xx
term_language/term_id: xxx
term_language/term_taxonomy_id: xxx
term_language/count: xx
```

And add a test.

### DIFF
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -318,7 +318,7 @@ class PLL_Admin_Site_Health {
 
 			$keys_with_language_taxonomy = array_map(
 				function ( $key, $language_taxonomy ) {
-					return $language_taxonomy . '/' . $key;
+					return "{$language_taxonomy}/{$key}";
 				},
 				array_keys( $item ),
 				$language_taxonomy_array
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -305,7 +305,7 @@ class PLL_Admin_Site_Health {
 	/**
 	 * Adds term props data to the info languages array.
 	 *
-	 * @since 3.3
+	 * @since 3.4
 	 *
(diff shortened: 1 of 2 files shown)


modules/site-health/admin-site-health.php
```diff
@@ -297,6 +302,36 @@ public function info_languages( $debug_info ) {
...
+	 */
+	protected function get_info_term_props( $value ) {
+		$return_value = array();
+
+		foreach ( $value as $language_taxonomy => $item ) {
+			$language_taxonomy_array = array_fill( 0, count( $item ), $language_taxonomy );
+
+			$keys_with_language_taxonomy = array_map(
+				function ( $key, $language_taxonomy ) {
+					return $language_taxonomy . '/' . $key;
```
Screenfeed, line 321: Not mandatory but when we have 2 or more concatenations, we should prefer double quoted strings. From what I know (I may be wrong), it is a bit more efficient.
```suggestion
					return "{$language_taxonomy}/{$key}";
```

modules/site-health/admin-site-health.php
```diff
@@ -297,6 +302,36 @@ public function info_languages( $debug_info ) {
...
+				},
+				array_keys( $item ),
+				$language_taxonomy_array
+			);
+
+			$value = array_combine( array_values( $keys_with_language_taxonomy ), array_values( $item ) );
```
Screenfeed, line 327: `array_values()` isn't required: `array_combine()` doesn't care about the array keys.
```suggestion
			$value = array_combine( $keys_with_language_taxonomy, $item );
```

Screenfeed (changes requested):
Nice job 👍

PR #1258, opened 2023-04-05
Fix infinite loop in wizard 🧙‍♂️

### LINKED ISSUE #1635: Media Library Language Dropdown
**Is your feature request related to a problem? Please describe.**
No. This is known behavior of the plugin based off documentation and user experience:
Image

Users are trying to use shared images between translated pages, such as stock images or limited time events where a translated image is not provided by the vendor. We would like to reduce duplicate resources and additional file space being used by media not being shared where possible. 

**Describe the solution you'd like**
We would like an additional dropdown beside the date dropdown in the media selection added (similar to screenshot below)
Image

This dropdown would display as follows:
- All Languages
- Language 1
- Language 2
- Language 3
- etc.

Ideally this dropdown would function similar to the language selector in the top menu (Example screenshot below)
Image
 
**Describe alternatives you've considered**
Adding an additional tab or Action option to the media selection screen are possible alternatives. Also updating the documentation with a guide for how to share images by uploading to the media library prior to adding to a post/page. This would allow technical staff to provide a walkthrough to users to adjust workflow.

**Additional context, screenshots**
Screenshots included where relevant.

Screenfeed:
Fixes [#1635](https://github.com/polylang/polylang-pro/issues/1635).

This PR fixes 2 bugs that can trigger an infinite loop when assigning a language to untranslated objects in the wizard.

## Cache issue

The main difference between `settings.php` and the wizard is this:
```php
while ( $nolang = $this->model->get_objects_with_no_lang( 1000 ) ) {
```
Since we expect lots of items in the wizard, we create chunks of 1000 items and loop through the chunks.
The issue is that `PLL_Translatable_Object::query_objects_with_no_lang()` stores the result into a cache but `set_language_in_mass()` doesn't clean this cache. This leads to `get_objects_with_no_lang()` to return the same results over and over instead of returning the values for the next chunk.

### Solution

Invalidate our cache at the end of `PLL_Translatable_Object::set_language_in_mass()`.

## Object types issue

#1242 re-introduced `PLL_Model::set_language_in_mass()` and used it in `settings.php`, but forgot to use it in the wizard as well. This can lead to `get_objects_with_no_lang()` to return results for a custom DB table, but these results would never be processed, leading again to an infinite loop.

### Solution

The wizard still needs to loop through chunks of IDs, so why not make it internal to `PLL_Model::get_objects_with_no_lang()`?
`PLL_Model::get_objects_with_no_lang()` and `PLL_Model::set_language_in_mass()` have been revamped:
- The `1000` limit is now internal.
- Instead of using `while`, `set_language_in_mass()` is now recursive. This prevents some unecessary SQL queries by using `get_objects_with_no_lang()` only for the types that may have untranslated items.

PS: I suggest to move all the "no_lang/in_mass" stuff from `PLL_Model` to a dedicated class in a future PR.

### DIFF
--- a/include/model.php
+++ b/include/model.php
@@ -775,14 +775,14 @@ class PLL_Model {
 	 * @param PLL_Language|null $lang  Optional. The language to assign to objects. Defaults to `null` (default language).
 	 * @param string[]          $types Optional. Types to handle (@see PLL_Translatable_Object::get_type()). Defaults
 	 *                                 to an empty array (all types).
-	 * @return string[] The types that may still have non-translated objects.
+	 * @return void
 	 */
 	public function set_language_in_mass( PLL_Language $lang = null, array $types = array() ) {
 		if ( empty( $lang ) ) {
 			$lang = $this->get_default_language();
 
 			if ( empty( $lang ) ) {
-				return array();
+				return;
 			}
 		}
 
@@ -790,7 +790,7 @@ class PLL_Model {
 		$nolang = $this->get_objects_with_no_lang( 1000, $types );
 
 		if ( empty( $nolang ) ) {
-			return array();
+			return;
 		}
 
 		/**
@@ -815,10 +815,10 @@ class PLL_Model {
 		}
 
 		if ( empty( $types_with_objects ) ) {
-			return array();
+			return;
 		}
 
-		return $this->set_language_in_mass( $lang, $types_with_objects );
+		$this->set_language_in_mass( $lang, $types_with_objects );
 	}
 
 	/**
--- a/include/model.php
+++ b/include/model.php
@@ -770,7 +770,6 @@ class PLL_Model {
 	 * @since 1.2
 	 * @since 3.4 Moved from PLL_Admin_Model class.
 	 *            Removed `$limit` parameter, added `$lang` and `$types` parameters.
-	 *            Returns the types that may still have non-translated objects.
 	 *
 	 * @param PLL_Language|null $lang  Optional. The language to assign to objects. Defaults to `null` (default language).
 	 * @param string[]          $types Optional. Types to handle (@see PLL_Translatable_Object::get_type()). Defaults
--- a/include/model.php
+++ b/include/model.php
@@ -776,8 +776,8 @@ class PLL_Model {
 	 *                                 to an empty array (all types).
 	 * @return void
 	 */
-	public function set_language_in_mass( PLL_Language $lang = null, array $types = array() ) {
-		if ( empty( $lang ) ) {
+	public function set_language_in_mass( $lang = null, array $types = array() ) {
+		if ( ! $lang instanceof PLL_Language ) {
 			$lang = $this->get_default_language();
 
 			if ( empty( $lang ) ) {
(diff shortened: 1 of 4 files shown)


include/model.php
```diff
@@ -747,28 +769,56 @@ public function get_terms_with_no_lang( $taxonomies, $limit ) {
...
-	 * @param int $limit Max number of posts or terms to return. Defaults to -1 (no limit).
-	 * @return void
+	 *            Removed `$limit` parameter, added `$lang` and `$types` parameters.
+	 *            Returns the types that may still have non-translated objects.
 	 *
-	 * @phpstan-param -1|positive-int $limit
+	 * @param PLL_Language|null $lang  Optional. The language to assign to objects. Defaults to `null` (default language).
+	 * @param string[]          $types Optional. Types to handle (@see PLL_Translatable_Object::get_type()). Defaults
+	 *                                 to an empty array (all types).
+	 * @return string[] The types that may still have non-translated objects.
```
Hug0-Drelon, line 778: Not sure, but returning `void` is enough, right?

include/translatable-object.php
```diff
@@ -485,20 +485,27 @@ public function set_language_in_mass( $ids, $lang ) {
 		if ( empty( $tt_id ) ) {
 			return;
 		}
+		$ids = array_map( 'intval', $ids );
```
Hug0-Drelon, line 488: Maybe the sanitization should be done outside `set_language_in_mass()` according to the documentation.

include/model.php
Screenfeed: No, that's what makes the method recursive while doing less db queries.

include/translatable-object.php
Screenfeed: That's how @Chouby reassures himself 😏

Hug0-Drelon: I see 😅

Hug0-Drelon (approved):

include/model.php
```diff
@@ -747,28 +769,55 @@ public function get_terms_with_no_lang( $taxonomies, $limit ) {
...
+	 * @param string[]          $types Optional. Types to handle (@see PLL_Translatable_Object::get_type()). Defaults
+	 *                                 to an empty array (all types).
 	 * @return void
-	 *
-	 * @phpstan-param -1|positive-int $limit
 	 */
-	public function set_language_in_mass( $limit = -1 ) {
-		$nolang = $this->get_objects_with_no_lang( $limit );
+	public function set_language_in_mass( PLL_Language $lang = null, array $types = array() ) {
+		if ( empty( $lang ) ) {
```
Chouby, line 780:
```suggestion
	public function set_language_in_mass( $lang = null, array $types = array() ) {
		if ( ! $lang instanceof PLL_Language ) {
```
Reasoning: This is a public method. If someone calls it with the legacy param, `set_language_in_mass( 1000 )`, we avoid to trigger a fatal.

Chouby (changes requested):

include/model.php
Screenfeed: Done in [acf206e](https://github.com/polylang/polylang/pull/1258/commits/acf206e1effdd2bb39ba2f4ec10c5431c4240fb2).

Chrystll:
I just tested quickly and your PR fixes the issue :+1:

PR #1259, opened 2023-04-07
Backward compatibility for `WP_Block_Editor_Context::$name` prop.

Hug0-Drelon:
## What?
Fixes https://app.travis-ci.com/github/polylang/polylang/jobs/599722687
`WP_Block_Editor_Context::$name` only available since WordPress 6.0. [See](https://github.com/WordPress/wordpress-develop/blob/6.2/src/wp-includes/class-wp-block-editor-context.php#L15-L58).

## How?
Remove test on the property, since if `WP_Block_Editor_Context::$post` is a `WP_Post` then the context is for post block editor.


Chouby (approved):

Chouby:
Folllow-up of #1243

PR #1263, opened 2023-04-13
Load settings views in an action

Chouby:
See https://secure.helpscout.net/conversation/2210277357 for reference.

Combining the filter `pll_settings_tabs` and the action `pll_settings_active_tab_{$active_tab}` allows to easily add a new page to the Polylang admin menu or to remove the existing ones. However, this combination has initially not been thought to replace an existing page.

Typically, if the goal is to replace the table in the translations page, it requires to
- Remove the current tab and add a new one (with a different slug as the current slug will automatically our current table) with the filter `pll_settings_tabs`.
- With the action `pll_settings_active_tab_{$active_tab}`:
  - Instanciate the new translations table.
  - Load the current strings view 

This is OK for Polylang but this doesn't load the import export metaboxes in Polylang Pro. So keeping them would require more uneasy steps.

In this PR, all the views are loaded using the action  `pll_settings_active_tab_{$active_tab}`. This allows to easily unhook them.

So if we take our example above to replace the table in the translations pages, this would require to:
- Unhook the new method `displays_strings_form()`
- With the action `pll_settings_active_tab_strings` (Keeping the current slug) :
  - Instanciate the new translations table.
  - Load the current strings view 

In this case, keeping the `strings` slug allows Polylang Pro to work as expected.

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -291,6 +291,12 @@ class PLL_Settings extends PLL_Admin_Base {
 	 * @return void
 	 */
 	public function display_languages_form() {
+		$action = isset( $_REQUEST['pll_action'] ) ? sanitize_key( $_REQUEST['pll_action'] ) : ''; // phpcs:ignore WordPress.Security.NonceVerification
+		if ( 'edit' === $action && ! empty( $_GET['lang'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			// phpcs:ignore WordPress.Security.NonceVerification, VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+			$edit_lang = $this->model->get_language( (int) $_GET['lang'] );
+		}
+
 		$list_table = new PLL_Table_Languages();
 		$list_table->prepare_items( $this->model->get_languages_list() );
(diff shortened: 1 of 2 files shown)


ouun:
Thanks a lot @Chouby. This would be a huge timesafer and simplify a lot how we can adjust the translations table.
Would be great to see this merged.

Screenfeed (approved):
Nice, backward compat' is kept 👍

Chouby:
@ouun Sorry we had to revert this PR as our QA detected that I broke forms submissions. We will see if we can find solution to this issue in the future. It's however too late for 3.4 now.

ouun:
@Chouby thank you very much for letting me know, what a pitty. However would be great to see it in one of the next versions, or at least a filter hooks that allows to hook into the output of the table rows.

ouun:
@Chouby I am wondering if there is any progress with that or if I can be of help somehow. It is required to extend/replace the translations table. Thanks

ouun:
@Chouby & @Screenfeed I am very sorry to bump this again. But I wonder if it makes to wait anytime longer to this getting merged with a proper solution. Would really like to finalize the package for enhancing the translations screen. 

Thanks and kind regards

PR #1264, opened 2023-04-13
Avoid a fatal error if the WPML registered string option is corrupted

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1654

### DIFF
--- a/modules/wpml/wpml-compat.php
+++ b/modules/wpml/wpml-compat.php
@@ -23,7 +23,7 @@ class PLL_WPML_Compat {
 	 *
 	 * @var array
 	 */
-	protected static $strings;
+	protected static $strings = array();
 
 	/**
 	 * @var PLL_WPML_API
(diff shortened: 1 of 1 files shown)


sebastienserre (approved):
This PR is solving the issue on customer website

modules/wpml/wpml-compat.php
```diff
@@ -23,7 +23,7 @@ class PLL_WPML_Compat {
 	 *
 	 * @var array
 	 */
-	protected static $strings;
+	protected static $strings = array();
```
Screenfeed, line 26: Doesn't hurt but wasn't required.

Screenfeed (approved):

modules/wpml/wpml-compat.php
Chouby: I thought the same ;)

PR #1268, opened 2023-04-28
Fix pages sorting with get_pages() in WP 6.3

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/1617

See https://core.trac.wordpress.org/ticket/12821.

Now that `get_pages()` uses a `WP_Query`, our pages are always filtered by the current language, even if we pass a language argument to `get_pages()`.

A [PR](https://github.com/WordPress/wordpress-develop/pull/4386) has been opened with a filter to allow us to filter the `$query_args`.

This PR uses this filter to add the language passed as argument to `get_pages()` to the `WP_Query`.

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -333,9 +333,13 @@ class PLL_Admin_Classic_Editor {
 	 */
 	public function page_attributes_dropdown_pages_args( $dropdown_args, $post ) {
 		$dropdown_args['lang'] = isset( $_POST['lang'] ) ? $this->model->get_language( sanitize_key( $_POST['lang'] ) ) : $this->model->post->get_language( $post->ID ); // phpcs:ignore WordPress.Security.NonceVerification
+
 		if ( ! $dropdown_args['lang'] ) {
 			$dropdown_args['lang'] = $this->pref_lang;
+			return $dropdown_args;
 		}
+
+		$dropdown_args['lang'] = $dropdown_args['lang']->slug;
 
 		return $dropdown_args;
 	}
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -334,11 +334,7 @@ class PLL_Admin_Classic_Editor {
 	public function page_attributes_dropdown_pages_args( $dropdown_args, $post ) {
 		$language = isset( $_POST['lang'] ) ? $this->model->get_language( sanitize_key( $_POST['lang'] ) ) : $this->model->post->get_language( $post->ID ); // phpcs:ignore WordPress.Security.NonceVerification
 
-		if ( empty( $language ) ) {
-			$dropdown_args['lang'] = $this->pref_lang->slug;
-		} else {
-			$dropdown_args['lang'] = $language->slug;
-		}
+		$dropdown_args['lang'] = empty( $language ) ? $this->pref_lang->slug : $language->slug;
 
 		return $dropdown_args;
 	}
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -335,13 +335,10 @@ class PLL_Admin_Classic_Editor {
 		$language = isset( $_POST['lang'] ) ? $this->model->get_language( sanitize_key( $_POST['lang'] ) ) : $this->model->post->get_language( $post->ID ); // phpcs:ignore WordPress.Security.NonceVerification
 
 		if ( empty( $language ) ) {
+			$language = $this->pref_lang;
+		}
 
-			if ( empty( $this->pref_lang ) ) {
-				return $dropdown_args;
-			}
-
-			$dropdown_args['lang'] = $this->pref_lang->slug;
-		} else {
+		if ( ! empty( $language ) ) {
 			$dropdown_args['lang'] = $language->slug;
 		}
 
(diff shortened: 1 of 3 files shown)


Chouby:
@spacedmonkey Thank you!

include/filters.php
```diff
@@ -58,7 +60,11 @@ public function __construct( &$polylang ) {
 		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 10, 2 );
 
 		// Filters the get_pages function according to the current language
-		add_filter( 'get_pages', array( $this, 'get_pages' ), 10, 2 );
+		if ( ! version_compare( $wp_version, '6.3.alpha', '>=' ) ) {
```
Screenfeed, line 63: No need for the negation :)
We should also add a comment about backward compatibility, to ease the removal of the code later.

```suggestion
		// Backward compatibility with WP < 6.3.
		if ( version_compare( $wp_version, '6.3.alpha', '<' ) ) {
```

include/filters.php
```diff
@@ -235,6 +241,25 @@ public function get_pages( $pages, $args ) {
...
+	 * @return array Array of arguments passed to WP_Query with the language.
+	 */
+	public function get_pages_query_args( $query_args, $parsed_args ) {
+		if ( isset( $parsed_args['lang'] ) ) {
+			if ( $parsed_args['lang'] instanceof PLL_Language ) {
+				$parsed_args['lang'] = $parsed_args['lang']->slug;
+			}
+			$query_args['lang'] = $parsed_args['lang'];
+		}
+		return $query_args;
```
Screenfeed, line 260: I think we should reproduce this behavior that we have in `get_pages()`:
```php
	public function get_pages( $pages, $args ) {
		if ( isset( $args['lang'] ) && empty( $args['lang'] ) ) {
			return $pages;
		}

		$language = empty( $args['lang'] ) ? $this->curlang : $this->model->get_language( $args['lang'] );
		// ...
	}
```
Maybe like this?
```suggestion
		if ( isset( $parsed_args['lang'] ) && empty( $parsed_args['lang'] ) ) {
			return $query_args;
		}

		$language = empty( $parsed_args['lang'] ) ? $this->curlang : $this->model->get_language( $parsed_args['lang'] );

		if ( empty( $language ) ) {
			return $query_args;
		}

		$query_args['lang'] = $language->slug;

		return $query_args;
```

What da ya think?

I'm not sure about `$this->model->is_translated_post_type( $args['post_type'] )` though 🤔

include/filters.php
Marianne380: Isn't the comment just below?

Screenfeed: I should buy a boat (and better glasses)

Screenfeed (approved):

Screenfeed (approved):

include/filters.php
```diff
@@ -58,7 +60,11 @@ public function __construct( &$polylang ) {
 		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 10, 2 );
 
 		// Filters the get_pages function according to the current language
-		add_filter( 'get_pages', array( $this, 'get_pages' ), 10, 2 );
+		if ( version_compare( $wp_version, '6.3.alpha', '<' ) ) {
```
Chouby, line 63:
```suggestion
		if ( version_compare( $wp_version, '6.3-alpha', '<' ) ) {
```

include/filters.php
```diff
@@ -235,6 +241,33 @@ public function get_pages( $pages, $args ) {
 		return $pages;
 	}
 
+	/**
+	 * Filters the WP_Query in get_pages() per language.
+	 *
+	 * @since 3.4
```
Chouby, line 247:
```suggestion
	 * @since 3.4.3
```

include/filters.php
```diff
@@ -235,6 +241,33 @@ public function get_pages( $pages, $args ) {
...
+		}
+
+		$language = empty( $parsed_args['lang'] ) ? $this->curlang : $this->model->get_language( $parsed_args['lang'] );
+
+		if ( empty( $language ) || ! $this->model->is_translated_post_type( $parsed_args['post_type'] ) ) {
+			$query_args['lang'] = '';
+			return $query_args;
+		}
+
+		$query_args['lang'] = $language->slug;
```
Chouby, line 266: This is far too complex.
- There is no need to make a specific case for `'lang' => ''`.
- `WP_Query` will check for translated post type iself as it won't handle the parameter if the query is not registered.
- The current language is already handled by our `WP_Query` filter
- If you get a `PLL_Language` object here, that's a bug and it should be fixed in [the right place](https://github.com/polylang/polylang/blob/3.4.2/admin/admin-classic-editor.php#L334) to receive only a language slug.

```suggestion
		if ( isset( $parsed_args['lang'] ) ) {
			$query_args['lang'] = $parsed_args['lang'];
		}
```

Chouby (changes requested):

admin/admin-classic-editor.php
```diff
@@ -332,9 +332,17 @@ public function ajax_posts_not_translated() {
...
+		if ( empty( $language ) ) {
+
+			if ( empty( $this->pref_lang ) ) {
+				return $dropdown_args;
+			}
+
+			$dropdown_args['lang'] = $this->pref_lang->slug;
+		} else {
+			$dropdown_args['lang'] = $language->slug;
 		}
```
Chouby, line 346: Keep it simple:
```suggestion
		if ( empty( $language ) ) {
			$language = $this->pref_lang;
		}

		if ( ! empty( $language ) ) {
			$dropdown_args['lang'] = $language->slug;
		}
```

Chouby (changes requested):

PR #1269, opened 2023-05-05
Fix 3.4 upgrade notice.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/1664

## Why?
`PLL_Admin_Base::init_user()` being hooked to `setup_theme` [here](https://github.com/polylang/polylang/blob/312c7d41e1c9a4abbc9c3ada6146bbf5cfceb07b/admin/admin-base.php#L107) and `PLL_Upgrade::_upgrade()` to `admin_init` [here](https://github.com/polylang/polylang/blob/312c7d41e1c9a4abbc9c3ada6146bbf5cfceb07b/install/upgrade.php#L58), `pll_languages_list` transient isn't deleted soon enough. (See [hook order of precedence](https://lancecleveland.com/2017/10/11/wordpress-hooks-and-filters-order-of-precedence/)).

## How?
Delete the transient directly in `Polylang::init()` when the upgrade is detected.

### DIFF
--- a/tests/phpunit/tests/test-upgrade.php
+++ b/tests/phpunit/tests/test-upgrade.php
@@ -1,0 +1,37 @@
+<?php
+
+class Upgrade_Test extends PLL_UnitTestCase {
+	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
+		parent::wpSetUpBeforeClass( $factory );
+
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR' );
+	}
+
+	public function test_upgrade() {
+		// Fake old version.
+		$options = get_option( 'polylang' );
+		$options['version'] = '3.3';
+		update_option( 'polylang', $options );
+
+		// Fake old transient.
+		self::$model->set_languages_ready();
+		self::$model->get_languages_list();
+		$languages = get_transient( 'pll_languages_list' );
+		foreach ( $languages as $i => $language ) {
+			unset( $language['term_props'] );
+			$languages[ $i ] = $language;
+		}
+		set_transient( 'pll_languages_list', $languages );
+
+		$pll = new Polylang();
+
+		try {
+			$pll->init();
+		} catch ( \Throwable $th ) {
+			$this->assertTrue( false, "Polylang upgrade failed with the error {$th}" );
+		}
+
+		$this->assertTrue( true );
+	}
+}
--- a/tests/phpunit/tests/test-upgrade.php
+++ b/tests/phpunit/tests/test-upgrade.php
@@ -22,21 +22,45 @@ class Upgrade_Test extends PLL_UnitTestCase {
 		$links_model             = new PLL_Links_Default( $model );
 		$admin                   = new PLL_Admin( $links_model );
 
-		// Fake old transient.
-		self::$model->set_languages_ready();
-		self::$model->get_languages_list();
-		$languages = get_transient( 'pll_languages_list' );
-		foreach ( $languages as $i => $language ) {
-			unset( $language['term_props'] );
-			$languages[ $i ] = $language;
-		}
+		// Old transient from 3.3.
+		$en = $admin->model->get_language( 'en' );
+		$transient_3_3 = array(
+			array(
+				'term_id'             => $en->term_id,
+				'name'                => 'English',
+				'slug'                => 'en',
+				'term_group'          => 0,
+				'term_taxonomy_id'    => $en->get_tax_prop( 'language', 'term_taxonomy_id' ),
+				'count'               => $en->get_tax_prop( 'language', 'count' ),
+				'tl_term_id'          => $en->get_tax_prop( 'term_language', 'term_id' ),
+				'tl_term_taxonomy_id' => $en->get_tax_prop( 'term_language', 'term_taxonomy_id' ),
+				'tl_count'            => $en->get_tax_prop( 'term_language', 'term_taxonomy_id' ),
+				'locale'              => 'en_US',
+				'is_rtl'              => 0,
+				'w3c'                 => 'en-US',
+				'facebook'            => 'en_US',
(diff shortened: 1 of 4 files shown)


Marianne380 (changes requested):
It looks like there is some debug code left, otherwise the fix seems good to me.

Marianne380 (approved):

Hug0-Drelon:
> OK to delete the transient at each update. But I would move this to `PLL_Upgrade`. And probably we can remove the current one (and adapt the description of `upgrade_3_4()` as it won't be relevant anymore. Is there a way to test?

Done in 9f7a7c2c.

tests/phpunit/tests/test-upgrade.php
```diff
@@ -0,0 +1,44 @@
...
+		parent::wpSetUpBeforeClass( $factory );
+
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR' );
+	}
+
+	/**
+	 * @ticket #1664 see {https://github.com/polylang/polylang-pro/issues/1664}.
+	 */
+	public function test_upgrade_to_3_4() {
```
Chouby, line 16: I suggest to rename it maybe to `test_delete_transient_at_upgrade_to_3_4()` as here you don't really test the upgrade to 3.4. 
This could however be an idea to test the upgrade to 3.4 in 2 different tests (one for the strings translations, one for the fallback), maybe in a different PR though.

tests/phpunit/tests/test-upgrade.php
```diff
@@ -0,0 +1,44 @@
...
+		new PLL_Upgrade( $admin->options );
+
+		try {
+			do_action( 'setup_theme' ); // See the issue, `PLL_Admin_Base::init_user()` being hooked to `setup_theme`.
+		} catch ( \Throwable $th ) {
+			$this->assertTrue( false, "Polylang admin failed with error: {$th}" );
+		}
+
+		$this->assertTrue( true );
+	}
```
Chouby, line 43: I suggest to simplifify and to add a true assertion even if assing code after the assertion is a bit hacky.
```php
new PLL_Upgrade( $admin->options );
$this->assertFalse( get_transient( 'pll_languages_list' ) );
do_action( 'setup_theme' ); // See the issue, `PLL_Admin_Base::init_user()` being hooked to `setup_theme`.
```

tests/phpunit/tests/test-upgrade.php
```diff
@@ -0,0 +1,44 @@
...
+
+	/**
+	 * @ticket #1664 see {https://github.com/polylang/polylang-pro/issues/1664}.
+	 */
+	public function test_upgrade_to_3_4() {
+		wp_set_current_user( 1 );
+		update_user_meta( get_current_user_id(), 'pll_filter_content', 'en' );
+		$links_model = self::$model->get_links_model();
+		$admin = new PLL_Admin( $links_model );
+		$admin->init();
```
Chouby, line 21: To follow the same process as Polylang, this should be moved between `new PLL_upgrade()` and the `setup_theme` action. Aslo please create a new model.

tests/phpunit/tests/test-upgrade.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+class Upgrade_Test extends PLL_UnitTestCase {
+	protected $options_backup;
+
+	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
+		parent::wpSetUpBeforeClass( $factory );
+
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR' );
```
Chouby, line 10: I guess you can remove one language.

tests/phpunit/tests/test-upgrade.php
```diff
@@ -0,0 +1,44 @@
...
+		// Fake old transient.
+		self::$model->set_languages_ready();
+		self::$model->get_languages_list();
+		$languages = get_transient( 'pll_languages_list' );
+		foreach ( $languages as $i => $language ) {
+			unset( $language['term_props'] );
+			$languages[ $i ] = $language;
+		}
+		$admin->model->clean_languages_cache();
+		set_transient( 'pll_languages_list', $languages );
```
Chouby, line 32: What about filling the the transient with a true array of 1 language from version 3.3?

Chouby (changes requested):

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: Done in 0374d8fe

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: Done in 0374d8fe.

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: Missed that, done in 0374d8fe.

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: Indeed, done in 0374d8fe.

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: One problem though, it does not show the error from the issue.

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: Good idea, done in 5f69d1fa.

tests/phpunit/tests/test-upgrade.php
Hug0-Drelon: I made some changes in 7f4155f6. I propose to test the new transient as well as the option version instead of using `expectNotToPerformAssertions`. What do you think?

PR #1270, opened 2023-05-05
Fix the lang extraction from URL

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1583.

## Problem

A php warning "Attempt to read property `slug` on bool" [here](https://github.com/polylang/polylang/blob/100883399d35b28cbcf8096483f9cea72ae54131/include/crud-terms.php#L197).

## Why this is happening

The language slug in `$args['lang']` is invalid, due to a wrong extraction of it from the current URL.

## Proposed solution

Fix the regex pattern in `PLL_Links_Default::get_language_from_url()`.
The following values will now be considered invalid: `fra`, `fr/`. Previously, those values were considered valid (`fr`).
[Test of the pattern](https://regex101.com/r/fS5mhm/2).

### DIFF
--- a/include/links-default.php
+++ b/include/links-default.php
@@ -84,10 +84,17 @@ class PLL_Links_Default extends PLL_Links_Model {
 	public function get_language_from_url( $url = '' ) {
 		if ( empty( $url ) ) {
 			$url = pll_get_requested_url();
+
+			if ( empty( $url ) ) {
+				return '';
+			}
 		}
 
-		$pattern = '#lang=(' . implode( '|', $this->model->get_languages_list( array( 'fields' => 'slug' ) ) ) . ')#';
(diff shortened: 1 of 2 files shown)


include/links-default.php
```diff
@@ -84,10 +84,17 @@ public function add_paged_to_link( $url, $page ) {
 	public function get_language_from_url( $url = '' ) {
 		if ( empty( $url ) ) {
 			$url = pll_get_requested_url();
+
+			if ( empty( $url ) ) {
```
Hug0-Drelon, line 88: You use a early return for performance? Otherwise it wouldn't pass the `preg_match` anyway, right?

Screenfeed: Yes, it prevents calling `get_languages_list()` + `preg_match()` + `implode()`.

Hug0-Drelon (approved):

PR #1273, opened 2023-05-10
Prevent importing/exporting empty strings

Screenfeed:
This PR prevents `PLL_MO::export_to_db()` from exporting to the DB a translation whose original text is an empty string.
This does the same for `PLL_MO::import_from_db()`, by preventing it from importing from the DB a translation whose original text is an empty string.

### DIFF
--- a/include/mo.php
+++ b/include/mo.php
@@ -29,7 +29,9 @@ class PLL_MO extends MO {
 		 */
 		$strings = array();
 		foreach ( $this->entries as $entry ) {
-			$strings[] = wp_slash( array( $entry->singular, $this->translate( $entry->singular ) ) );
+			if ( '' !== $entry->singular ) {
+				$strings[] = wp_slash( array( $entry->singular, $this->translate( $entry->singular ) ) );
+			}
 		}
 
 		update_term_meta( $lang->term_id, '_pll_strings_translations', $strings );
@@ -53,7 +55,11 @@ class PLL_MO extends MO {
 		}
 
 		foreach ( $strings as $msg ) {
-			$this->add_entry( $this->make_entry( $msg[0], $msg[1] ) );
+			$entry = $this->make_entry( $msg[0], $msg[1] );
+
+			if ( '' !== $entry->singular ) {
+				$this->add_entry( $entry );
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


include/mo.php
```diff
@@ -53,7 +55,11 @@ public function import_from_db( $lang ) {
 		}
 
 		foreach ( $strings as $msg ) {
-			$this->add_entry( $this->make_entry( $msg[0], $msg[1] ) );
+			$entry = $this->make_entry( $msg[0], $msg[1] );
```
Hug0-Drelon, line 58: Is checking `'' !== $msg[0]` enough?

tests/phpunit/tests/test-strings.php
Hug0-Drelon, line 1: Not sure it would be better, but we could create a data provider for the two created tests. What do you think?

include/mo.php
Screenfeed: I prefer to use `make_entry()` because of [this](https://github.com/WordPress/WordPress/blob/fe6b65c44b929ee5583f6146d92388ef43061f9a/wp-includes/pomo/mo.php#L322-L329). Do you think I'm going too far?

tests/phpunit/tests/test-strings.php
Screenfeed: How would it work? I don't see how it could 🤔

include/mo.php
Hug0-Drelon: Oh I see, maybe a bit but we never know 😅

tests/phpunit/tests/test-strings.php
Hug0-Drelon: With a pair of key/value for source/translation and an associated pair of expected result for import/export.
Your assertions might change a bit.
For instance:
```php
return array(
    'with_empty_source`         => array( '', 'empty string fr', array() ),
    'with_non_empty_strings' => array( 'test', 'test fr', array( 'test', 'test fr' ),
);

// Giving the following parameters to tests:
( $source_string, $translated_string, $expected_entry )
```
`$expected_entry` being both `$mo->entries` and the result of `get_term_meta()`.

Screenfeed: I still don't see it XD

Screenfeed: OK I think I see what you mean, but I think it complicates things for no reason by doing twice the same job instead of once.

Hug0-Drelon: As I sais, not sure it'd be better 😅 
The only benefit being to add string to test easily.
But it is indeed complicated.

Hug0-Drelon (approved):

PR #1278, opened 2023-05-15
Fix PHPStan errors related to intval usage

Chouby:
Since the version 1.10.15, PHPStan reports an error when passing `mixed` to `intval()`.
So this PR replaces this function by an explicit type check of th `get_option()` return value.

### DIFF
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -634,46 +634,48 @@ class PLL_Admin_Filters_Term {
 			$lang_slug = sanitize_key( $_POST['term_lang_choice'] ); // phpcs:ignore WordPress.Security.NonceVerification
 			if ( ! empty( $lang_slug ) ) {
 				$lang = $this->model->get_language( $lang_slug );
+				return $lang instanceof PLL_Language ? $lang : null;
 			}
 		}
 
-		elseif ( ! empty( $_POST['inline_lang_choice'] ) && is_string( $_POST['inline_lang_choice'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		if ( ! empty( $_POST['inline_lang_choice'] ) && is_string( $_POST['inline_lang_choice'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 			$lang_slug = sanitize_key( $_POST['inline_lang_choice'] ); // phpcs:ignore WordPress.Security.NonceVerification
 			if ( ! empty( $lang_slug ) ) {
 				$lang = $this->model->get_language( $lang_slug );
+				return $lang instanceof PLL_Language ? $lang : null;
 			}
 		}
 
 		// *Post* bulk edit, in case a new term is created
-		elseif ( isset( $_GET['bulk_edit'], $_GET['inline_lang_choice'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		if ( isset( $_GET['bulk_edit'], $_GET['inline_lang_choice'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 			// Bulk edit does not modify the language
 			if ( -1 === (int) $_GET['inline_lang_choice'] ) { // phpcs:ignore WordPress.Security.NonceVerification
 				$lang = $this->model->post->get_language( $this->post_id );
+				return $lang instanceof PLL_Language ? $lang : null;
 			} elseif ( is_string( $_GET['inline_lang_choice'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 				$lang_slug = sanitize_key( $_GET['inline_lang_choice'] ); // phpcs:ignore WordPress.Security.NonceVerification
 				if ( ! empty( $lang_slug ) ) {
 					$lang = $this->model->get_language( $lang_slug );
+					return $lang instanceof PLL_Language ? $lang : null;
 				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


admin/admin-filters-term.php
```diff
@@ -658,12 +658,18 @@ public function get_inserted_term_language( $lang ) {
 		}
 
 		// Special cases for default categories as the select is disabled.
-		elseif ( ! empty( $_POST['tag_ID'] ) && in_array( intval( get_option( 'default_category' ) ), $this->model->term->get_translations( (int) $_POST['tag_ID'] ), true ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$lang = $this->model->term->get_language( (int) $_POST['tag_ID'] ); // phpcs:ignore WordPress.Security.NonceVerification
+		elseif ( ! empty( $_POST['tag_ID'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$default_term = get_option( 'default_category' );
+			if ( is_int( $default_term ) && in_array( $default_term, $this->model->term->get_translations( (int) $_POST['tag_ID'] ), true ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+				$lang = $this->model->term->get_language( (int) $_POST['tag_ID'] ); // phpcs:ignore WordPress.Security.NonceVerification
+			}
```
Screenfeed, line 665: I'm sure you're aware of this but we're changing the behavior here.
Previously we were entering this statement if we add a `tag_ID` AND if it was in the list of translations. Now it only requires `tag_ID` to not be empty.

Meaning:
Previously: non empty `tag_ID` + not a translation = go to next test about `tax_ID`.
Now: non empty `tag_ID` + not a translation = enter this test and do nothing.

To trigger a bug it would require to have both `tag_ID` and `tax_ID` so I suppose you're expecting this not to happen?

admin/admin-filters-term.php
```diff
@@ -658,12 +658,18 @@ public function get_inserted_term_language( $lang ) {
 		}
 
 		// Special cases for default categories as the select is disabled.
-		elseif ( ! empty( $_POST['tag_ID'] ) && in_array( intval( get_option( 'default_category' ) ), $this->model->term->get_translations( (int) $_POST['tag_ID'] ), true ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$lang = $this->model->term->get_language( (int) $_POST['tag_ID'] ); // phpcs:ignore WordPress.Security.NonceVerification
+		elseif ( ! empty( $_POST['tag_ID'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$default_term = get_option( 'default_category' );
+			if ( is_int( $default_term ) && in_array( $default_term, $this->model->term->get_translations( (int) $_POST['tag_ID'] ), true ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
Screenfeed, line 663: `is_int( $default_term )` so we're neglecting the case where the option is filtered and the returned value is a numeric string?

Chouby: You are right. I changed this in https://github.com/polylang/polylang/pull/1278/commits/4ae13f8681c683073e39548fac36964e3f4aff1a

admin/admin-filters-term.php
Chouby: I rewrote this differently in https://github.com/polylang/polylang/pull/1278/commits/4ae13f8681c683073e39548fac36964e3f4aff1a

admin/admin-filters-term.php
```diff
@@ -634,40 +634,48 @@ public function get_inserted_term_language( $lang ) {
 			$lang_slug = sanitize_key( $_POST['term_lang_choice'] ); // phpcs:ignore WordPress.Security.NonceVerification
 			if ( ! empty( $lang_slug ) ) {
 				$lang = $this->model->get_language( $lang_slug );
+				return $lang instanceof PLL_Language ? $lang : null;
 			}
```
Screenfeed, line 638: If we want to reproduce the exact same behavior, I think we should do like this, here and in the other cases:

```suggestion
			if ( empty( $lang_slug ) ) {
				return null;
			}
			$lang = $this->model->get_language( $lang_slug );
			return $lang instanceof PLL_Language ? $lang : null;
```

include/static-pages.php
```diff
@@ -76,12 +76,21 @@ public function __construct( &$polylang ) {
...
+		}
+
+		$page_on_front = get_option( 'page_on_front' );
+		if ( is_int( $page_on_front ) ) {
+			$this->page_on_front = $page_on_front;
+		}
+
+		$page_for_posts = get_option( 'page_for_posts' );
+		if ( is_int( $page_for_posts ) ) {
+			$this->page_for_posts = $page_for_posts;
```
Screenfeed, line 93: What do you think about this? (just in case it is filtered and returns a numeric string)
```suggestion
		if ( is_numeric( $page_on_front ) ) {
			$this->page_on_front = (int) $page_on_front;
		}

		$page_for_posts = get_option( 'page_for_posts' );
		if ( is_numeric( $page_for_posts ) ) {
			$this->page_for_posts = (int) $page_for_posts;
```

Screenfeed (changes requested):

Screenfeed (approved):

admin/admin-filters-term.php
Screenfeed: Even better indeed.

PR #1280, opened 2023-05-17
Use `sanitize_locale_name()` when available

Screenfeed:
WP 6.2.1 introduced `sanitize_locale_name()` to sanitize a locale code. This PR uses it if it is available.

This PR introduces the function `pll_sanitize_locale_code()`.
This function uses `sanitize_locale_name()` if available. However, the filter `'sanitize_locale_name'` is not polyfilled because it didn't exist before the introduction of the function, so it wouldn't make sense to have it before WP 6.2.1.
This function uses a regex pattern that removes slashes, so it is considered as a sanitization AND an unslashing function.
It is a function instead of a method so it can be liste in PHPCS (see next point).

- Tell PHPCS that `pll_sanitize_locale_code()` is a sanitization + unslashing function.
- Watch `sanitize_locale_name()` for changes.
- `wp_unslash()` is not needed anymore.
- In `PLL_Settings::handle_actions()`, ignore only this specific sniff:

```
Detected usage of a possibly undefined superglobal array index: $_POST['locale']. Use isset() or empty() to check the index exists before using it
```

### DIFF
--- a/include/functions.php
+++ b/include/functions.php
@@ -91,3 +91,20 @@ function pll_use_block_editor_plugin() {
 	 */
 	return class_exists( 'PLL_Block_Editor_Plugin' ) && apply_filters( 'pll_use_block_editor_plugin', ! defined( 'PLL_USE_BLOCK_EDITOR_PLUGIN' ) || PLL_USE_BLOCK_EDITOR_PLUGIN );
 }
+
+/**
+ * Strips out all characters not allowed in a locale code.
+ *
+ * @since 3.5
+ *
+ * @param string $locale_code The locale code to be sanitized.
+ * @return string The sanitized value.
+ */
+function pll_sanitize_locale_code( $locale_code ) {
+	if ( function_exists( 'sanitize_locale_name' ) ) {
+		return (string) sanitize_locale_name( $locale_code );
+	}
+
+	// Limit to A-Z, a-z, 0-9, '_', '-'.
+	return (string) preg_replace( '/[^A-Za-z0-9_-]/', '', $locale_code );
+}
--- a/include/functions.php
+++ b/include/functions.php
@@ -29,9 +29,9 @@ if ( ! function_exists( 'wpcom_vip_get_page_by_path' ) ) {
 if ( ! function_exists( 'sanitize_locale_name' ) ) {
 	/**
 	 * Strips out all characters not allowed in a locale code.
+	 * Backward compatibility with WP < 6.2.1.
 	 *
 	 * @since 3.5
-	 * @since WP 6.2.1
 	 *
 	 * @param string $locale_name The locale name to be sanitized.
 	 * @return string The sanitized value.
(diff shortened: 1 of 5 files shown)


include/functions.php
```diff
@@ -91,3 +91,20 @@ function pll_use_block_editor_plugin() {
...
+
+/**
+ * Strips out all characters not allowed in a locale code.
+ *
+ * @since 3.5
+ *
+ * @param string $locale_code The locale code to be sanitized.
+ * @return string The sanitized value.
+ */
+function pll_sanitize_locale_code( $locale_code ) {
```
Chouby, line 103: I would prefer a polyfill. This is then easier to maintain when we stop supporting old versions. I just have to remove the polyfill without caring wher the function's used as I did in https://github.com/polylang/polylang/commit/925366af3746e575660816911578339136652aab

Chouby (changes requested):

include/functions.php
Screenfeed: Makes sense.
Done in [08238b6](https://github.com/polylang/polylang/pull/1280/commits/08238b644d453f4bb4e45d9635eb5df5a1803a5a).

However, I didn't strictly copied the original function, I added a type-cast `(string) preg_replace( ... )` because PHPStan wasn't happy: `preg_replace()` can return `null` on error.
I chose this instead of adding an exception in `phpstan.neon.dist` because:
- There shouldn't have such errors.
- Easier to remove.

include/functions.php
```diff
@@ -26,6 +26,33 @@ function wpcom_vip_get_page_by_path( $page_path, $output = OBJECT, $post_type =
 	}
 }
 
+if ( ! function_exists( 'sanitize_locale_name' ) ) {
+	/**
+	 * Strips out all characters not allowed in a locale code.
+	 *
+	 * @since 3.5
+	 * @since WP 6.2.1
```
Chouby, line 34: I'd like to add the backward compatibility comment that I usually serch for when I remove old code.
Also, I propose to keep `@since` tag for our own versions.
```suggestion
	 * Strips out all characters not allowed in a locale code.
	 *
	 * Backward compatibility with WP < 6.2.1.
	 *
	 * @since 3.5
```

include/functions.php
```diff
@@ -26,6 +26,33 @@ function wpcom_vip_get_page_by_path( $page_path, $output = OBJECT, $post_type =
...
+	 */
+	function sanitize_locale_name( $locale_name ) {
+		// Limit to A-Z, a-z, 0-9, '_', '-'.
+		$sanitized = (string) preg_replace( '/[^A-Za-z0-9_-]/', '', $locale_name );
+
+		/**
+		 * Filters a sanitized locale name string.
+		 *
+		 * @since 3.5
+		 * @since WP 6.2.1
```
Chouby, line 47:
```suggestion
		 * Filters a sanitized locale name string.
		 *
		 * Backward compatibility with WP < 6.2.1.
		 *
		 * @since 3.5
```

Chouby (changes requested):

include/functions.php
Screenfeed: Done in [92f0b2f](https://github.com/polylang/polylang/pull/1280/commits/92f0b2f013f5527be165cb8f0c1cf3bd73e2f458).

Chouby (approved):

PR #1284, opened 2023-05-24
Use PLL_Cache to cache translated options

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1685

In #1196, I introduced a cache for translated options. This causes a bug as the cache is the same for all sites on a multisite.

In this PR, I use `PLL_Cache` (which is natively site dependent) to handle the cache instead of a simple array.

### DIFF
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -111,7 +111,7 @@ class PLL_Translate_Option {
 		$lang = $GLOBALS['l10n']['pll_string']->get_header( 'Language' );
 
 		$cache = $this->cache->get( $lang );
-		if ( empty( $cache ) ) {
+		if ( false === $cache ) {
 			$cache = $this->translate_string_recursive( $value, $this->keys );
 			$this->cache->set( $lang, $cache );
 		}
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -40,9 +40,9 @@ class PLL_Translate_Option {
 	/**
 	 * Cache for the translated values.
 	 *
-	 * @var array
+	 * @var PLL_Cache
 	 */
-	private $translated_values;
+	private $cache;
 
 	/**
 	 * Constructor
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-multisite-option.php
Hug0-Drelon, line 1: Should we add another language and test the blog name is also translated correctly? Just to cover more cases (maybe in another PR since it's urgent).

include/translate-option.php
Hug0-Drelon, line 1: Should we clean the case for `delete_option_{$option}` too?

include/translate-option.php
```diff
@@ -108,11 +110,13 @@ public function translate( $value ) {
 
 		$lang = $GLOBALS['l10n']['pll_string']->get_header( 'Language' );
 
-		if ( ! isset( $this->translated_values[ $lang ] ) ) {
-			$this->translated_values[ $lang ] = $this->translate_string_recursive( $value, $this->keys );
+		$cache = $this->cache->get( $lang );
+		if ( empty( $cache ) ) {
```
Screenfeed, line 114: Shouldn’t we compare to `false` instead?
```suggestion
		if ( false === $cache ) {
```

tests/phpunit/tests/test-multisite-option.php
```diff
@@ -0,0 +1,50 @@
...
+		}
+
+		/**
+		 * @ticket #1685
+		 * @see https://github.com/polylang/polylang-pro/issues/1685
+		 */
+		public function test_translate_blogname() {
+			$blog_id = self::factory()->blog->create( array( 'title' => 'The new blog to test' ) );
+
+			$language = self::$model->get_language( 'en' );
```
Screenfeed, line 35: The intermediate variable `$language` doesn’t seem necessary.

Screenfeed (changes requested):

tests/phpunit/tests/test-multisite-option.php
Chouby: Yes. I also removed other useless code in https://github.com/polylang/polylang/pull/1284/commits/6e21e008d4b295fd0e5ef4eebb1efb67b746532b

include/translate-option.php
Chouby: Yes! Fixed in https://github.com/polylang/polylang/pull/1284/commits/b5537ff7b1f34b69010068ab7acd5a794cc3b4d0

Screenfeed (approved):

Hug0-Drelon (approved):
One remaining question though. But not blocking.

PR #1285, opened 2023-05-25
Override `set_language_home_urls()` for domains and subdomains

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1688

This PR fixes a regresion introduced in version 3.4.

We used to have a [`PLL_Links_Model::set_home_urls()`](https://github.com/polylang/polylang/blob/3.3.3/include/links-model.php#L167-L170) method calling `front_page_url()` when using a static front page. It used to be overriden in [`PLL_Links_Abstract_Domain`](https://github.com/polylang/polylang/blob/3.3.3/include/links-abstract-domain.php#L56-L59) to avoid calling `front_page_url()` in this specific case.

In version 3.4, we replaced `set_home_urls()` by `set_language_home_urls()` but we forgot to override the method in `PLL_Links_Abstract_Domain`.

### DIFF
--- a/include/links-abstract-domain.php
+++ b/include/links-abstract-domain.php
@@ -81,7 +81,7 @@ abstract class PLL_Links_Abstract_Domain extends PLL_Links_Permalinks {
 	/**
 	 * Adds home and search URLs to language data before the object is created.
 	 *
-	 * @since 3.4
+	 * @since 3.4.1
 	 *
 	 * @param array $additional_data Array of language additional data.
 	 * @param array $language        Language data.
--- a/include/links-abstract-domain.php
+++ b/include/links-abstract-domain.php
@@ -77,4 +77,20 @@ abstract class PLL_Links_Abstract_Domain extends PLL_Links_Permalinks {
 		$uploads['baseurl'] = $this->add_language_to_link( $uploads['baseurl'], $lang );
 		return $uploads;
 	}
+
+	/**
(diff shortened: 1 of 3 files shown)


include/links-abstract-domain.php
```diff
@@ -77,4 +77,20 @@ public function upload_dir( $uploads ) {
...
+	 *
+	 * @param array $additional_data Array of language additional data.
+	 * @param array $language        Language data.
+	 * @return array Language data with home and search URLs added.
+	 */
+	public function set_language_home_urls( $additional_data, $language ) {
+		$language = array_merge( $language, $additional_data );
+		$additional_data['search_url'] = $this->home_url( $language['slug'] );
+		$additional_data['home_url']   = $additional_data['search_url'];
+		return $additional_data;
```
Hug0-Drelon, line 94:
```suggestion
		$additional_data = parent::set_language_home_urls( $additional_data, $language );
		$additional_data['home_url']   = $additional_data['search_url'];
		return $additional_data;
```
Does it work?

Screenfeed: We may call `$this->front_page_url()` for nothing.

Chouby: I agree with @Screenfeed.

Screenfeed (approved):

include/links-abstract-domain.php
Hug0-Drelon: 👍

PR #1288, opened 2023-05-26
Fix warning on term_props

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1689.

After upgrading to Polylang 3.4, some users have a `Undefined array key "term_props"` php warning:
```
Warning: Undefined array key "term_props" in include/language-factory.php on line 161
Warning: foreach() argument must be of type array|object, null given in include/language-factory.php on line 161
Warning: Trying to access array offset on value of type null in include/language.php on line 300
Warning: Undefined array key "term_props" in include/language-factory.php on line 161
Warning: foreach() argument must be of type array|object, null given in include/language-factory.php on line 161
Warning: Trying to access array offset on value of type null in include/language.php on line 300
Warning: foreach() argument must be of type array|object, null given in include/model.php on line 283
```

For some reason, the data sent to `PLL_Language` is incorrect. The only possibility we can think of is the cache transient not being cleared during the upgrade process.

This PR does a quick test to make sure the languages data in the transient is in the new format before using it.

### DIFF
--- a/tests/phpunit/tests/test-model.php
+++ b/tests/phpunit/tests/test-model.php
@@ -270,5 +270,30 @@ class Model_Test extends PLL_UnitTestCase {
 		);
 		$this->assertSameSetsWithIndex( $slugs, $terms );
 	}
+
+	public function test_dont_use_old_format_languages_list() {
+		// Build cache.
+		self::$model->set_languages_ready();
+		self::$model->get_languages_list();
+		// Modify the transient.
+		$languages = get_transient( 'pll_languages_list' );
+		foreach ( $languages as &$language ) {
+			unset( $language['term_props'] );
+		}
+		// Inject our transient.
+		self::$model->clean_languages_cache();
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-model.php
```diff
@@ -270,5 +270,30 @@ public function test_maybe_create_language_terms() {
 		);
 		$this->assertSameSetsWithIndex( $slugs, $terms );
 	}
+
+	public function test_dont_use_old_format_languages_list() {
```
Chouby, line 274:
```suggestion
	/**
	 * @ticket #1689
	 * @see https://github.com/polylang/polylang-pro/issues/1689
	 */
	public function test_dont_use_languages_list_format_older_than_3_4() {
```

Chouby (changes requested):

Screenfeed:
Only the warning.

PR #1289, opened 2023-05-29
Fix customizer switching from the static front page to the blog

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1692
See also https://wordpress.org/support/topic/cant-make-changes-refresh-redirects-to-latest-posts/

NB: While doing tests for this PR, I noticed that the Homepage Settings panel was removed from the customizer with Polylang versions < 3.4. The correct behavior was restored in Polylang 3.4.

The reported issue was due to an incorrect usage of the filter `pre_option_page_on_front`. If this setting is not modified, then the filter param is `false` which was transformed to `0` by `pll_get_post()` and thus WP always displayed the blog page.

In the bug fix (f054942768c96330b90f89d6923a0ca7d70a1bb2), I preferred to use the `customize_register` action instead of reading directly the data passed to the HTTP request. This required to (conditionnally) increase our counter of our `customize_register` actions (00b51046489a81467d103b65ac43d0dfe375a5b7).

Once done, I fixed several other issues:
- If the setting was set to a static front page and we switched to the blog page in the customizer, the language switcher url remained to the static front page (visible when the page slug is in the url). See d71ffbf5d354b24275436e05dce2a37804e4bc9b
- Then the page displayed after the click in the language switcher was the translation of the previous static front page page instead of the translated list of posts. See 0926e21d7b96cb3d39769711ae0b4cd54108d423
- On the opposite, if the setting was set to the blog page and switched to a static front page in the customizer, the translation mechanism did not work at all. See ae3a0c13d4959fbde54a1c6059737c3a545ecb89

### DIFF
--- a/frontend/frontend-static-pages.php
+++ b/frontend/frontend-static-pages.php
@@ -29,7 +29,7 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 	protected $options;
 
 	/**
-	 * Constructor: setups filters and actions
+	 * Constructor: setups filters and actions.
 	 *
 	 * @since 1.8
 	 *
@@ -44,7 +44,7 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 
 		add_action( 'pll_home_requested', array( $this, 'pll_home_requested' ) );
 
-		// Manages the redirection of the homepage
+		// Manages the redirection of the homepage.
 		add_filter( 'redirect_canonical', array( $this, 'redirect_canonical' ) );
 
 		add_filter( 'pll_pre_translation_url', array( $this, 'pll_pre_translation_url' ), 10, 3 );
@@ -52,23 +52,9 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 
 		add_filter( 'pll_set_language_from_query', array( $this, 'page_on_front_query' ), 10, 2 );
 		add_filter( 'pll_set_language_from_query', array( $this, 'page_for_posts_query' ), 10, 2 );
-	}
 
-	/**
-	 * Init the filters
-	 *
-	 * @since 1.8
-	 *
-	 * @return void
-	 */
-	public function pll_language_defined() {
-		parent::pll_language_defined();
-
-		// Support theme customizer
-		if ( isset( $_POST['wp_customize'], $_POST['customized'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			add_filter( 'pre_option_page_on_front', 'pll_get_post', 20 );
-			add_filter( 'pre_option_page_for_post', 'pll_get_post', 20 );
-		}
+		// Specific cases for the customizer.
+		add_action( 'customize_register', array( $this, 'filter_customizer' ) );
 	}
 
 	/**
@@ -277,5 +263,29 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 		}
 
 		return 0; // No page queried.
+	}
+
+	/**
+	 * Add support for the theme customizer.
+	 *
+	 * @since 3.4.2
+	 *
+	 * @return void
+	 */
+	public function filter_customizer() {
+		add_filter( 'pre_option_page_on_front', array( $this, 'customize_page' ), 20 ); // After the customizer.
+		add_filter( 'pre_option_page_for_post', array( $this, 'customize_page' ), 20 );
+	}
+
+	/**
+	 * Translate the page id when customized
+	 *
+	 * @since 3.4.2
+	 *
+	 * @param int|false $pre A page id if the setting is customized, false otherwise.
+	 * @return int|false
+	 */
+	public function customize_page( $pre ) {
+		return is_numeric( $pre ) ? pll_get_post( (int) $pre ) : $pre;
 	}
 }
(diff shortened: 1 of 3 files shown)


frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
 
 		return 0; // No page queried.
 	}
+
+	/**
+	 * Add support for the theme customizer.
```
Screenfeed, line 269:
```suggestion
	 * Adds support for the theme customizer.
```

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
+	 */
+	public function filter_customizer() {
+		add_filter( 'pre_option_page_on_front', array( $this, 'customize_page' ), 20 ); // After the customizer.
+		add_filter( 'pre_option_page_for_post', array( $this, 'customize_page' ), 20 );
+
+		add_filter( 'pll_pre_translation_url', array( $this, 'customize_translation_url' ), 20, 2 ); // After the generic hook in this class.
+	}
+
+	/**
+	 * Translate the page id when customized.
```
Screenfeed, line 283:
```suggestion
	 * Translates the page ID when customized.
```

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
+	 *
+	 * @since 3.4.2
+	 *
+	 * @param int|false $pre A page id if the setting is customized, false otherwise.
```
Screenfeed, line 287:
```suggestion
	 * @param int|false $pre A page ID if the setting is customized, false otherwise.
```

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
+	 * @return int|false
+	 */
+	public function customize_page( $pre ) {
+		return is_numeric( $pre ) ? pll_get_post( (int) $pre ) : $pre;
+	}
+
+	/**
+	 * Fix the translation url the option 'show_on_front' is customized.
```
Screenfeed, line 295:
```suggestion
	 * Fixes the translation URL if the option 'show_on_front' is customized.
```

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
+	 *
+	 * @since 3.4.2
+	 *
+	 * @param string       $url      An empty string or the url of the translation of the current page.
```
Screenfeed, line 299:
```suggestion
	 * @param string       $url      An empty string or the URL of the translation of the current page.
```

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
+	 * @param PLL_Language $language The language of the translation.
+	 * @return string
+	 */
+	public function customize_translation_url( $url, $language ) {
+		if ( 'posts' === get_option( 'show_on_front' ) && is_front_page() ) {
+			// When the page on front displays posts, the home url is the same as the search url.
```
Screenfeed, line 305:
```suggestion
			// When the page on front displays posts, the home URL is the same as the search URL.
```

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
+			$url = $language->get_search_url();
```
Screenfeed, line 306: There is no need for this assignation, we can simply return the value.
```suggestion
			return $language->get_search_url();
```

include/base.php
```diff
@@ -188,7 +188,7 @@ public function should_customize_menu_be_removed() {
 		}
 
 		$customize_register_hooks = count( array_merge( ...array_values( $wp_filter['customize_register']->callbacks ) ) );
-		if ( $customize_register_hooks > 1 ) {
+		if ( $customize_register_hooks > 1 + (int) ( 'page' === get_option( 'show_on_front' ) ) ) { // Are there other hooks than our own?
```
Screenfeed, line 191: Maybe we could add a comment similar to this:
```suggestion
		/*
		 * 'customize_register' is hooked by:
		 * @see PLL_Nav_Menu::create_nav_menu_locations()
		 * @see PLL_Frontend_Static_Pages::filter_customizer()
		 */
		if ( $customize_register_hooks > 1 + (int) ( 'page' === get_option( 'show_on_front' ) ) ) { // Are there other hooks than our own?
```

Screenfeed (changes requested):

include/base.php
Hug0-Drelon, line 191: Why not checking our callbacks are set?

frontend/frontend-static-pages.php
```diff
@@ -278,4 +264,47 @@ protected function get_page_id( $query ) {
...
 	}
+
+	/**
+	 * Add support for the theme customizer.
+	 *
+	 * @since 3.4.2
+	 *
+	 * @return void
+	 */
+	public function filter_customizer() {
```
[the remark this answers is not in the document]
Chouby: I did not look at this direction, keeping the more or less the way it was done previously.

Hug0-Drelon (approved):

Screenfeed (approved):

PR #1292, opened 2023-05-31
Add tests for Customizer.

Hug0-Drelon:
All in the title.

PR #1293, opened 2023-05-31
Check Polylang's callbacks are hooked before removing removing customize menu.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/1696

## Why?
In some context (e.g. `PLL_Settings`) our callbacks might not be hooked.

## How?
By checking our callbacks are really hooked, we can count `customize_register` callbacks properly.

*Note: `Customize_Test` class being introduced by #1292, we should wait until it's merge before.*

### DIFF
--- a/include/base.php
+++ b/include/base.php
@@ -182,22 +182,40 @@ abstract class PLL_Base {
 			return false;
 		}
 
+		return ! $this->is_customize_register_hooked();
+	}
+
+	/**
+	 * Tells whether or not Polylang's callbacks are hooked to `customize_register`.
+	 * Overridden by child classes.
+	 *
+	 * @since 3.4.3
+	 *
+	 * @global $wp_filter
+	 *
+	 * @return boolean True if Polylang's callbacks are hooked, false otherwise.
+	 */
+	protected function is_customize_register_hooked() {
 		global $wp_filter;
-		if ( empty( $wp_filter['customize_register'] ) ) {
+
+		if ( empty( $wp_filter['customize_register'] ) || ! $wp_filter['customize_register'] instanceof WP_Hook ) {
 			return false;
 		}
-
-		$customize_register_hooks = count( array_merge( ...array_values( $wp_filter['customize_register']->callbacks ) ) );
 
 		/*
 		 * 'customize_register' is hooked by:
 		 * @see PLL_Nav_Menu::create_nav_menu_locations()
 		 * @see PLL_Frontend_Static_Pages::filter_customizer()
 		 */
-		if ( $customize_register_hooks > 1 + (int) ( 'page' === get_option( 'show_on_front' ) ) ) { // Are there other hooks than our own?
-			return false;
+		$floor = 0;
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/test-customizer.php
```diff
@@ -0,0 +1,133 @@
...
+		$_wp_theme_features['block-template-parts'] = true;
+
+		$_POST['wp_customize'] = 'on';
+
+		$links_model   = self::$model->get_links_model();
+		$this->pll_env = new PLL_Frontend( $links_model );
+		$this->pll_env->init(); // Implicit `PLL_Frontend_Nav_Menu` instance.
+		$GLOBALS['wp_customize'] = new WP_Customize_Manager();
+		$this->wp_customize      = $GLOBALS['wp_customize'];
+		do_action( 'customize_register', $this->wp_customize );
```
Screenfeed, line 129: After discussing with @Hug0-Drelon: creating an instance of `WP_Customize_Manager` adds new hooks to `customize_register`, meaning that in that case we should not remove the menu item.

However, to make the test more clear about its intend, this can be simplified:
```suggestion
		add_action( 'customize_register', '__return_false' );
```

Screenfeed (changes requested):

Screenfeed:
Why are you betraying me Travis?!

Screenfeed (approved):

include/base.php
```diff
@@ -182,22 +182,41 @@ public function should_customize_menu_be_removed() {
...
+	}
+
+	/**
+	 * Tells whether or not Polylang or third party callbacks are hooked to `customize_register`.
+	 *
+	 * @since 3.4.3
+	 *
+	 * @global $wp_filter
+	 *
+	 * @return boolean True if Polylang's callbacks are hooked, false otherwise.
```
Chouby, line 195:
```suggestion
	 * @return bool True if Polylang's callbacks are hooked, false otherwise.
```

Chouby (changes requested):

Chouby:
I attempted to propose something more generic, in case we add more filters in the future. This resulted in something that it's not very readable so I preferred to keep your proposal. 
For the record:
```php
$filters = array(
	array( 'nav_menu', 'create_nav_menu_locations' ),
	array( 'static_pages', 'filter_customizer' ),
);

$floor = array_reduce(
	$filters,
	function( $carry, $item ) {
		$has_filter = ! empty( $this->{$item[0]} ) && (bool) wp_filter['customize_register']->has_filter( 'customize_register', array( $this->{$item[0]}, $item[1] ) );
		return $carry + (int) $has_filter;
	}
);
```

Screenfeed:
Actually I thought about some sort of small registry (either in a separate class or in `PLL_Base`) where the 2 classes (and more) could register against, and we could loop through the registered hooks (yes, I know, I love this mechanics).

Screenfeed:
> I like the idea, we could create a separate class for that. Something like `PLL_Tools` if you like. Even though it might be a bit complex, it'd be hidden and reusable!

Actually I thought about some sort of small registry (either in a separate class or in PLL_Base) where the 2 classes (and more) could register against, and we could loop through the registered hooks (yes, I know, I love this mechanics).

Chouby:
Maybe some day if we need to hide other cutomizer actions.

PR #1295, opened 2023-05-31
Assign static pages to `PLL_Language` even if no translation exist.

### LINKED ISSUE #1291: Static front page does not work in 3.4.2 with only one language
## Prerequisites

1. Install the Polylang plugin
2. Add **one** language
3. Add a static front page
4. Configure the static front page in the reading settings
5. Visit the front page

When I follow those steps, I see a list of posts on the front page.

I was expecting that I see the static front page.

## Workaround

Everything worked as expected after adding another language and translating the static front page to this new language.

But with only one language added, the static front page does not work.

## Explantion

I have a website where a customer planned to have multiple languages later. To avoid any issues with the URLs later, we added multilanguage with just one language before go-live.

Today, I updated from 3.3.3 to 3.4.2. After the update, the front page was empty since we were using ACF on the front page, and the fields were not available anymore. I investigated the issue and found out that the wrong posts were loaded. After disabling the Polylang plugin, everything worked as expected again.

So I started to investigate the issue in a new WordPress installation and was able to reproduce the issue, as long as there is only one language added to Polylang (or the front page of the second language is not translated).

In my eyes, this is a very rare use case. But if you have an idea how to fix it, it would be great to have a solution for that.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/1697
Fixes #1291 

## Why?
When a static page has no translation, trying to translate it before assigning it to a `PLL_Language` object forces the value to `0`.

## How?
If a static page hasn't any translation, check if it's language correspond to the currently created language.

### DIFF
--- a/include/static-pages.php
+++ b/include/static-pages.php
@@ -105,6 +105,21 @@ class PLL_Static_Pages {
 	 */
 	protected function get_translation( $static_page, $language ) {
 		$translations = $this->model->post->get_raw_translations( $this->$static_page );
+
+		if ( empty( $translations ) ) {
+			$page_lang = $this->model->post->get_object_term( $this->$static_page, $this->model->post->get_tax_language() );
+
+			if ( empty( $page_lang ) || $page_lang->slug !== $language['slug'] ) {
+				return 0;
+			}
+
+			/*
+			 * Current page doesn't have any translation,
+			 * but has a language defined corresponding to the current one.
+			 */
+			return $this->$static_page;
+		}
+
 		if ( ! isset( $translations[ $language['slug'] ] ) ) {
 			return 0;
 		}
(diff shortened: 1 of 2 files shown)


Chrystll:
Just to let you know that I tested on my side and that's OK for me.

Screenfeed (approved):

include/static-pages.php
```diff
@@ -105,6 +105,21 @@ public function init() {
...
+			if ( empty( $page_lang ) || $page_lang->slug !== $language['slug'] ) {
+				return 0;
+			}
+
+			/*
+			 * Current page doesn't have any translation,
+			 * but has a language defined corresponding to the current one.
+			 */
+			return $this->$static_page;
+		}
```
Chouby, line 121: I propose something slightly simpler with a different comment:
```suggestion
		// When the current static page doesn't have any translation, we must return itself for its language.
		if ( empty( $translations ) ) {
			$page_lang = $this->model->post->get_object_term( $this->$static_page, $this->model->post->get_tax_language() );

			if ( ! empty( $page_lang ) && $page_lang->slug === $language['slug'] ) {
				return $this->$static_page;
			}
		}
```

PR #1305, opened 2023-06-22
Don't set $_column_headers in list tables

Chouby:
This fixes the error `Undefined array key 2` reported by our [PHPUnit tests](https://app.travis-ci.com/github/polylang/polylang/jobs/604689796) 

See https://core.trac.wordpress.org/ticket/32170#comment:65 for the context
Maybe the backward compatibility issue will be fixed by WordPress before the final release. However, there is no point to keep the legacy way to define column headers.


Screenfeed (approved):
Could not test with the bugged version since it's already patched in core but:
- 6.2.3-alpha-55925 ✅
- 6.3-alpha-55991 ✅

PR #1307, opened 2023-06-26
Revert "Don't set $_column_headers in list tables"

Chouby:
This reverts #1305, because:
- It is now useless since WP fixed the issue on its side.
- This totally breaks the tables.

Unfortunately, things are not as simple as I imagined.
The columns are added by `WP_List_Table` with `get_column_headers()` using the filter `manage_{$screen->id}_columns`. So there are several steps to make things happen the same was as in WP:
1. Pass the `screen` param to the `WP_List_Table` constructor.
2. Instantiate our `WP_List_Table` child class early enough (probably hooked to `load-{$screen->id}` otherwise `get_column_headers()` is called before the filter is added in the  the `WP_List_Table` constructor.

### DIFF
--- a/tests/phpunit/tests/test-settings-list-tables.php
+++ b/tests/phpunit/tests/test-settings-list-tables.php
@@ -12,10 +12,13 @@ class Settings_List_Tables_Test extends PLL_UnitTestCase {
 		self::create_language( 'fr_FR' );
 	}
 
-	public function test_display_languages_table() {
-		wp_set_current_user( 1 );
+	public function set_up() {
+		parent::set_up();
 
-		$GLOBALS['hook_suffix'] = 'settings_page_mlang';
+		wp_set_current_user( 1 );
+	}
+
+	public function test_display_languages_table() {
 		$_GET['page'] = 'mlang';
 
 		$links_model = self::$model->get_links_model();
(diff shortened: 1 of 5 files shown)


tests/phpunit/tests/test-settings-list-tables.php
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	public function test_display_languages_table() {
+		wp_set_current_user( 1 );
+
+		$GLOBALS['hook_suffix'] = 'settings_page_mlang';
+		$_GET['page'] = 'mlang';
+
+		$links_model = self::$model->get_links_model();
+		$pll_env = new PLL_Settings( $links_model );
```
Hug0-Drelon, line 22: I think it could be moved into `set_up()` using predefined prop `$this->pll_env`.

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-settings-list-tables.php
Hug0-Drelon: Oh I see, maybe with a method like `init_pll_settings( $page_name )` then?

Chouby: Good idea.

Hug0-Drelon (approved):

1ucay:
Hello, how can I add custom column to admin.php?page=mlang. I just tried 

    add_filter( 'manage_toplevel_page_mlang_columns', function( $columns ) {

but no luck. Thx

PR #1311, opened 2023-06-29
Don't reset the"static front page" option on deletion of a home page translation

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1701.

This PR allows to disable/enable the translation of the options `page_for_posts` and `page_on_front` whenever we want. This is done with the methods `PLL_Static_Pages::disable_page_on_front_and_page_for_posts_translation()` and `PLL_Static_Pages::enable_page_on_front_and_page_for_posts_translation()`.

This PR also disables these translations automatically when deleting the page for posts, page on front, or one of their translations. This is done by "surrounding" `_reset_front_page_settings_for_post()`, which is hooked to `before_delete_post` and `wp_trash_post`.

Note: maybe `disable_page_on_front_and_page_for_posts_translation()` could be used directly in place of `maybe_disable_page_on_front_and_page_for_posts_translation()`, but it feels safer (at least to me) to keep it.

### DIFF
--- a/include/static-pages.php
+++ b/include/static-pages.php
@@ -36,6 +36,13 @@ class PLL_Static_Pages {
 	protected $curlang;
 
 	/**
+	 * Used to temporarilly disable the static page translation.
+	 *
+	 * @var bool
+	 */
+	protected $page_translation_disabled = false;
+
+	/**
 	 * Constructor: setups filters and actions.
 	 *
 	 * @since 1.8
@@ -161,6 +168,16 @@ class PLL_Static_Pages {
 		// Translates page for posts and page on front.
 		add_filter( 'option_page_on_front', array( $this, 'translate_page_on_front' ) );
 		add_filter( 'option_page_for_posts', array( $this, 'translate_page_for_posts' ) );
+
+		/*
+		 * Disables the page translation when deleting the page for posts, page on front, or one of their translations.
+		 * The idea is to disable the translation when `_reset_front_page_settings_for_post()` is used, which is hooked
+		 * to `before_delete_post` and `wp_trash_post` at priority 10.
+		 */
+		add_action( 'before_delete_post', array( $this, 'maybe_disable_page_on_front_and_page_for_posts_translation' ), -10000 );
+		add_action( 'before_delete_post', array( $this, 'enable_page_on_front_and_page_for_posts_translation' ), 10000 );
+		add_action( 'wp_trash_post', array( $this, 'maybe_disable_page_on_front_and_page_for_posts_translation' ), -10000 );
+		add_action( 'wp_trash_post', array( $this, 'enable_page_on_front_and_page_for_posts_translation' ), 10000 );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-static-pages.php
Hug0-Drelon, line 1: Should we add a test for page deletion without translation?

Hug0-Drelon (changes requested):

include/static-pages.php
```diff
@@ -186,8 +211,62 @@ public function translate_page_on_front( $page_id ) {
...
+	}
+
+	/**
+	 * Disables the translation of the "page on front" and "page for posts".
+	 *
+	 * @since 3.4.4
+	 *
+	 * @return void
+	 */
+	public function disable_page_on_front_and_page_for_posts_translation() {
```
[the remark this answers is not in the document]
Screenfeed: Not strictly necessary but it felt weird to me to have one public and the other private.

include/static-pages.php
```diff
@@ -161,6 +168,16 @@ public function pll_language_defined() {
...
 		add_filter( 'option_page_on_front', array( $this, 'translate_page_on_front' ) );
 		add_filter( 'option_page_for_posts', array( $this, 'translate_page_for_posts' ) );
+
+		/*
+		 * Disables the page translation when deleting the page for posts, page on front, or one of their translations.
+		 * The idea is to disable the translation when `_reset_front_page_settings_for_post()` is used, which is hooked
+		 * to `before_delete_post` and `wp_trash_post` at priority 10.
+		 */
+		add_action( 'before_delete_post', array( $this, 'maybe_disable_page_on_front_and_page_for_posts_translation' ), -10000 );
+		add_action( 'before_delete_post', array( $this, 'enable_page_on_front_and_page_for_posts_translation' ), 10000 );
```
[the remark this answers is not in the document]
Screenfeed: Won't work, we can't add a hook from inside this hook.

tests/phpunit/tests/test-static-pages.php
Screenfeed: Would you expect a different behavior?

include/static-pages.php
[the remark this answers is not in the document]
Hug0-Drelon: You're right, let's be consistent.

tests/phpunit/tests/test-static-pages.php
Hug0-Drelon: No, but I would prevent to have a different one in the future having this case covered.

Screenfeed: Done in [f285bac](https://github.com/polylang/polylang/pull/1311/commits/f285bac386768202384dd2b0f332db11019fa954).

Hug0-Drelon (approved):

Chouby:
Would it work to simply test `doing_action( 'before_delete_post' )` and `doing_action( 'wp_trash_post' )` in `translate_page_on_front()` and `translate_page_for_posts()` as we already do for `switch_blog`? This would be much simpler. This would also be less precise but I expect side effects to be unlikely.

Screenfeed:
> Would it work to simply test `doing_action( 'before_delete_post' )` and `doing_action( 'wp_trash_post' )` in `translate_page_on_front()` and `translate_page_for_posts()` as we already do for `switch_blog`? This would be much simpler. This would also be less precise but I expect side effects to be unlikely.

Indeed, less precise but much simpler. Done in [7ad209f](https://github.com/polylang/polylang/pull/1311/commits/7ad209fc7ee8be4d9e71d4076bf3274ab1a8bfb7).

PR #1312, opened 2023-06-29
Fix tags translation during post update.

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/1712

## What?
`wp_update_post()` is removing `tags_input` param when there is no change. But in our case we do need it so `PLL_CRUD_Posts::set_object_terms()` can update the post tags according to the new language.

## How?
- Hook to `post_updated` and force `wp_insert_post()` with the `tags_input` parameter.

## And?
- Add tests for tag, category and custom taxonomy just for safety.

*Note: I previously wanted to use `wp_insert_post_data` to filter the arguments but it can't be done since the first parameter is a post data array containing only DB columns. As I see it there is no way to filter `$postarr` in `wp_insert_post()`...*

### DIFF
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -337,26 +337,11 @@ class PLL_CRUD_Posts {
 	 * @return void
 	 */
 	public function force_tags_translation( $post_id, $post_after, $post_before ) {
-		static $avoid_recursion = false;
-
-		if ( $avoid_recursion ) {
-			return;
-		}
-
-		$avoid_recursion = true;
-
 		$post_before = $post_before->to_array();
 
-		if ( empty( $post_before['tags_input'] ) ) {
-			$avoid_recursion = false;
-			return;
+		if ( ! empty( $post_before['tags_input'] ) ) {
+			// Let's ensure that  `PLL_CRUD_Posts::set_object_terms()` will do its job.
+			wp_set_post_tags( $post_id, $post_before['tags_input'] );
 		}
-
-		$post_after               = $post_after->to_array();
-		$post_after['tags_input'] = $post_before['tags_input'];  // Will force `PLL_CRUD_Posts::set_object_terms()` to do its job.
-
-		wp_insert_post( $post_after );
-
-		$avoid_recursion = false;
 	}
 }
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -337,12 +337,18 @@ class PLL_CRUD_Posts {
 	 * @return void
 	 */
 	public function force_tags_translation( $post_id, $post_after, $post_before ) {
-		remove_action( 'post_updated', array( $this, 'force_tags_translation' ) );
+		static $avoid_recursion = false;
+
+		if ( $avoid_recursion ) {
+			return;
+		}
+
+		$avoid_recursion = true;
 
 		$post_before = $post_before->to_array();
 
 		if ( empty( $post_before['tags_input'] ) ) {
-			add_action( 'post_updated', array( $this, 'force_tags_translation' ), 10, 3 );
+			$avoid_recursion = false;
 			return;
 		}
 
@@ -351,6 +357,6 @@ class PLL_CRUD_Posts {
 
 		wp_insert_post( $post_after );
 
-		add_action( 'post_updated', array( $this, 'force_tags_translation' ), 10, 3 );
+		$avoid_recursion = false;
 	}
 }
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/test-crud-posts.php
```diff
@@ -0,0 +1,88 @@
...
+
+		// Pass the term in previous language on purpose.
+		if ( ! in_array( $taxonomy, array( 'category', 'post_tag' ), true ) ) {
+			wp_set_current_user( 1 ); // Current user should have the proper capability to set custom taxonomy terms.
+			$postarr[ $post_tax_arg ] = array( $taxonomy => array( $term_en ) ); // Special case fo custom taxonomies wich expects an array of array.
+		} elseif ( 'post_tag' === $taxonomy ) {
+			$postarr[ $post_tax_arg ] = array( get_term( $term_en, $taxonomy )->name ); // Special case for tags where `wp_update_post()` removes existing one by names.
+		} else {
+			$postarr[ $post_tax_arg ] = array( $term_en ); // Pass the term in previous language on purpose.
+		}
```
manooweb, line 74: It seems useless for me here because `tags_input` and `post_category` are already correctly set at this point.

manooweb, line 74: It seems useless for me here because tags_input and post_category are already correctly set at this point.
The test passes without these lines.

manooweb (changes requested):

tests/phpunit/tests/test-crud-posts.php
manooweb: Ok. As we discussed together it is for consistency with custom taxonomy case. Let's keep like this

manooweb (approved):
LTGM 👌

include/crud-posts.php
```diff
@@ -324,4 +325,32 @@ public function create_media_translation( $post_id, $lang ) {
...
+	 *
+	 * @since 3.4.4
+	 *
+	 * @param int     $post_id      Post ID, unused.
+	 * @param WP_Post $post_after   Post object following the update.
+	 * @param WP_Post $post_before  Post object before the update.
+	 * @return void
+	 */
+	public function force_tags_translation( $post_id, $post_after, $post_before ) {
+		remove_action( 'post_updated', array( $this, 'force_tags_translation' ) );
```
Chouby, line 340: Please don't do that. 
Due to [a bug in WP](https://core.trac.wordpress.org/ticket/17817) that I'm quite sure I still encountered the issue after it was supposedly fixed, you can't remove an action in itself. All actions after yours will never be executed. 
Using a static variable allows you to achieve the same goal without any risk. See for example https://github.com/polylang/polylang/blob/3.4.4/include/filters-sanitization.php#L67-L78
It may even simplify your code.

Chouby (changes requested):

include/crud-posts.php
Chouby: For the reference: https://core.trac.wordpress.org/ticket/40393

Hug0-Drelon: Wow great to know, thanks!
Fixed in 4e8bfe69.

include/crud-posts.php
```diff
@@ -324,4 +325,38 @@ public function create_media_translation( $post_id, $lang ) {
...
+			$avoid_recursion = false;
+			return;
+		}
+
+		$post_after               = $post_after->to_array();
+		$post_after['tags_input'] = $post_before['tags_input'];  // Will force `PLL_CRUD_Posts::set_object_terms()` to do its job.
+
+		wp_insert_post( $post_after );
+
+		$avoid_recursion = false;
```
Chouby, line 360: I'm looking for a way not to execute the complete `wp_insert_post()` a second time.
Would this work?
```suggestion
		if ( ! empty( $post_before['tags_input'] ) ) {
			// Let  ensure that  `PLL_CRUD_Posts::set_object_terms()` will do its job.
			wp_set_post_tags( $post_id, $post_before['tags_input'] ) );
		}
```

Chouby (changes requested):

include/crud-posts.php
Hug0-Drelon: Yeah! It works! Done in 174e49ee 🚀

include/crud-posts.php
```diff
@@ -324,4 +325,23 @@ public function create_media_translation( $post_id, $lang ) {
...
+	 * @param WP_Post $post_after   Post object following the update.
+	 * @param WP_Post $post_before  Post object before the update.
+	 * @return void
+	 */
+	public function force_tags_translation( $post_id, $post_after, $post_before ) {
+		$post_before = $post_before->to_array();
+
+		if ( ! empty( $post_before['tags_input'] ) ) {
+			// Let's ensure that  `PLL_CRUD_Posts::set_object_terms()` will do its job.
+			wp_set_post_tags( $post_id, $post_before['tags_input'] );
```
ve3, line 344: This may cause another problem. Please see my issue https://github.com/polylang/polylang/issues/1401

PR #1319, opened 2023-07-17
[MS] Fix option creation on blog creation

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1721.

Everything is explained in the issue. This PR:
1. Adds the missing `switch_to_blog()` and `restore_current_blog()` in `Multisite_Test::test_new_site()`, so the option tested is the one from the new blog.
2. Hooks `'wp_initialize_site'` instead of `'wp_insert_site'`, so the option is created after the DB tables are created when a new blog is created.

### DIFF
--- a/tests/phpunit/tests/test-multisite.php
+++ b/tests/phpunit/tests/test-multisite.php
@@ -1,15 +1,20 @@
 <?php
 
-if ( is_multisite() ) :
+class Multisite_Test extends PLL_UnitTestCase {
 
-	class Multisite_Test extends PLL_UnitTestCase {
-		public function test_new_site() {
-			$site_id = self::factory()->blog->create();
-			switch_to_blog( $site_id );
-			$options = get_option( 'polylang' );
-			restore_current_blog();
-			$this->assertNotFalse( $options );
+	public function set_up() {
+		if ( ! is_multisite() ) {
+			$this->markTestSkipped( 'This test requires WordPress multisite.' );
 		}
+
+		parent::set_up();
 	}
 
-endif;
+	public function test_new_site() {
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-multisite.php
```diff
@@ -1,13 +1,20 @@
...
+class Multisite_Test extends PLL_UnitTestCase {
 
-	class Multisite_Test extends PLL_UnitTestCase {
-		public function test_new_site() {
-			$site_id = self::factory()->blog->create();
-			$options = get_option( 'polylang' );
-			$this->assertNotFalse( $options );
+	public function set_up() {
+		if ( ! is_multisite() ) {
+			$this->markTestSkipped( 'This test requires WordPress multisite.' );
```
Chouby, line 7: I would prefer not to change this. At least not as proposed and not in this PR.

The reason it is written like this is to avoid skipped tests for a normal install (to avoid the message "OK, but incomplete, skipped, or risky tests!" when testing the stable version with the the stable WordPress. Yes I like green ;)

However, I am ok to improve this but:
1. We must do this in a separate PR
2. All mulitisite test classes must be modified for consistency
3. I prpose to create a multisite group which we could exclude by default and include when testing a multisite.

Chouby (changes requested):

tests/phpunit/tests/test-multisite.php
Screenfeed: Done in [131c8f0](https://github.com/polylang/polylang/pull/1319/commits/131c8f0a584c7a5dc3ee54cb6c0eb68ba7f8ba9c).

PR #1321, opened 2023-07-18
Require PHP 7 or higher

Chouby:
[WP 6.3 will drop the support of PHP 5.6](https://make.wordpress.org/core/2023/07/05/dropping-support-for-php-5/). To potentially avoid a drama similar to #415 planned for 2.7 and moved to 2.6.7 in urgence, I propose to move to PHP 7 more or less at the same time as WP. Even if we still support WP 5.8.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
 	"homepage": "https://polylang.pro",
 	"type": "wordpress-plugin",
 	"require": {
-		"php": ">=5.6"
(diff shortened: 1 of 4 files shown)


manooweb (approved):
OK 👌

Chouby:
Force-pushed a new branch with chnages to Travis config as it is now removed.

PR #1322, opened 2023-07-18
Allow to mock some constants in tests

Screenfeed:
This PR introduces 2 functions that will allow us to mock constants in tests.

1. `pll_has_constant()` is basically an alias for `defined()`.
2. `pll_get_constant()` returns the value of a constant if it is defined. The 2nd argument, `$default`, allows to simplify the code a bit. For example, this:
```php
if ( ! defined( 'FOOBAR' ) || FOOBAR ) {
```
Can be replaced by this:
```php
if ( pll_get_constant( 'FOOBAR', true ) ) {
```

## Trick

If `! defined( 'FOOBAR' ) || FOOBAR` is equal to `pll_get_constant( 'FOOBAR', true )`, then its opposite, `defined( 'FOOBAR' ) && ! FOOBAR`, is equal to `! pll_get_constant( 'FOOBAR', true )`.

I first created a function for that, `pll_has_falsy_constant()`: it seems more readable but redundant, so [I removed it](https://github.com/polylang/polylang/pull/1322/commits/f0f9ce9525c699b67fe2694652443799524fba5d).

## Compat

Ready for php 7.0.


include/language.php
Hug0-Drelon, line 1: Maybe we can use `pll_get_constant` for `PLL_ENCODED_FLAGS`? Since this file is modified...

include/constants.php
Hug0-Drelon, line 1: I think this file doesn't belong to `/include` where we should keep business logic.
Maybe we could create a new folder named `/tools`?

Hug0-Drelon: Also, even if we do not use it right now, it could benefit from another function `pll_set_constant` which would check if a constant is defined before setting it. What do you think?

include/language.php
Screenfeed: I'm not against, for consistency.

Screenfeed: Done in [122f306](https://github.com/polylang/polylang/pull/1322/commits/122f306a8595addce037e39dbc79630ef706d856).

include/constants.php
Screenfeed: @Chouby what do you think about all this?

Hug0-Drelon (approved):

include/constants.php
Chouby: I would move this to functions.php.

integrations/domain-mapping/domain-mapping.php
```diff
@@ -17,7 +17,7 @@ class PLL_Domain_Mapping {
 	 */
 	public function __construct() {
 		if ( function_exists( 'redirect_to_mapped_domain' ) ) {
-			if ( ! defined( 'PLL_CACHE_HOME_URL' ) && ( $options = get_option( 'polylang' ) ) && $options['force_lang'] < 2 ) {
+			if ( ! pll_has_constant( 'PLL_CACHE_HOME_URL' ) && ( $options = get_option( 'polylang' ) ) && $options['force_lang'] < 2 ) {
```
Chouby, line 20: Would it be interesting to introduce a function like `pll_maybe_define_constant()`

Screenfeed: You mean like @Hug0-Drelon suggested [here](https://github.com/polylang/polylang/pull/1322#discussion_r1269506955)? 😅

integrations/domain-mapping/domain-mapping.php
```diff
@@ -17,7 +17,7 @@ class PLL_Domain_Mapping {
...
 				define( 'PLL_CACHE_HOME_URL', false );
 			}
```
Chouby, line 22:
```suggestion
			$options = get_option( 'polylang' );

			if ( is_array( $options ) && $options['force_lang'] < 2 ) {
				pll_set_constant( 'PLL_CACHE_HOME_URL', false );
			}
```

PR #1300, opened 2023-06-02
New filter for `PLL_Language` URLs instead of `site_url()`.

Hug0-Drelon:
## What?
Fixes #1296 

## Why?
See https://github.com/polylang/polylang/issues/1296#issuecomment-1572107562
`site_url()` adds an extra `/subdir/`.

## How?
⚠️ ~~Introduce `pll_language_url` to filter language URL (especially for domains).~~ changed the way to fix the issue, see bellow.
- Add `pll_language_flag_url` filter hooked by `PLL_Links_Abstract_Domain::site_url()`
- Add `pll_home_url` filter hooked by `PLL_Links_Model::set_language_home_url()`
- Add `pll_search_url` filter hooked by `PLL_Links_Model::set_language_search_url()`

## Test instruction
Add the following code to your `pll-config.php` file:
```php
<?php

function pll_log_lang_urls() {
    $languages = PLL()->model->get_languages_list();

    if ( ! defined( 'PLL_CACHE_LANGUAGES' ) ) {
        define( 'PLL_CACHE_LANGUAGES', false );
    }

    foreach ( $languages as $language ) {
        error_log( "{$language->name} Flag URL: {$language->get_display_flag_url()}" );
        error_log( "{$language->name} Home URL: {$language->get_home_url()}" );
        error_log( "{$language->name} Search URL: {$language->get_search_url()}" );
    }
}
add_action( 'pll_init', 'pll_log_lang_urls' );
```

Clean the transient and visit any page of your site. Check your `debug.log`, the logged URL should be correct.

*Note: Allow `.jpeg` extension for custom flags. And clean `PLL_Domain_UnitTestCase`*

PR #1324, opened 2023-07-20
Mock language cache constants

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/1708.

This PR mocks the constants `PLL_CACHE_LANGUAGES` and `PLL_CACHE_HOME_URL`, and adds more cases to the tests introduced in #1300.

## Some details

- This replaces the dev-dependency `yoast/phpunit-polyfills` by `yoast/wp-test-utils` to get access to `brain/monkey`. Note: `wpsyntex/wp-phpunit` does the same.
- Requiring Patchwork in integration tests broke the "ajax on front" tests: replacing the file `fr_FR.mo` fixed the issue. The file `fr_FR.po` is now also included so `fr_FR.mo` can be modified more easily.


tests/phpunit/tests/test-links-directory.php
```diff
@@ -202,10 +211,30 @@ public function test_get_language_from_url_with_wrong_ssl() {
...
 
+		Functions\when( 'pll_get_constant' )->alias(
+			function ( $constant_name ) use ( $cache_languages, $cache_home_url ) {
+				switch ( $constant_name ) {
+					case 'PLL_CACHE_LANGUAGES':
+						return $cache_languages;
+					case 'PLL_CACHE_HOME_URL':
+						return $cache_home_url;
+					default:
+						return null;
```
Hug0-Drelon, line 233: Shouldn't default be?
```php
return constant( $constant_name );
```
Otherwise if the function is called during the code execution for something different than `PLL_CACHE_LANGUAGES` or `PLL_CACHE_HOME_URL` it might return a wrong value, am I right?

tests/phpunit/tests/test-links-multi-domains-to-one.php
```diff
@@ -42,8 +47,28 @@ public function tear_down() {
...
+	 * @param bool $cache_home_url  Value of the constant `PLL_CACHE_HOME_URL`.
+	 *
+	 * @testWith [true, false]
+	 *           [false, true]
+	 *           [false, false]
 	 */
-	public function test_home_and_search_urls() {
+	public function test_home_and_search_urls( $cache_languages, $cache_home_url ) {
+		Functions\when( 'pll_get_constant' )->alias(
+			function ( $constant_name ) use ( $cache_languages, $cache_home_url ) {
```
Hug0-Drelon, line 60: Is it the `use` keyword that prevent you from avoid code duplication?

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-links-multi-domains-to-one.php
Screenfeed: Are you suggesting to to put this mock in a trait?

Hug0-Drelon: If it's doable, I think we should. Or even in the testcase.

Screenfeed: Yes, I think it's a good idea.
Done in [a549d3c](https://github.com/polylang/polylang/pull/1324/commits/a549d3c6e01a50a75a9a866df7ac332d09a2dfda).

Hug0-Drelon (approved):

Chouby:
Closed by https://github.com/polylang/polylang-pro/issues/1703

PR #1325, opened 2023-07-25
Make testcase classes abstract and autoloaded

Screenfeed:
Making those classes abstract allows us to autoload them here and in other plugins like Polylang Pro.

This PR also autoloads all test classes, it could be useful in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,11 @@
 			"settings/"
 		]
 	},
+	"autoload-dev": {
+		"classmap": [
+			"tests/phpunit/includes/"
+		]
+	},
 	"scripts": {
 		"test":"vendor/bin/phpunit",
 		"cs":"vendor/bin/phpcs",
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -29,6 +29,11 @@
 			"settings/"
 		]
 	},
+	"autoload-dev": {
+		"classmap": [
+			"tests/phpunit/includes/"
```
Hug0-Drelon, line 34:
```suggestion
			"tests/phpunit/"
```
Let's autoload all test classes to be able to extend one from another in different folders, what do you think?

Screenfeed: OK. Not useful for now but may be in the future.
Done in [616794b](https://github.com/polylang/polylang/pull/1325/commits/616794b2f9d955ac2938e4eedfb29c8bff31d6d0).

PR #1326, opened 2023-07-27
Add pll_admin_ajax_params filter

manooweb:
To fix the Polylang for WooCommerce PR https://github.com/polylang/polylang-wc/issues/667 we need to introduce this filter to be able to set the order id in any ajax request and thus filter correctly by the order language.

Related to a Polylang for WooCommerce PR https://github.com/polylang/polylang-wc/pull/672

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -302,6 +302,15 @@ abstract class PLL_Admin_Base extends PLL_Base {
 			$params = array_merge( $params, array( 'pll_term_id' => (int) $tag_ID ) );
 		}
 
+		/**
+		 * Filters the list of parameters to add to ajax request.
+		 *
+		 * @since 3.5
+		 *
+		 * @param array $params List of parameters.
+		 */
+		$params = apply_filters( 'pll_ajax_params', $params );
+
 		$str = http_build_query( $params );
 		$arr = wp_json_encode( $params );
 		?>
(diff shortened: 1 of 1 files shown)


admin/admin-base.php
```diff
@@ -302,6 +302,15 @@ public function admin_print_footer_scripts() {
...
 		}
 
+		/**
+		 * Filters the list of parameters to add to ajax request.
+		 *
+		 * @since 3.5
+		 *
+		 * @param array $params List of parameters.
+		 */
+		$params = apply_filters( 'pll_ajax_params', $params );
```
Screenfeed, line 312: If this filter is meant to be generic, maybe we should add a second parameter for the context. Maybe something like `'request prefilter'`? What do you think?

Screenfeed (approved):

admin/admin-base.php
manooweb: Generic mean that we can put some other parameters not that we want to apply the filter elsewhere.
So, I think we can keep only one parameter.

Chouby, line 312:
```suggestion
		 * @since 3.4.5
		 *
		 * @param array $params List of parameters to add to the admin ajax request.
		 */
		$params = apply_filters( 'pll_admin_ajax_params', $params );
```

Chouby (changes requested):

admin/admin-base.php
manooweb: Done in https://github.com/polylang/polylang/pull/1326/commits/4c07b6498b5d308fef80b92c3c756be183d2ef65

PR #1332, opened 2023-08-21
Javascript type is not required for HTML5

Chouby:
For the context: https://wordpress.org/support/topic/the-type-attribute-is-unnecessary-for-javascript-resources-3/

This PR handles two cases:
- On front, it honors `current_theme_support( 'html5', 'script' )` to output javascript type (and optionally `CDATA`) or not for inline scripts, as done by WP since the verson 5.3. See:  https://core.trac.wordpress.org/ticket/42804
- On admin, the javascript type and CDATA are removed as the doctype was switched to html5 since WP 3.3. See https://core.trac.wordpress.org/ticket/18202

### DIFF
--- a/frontend/frontend-filters-search.php
+++ b/frontend/frontend-filters-search.php
@@ -132,20 +132,21 @@ class PLL_Frontend_Filters_Search {
 	}
 
 	/**
-	 * Allows modifying the search form if it does not pass get_search_form
+	 * Allows modifying the search form if it does not pass get_search_form.
 	 *
 	 * @since 0.1
 	 *
 	 * @return void
 	 */
 	public function wp_print_footer_scripts() {
-		// Don't use directly e[0] just in case there is somewhere else an element named 's'
-		// Check before if the hidden input has not already been introduced by get_search_form ( FIXME: is there a way to improve this ) ?
-		// Thanks to AndyDeGroo for improving the code for compatibility with old browsers
-		// http://wordpress.org/support/topic/development-of-polylang-version-08?replies=6#post-2645559
+		/*
+		 * Don't use directly e[0] just in case there is somewhere else an element named 's'
+		 * Check before if the hidden input has not already been introduced by get_search_form ( FIXME: is there a way to improve this ) ?
+		 * Thanks to AndyDeGroo for improving the code for compatibility with old browsers
+		 * http://wordpress.org/support/topic/development-of-polylang-version-08?replies=6#post-2645559
+		 */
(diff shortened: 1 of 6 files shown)


frontend/frontend-filters-search.php
```diff
@@ -159,12 +160,18 @@ public function wp_print_footer_scripts() {
...
+
+		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
+
+		if ( $type_attr ) {
+			$js = "/* <![CDATA[ */\n{$js}\n/* ]]> */";
 		}
-		//]]>";
-		echo '<script type="text/javascript">' . $js . '</script>'; // phpcs:ignore WordPress.Security.EscapeOutput
+
+		echo "<script{$type_attr}>\n{$js}\n</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
```
manooweb, line 175: Same as in Polylang Pro https://github.com/polylang/polylang-pro/pull/1779#discussion_r1301437991
Couldn't we use printf instead?

integrations/cache/cache-compat.php
```diff
@@ -67,7 +67,10 @@ public function add_cookie_script() {
 			'; SameSite=' . $samesite,
 			esc_js( $expiration )
 		);
-		echo "<script type='text/javascript'>\n" . $js . "</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
+
+		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
+
+		echo "<script{$type_attr}>\n{$js}</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
```
manooweb, line 73: Same as in Polylang Pro https://github.com/polylang/polylang-pro/pull/1779#discussion_r1301437991
Couldn't we use printf instead?

manooweb (changes requested):

integrations/cache/cache-compat.php
manooweb: it seems that it misses a `\n` character just behind `{$js}` as in `PLL_Frontend_Filters_Search`
 https://github.com/polylang/polylang/blob/6ecea13398ad11cbd7dbbce5a01cfd04fbd781ec/frontend/frontend-filters-search.php#L175

manooweb (changes requested):

manooweb (approved):
ok 👌

PR #1336, opened 2023-08-30
Assign terms in the correct language

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1766.
Related: #1312, https://github.com/polylang/polylang-pro/issues/1712, https://github.com/polylang/polylang-pro/issues/1705.

This PR makes sure that terms that are assigned to a post are in the correct language.
This is done by `PLL_CRUD_Posts::set_object_terms()`, which was introduced in 2.3, and improved here.

### Note about the issue

The original issue may not find its source here, but this method was failing when, precisely, terms had to be translated. Maybe there's an external factor that changed (filters for `get_terms()` or `term_exists()`?).
I noticed that the tags received when saving the post are in the default language (while the post is not), so maybe the post's language is not conveyed all along anymore.

### Notes about the solution introduced here

- This solution should work in any context, as long as the terms are assigned to the post. Which means:
- This solution works best in the block editor, since it triggers a post save after switching the post's language.
- Which means we can't select a non-translated category in the classic editor: no post save after switching the post's language => categories and tags are not translated/created => non-translated categories cannot be selected.
- I wasn't sure of where to put the 2 new private methods.

### Issue 1: terms not translated

Originally, the method was fetching the post's terms (`wp_get_object_terms()`) by name. The problem is that terms may not have the same name in different languages, which leads to terms not being translated.
The solution used here is to fetch all terms by their `term_taxonomy_ids`, and independently of the post. Then it's a `array_diff()` game with the post's terms to find which one must be translated/added to the post and which one must be removed.

### Issue 2: translation group when creating a missing term

In the way I coded this PR, there is a behavior change, because at first I thought it was a bug (but maybe it isn't). After discussing with the team, we're keeping the new behavior.

#### New behavior

When a term translation must be created (based on the name of the term in the default language (block editor) or previous language (classic editor)), the post's language is assigned to it, and its `term_id` is added as a translation.

#### Previous behavior

When a term translation must be created, it is not added as a translation of the previous term, it has its own translation group. Note: the language was assigned somewhere else.

### Issue 3: creation of a missing term in a hierarchical taxonomy

The previous version did not take care about the term's parent. The new version creates the term's parent if it's missing, recursively.

### Unit Tests

1. `Admin_Filters_Term_Test::test_create_term_from_post_bulk_edit` was breaking and has been fixed in [a2b7c11](https://github.com/polylang/polylang/pull/1336/commits/a2b7c1199d3869440f6926745cb427e7b0cff9af).
```
Attempt to read property "slug" on bool
/polylang/admin/admin-filters-term.php:327
```

2. `Admin_Filters_Term_Test::test_change_language_bulk_edit_translated_post_with_same_named_categories` was failing because of the behavior change in Issue 2 and has been fixed in [55bfaa5](https://github.com/polylang/polylang/pull/1336/commits/55bfaa556ae701290ed0a7c4c34fbf3ea1c95789).

3. The following tests already trigger this mechanic:
    - `Admin_Filters_Post_Test->test_save_post_with_categories()`.
    - `Admin_Filters_Post_Test->test_save_post_with_tags()`.
    - `Admin_Filters_Term_Test->test_create_term_from_post_bulk_edit()`.
    - `Admin_Filters_Term_Test->test_change_language_bulk_edit_translated_post_with_same_named_categories()`.
    - `CRUD_Posts_Test->test_language_change_with_taxonomy()`.
    - `Query_Test->test_single_post_private_translation()`.

5. A new test `CRUD_Posts_Test::test_terms_translation_on_post_save()` has been added.

### Manual Tests

This PR fixes the language of the categories and tags assigned to a post, to prevent having a post in one language and terms assigned to it in another one.
Things that can be tested:
- Post creation, then assign terms.
- Change the language of an existing post with terms assigned to it.
- Bulk editor for posts.

Temporary branch to test with Polylang Pro: https://github.com/polylang/polylang-pro/tree/fix-1766-post-term-lang.

#### Notes about the manual tests

- Imagine having a post in English that have a tag in English. Then change the post's language to French: if the tag doesn't have a French translation, it should be created. For categories and other hierarchical taxonomies, translation for the parent terms should also be created.
- With the classic editor, change the post's language:
    1. The tags in the metabox shouldn't change, but they will be translated after saving the post.
    2. The categories should be unselected because the categories are still in the old language while the metabox now displays the categories in the new language. Re-select them in the new language, they will be in the new language after saving the post. However, categories that don't exist in the new language cannot be selected.


include/crud-posts.php
```diff
@@ -340,8 +347,111 @@ public function force_tags_translation( $post_id, $post_after, $post_before ) {
...
+			}
+		}
+
+		$callback = function ( $lang, $tax, $slug ) use ( $language, $term, $taxonomy ) {
+			if ( ! $lang instanceof PLL_Language && $tax === $taxonomy && $slug === $term->slug ) {
+				return $language;
+			}
+			return $lang;
+		};
+		add_filter( 'pll_inserted_term_language', $callback, 10, 3 );
```
Hug0-Drelon, line 425: Do you think we have to filter `pll_inserted_term_parent` as well here?

include/crud-posts.php
```diff
@@ -340,8 +347,111 @@ public function force_tags_translation( $post_id, $post_after, $post_before ) {
...
+	/**
+	 * Translates the given term into the given language.
+	 * If the translation doesn't exist, it is created (for a hierarchical taxonomy, terms are created recursively).
+	 *
+	 * @since 3.5
+	 *
+	 * @param WP_Term      $term     The term to translate.
+	 * @param string       $taxonomy The term's taxonomy.
+	 * @param PLL_Language $language The language to translate the term into.
+	 * @return int
```
Hug0-Drelon, line 387: Maybe you could add doc for `@return`?

tests/phpunit/tests/test-admin-filters-term.php
```diff
@@ -178,6 +185,168 @@ public function test_create_term_from_post_bulk_edit() {
 		$this->assertEqualSets( array( $new_fr, $test_fr, $third ), $tags );
 	}
 
+	/**
+	 * @ticket #1766 see {https://github.com/polylang/polylang-pro/issues/1766}.
+	 */
+	public function test_terms_translation_on_post_save() {
```
Hug0-Drelon, line 191: I'd rather split the test, at least one for categories and another for tags (you could still reuse some code between tests).
Also I wonder if the test belongs to `Admin_Filters_Term_Test` since it's not only related to admin.

include/crud-posts.php
Screenfeed: Done in [8d687bf](https://github.com/polylang/polylang/pull/1336/commits/8d687bfbedef0d65af4cae96d8bf64b16c583bab).

tests/phpunit/tests/test-admin-filters-term.php
Screenfeed: Moved to `CRUD_Posts_Test` in [1259983](https://github.com/polylang/polylang/pull/1336/commits/125998325cdcd4f7cc7934a0fd45294bce8364cb).
However, what would be the benefits of splitting it?

Chrystll:
Here are my tests. It just to remember what I tested.

**Tags**
EN Created first tag EN - slug => toto-tag (default language)
FR Toto tag - slug => toto-tag
DE Toto tag - slug => toto-tag
ES no tag created

Block editor
Asssign a translated tag "Created first tag EN" toto-tag to the EN post > publish
-> switch to the FR
--> Toto tag is automatically checked, in DB this is well the tag in FR
-> switch to the DE
--> Toto tag is automatically checked, in DB this is well the tag in DE
-> switch to the ES
--> a tag is automatically created in ES and linked by the translation as here "Created first tag EN"

=> I get the same behavior when changing the language of a translated post.

Edit a post in EN > Add new tag > add "News" > save
-> switch to FR
--> "News" is automatically created on the fly and linked by the translation.

But I don't get the same behavior when doing
-> make sure that the duplicate tool is deactivated > then from the post list click on the DE plus sign > "News" is *not* created on the fly 
=> this is the expected behavior since this part is handled by the duplicate tool.

Bulk edit a post in FR and in DE (not linked by the translation)
-> create a new tag "Hello" > save
--> the tag has been added to these 2 posts and it has been created in the Tag interface in FR and DE and linked by the translation.

Chrystll:
I retested the issues below to be sure that this PR hasn't impacted the fixes. All is fine.

- https://github.com/polylang/polylang-pro/issues/1705
- https://github.com/polylang/polylang-pro/issues/1712

tests/phpunit/tests/test-admin-filters-term.php
Screenfeed: Done in [dd4b6c1](https://github.com/polylang/polylang/pull/1336/commits/dd4b6c11d718741d74821b029c88d1ab045e05b1).

Hug0-Drelon (approved):

PR #1339, opened 2023-09-01
Replaces pre_get_posts by parse_query in PLL_Frontend_Auto_Translate

Marianne380:
In some cases, we don't pass through `pre_get_posts` but through `parse_query`, so the query was not filtered.

By passing through `parse_query`, we can always filter the query.

For instance, for WooCommerce blocks of type `AbstractProductGrid`, we passed through a method that creates a `BlocksWpQuery` and therefore passes through `parse_query()` without passing through a `get_posts()`, which meant that our query wasn't filtered ( since we didn't pass through the `pre_get_posts` hook).
So the products in the block were not filtered in the right language.

### DIFF
--- a/frontend/frontend-auto-translate.php
+++ b/frontend/frontend-auto-translate.php
@@ -33,7 +33,7 @@ class PLL_Frontend_Auto_Translate {
 		$this->model = &$polylang->model;
 		$this->curlang = &$polylang->curlang;
 
-		add_action( 'pre_get_posts', array( $this, 'pre_get_posts' ) ); // after main Polylang filter
+		add_action( 'parse_query', array( $this, 'pre_get_posts' ), 20 ); // after main Polylang filter
 		add_filter( 'get_terms_args', array( $this, 'get_terms_args' ), 20, 2 );
 	}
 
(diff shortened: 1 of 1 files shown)


frontend/frontend-auto-translate.php
```diff
@@ -33,7 +33,7 @@ public function __construct( &$polylang ) {
 		$this->model = &$polylang->model;
 		$this->curlang = &$polylang->curlang;
 
-		add_action( 'pre_get_posts', array( $this, 'pre_get_posts' ) ); // after main Polylang filter
+		add_action( 'parse_query', array( $this, 'pre_get_posts' ), 100 ); // After all Polylang filters.
```
Hug0-Drelon, line 36: I think it worths renaming `$this->pre_get_posts`, don't you think?
Also, do you see any edge case (like WC block that should be translated on the fly on front) to test?

Marianne380: I had the same thought when I changed the filter, isn't it risky to rename a public method?

Hug0-Drelon: We could still implement a depreciation, what do you think @Chouby ?

PR #1340, opened 2023-09-04
Prevent Patchwork to print text during init

Screenfeed:
This is a quick and durty patch to prevent Patchwork to print `;\Patchwork\CodeManipulation\Stream::reinstateWrapper();` in CLI during the tests init.

It seems that requiring the empty file `wp-content/db.php` triggers the issue. I couldn't find any ticket about it on Patchwork's side. So, since it isn't a big issue and since we may someday rework the bootstrap (like we did in Polylang Pro), I opted for a quick and dirty fix.

Note: the issue is not triggered with Polylang Pro.

### DIFF
--- a/tests/bin/install-wp-tests.sh
+++ b/tests/bin/install-wp-tests.sh
@@ -91,8 +91,6 @@ install_wp() {
 		download https://wordpress.org/${ARCHIVE_NAME}.tar.gz  $TMPDIR/wordpress.tar.gz
 		tar --strip-components=1 -zxmf $TMPDIR/wordpress.tar.gz -C $WP_CORE_DIR
 	fi
-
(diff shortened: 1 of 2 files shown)


Chouby (changes requested):
This file is useless. In fact it's downloaded by `install-wp-tests.sh`. See https://github.com/polylang/polylang/blob/master/tests/bin/install-wp-tests.sh#L95
I did not attempt to find out why it's empty when it should not.
But in my opinion, the quickest way to solve the issue would be to remove this line (I tested and it works).
It's worthy to note that [this is proposed for the original file too](https://github.com/wp-cli/scaffold-command/issues/322).

Screenfeed:
> the quickest way to solve the issue would be to remove this line (I tested and it works).

Done in [18541f3](https://github.com/polylang/polylang/pull/1340/commits/18541f3e5b524095b9bd9a56e999222dcf22da2e).

PR #1344, opened 2023-09-12
Use dynamic filter for screen options

Chouby:
This PR uses dedicated filters to save our screen options instead of the legacy generic filter.
This minor improvement is possible since the dynamic filter has been introduced in WP 5.4.2.


manooweb (approved):
👌Tests with WordPress nightly has been fixed since this PR has been proposed

PR #1345, opened 2023-09-13
Fix rewrite rules for WordPress 6.4.

Hug0-Drelon:
## What?
Some changes have been made in the way WordPress flushes rewrite rules for 6.4 version.
See https://github.com/WordPress/wordpress-develop/commit/c10caf93d5ece0b73bd25d9ecad431ebc6891b41#diff-9cf1519e03e09d5698082a4594d70539d9ce18aa16defbbae15a982583dec4efL1867
Now, `update_options( 'rewrite_rules', '' )` isn't called anymore, preventing us to set up our [filters](https://github.com/polylang/polylang/blob/3fbfa8000512f449a23e3b68b2b9f754d583065a/include/links-directory.php#L54) when `pre_option_rewrite_rules` is called.
This means rewrite rules are broken when flushed from WordPress settings (not ours).

## How?
- Hook to `wp_loaded` to tell rewrite rules filters can be prepared (the same way as WordPress does now).
- Create a new action `pll_prepare_rewrite_rules` telling when Polylang is able to prepare rewrite rules filters.

*Note: Here is the WordPress ticket motivating the changes https://core.trac.wordpress.org/ticket/58998.*


Screenfeed (changes requested):

include/base.php
```diff
@@ -144,7 +144,7 @@ public function load_strings_translations( $locale = '' ) {
 	public function switch_blog( $new_blog_id, $prev_blog_id ) {
 		if ( $this->is_active_on_new_blog( $new_blog_id, $prev_blog_id ) ) {
 			$this->options = get_option( 'polylang' ); // Needed for menus.
-			remove_action( 'pre_option_rewrite_rules', array( $this->links_model, 'prepare_rewrite_rules' ) );
+			remove_action( 'wp_loaded', array( $this->links_model, 'prepare_rewrite_rules' ) );
```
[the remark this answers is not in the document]
Hug0-Drelon: As of today, it doesn't seem to be removed in Pro or WC when blog is switched.
Instead it seems to be re-init in Pro. See https://github.com/polylang/polylang-pro/blob/fcb3c9b06c5dfb3b4ac44852f5fc8dabc68b20d9/modules/translate-slugs/translate-slugs-model.php#L75.

include/links-directory.php
```diff
@@ -34,6 +34,8 @@ public function __construct( &$model ) {
 		} else {
 			add_action( 'pll_init', array( $this, 'init' ) );
 		}
+
+		add_action( 'pll_prepare_rewrite_rules', array( $this, 'prepare_rewrite_rules' ) );
```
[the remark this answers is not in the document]
Hug0-Drelon: Nice catch, done in 14c86935.

include/base.php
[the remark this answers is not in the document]
Screenfeed: Possibly an issue in PLLWC 🤔

Screenfeed (approved):

include/links-directory.php
```diff
@@ -44,14 +40,13 @@ public function __construct( &$model ) {
 	 * @return void
 	 */
 	public function init() {
```
Chouby, line 42: I believe that we can simplify and suppress this method to avoid one function call. `add_permastruct()` could be called directly from the constructor.

include/links-permalinks.php
```diff
@@ -60,6 +60,47 @@ public function __construct( &$model ) {
 		$permalink_structure = get_option( 'permalink_structure' );
 		$this->root = preg_match( '#^/*' . $this->index . '#', $permalink_structure ) ? $this->index . '/' : '';
 		$this->use_trailing_slashes = ( '/' == substr( $permalink_structure, -1, 1 ) );
+
+		if ( did_action( 'pll_init' ) ) {
+			$this->init();
+		} else {
+			add_action( 'pll_init', array( $this, 'init' ) );
+		}
```
Chouby, line 68: We should simplify and remove this.

include/links-permalinks.php
```diff
@@ -60,6 +60,47 @@ public function __construct( &$model ) {
...
+	 *
+	 * @since 3.5
+	 *
+	 * @return void
+	 */
+	public function init() {
+		if ( did_action( 'wp_loaded' ) ) {
+			$this->can_prepare_rewrite_rules();
+		} else {
+			add_action( 'wp_loaded', array( $this, 'can_prepare_rewrite_rules' ), 9 ); // Just before WordPress callback `WP_Rewrite::flush_rules()`.
```
Chouby, line 82: This could be moved to the constructor.

include/base.php
```diff
@@ -136,17 +144,66 @@ public function load_strings_translations( $locale = '' ) {
...
 			$this->options = get_option( 'polylang' ); // Needed for menus.
-			remove_action( 'pre_option_rewrite_rules', array( $this->links_model, 'prepare_rewrite_rules' ) );
 			$this->links_model = $this->model->get_links_model();
 		}
+
+		if ( 'restore' === $context ) {
+			$this->restore_prepare_rewrite_rules_actions();
+		} else {
+			$this->remove_prepare_rewrite_rules_actions();
+		}
```
Chouby, line 164: This should not be needed but replaced by `remove_all_actions( 'pll_prepare_rewrite_rules' )` which removes all rewrite rules filters related to the old site. The new links mode instance on the new site will fire an action which will add the filters related to the new site. Restoring the current site is not a spacial case as the same process applies.

Chouby (changes requested):

include/base.php
Hug0-Drelon: Indeed, this is stil a WIP. Since the test doesn't behave as I expected at the moment ...

include/links-permalinks.php
Hug0-Drelon: Done in b6eb780a.

include/links-permalinks.php
Hug0-Drelon: Done in b6eb780a.

include/links-directory.php
```diff
@@ -285,4 +268,21 @@ public function rewrite_rules( $rules ) {
...
+	 * @return void
+	 */
+	public function remove_hooks() {
+		remove_filter( 'language_rewrite_rules', '__return_empty_array' ); // Suppress the rules created by WordPress for our taxonomy.
+
+		foreach ( $this->get_rewrite_rules_filters() as $type ) {
+			remove_filter( $type . '_rewrite_rules', array( $this, 'rewrite_rules' ) );
+		}
+
+		remove_filter( 'rewrite_rules_array', array( $this, 'rewrite_rules' ) ); // Needed for post type archives.
```
Chouby, line 286: To avoid duplicate code, what do you think to move `get_rewrite_rules_filters()` to this class (as it's not used anywhere else than here) and rewrite it to include all filters like this?
```php
	public function get_rewrite_rules_filters() {
		// Make sure that we have the right post types and taxonomies.
		$types = array_values( array_merge( $this->model->get_translated_post_types(), $this->model->get_translated_taxonomies(), $this->model->get_filtered_taxonomies() ) );
		$types = array_merge( $this->always_rewrite, $types );

		/**
		 * Filters the list of rewrite rules filters to be used by Polylang.
		 *
		 * @since 0.8.1
		 *
		 * @param array $types The list of filters (without '_rewrite_rules' at the end).
		 */
		$types = apply_filters( 'pll_rewrite_rules', $types );
		
		filters = array(
			'language_rewrite_rules' => '__return_empty_array',
			rewrite_rules_array'     => array( $this, 'rewrite_rules' ),
		);
		
		foreach ( $types as $type ) {
			$filters[ $type . '_rewrite_rules' ] = array( $this, 'rewrite_rules' );
		}
		
		return $filters;
	}
	
	
	public function prepare_rewrite_rules() {
		if ( ! $this->model->has_languages() || ! did_action( 'wp_loaded' ) || has_filter( 'language_rewrite_rules', '__return_empty_array' ) {
			return;
		}
		
		foreach ( $this->get_rewrite_rules_filters() as $rule => $callback ) {
			add_filter( $rule, $callback );
		}
	}
	
	public function remove_hooks() {
		foreach ( $this->get_rewrite_rules_filters() as $rule => $callback ) {
			remove_filter( $rule, $callback );
		}	
	}
```

Hug0-Drelon: Good idea, done in 43ab719d.

include/links-permalinks.php
Hug0-Drelon: Found that's needed for our other plugins (especially for the test suites).

include/base.php
```diff
@@ -142,10 +142,13 @@ public function load_strings_translations( $locale = '' ) {
 	 * @return void
 	 */
 	public function switch_blog( $new_blog_id, $prev_blog_id ) {
+		remove_all_actions( 'pll_prepare_rewrite_rules' );
+		$this->links_model->remove_hooks();
+
 		if ( $this->is_active_on_new_blog( $new_blog_id, $prev_blog_id ) ) {
 			$this->options = get_option( 'polylang' ); // Needed for menus.
-			remove_action( 'pre_option_rewrite_rules', array( $this->links_model, 'prepare_rewrite_rules' ) );
 			$this->links_model = $this->model->get_links_model();
```
Screenfeed, line 150: Now these can be aligned 🌈 
```php
			$this->options     = get_option( 'polylang' ); // Needed for menus.
			$this->links_model = $this->model->get_links_model();
```

include/links-permalinks.php
```diff
@@ -62,6 +62,41 @@ public function __construct( &$model ) {
...
+
+	/**
+	 * Fires our own action telling Polylang plugins
+	 * and third parties are able to prepare rewrite rules.
+	 *
+	 * @since 3.5
+	 *
+	 * @return void
+	 */
+	public function can_prepare_rewrite_rules() {
```
Screenfeed, line 88: This method name makes no sense to me 😅
Since it starts with `can_`, I expect it to return a boolean 😖

```
init()
prepare_rewrite_rules()
can_prepare_rewrite_rules()
remove_hooks()
```

Am I the only one being confused?

Screenfeed (changes requested):

include/links-permalinks.php
Hug0-Drelon: Indeed, it's a bit confusing.
What do you suggest?

Hug0-Drelon: For instance:
```php
init();
prepare_rewrite_rules();
fire_prepare_rewrite_rules();
unprepare_rewrite_rules();
```
?

Screenfeed: Feels better indeed.

Hug0-Drelon: Done in f8a2ee95.

include/base.php
Hug0-Drelon: Aligned in 60cd8fed.

Screenfeed (approved):

Chouby (changes requested):

include/links-permalinks.php
```diff
@@ -62,6 +62,41 @@ public function __construct( &$model ) {
...
+
+	/**
+	 * Fires our own action telling Polylang plugins
+	 * and third parties are able to prepare rewrite rules.
+	 *
+	 * @since 3.5
+	 *
+	 * @return void
+	 */
+	public function fire_prepare_rewrite_rules() {
```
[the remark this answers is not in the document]
Screenfeed: 🔥🔥

include/base.php
```diff
@@ -142,10 +142,12 @@ public function load_strings_translations( $locale = '' ) {
...
 	 */
 	public function switch_blog( $new_blog_id, $prev_blog_id ) {
+		$this->links_model->remove_filters();
+
 		if ( $this->is_active_on_new_blog( $new_blog_id, $prev_blog_id ) ) {
-			$this->options = get_option( 'polylang' ); // Needed for menus.
-			remove_action( 'pre_option_rewrite_rules', array( $this->links_model, 'prepare_rewrite_rules' ) );
+			$this->options     = get_option( 'polylang' ); // Needed for menus.
 			$this->links_model = $this->model->get_links_model();
+			$this->links_model->init();
```
Chouby, line 150: It shouldn't be useful anymore.
```suggestion
```

include/links-directory.php
```diff
@@ -28,30 +28,25 @@ public function __construct( &$model ) {
...
-			add_action( 'pll_init', array( $this, 'init' ) );
-		}
 	}
 
 	/**
-	 * Called only at first object creation to avoid duplicating filters when switching blog.
+	 * Initializes links.
 	 *
-	 * @since 1.6
+	 * @since 3.5
```
Chouby, line 36:
```suggestion
	 * @since 1.6
```

include/links-directory.php
```diff
@@ -28,30 +28,25 @@ public function __construct( &$model ) {
...
-		if ( did_action( 'pll_init' ) ) {
-			$this->init();
-		} else {
-			add_action( 'pll_init', array( $this, 'init' ) );
-		}
 	}
 
 	/**
-	 * Called only at first object creation to avoid duplicating filters when switching blog.
+	 * Initializes links.
```
Chouby, line 34:
```suggestion
	 * Adds hooks for rewrite rules.
```

include/links-permalinks.php
```diff
@@ -164,4 +201,17 @@ public function get_rewrite_rules_filters() {
...
+
+	/**
+	 * Removes hooks to filter rewrite rules, called when switching blog @see {PLL_Base::switch_blog()}.
+	 *
+	 * @since 3.5
+	 *
+	 * @return void
+	 */
+	public function remove_filters() {
+		parent::remove_filters();
```
Chouby, line 213: Not needed as the parent method does nothing. 
Note for my future self. It's useful anyway if someday the parent method does something as we may not think to look at this child method at this future time.

tests/phpunit/includes/links-trait.php
```diff
@@ -55,6 +55,7 @@ protected function init_links_model() {
 		$wp_rewrite->init();
 		$wp_rewrite->set_permalink_structure( $this->structure );
 		$this->links_model = self::$model->get_links_model();
+		$this->links_model->init();
```
Chouby, line 58: Not needed.

include/links-directory.php
```diff
@@ -285,4 +277,38 @@ public function rewrite_rules( $rules ) {
...
+	 * See `self::prepare_rewrite_rules()` for added hooks.
+	 *
+	 * @since 3.5
+	 *
+	 * @return void
+	 */
+	public function remove_filters() {
+		parent::remove_filters();
+
+		remove_all_actions( 'pll_prepare_rewrite_rules' );
```
Chouby, line 292: Something has been lost in the loop.
This part is already done in the parent method.

include/links-model.php
```diff
@@ -48,9 +48,24 @@ public function __construct( &$model ) {
 
 		$this->home = home_url();
 
-		// Hooked with normal priority because it needs to be run after static pages is set in language data.
+		// Hooked with normal priority because it needs to be run after static pages is set in language data. Must be done early (before languages objects are created).
```
Chouby, line 51: This additional comment makes me fear that moving the other filters to `init()` is also a bad idea. Sorry for the wrong path I pushed you.

include/links-model.php
```diff
@@ -48,9 +48,24 @@ public function __construct( &$model ) {
...
+	}
+
+	/**
+	 * Initializes links.
+	 *
+	 * @since 3.5
+	 *
+	 * @return void
+	 */
+	public function init() {
```
Chouby, line 68: Based on the comment above, we should go back and let this method do nothing.
```suggestion
	/**
	 * Initializes the links model.
	 * Does nothing by default.
	 *
	 * @since 3.5
	 *
	 * @return void
	 */
	public function init() {}
```

include/links-model.php
```diff
@@ -226,4 +241,14 @@ public function set_language_home_url( $url, $language ) {
 	public function set_language_search_url( $url, $language ) {
 		return $this->home_url( $language['slug'] );
 	}
+
+	/**
+	 * Removes hooks to filter rewrite rules, called when switching blog @see {PLL_Base::switch_blog()}.
```
Chouby, line 246:
```suggestion
	 * Used to remove hooks in child classes, called when switching blog @see {PLL_Base::switch_blog()}.
```

Chouby (changes requested):

include/links-permalinks.php
Hug0-Drelon: That's what I had in mind.

include/links-directory.php
Hug0-Drelon: Fixed in aa8be6e7.

Chrystll:
Fatal error on the front when testing multisite:

```
PHP Fatal error:  Uncaught Error: Call to undefined function is_plugin_active() in /polylang-pro/vendor/wpsyntex/polylang/include/base.php:167
Stack trace:
#0 /polylang-pro/vendor/wpsyntex/polylang/include/base.php(147): PLL_Base->is_active_on_new_blog('1', 2)
#1 /polylang-pro/vendor/wpsyntex/polylang/frontend/frontend.php(234): PLL_Base->switch_blog('1', 2)
#2 /multisite-1/wp-includes/class-wp-hook.php(312): PLL_Frontend->switch_blog('1', 2)
#3 /multisite-1/wp-includes/class-wp-hook.php(334): WP_Hook->apply_filters(NULL, Array)
#4 /multisite-1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#5 /multisite-1/wp-includes/ms-blogs.php(585): do_action('switch_blog', '1', 2, 'switch')
#6 /multisite-1/wp-includes/class-wp-site.php(324): switch_to_blog('1')
#7 /multisite-1/wp-includes/class-wp-site.php(238): WP_Site->get_details()
#8 /multisite-1/wp-includes/user.php(1021): WP_Site->__get('blogname')
#9 /multisite-1/wp-includes/class-wp-admin-bar.php(41): get_blogs_of_user(1)
#10 /multisite-1/wp-includes/admin-bar.php(49): WP_Admin_Bar->initialize()
#11 /multisite-1/wp-includes/class-wp-hook.php(310): _wp_admin_bar_init('')
#12 /multisite-1/wp-includes/class-wp-hook.php(334): WP_Hook->apply_filters('', Array)
#13 /multisite-1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#14 /multisite-1/wp-includes/template-loader.php(13): do_action('template_redire...')
#15 /multisite-1/wp-blog-header.php(19): require_once('/home/christell...')
#16 /multisite-1/index.php(17): require('/home/christell...')
#17 {main}
  thrown in /polylang-pro/vendor/wpsyntex/polylang/include/base.php on line 167
```

include/functions.php
```diff
@@ -172,3 +172,20 @@ function pll_set_constant( string $constant_name, $value ): bool {
 
 	return define( $constant_name, $value ); // phpcs:ignore WordPressVIPMinimum.Constants.ConstantString.NotCheckingConstantName
 }
+
+/**
+ * Determines whether a plugin is active.
```
Chouby, line 177: I propose to add a comment on why this function is useful:
```suggestion
 * Determines whether a plugin is active.
 *
 * We define our own function because `is_plugin_active()` is available only in the backend.
```

PR #1352, opened 2023-09-22
Do not reset filters and `PLL_Links_Model` when switching toward same blog as current one.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/1867.

Also, PR #1345 showed up an issue when switching blogs, from one with pretty permalinks and language as directory to another with domains.
Indeed, when switching, and instance of `PLL_Links_Directory` is already hooked to filter the rewrite rules and thus when flushing them on the next blog, new rules are filtered too.

## How?
- Do not perform any processing when switching to the same blog as the current one.
- As for the test following PR #1345, it took care to remove filters properly when switching blogs, but it lacked of tests. This is the purpose of this PR.
- It also provides a test case to be used in all our plugins.

### DIFF
--- a/tests/phpunit/includes/testcase-multisites.php
+++ b/tests/phpunit/includes/testcase-multisites.php
@@ -155,6 +155,8 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 
 		// Set up blog with Polylang not activated and pretty permalinks.
 		$this->set_up_blog_without_pll( $this->blog_without_pll_pretty_links, $this->pretty_structure );
+
+		$this->clean_up_filters();
 	}
 
 	public function tear_down() {
--- a/tests/phpunit/includes/testcase-multisites.php
+++ b/tests/phpunit/includes/testcase-multisites.php
@@ -57,7 +57,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 	protected $plain_structure = '';
 
 	/**
-	 * Languages data for their creation keyed by language slug
+	 * Languages data for their creation keyed by language slug.
 	 *
 	 * @var array
 	 */
@@ -142,7 +142,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 			$this->pretty_structure
 		);
 
-		// Set up blog with Polylang activated, plain (i.e. default) permalinks , English and French created.
+		// Set up blog with Polylang activated, plain (i.e. default) permalinks, English and French created.
 		$this->set_up_blog_with_pll(
 			self::$blog_with_pll_default_links,
 			array( $this->languages['en'], $this->languages['fr'] ),
@@ -182,7 +182,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 	 * @param string  $structure Permalink structure to use.
 	 * @return void
 	 */
-	protected function set_up_blog_with_pll( $blog, $languages, $options, $structure ) {
+	protected function set_up_blog_with_pll( WP_Site $blog, array $languages, array $options, string $structure ) {
 		global $wp_rewrite;
 
 		switch_to_blog( $blog->blog_id );
@@ -220,7 +220,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 	 * @param string  $structure Permalinks structure to use.
 	 * @return void
 	 */
-	protected function set_up_blog_without_pll( $blog, $structure ) {
+	protected function set_up_blog_without_pll( WP_Site $blog, string $structure ) {
 		global $wp_rewrite;
 
 		switch_to_blog( $blog->blog_id );
@@ -257,7 +257,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 	 * Returns an instance of the main Polylang object along required instanciated classes for the tests.
 	 *
 	 * @param array $options Plugin options.
-	 * @return PLL_Base Polylang main class instance.
+	 * @return PLL_Admin_Base Polylang main class instance.
 	 */
-	abstract protected function get_pll_env( $options );
+	abstract protected function get_pll_env( array $options ): PLL_Admin_Base;
 }
--- a/tests/phpunit/includes/testcase-multisites.php
+++ b/tests/phpunit/includes/testcase-multisites.php
@@ -110,13 +110,13 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 
 		$this->blog_without_pll_plain_links = $factory->blog->create_and_get(
 			array(
-				'domain' => 'fake.com',
+				'domain' => 'plain.org',
 			)
 		);
 
 		$this->blog_without_pll_pretty_links = $factory->blog->create_and_get(
 			array(
-				'domain' => 'wordpress.org',
+				'domain' => 'pretty.org',
 			)
 		);
 
--- a/tests/phpunit/includes/testcase-multisites.php
+++ b/tests/phpunit/includes/testcase-multisites.php
@@ -94,31 +94,11 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 		// Create all sites.
 		$factory = $this->factory();
 
-		$this->blog_with_pll_directory = get_site( 1 );
-
-		$this->blog_with_pll_domains = $factory->blog->create_and_get(
-			array(
-				'domain' => 'polylang-domains.org',
-			)
-		);
-
-		$this->blog_with_pll_plain_links = $factory->blog->create_and_get(
-			array(
-				'domain' => 'polylang-plain.org',
-			)
-		);
-
-		$this->blog_without_pll_plain_links = $factory->blog->create_and_get(
-			array(
-				'domain' => 'plain.org',
-			)
-		);
-
-		$this->blog_without_pll_pretty_links = $factory->blog->create_and_get(
-			array(
-				'domain' => 'pretty.org',
-			)
-		);
+		$this->blog_with_pll_directory       = get_site( 1 );
+		$this->blog_with_pll_domains         = $factory->blog->create_and_get();
+		$this->blog_with_pll_plain_links     = $factory->blog->create_and_get();
+		$this->blog_without_pll_plain_links  = $factory->blog->create_and_get();
+		$this->blog_without_pll_pretty_links = $factory->blog->create_and_get();
 
 		// Set up blog with Polylang activated, permalinks as directory, English and French created.
 		$this->set_up_blog_with_pll(
--- a/tests/phpunit/includes/testcase-multisites.php
+++ b/tests/phpunit/includes/testcase-multisites.php
@@ -9,7 +9,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 	 *
 	 * @var WP_Site
 	 */
-	protected static $blog_without_pll_plain_links; // Default test website.
+	protected static $blog_without_pll_plain_links;
 
 	/**
 	 * Blog in pretty permalinks without Polylang.
@@ -32,7 +32,7 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 	 *
 	 * @var WP_Site
 	 */
-	protected static $blog_with_pll_directory;
+	protected static $blog_with_pll_directory; // Main blog.
 
 	/**
 	 * Blog in pretty permalinks with Polylang and language as domains.
@@ -93,19 +93,8 @@ abstract class PLL_Multisites_TestCase extends WP_UnitTestCase {
 
 		// Create all sites.
 		$factory = $this->factory();
-		self::$blog_without_pll_plain_links = get_blog_details();
 
-		self::$blog_without_pll_pretty_links = $factory->blog->create_and_get(
-			array(
-				'domain' => 'wordpress.org',
-			)
-		);
-
-		self::$blog_with_pll_directory = $factory->blog->create_and_get(
-			array(
-				'domain' => 'polylang-dir.org',
-			)
-		);
+		self::$blog_with_pll_directory = get_site( 1 );
 
 		self::$blog_with_pll_domains = $factory->blog->create_and_get(
 			array(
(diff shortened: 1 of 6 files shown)


Screenfeed (approved):

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	/**
+	 * Returns an instance of the main Polylang object along required instanciated classes for the tests.
+	 *
+	 * @param array $options Plugin options.
+	 * @param bool  $init    Trigger `PLL_Links_Model`'s and `PLL_Admin_Base`'s init or not. Default is `true`.
+	 * @return PLL_Admin_Base Polylang main class instance.
+	 */
+	protected function get_pll_admin_env( array $options = array(), bool $init = true ): PLL_Admin_Base {
```
Chouby, line 268: In fact, in always return `PLL_Admin`.

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -0,0 +1,312 @@
...
+		),
+		'de' => array(
+			'name'       => 'Deutsch',
+			'slug'       => 'de',
+			'locale'     => 'de_DE',
+			'rtl'        => 0,
+			'flag'       => 'de',
+			'term_group' => 2,
+		),
+	);
```
Chouby, line 89: Why not to extend `PLL_Unit_Test_Case` and use its `create_language()` method?

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -0,0 +1,312 @@
...
+
+	public function set_up() {
+		parent::set_up();
+
+		// Create all sites.
+		$factory = $this->factory();
+
+		self::$blog_with_pll_directory = get_site( 1 );
```
Chouby, line 97: Why using static properties if they are not used in static methods?
Here 4 sites are created per test (so 24 for all 6 tests). Is it intentional or did  you want to create 4 tests per test class?

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -0,0 +1,312 @@
...
+			array(
+				'domain' => 'fake.com',
+			)
+		);
+
+		self::$blog_without_pll_pretty_links = $factory->blog->create_and_get(
+			array(
+				'domain' => 'wordpress.org',
+			)
+		);
```
Chouby, line 121: To improve readability in case we use these domains in tests:
```suggestion
		self::$blog_without_pll_plain_links = $factory->blog->create_and_get(
			array(
				'domain' => 'plain.org',
			)
		);

		self::$blog_without_pll_pretty_links = $factory->blog->create_and_get(
			array(
				'domain' => 'pretty.org',
			)
		);
```

tests/phpunit/tests/test-switch-blog-query.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+if ( is_multisite() ) :
+
+	class Switch_Blog_Urls_Test extends PLL_Multisites_TestCase {
```
Chouby, line 5: To match the file name:
```suggestion
	class Switch_Blog_Query_Test extends PLL_Multisites_TestCase {
```

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -0,0 +1,312 @@
...
+		update_option( 'active_plugins', array_diff( $plugins, $this->get_plugin_names() ) ); // Ensure Polylang plugins are deactivated.
+
+		restore_current_blog();
+	}
+
+	/**
+	 * Removes *all* filters that could interfere with a test.
+	 * Must be called *before* creating a new Polylang environment in a test.
+	 */
+	protected function clean_up_filters() {
```
Chouby, line 239: Why is this necessary? I expect that WordPress will clean up all filters added during the tests.
NB: when I run all 6 tests without the method, only the last one fails.

tests/phpunit/tests/test-switch-blog-query.php
```diff
@@ -0,0 +1,83 @@
...
+
+			$this->clean_up_filters();
+
+			switch_to_blog( (int) self::$blog_with_pll_directory->blog_id );
+
+			$pll_admin = $this->get_pll_admin_env();
+			do_action_ref_array( 'pll_init', array( &$pll_admin ) );
+
+			$post = $this->factory()->post->create();
+			$pll_admin->model->post->set_language( $post, 'fr' );
```
Chouby, line 63: What's the purpose? When I read the test title, we just need to switch the blog twice. No need to create a new post.

tests/phpunit/tests/test-switch-blog-query.php
```diff
@@ -0,0 +1,83 @@
...
+			restore_current_blog(); // Restore to switch back, to ensure rewrite rules filters are set back correctly.
+
+			$wp_rewrite->init();
+			flush_rewrite_rules();
+
+			$pll_admin->curlang = $pll_admin->model->get_language( 'fr' ); // Force current language.
+
+			$this->go_to( 'http://example.org/fr' );
+
+			$this->assertQueryTrue( 'is_home', 'is_front_page' );
```
Chouby, line 79: It's not consistent to be on admin side of Polylang and to query a frontend url.

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
```diff
@@ -0,0 +1,121 @@
...
+			$this->assertArrayHasKey( '(fr)/?$', $rules );
+			$this->assertArrayNotHasKey( '(fr)/(fr)/?$', $rules );
+
+			$languages = $pll_admin->model->get_languages_list();
+
+			$this->assertCount( 2, $languages );
+			$this->assertSame( 'en', $languages[0]->slug );
+			$this->assertSame( 'http://' . self::$blog_with_pll_directory->domain . self::$blog_with_pll_directory->path, $languages[0]->get_home_url() );
+			$this->assertSame( 'fr', $languages[1]->slug );
+			$this->assertSame( 'http://' . self::$blog_with_pll_directory->domain . self::$blog_with_pll_directory->path . 'fr/', $languages[1]->get_home_url() );
```
Chouby, line 34: If you are obliged to add the path in the url, then it means that there is no domain mapping. So I expect that it's useless to create sites on a specific domain and you can keep the domain param empty to keep the default one.

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
```diff
@@ -0,0 +1,121 @@
...
+			 */
+			switch_to_blog( (int) self::$blog_with_pll_domains->blog_id );
+
+			$wp_rewrite->init();
+			$wp_rewrite->flush_rules();
+			$rules = $wp_rewrite->wp_rewrite_rules();
+
+			$this->assertNotEmpty( $rules );
+			$this->assertArrayNotHasKey( '(en)/?$', $rules );
+			$this->assertArrayNotHasKey( '(de)/?$', $rules );
```
Chouby, line 49: We have no FR language on this site, but we can check that there is no leak from the previous site: 
```suggestion
			$this->assertArrayNotHasKey( '(de)/?$', $rules );
			$this->assertArrayNotHasKey( '(fr)/?$', $rules );
```

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
```diff
@@ -0,0 +1,121 @@
...
+
+			$this->assertEmpty( $rules );
+
+			$languages = $pll_admin->model->get_languages_list();
+
+			$this->assertCount( 2, $languages );
+			$this->assertSame( 'en', $languages[0]->slug );
+			$this->assertSame( 'http://' . self::$blog_with_pll_default_links->domain . self::$blog_with_pll_default_links->path, $languages[0]->get_home_url() );
+			$this->assertSame( 'fr', $languages[1]->slug );
+			$this->assertSame( 'http://' . self::$blog_with_pll_default_links->domain . self::$blog_with_pll_default_links->path . '?lang=fr', $languages[1]->get_home_url() );
```
Chouby, line 78: Same as above. A custom domain doesn't seem to be useful.

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
```diff
@@ -0,0 +1,121 @@
...
+			switch_to_blog( (int) self::$blog_with_pll_directory->blog_id );
+
+			$wp_rewrite->init();
+			$wp_rewrite->flush_rules();
+			$rules = $wp_rewrite->wp_rewrite_rules();
+
+			$this->assertNotEmpty( $rules );
+			$this->assertArrayNotHasKey( '(en)/?$', $rules );
+			$this->assertArrayHasKey( '(fr)/?$', $rules );
+			$this->assertArrayNotHasKey( '(fr)/(fr)/?$', $rules );
```
Chouby, line 26: I would not only check rules from the home group but also at least rules from another group like 'post' or 'category'. The same for the other sites lower.

Chouby (changes requested):

Chouby (changes requested):
Additionally, could you add a test for #1375?
When I remove https://github.com/polylang/polylang/pull/1375/files#diff-e3279fb11a89ee9bd6193072d031a98f3ab53e4c420168cf7cc98fd763edb0e8R234-R237 and https://github.com/polylang/polylang/pull/1375/files#diff-7bcdbc1cb24cba2bfcef22987b4a3bc3910e4972ac2cbf5a219a960327d8c939R145-R148, no test is currently failing.

Screenfeed:
> Additionally, could you add a test for #1375?

Done in [b35ba2d](https://github.com/polylang/polylang/pull/1352/commits/b35ba2d8d598ca9ff865b1cb25ec29d004917761).

tests/phpunit/tests/test-switch-blog-query.php
Screenfeed: Done in [1ef9372](https://github.com/polylang/polylang/pull/1352/commits/1ef93729e1b5fc6bab6a8af44a0cc6c96c41c3d2).

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -0,0 +1,312 @@
...
+	 */
+	protected static $blog_without_pll_pretty_links;
+
+	/**
+	 * Blog in plain permalinks with Polylang (i.e. default installation).
+	 * Created languages are English and French.
+	 *
+	 * @var WP_Site
+	 */
+	protected static $blog_with_pll_default_links;
```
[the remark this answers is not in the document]
Screenfeed: Done in [dc557eb](https://github.com/polylang/polylang/pull/1352/commits/dc557eb07144c516e66acbc4357d02eb769808d0).

tests/phpunit/includes/testcase-multisites.php
Screenfeed: Done in [6d6c6e7](https://github.com/polylang/polylang/pull/1352/commits/6d6c6e7c6ab110dbd623ff1a5685d597404d1bd7).

tests/phpunit/includes/testcase-multisites.php
Screenfeed: Static to instance props in [7a59368](https://github.com/polylang/polylang/pull/1352/commits/7a593681cd2fc84904eed42625320f728ce60bcb).
I don't know about the original intent, but the sites are deleted in `tear_down()`.

tests/phpunit/tests/test-switch-blog-query.php
Screenfeed: Changed in [5fc4b49](https://github.com/polylang/polylang/pull/1352/commits/5fc4b49393f5eaa534023b4d2fd9feee9fd88804).

tests/phpunit/includes/testcase-multisites.php
Screenfeed: Changed in [5732659](https://github.com/polylang/polylang/pull/1352/commits/57326590f4a2234c7f9e06f98a9e7c32b64ab781).

tests/phpunit/includes/testcase-multisites.php
Screenfeed: I just tried, it's a mess. The usual `Notice: Undefined index: default_lang`.
I tried to create and delete sites in `wpSetUpBeforeClass()` and `wpTearDownAfterClass()`: it improved things, but one test still breaks, along with the database itself (`One or more database tables are unavailable.`).

tests/phpunit/tests/test-switch-blog-query.php
Screenfeed: I wondered the same thing. Removed in [7a6d02b](https://github.com/polylang/polylang/pull/1352/commits/7a6d02bfc731ee2a49db6313955b6ed855f031da).

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
Screenfeed: Added in [a9273e3](https://github.com/polylang/polylang/pull/1352/commits/a9273e35d2ea81f1487274a4a74837a4f6712e87).

tests/phpunit/includes/testcase-multisites.php
Screenfeed: Because of [this](https://github.com/polylang/polylang/blob/a9273e35d2ea81f1487274a4a74837a4f6712e87/tests/phpunit/includes/testcase-multisites.php#L197). If we place a `$this->clean_up_filters();` right after this line, all fine; and if we place it before, it fails.

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
Screenfeed: Done in [d0e3811](https://github.com/polylang/polylang/pull/1352/commits/d0e3811ae3bb8c19f3cb6564e0e118471a5dfbcb).

tests/phpunit/tests/test-switch-blog-rewrite-rules.php
Screenfeed: Domains removed in [4b3bdc6](https://github.com/polylang/polylang/pull/1352/commits/4b3bdc6ecc7ee94cd8641d7270528c541266ed17).

PR #1355, opened 2023-09-28
Unhook `print_emoji_styles()` in tests.

Hug0-Drelon:
## What?
`print_emoji_styles()` is deprecated since WP 6.4, but is still hooked for backward compatibility.

## Why?
This triggers a depreciation warning when `wp_head()` is called in tests.

## How?
Remove the action in `PLL_UnitTestCase_Trait::wpSetUpBeforeClass()` so our other plugins benefit from it.

*Note: depreciation warning isn't triggered in Polylang test suite, only on Pro and WC.*

### DIFF
--- a/tests/phpunit/includes/testcase-trait.php
+++ b/tests/phpunit/includes/testcase-trait.php
@@ -65,6 +65,13 @@ trait PLL_UnitTestCase_Trait {
 		remove_action( 'current_screen', '_load_remote_block_patterns' );
 		remove_action( 'current_screen', '_load_remote_featured_patterns' );
 
+		if ( version_compare( $GLOBALS['wp_version'], '6.4', '>=' ) ) {
+			/*
+			 * `print_emoji_styles()` deprecated since WP 6.4, still hooked for backaward compatibility for other plugins @see {https://github.com/WordPress/wordpress-develop/commit/54c4de13edfaf3b9fbc1b67d6ba384618614d59e}.
+			 */
+			remove_action( 'wp_print_styles', 'print_emoji_styles' );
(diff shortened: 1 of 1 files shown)


tests/phpunit/includes/testcase-trait.php
```diff
@@ -65,6 +65,13 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) { // p
 		remove_action( 'current_screen', '_load_remote_block_patterns' );
 		remove_action( 'current_screen', '_load_remote_featured_patterns' );
 
+		if ( version_compare( $GLOBALS['wp_version'], '6.4', '>=' ) ) {
```
Chouby, line 68: We don't really need this check. Do we ? As far as I know we never tests emojis :wink: :roll_eyes:

Screenfeed (commented):
I'm not sure `version_compare()` is useful, we could probably remove it for any version.
While at it, maybe we should also remove it from `admin_print_styles`.
What do you think?

tests/phpunit/includes/testcase-trait.php
Hug0-Drelon: I thought it could be used by a third party running along our test suite for WP < 6.4. But maybe a bit overkill.
Removed in 95cc1901.

Hug0-Drelon:
> I'm not sure `version_compare()` is useful, we could probably remove it for any version. While at it, maybe we should also remove it from `admin_print_styles`. What do you think?

Good idea! Done in e0993503.

Screenfeed (approved):

PR #1359, opened 2023-09-29
Register the language taxonomy only once

Chouby:
Working on #1345 and more specifically on b294b355c428550847fb85c58f2c4d0b826c95fc, it looked like (since the beginning!) we are adding language rewrite rules to remove them later.

Moreover, we register the language taxonomy 2 times (as WP does) because of the `rewrite` param which cannot be added before `setup_theme` is fired.

#1457 removed the `rewrite` param from the language taxonomy and the related code now useless. This PR focuses on registering the taxonomy only once.

Tests are adapted. Probably it will be necessary to do the same in Polylang Pro and Polylang for WooCommerce.

### DIFF
--- a/include/translatable-object.php
+++ b/include/translatable-object.php
@@ -84,8 +84,18 @@ abstract class PLL_Translatable_Object {
 
 		/*
 		 * Register our taxonomy as soon as possible.
-		 * This is early registration, not ready for rewrite rules as $wp_rewrite will be setup later.
 		 */
+		$this->register_language_taxonomy();
+	}
+
+	/**
+	 * Registers the language taxonomy.
+	 *
+	 * @since 3.6
+	 *
+	 * @return void
+	 */
+	protected function register_language_taxonomy() {
 		register_taxonomy(
 			$this->tax_language,
 			(array) $this->object_type,
--- a/include/translatable-object.php
+++ b/include/translatable-object.php
@@ -95,7 +95,7 @@ abstract class PLL_Translatable_Object {
 	 *
 	 * @return void
 	 */
-	protected function register_language_taxonomy() {
+	protected function register_language_taxonomy(): void {
 		register_taxonomy(
 			$this->tax_language,
 			(array) $this->object_type,
--- a/include/translatable-object.php
+++ b/include/translatable-object.php
@@ -91,7 +91,7 @@ abstract class PLL_Translatable_Object {
 	/**
 	 * Registers the language taxonomy.
 	 *
-	 * @since 3.6
+	 * @since 3.7
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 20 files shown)


include/translatable-object.php
```diff
@@ -84,8 +84,18 @@ public function __construct( PLL_Model &$model ) {
...
+	}
+
+	/**
+	 * Registers the language taxonomy.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	protected function register_language_taxonomy() {
```
Hug0-Drelon, line 98:
```suggestion
	protected function register_language_taxonomy(): void {
```

include/translated-object.php
```diff
@@ -36,6 +36,17 @@ public function __construct( PLL_Model &$model ) {
...
+	}
+
+	/**
+	 * Registers the translations taxonomy.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	protected function register_translations_taxonomy() {
```
Hug0-Drelon, line 49:
```suggestion
	protected function register_translations_taxonomy(): void {
```

include/translated-post.php
```diff
@@ -66,6 +66,35 @@ public function __construct( PLL_Model &$model ) {
...
 
+	/**
+	 * Registers the language taxonomy.
+	 *
+	 * @since 1.2
+	 * @since 3.7 Protected and renamed from `register_taxonomy()`.
+	 *
+	 * @return void
+	 */
+	protected function register_language_taxonomy() {
```
Hug0-Drelon, line 77:
```suggestion
	protected function register_language_taxonomy(): void {
```

include/translated-post.php
Hug0-Drelon, line 1: This is not related to this PR, but reading the code I wonder if `show_in_rest` could benefit us some day.

include/translatable-object.php
Hug0-Drelon, line 1: I wonder if it's a good idea to take the opportunity to improve uninstall.
See [here](https://github.com/polylang/polylang/blob/dc074362e6e795090534cab59b19aee629c67bb9/uninstall.php#L56-L59), where we register taxonomies to remove data. Maybe we could find a way to reuse what's in `PLL_Translatable_Objects` to ensure we got all taxonomies. What do you think? Maybe it's too out of scope though.

Hug0-Drelon (changes requested):

include/translated-post.php
Chouby: I don't think so. We'll want to expose complete languages, not separated taxonomies.

Hug0-Drelon: Gotcha!

Hug0-Drelon (approved):

PR #1328, opened 2023-08-08
Anticipate WPCS 3.0

Chouby:
See https://github.com/polylang/polylang-pro/pull/1753 for more context.

PR #1362, opened 2023-10-05
Bump WPCS and VIPWCS to version 3.0

Chouby:
This is a follow up of #1328


phpcs.xml.dist
```diff
@@ -37,6 +37,9 @@
 	</rule>
 
 	<rule ref="WordPress">
+		<exclude name="Generic.CodeAnalysis.AssignmentInCondition.Found"/>
```
Hug0-Drelon, line 40: Do we really want to exclude this kind of sniff?
Adding it requires to many changes?

Chouby: No I don't want to exclude it in the long term. Yes, it's a lot of work to follow this rule. Especially in Polylang as it's the oldest plugin and I used this a lot.

Hug0-Drelon (approved):

PR #1370, opened 2023-10-17
Fix phpstan errors

manooweb:
To follow up https://github.com/polylang/polylang-pro/pull/1864
This PR propose to do the same kind of fixes.


Screenfeed (approved):
Not always easy to decide what to do if `get_term()` doesn't return a `WP_Term`.

PR #1371, opened 2023-10-18
Test rest routes for custom taxonomies

Marianne380:
Add a test following https://github.com/polylang/polylang-pro/issues/1861.

Check the rest filtered route for a custom taxonomy.

### DIFF
--- a/tests/phpunit/includes/testcase-preload-paths.php
+++ b/tests/phpunit/includes/testcase-preload-paths.php
@@ -13,10 +13,7 @@ abstract class PLL_Preload_Paths_TestCase extends PLL_UnitTestCase {
 
 		register_post_type( 'custom', array( 'public' => true, 'show_in_rest' => true ) ); // Untranslatable CPT.
 		register_post_type( 'trcpt', array( 'public' => true, 'show_in_rest' => true ) ); // Translated CPT.
-		register_taxonomy( 'trtax', 'post', array( 'show_in_rest' => true ) ); // Translated custom taxonomy.
-
-		register_taxonomy_for_object_type( 'trtax', 'trcpt' );
-
+		register_taxonomy( 'trtax', 'trcpt', array( 'show_in_rest' => true ) ); // Translated custom 
 		$options = array_merge(
 			PLL_Install::get_default_options(),
 			array(
--- a/tests/phpunit/includes/testcase-preload-paths.php
+++ b/tests/phpunit/includes/testcase-preload-paths.php
@@ -15,6 +15,8 @@ abstract class PLL_Preload_Paths_TestCase extends PLL_UnitTestCase {
 		register_post_type( 'trcpt', array( 'public' => true, 'show_in_rest' => true ) ); // Translated CPT.
 		register_taxonomy( 'trtax', 'post', array( 'show_in_rest' => true ) ); // Translated custom taxonomy.
 
+		register_taxonomy_for_object_type( 'trtax', 'trcpt' );
+
 		$options = array_merge(
 			PLL_Install::get_default_options(),
 			array(
@@ -44,8 +46,6 @@ abstract class PLL_Preload_Paths_TestCase extends PLL_UnitTestCase {
 		_unregister_post_type( 'custom' );
 		_unregister_post_type( 'trcpt' );
 		_unregister_taxonomy( 'trtax' );
-		_unregister_post_type( 'book' );
-		_unregister_taxonomy( 'genre' );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-preload-paths.php
```diff
@@ -89,4 +89,48 @@ public function test_preload_path_for_translatable_media() {
...
+		$get_method   = $filter_route->getMethod( 'get' );
+		$get_method->setAccessible( true );
+
+		$routes = $get_method->invokeArgs( new PLL_Filter_REST_Routes( $this->pll_admin->model ), array() );
+
+		$this->assertArrayHasKey( 'genre', $routes );
+		$this->assertSame( 'wp/v2/genre', $routes['genre'] );
+
+		_unregister_post_type( 'book' );
+		_unregister_taxonomy( 'genre' );
```
Hug0-Drelon, line 134: It must be done in `tear_down()` to avoid polluting other tests in case this one fails.

tests/phpunit/tests/test-preload-paths.php
```diff
@@ -89,4 +89,48 @@ public function test_preload_path_for_translatable_media() {
...
+		);
+
+		$filter_route = new ReflectionClass( PLL_Filter_REST_Routes::class );
+		$get_method   = $filter_route->getMethod( 'get' );
+		$get_method->setAccessible( true );
+
+		$routes = $get_method->invokeArgs( new PLL_Filter_REST_Routes( $this->pll_admin->model ), array() );
+
+		$this->assertArrayHasKey( 'genre', $routes );
+		$this->assertSame( 'wp/v2/genre', $routes['genre'] );
```
Hug0-Drelon, line 131: I suggest to use a public method to test the behavior. In our case `PLL_Filter_REST_Routes::add_query_parameters()` and passing a query parameter and its value, then ensure it's really added to the route.

If you really want to test a private method, maybe test `PLL_Filter_REST_Routes::is_filtered()` which returns a boolean. It'd simplify the code IMHO.

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-preload-paths.php
```diff
@@ -121,13 +121,11 @@ function ( $taxonomies ) {
...
-		$get_method->setAccessible( true );
+		$filter_route  = new PLL_Filter_REST_Routes( $this->pll_admin->model );
+		$preload_paths = $filter_route->add_query_parameters( array( '/wp/v2/genre' ), array( 'test' => 'something' ) );
 
-		$routes = $get_method->invokeArgs( new PLL_Filter_REST_Routes( $this->pll_admin->model ), array() );
-
-		$this->assertArrayHasKey( 'genre', $routes );
-		$this->assertSame( 'wp/v2/genre', $routes['genre'] );
+		// If the parameter is added to the route, this means that the route is one of the filterable routes.
+		$this->assertNotEmpty( $preload_paths );
```
Hug0-Drelon, line 128:
```suggestion
		$this->assertNotEmpty( $preload_paths );
		$this->assertCount( 1, $preload_paths );
```
Just to make sure nothing's added.

Hug0-Drelon (approved):

tests/phpunit/tests/test-preload-paths.php
```diff
@@ -89,4 +89,43 @@ public function test_preload_path_for_translatable_media() {
...
+				return array_merge( $post_types, array( 'book' => 'book' ) );
+			}
+		);
+
+		add_filter(
+			'pll_get_taxonomies',
+			function ( $taxonomies ) {
+				return array_merge( $taxonomies, array( 'genre' => 'genre' ) );
+			}
+		);
```
Chouby, line 122:
```suggestion
		register_taxonomy_for_object_type( 'trtax', 'trcpt' );
```
The test case already defines translated post types and a translated taxonomy with the same parameter so we should not need to add new ones.
Here I propose to redefine to add the previously defined taxonomy to the previously defined cpt instead of creating new ones. However, it my even be better to directly define the taxonomy for this cpt in the test case if doesn't break other tests or require only small adaptation.

Chouby (changes requested):

tests/phpunit/includes/testcase-preload-paths.php
```diff
@@ -15,6 +15,8 @@ public function set_up() {
 		register_post_type( 'trcpt', array( 'public' => true, 'show_in_rest' => true ) ); // Translated CPT.
 		register_taxonomy( 'trtax', 'post', array( 'show_in_rest' => true ) ); // Translated custom taxonomy.
 
+		register_taxonomy_for_object_type( 'trtax', 'trcpt' );
+
```
Chouby, line 19:
```suggestion
		register_taxonomy( 'trtax', 'trcpt', array( 'show_in_rest' => true ) ); // Translated custom 
```
or if you absolutely need to register it on 'post' because it would require too many changes:
```suggestion
		register_taxonomy( 'trtax', array( 'post, 'trcpt' ), array( 'show_in_rest' => true ) ); // Translated custom 
```

Chouby (changes requested):

PR #1376, opened 2023-10-24
Sets rest_namespace to default wp/v2 if not set.

Marianne380:
Issue with CPT UI, we had a `rest_namespace` set to `null`, so the filtered route was wrong.

This PR sets the default route if the `rest_namespace` is not defined, as [WP](https://github.com/WordPress/WordPress/blob/6.3.2/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php#L70).

### DIFF
--- a/include/filter-rest-routes.php
+++ b/include/filter-rest-routes.php
@@ -157,8 +157,10 @@ class PLL_Filter_REST_Routes {
 		$this->filtered_entities = array();
 		foreach ( $rest_entities as $rest_entity ) {
 			if ( in_array( $rest_entity->name, $translatable_entities, true ) ) {
-				$rest_base = empty( $rest_entity->rest_base ) ? $rest_entity->name : $rest_entity->rest_base;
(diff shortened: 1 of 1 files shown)


Chouby:
See how CPTUI explicitly sets the namespace to null [for post types](https://github.com/WebDevStudios/custom-post-type-ui/blob/1.14.0/custom-post-type-ui.php#L505-L508) and [for taxonomies](https://github.com/WebDevStudios/custom-post-type-ui/blob/1.14.0/custom-post-type-ui.php#L741-L743).
Meanwhile WordPress [compares the param only to `false`](https://github.com/WordPress/wordpress-develop/blob/6.3.2/src/wp-includes/class-wp-taxonomy.php#L425-L428). So the taxonomy (or post type) is stored with a null namespace instead of the expected `wp/v2`. There is no bug in the route due to [the second check in the REST controller](https://github.com/WordPress/wordpress-develop/blob/6.3.2/src/wp-includes/rest-api/endpoints/class-wp-rest-terms-controller.php#L70)

PR #1377, opened 2023-10-25
Add a test for filtered routes with a null rest_namespace.

Marianne380:
Following https://github.com/polylang/polylang/pull/1376.

Adds a test case with a taxonomy with a null `rest_namespace`.


tests/phpunit/includes/testcase-preload-paths.php
```diff
@@ -14,6 +14,7 @@ public function set_up() {
 		register_post_type( 'custom', array( 'public' => true, 'show_in_rest' => true ) ); // Untranslatable CPT.
 		register_post_type( 'trcpt', array( 'public' => true, 'show_in_rest' => true ) ); // Translated CPT.
 		register_taxonomy( 'trtax', 'trcpt', array( 'show_in_rest' => true ) ); // Translated custom taxonomy.
+		register_taxonomy( 'trtax_with_no_namespace', 'trcpt', array( 'show_in_rest' => true, 'rest_namespace' => null ) ); // Translated custom taxonomy with no namespace defined.
```
Chouby, line 17:
```suggestion
		register_taxonomy( 'trtax_with_no_namespace', 'trcpt', array( 'show_in_rest' => true, 'rest_namespace' => null ) ); // Translated custom taxonomy with namespace explicitly set to null as in CPTUI.
```

PR #1378, opened 2023-10-25
Force the current language in the switch_blog().

sebastienserre:
Force the current language in the switch_blog()
This commit should fix https://github.com/polylang/polylang-pro/issues/1870

Add a test.

### DIFF
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -259,6 +259,7 @@ class PLL_Frontend extends PLL_Base {
 			$this->static_pages->init();
 		}
 
+		// Send the slug instead of the locale here to avoid conflicts with same locales.
 		$this->load_strings_translations( $this->curlang->slug );
 	}
 
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -256,7 +256,7 @@ class PLL_Frontend extends PLL_Base {
 			$this->static_pages->init();
 		}
 
-		$this->load_strings_translations();
+		$this->load_strings_translations( $this->curlang );
 	}
 
 	/**
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -256,7 +256,7 @@ class PLL_Frontend extends PLL_Base {
 			$this->static_pages->init();
 		}
 
-		$this->load_strings_translations( $this->curlang );
+		$this->load_strings_translations( $this->curlang->slug );
 	}
 
 	/**
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -249,8 +249,11 @@ class PLL_Frontend extends PLL_Base {
 			$restore_curlang = $this->curlang->slug; // To always remember the current language through blogs.
 		}
 
-		$lang = $this->model->get_language( $restore_curlang );
+		$lang          = $this->model->get_language( $restore_curlang );
 		$this->curlang = $lang ? $lang : $this->model->get_default_language();
+		if ( empty( $this->curlang ) ) {
+			return;
+		}
 
 		if ( isset( $this->static_pages ) ) {
 			$this->static_pages->init();
--- a/include/base.php
+++ b/include/base.php
@@ -112,7 +112,7 @@ abstract class PLL_Base {
 	 * @since 1.2
 	 * @since 2.1.3 $locale parameter added.
 	 *
-	 * @param string $locale Locale. Defaults to current locale.
+	 * @param string $locale Locale. Defaults to current locale. It can also be a slug in some cases.
 	 * @return void
 	 */
 	public function load_strings_translations( $locale = '' ) {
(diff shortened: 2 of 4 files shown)


frontend/frontend.php
```diff
@@ -249,14 +249,17 @@ public function switch_blog( $new_blog_id, $prev_blog_id ) {
...
+		if ( empty( $this->curlang ) ) {
+			return;
+		}
 
 		if ( isset( $this->static_pages ) ) {
 			$this->static_pages->init();
 		}
 
-		$this->load_strings_translations();
+		$this->load_strings_translations( $this->curlang->slug );
```
Hug0-Drelon, line 262:
```suggestion
		$this->load_strings_translations( $this->curlang->locale );
```
Shouldn't it be the locale instead of the slug?

tests/phpunit/tests/test-switch-blog.php
```diff
@@ -0,0 +1,70 @@
...
+		 * @see https://github.com/polylang/polylang-pro/issues/1870.
+		 */
+		public function test_strings_translation_with_same_language_locale_with_multisite() {
+			$pll_frontend          = $this->get_pll_frontend_env( array( 'force_lang' => 1 ) );
+			$pll_frontend->curlang = $pll_frontend->model->get_language( 'en' );
+			do_action( 'pll_language_defined' );
+
+			switch_to_blog( (int) $this->blog_with_pll_plain_links->blog_id );
+
+			restore_current_blog();
```
Hug0-Drelon, line 63: Do you really need to do this?

Marianne380: Yes to pass through `PLL_Frontend::switch_blog()`.

Hug0-Drelon: OK I see, then the next `switch_to_blog` should be sufficient? Since it will `PLL_Frontend::switch_blog()`.
Or maybe I missed something and you need to call it twice?

Marianne380: If you do a `switch_to_blog()` from the same site (prev and new), you pass [here](https://github.com/polylang/polylang/blob/606dbeb33e372277b512855c7f012728a4871312/frontend/frontend.php#L235-L236) into Polylang.

frontend/frontend.php
Marianne380: Yes, my mistake.

That said, it does cause an issue.
As the curlang is **English**, when you pass the `en_US` locale, the `load_strings_translations()` does the `get_language()` on this locale and it gets the **French** language. (because it's the first language with `en_US` in the cache). :eyes:

tests/phpunit/tests/test-switch-blog.php
Hug0-Drelon: And if you use `self::$blog_with_pll_plain_links` to perform the test?

Hug0-Drelon: Since you change the french locale for the whole network.

Marianne380: As discussed together, since the string has been added to the default site (`blog_with_pll_directory`), we need to return to this site to test it.

Hug0-Drelon (approved):

include/base.php
```diff
@@ -112,7 +112,7 @@ public function widgets_init() {
 	 * @since 1.2
 	 * @since 2.1.3 $locale parameter added.
 	 *
-	 * @param string $locale Locale. Defaults to current locale.
+	 * @param string $locale Locale. Defaults to current locale. It can also be a slug in some cases.
```
Chouby, line 115:
```suggestion
	 * @param string $locale Language locale or slug. Defaults to current locale.
```

tests/phpunit/tests/test-switch-blog.php
```diff
@@ -0,0 +1,70 @@
...
+
+		public function set_up() {
+			parent::set_up();
+
+			$pll_admin = $this->get_pll_admin_env();
+
+			$mo = new PLL_MO();
+			$mo->add_entry( $mo->make_entry( 'Test en', 'Test fr' ) );
+			$mo->export_to_db( $pll_admin->model->get_language( 'fr' ) );
+		}
```
Chouby, line 48: This can be moved to the test.

tests/phpunit/tests/test-switch-blog.php
```diff
@@ -0,0 +1,70 @@
...
+
+	class Switch_Blog_Test extends PLL_Multisites_TestCase {
+
+		/**
+		 * Languages data for their creation keyed by language slug.
+		 * The French language has the same locale as the English one ('en_US').
+		 *
+		 * @var array
+		 */
+		protected $languages = array(
```
Chouby, line 13: The critical change is not very visible if we don't read the above comment. I propose 2 solutions to improve this:
1. Don't copy anything and just modify the locale in `set_up()`.
2. Or keep overriding the property and use only 2 languages, let say English US and English India(in this case modify the test string accordingly). 
Moreover the comment could be moved in front of the local to increase the visibility.

Chouby (changes requested):

PR #1349, opened 2023-09-19
Remove unused variables in tests

Chouby:
Due to our current PHPCS config, this check is currently deactivated for tests.
This PR fixes the issues. The PHPCS config will be fixed in a separated PR.

PR #1380, opened 2023-10-26
Improve PHPCS config

Chouby:
This PR is a follow-up to #1328, #1333, #1347, #1349, #1362

- Fixes PHPCompatibilityWP config which was inoperative
- Fixes WordPressVIPMinimum config which used to exclude more sniffs in tests than expected
- Combines all `WordPressVIPMinimum.Functions.RestrictedFunctions` exclusions for tests
- Removes some useless config lines
- Adds `sanitize_locale_name()` to  auto escaped functions
- Attempts to clarify the structure of phpcs.xml.dist 
- Fixes some new reported errors and locally ignore others

### DIFF
--- a/tests/phpunit/includes/polyfills.php
+++ b/tests/phpunit/includes/polyfills.php
@@ -12,7 +12,7 @@ if ( ! function_exists( 'did_filter' ) ) {
 
 	add_filter(
 		'all',
-		function ( $hook_name ) { // phpcs:ignore WordPressVIPMinimum.Hooks.AlwaysReturnInFilter.MissingReturnStatement
+		function ( $hook_name ) {
 			global $wp_filters;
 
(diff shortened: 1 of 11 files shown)


tests/phpunit/includes/polyfills.php
```diff
@@ -12,7 +12,7 @@
 
 	add_filter(
 		'all',
-		function ( $hook_name ) {
+		function ( $hook_name ) { // phpcs:ignore WordPressVIPMinimum.Hooks.AlwaysReturnInFilter.MissingReturnStatement
```
Screenfeed, line 15: Why not just `return $hook_name;` instead of a PHPCS exclusion?

Screenfeed:
We could also get rid of the issue reported by PHPCS in `base.php`.

Screenfeed (approved):

Issue #133, opened 2017-05-17
The number of posts in the posts list table is not filtered per language


Chouby:
In the posts list table, the number of posts is displayed as:
`All (4) | Published (1) | Draft (3)`
When the admin language filter is set to display only one language, the numbers of posts are still displayed for all languages.

chesio:
"At a Glance" box on Dashboard page has the same problem: number of posts, pages and comments is always displayed for all languages.

szepeviktor:
@Chouby If you spend 1 day/month cleaning up issues/PR-s you will have an exceptionally clean repo!

Chouby:
We won't filter the number of posts per language to match the WP behavior which for example keeps displaying the total number of posts when the category filter or the date filter is active.

PR #1381, opened 2023-11-09
Ignore `never` return type issues reported by PHPStan

Screenfeed:
See https://github.com/php-stubs/wordpress-stubs/pull/110.

Note: this should be fixed by their PR #133 but, while it is merged into `master`, we still don't have it. So I guess we'll receive it soon or later and we'll be able to trash this PR.


Screenfeed:
WordPress stubs 6.4.1 have been released, including the patch.

PR #1395, opened 2023-12-07
Fix translations not loaded when the language is set from the content

Chouby:
To reproduce:
- Setup a site with Twenty Twenty One.
- Keep en_US as locale in Settings > General.
- Set the language from the content in Polylang settings.
- Visit the homepage in a secondary language.
- Check that Twenty Twenty One translations are not loaded.

The bug happens for all translation loaded by `_load_textdomain_just_in_time()` when the original WordPress locale is `en_US`, probably since WP 6.1. A quick fix for users is to set any locale except `en_US` in Settings > General.

This PR takes the opportunity to greatly simplify the process relying completely internal WP mechanisms (mainly `_load_textdomain_just_in_time()` working correctly since WP 6.1.

It does not include any backward compat for WP < 6.1.

### DIFF
--- a/include/olt-manager.php
+++ b/include/olt-manager.php
@@ -36,7 +36,7 @@ class PLL_OLT_Manager {
 		}
 
 		// Filters for text domain management.
-		add_filter( 'load_textdomain_mofile', array( $this, 'load_textdomain_mofile' ) );
+		add_filter( 'load_textdomain_mofile', array( $this, 'bypass_load_textdomain_mofile' ) );
 
 		// Loads text domains.
 		add_action( 'pll_language_defined', array( $this, 'load_textdomains' ), 2 ); // After PLL_Frontend::pll_language_defined.
@@ -88,10 +88,11 @@ class PLL_OLT_Manager {
 	 * Prevents WP loading textdomains before we set the locale ourselves.
 	 *
 	 * @since 2.0.4
+	 * @since 3.6 Renamed from `load_textdomain_mofile()`.
 	 *
 	 * @return string
 	 */
-	public function load_textdomain_mofile() {
+	public function bypass_load_textdomain_mofile() {
 		return '';
 	}
 
--- a/include/olt-manager.php
+++ b/include/olt-manager.php
@@ -11,15 +11,9 @@
  * or in a `wp` action (when the language is set from content on frontend).
  *
  * @since 1.2
+ * @since 3.6 Singleton removed, instantiate at your own risk!
  */
 class PLL_OLT_Manager {
-	/**
-	 * Singleton instance
-	 *
-	 * @var PLL_OLT_Manager|null
-	 */
-	protected static $instance;
-
 	/**
 	 * Constructor: setups relevant filters.
 	 *
@@ -41,21 +35,6 @@ class PLL_OLT_Manager {
 		// Loads text domains.
 		add_action( 'pll_language_defined', array( $this, 'load_textdomains' ), 2 ); // After PLL_Frontend::pll_language_defined.
 		add_action( 'pll_no_language_defined', array( $this, 'load_textdomains' ) );
-	}
-
-	/**
-	 * Access to the single instance of the class.
-	 *
-	 * @since 1.7
-	 *
-	 * @return PLL_OLT_Manager
-	 */
-	public static function instance() {
-		if ( empty( self::$instance ) ) {
-			self::$instance = new self();
-		}
-
-		return self::$instance;
 	}
 
 	/**
--- a/include/olt-manager.php
+++ b/include/olt-manager.php
@@ -11,7 +11,6 @@
  * or in a `wp` action (when the language is set from content on frontend).
  *
  * @since 1.2
- * @since 3.6 Singleton removed, instantiate at your own risk!
  */
 class PLL_OLT_Manager {
 	/**
@@ -37,7 +36,7 @@ class PLL_OLT_Manager {
 		}
 
 		// Filters for text domain management.
-		add_filter( 'load_textdomain_mofile', array( $this, 'bypass_load_textdomain_mofile' ) );
+		add_filter( 'load_textdomain_mofile', '__return_empty_string' );
 
 		// Loads text domains.
 		add_action( 'pll_language_defined', array( $this, 'load_textdomains' ), 2 ); // After PLL_Frontend::pll_language_defined.
@@ -67,8 +66,8 @@ class PLL_OLT_Manager {
 	 * @return void
 	 */
 	public function load_textdomains() {
-		// Our load_textdomain_mofile filter has done its job. let's remove it before calling load_textdomain.
-		remove_filter( 'load_textdomain_mofile', array( $this, 'bypass_load_textdomain_mofile' ) );
+		// Our load_textdomain_mofile filter has done its job. let's remove it to enable translation.
+		remove_filter( 'load_textdomain_mofile', '__return_empty_string' );
 
 		$GLOBALS['l10n'] = array();
 		$new_locale      = get_locale();
@@ -82,9 +81,10 @@ class PLL_OLT_Manager {
 			$GLOBALS['wp_locale'] = new WP_Locale();
 		}
 
+		/** This action is documented in wp-includes/class-wp-locale-switcher.php */
 		do_action( 'change_locale', $new_locale );
 
-		do_action_deprecated( 'pll_translate_labels', array(), '3.6', 'change_locale' );
+		do_action_deprecated( 'pll_translate_labels', array(), '3.7', 'change_locale' );
 	}
 
 	/**
--- a/include/olt-manager.php
+++ b/include/olt-manager.php
@@ -38,6 +38,21 @@ class PLL_OLT_Manager {
 	}
 
 	/**
+	 * Access to the single instance of the class
+	 *
+	 * @since 1.7
+	 *
+	 * @return PLL_OLT_Manager
+	 */
+	public static function instance() {
+		if ( empty( self::$instance ) ) {
+			self::$instance = new self();
+		}
+
+		return self::$instance;
+	}
+
+	/**
 	 * Loads textdomains.
 	 *
 	 * @since 0.1
--- a/include/olt-manager.php
+++ b/include/olt-manager.php
@@ -15,6 +15,13 @@
  */
 class PLL_OLT_Manager {
 	/**
+	 * Singleton instance
+	 *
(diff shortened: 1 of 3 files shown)


include/olt-manager.php
Hug0-Drelon, line 1: `$default_locale` property definition can be removed as well as `$list_texdomains` ans `$labels` , aren't they?

Hug0-Drelon (changes requested):
Do you think it worths adding some tests?
I'd be glad to help.

include/olt-manager.php
```diff
@@ -85,155 +80,40 @@ public static function instance() {
...
 
-		/**
-		 * Fires after the post types and taxonomies labels have been translated
-		 * This allows plugins to translate text the same way we do for post types and taxonomies labels
-		 *
-		 * @since 1.2
-		 *
-		 * @param array $labels list of strings to trnaslate
-		 */
-		do_action_ref_array( 'pll_translate_labels', array( &$this->labels ) );
```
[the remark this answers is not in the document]
Chouby: I believe that we have removed our unique usage of this action in https://github.com/polylang/polylang-wc/pull/733.
But why not? We won't pass anything to this action though.

Chouby:
> Do you think it worths adding some tests?
> I'd be glad to help.

Yes I am interested in your proposal. Maybe it will require to remove the singleton.

tests/phpunit/tests/test-olt-manager.php
```diff
@@ -21,4 +30,44 @@ public function test_polylang_first() {
...
+		$options                 = PLL_Install::get_default_options();
+		$options['force_lang']   = 0;
+		$options['default_lang'] = 'en';
+		$model                   = new PLL_Model( $options );
+		$links_model             = $model->get_links_model();
+		$frontend                = new PLL_Frontend( $links_model );
+
+		$is_loaded = load_plugin_textdomain( 'foo' );
+
+		$this->assertFalse( $is_loaded, 'textdomain should be bypassed by PLL_OLT_Manager.' );
```
Chouby, line 61: Using `load_plugin_texdomain()` disallows to treproduce the bug. We should always use `_load_textdomain_just_in_time()` even before the language is defined. As far as I remember this the removal of `load_theme_texdomain()` in Twenty Twenty one which revealed the bug:
 ```suggestion
		$this->assertSame( 'Dashboard', __( 'Dashboard', 'foo' ) );
```

File: include/olt-manager.php [hunk 3]
```diff
@@ -56,184 +29,52 @@ public function __construct() {
...
-	 *
-	 * @return PLL_OLT_Manager
-	 */
-	public static function instance() {
-		if ( empty( self::$instance ) ) {
-			self::$instance = new self();
-		}
-
-		return self::$instance;
-	}
```
Chouby, line 85: I propose to keep this method and keep using it in the production code (to keep access to the filters if some dev want to remove them). I did not remember but this is not a true singleton, so we can keep creating new instances in tests.

tests/phpunit/tests/test-olt-manager.php
Hug0-Drelon: Indeed, I managed to reproduce the issue with the old version of `PLL_OLT_Manager` and a call to `__()` before the language is defined!
See 0774f4e6
Even though it's weird that `load_theme_textdomain()` is still called in TT1 from the beginning ([cf](https://github.com/WordPress/twentytwentyone/blame/ba9f20cad89163761185c0467b346ba42541ae22/functions.php#L36)

include/olt-manager.php
Hug0-Drelon: Done in 3a60b577.

tests/phpunit/tests/test-olt-manager.php
Chouby: It used to be the case at the beginning. But [not anymore](https://github.com/WordPress/wordpress-develop/commit/1d1a2e3dec64bca6b94abd0eaa871f3916bbc11b).

include/olt-manager.php
```diff
@@ -4,13 +4,14 @@
...
- * this class differs text domain loading until the language is defined
- * either in a plugins_loaded action or in a wp action ( when the language is set from content on frontend )
+ * It is best practice that plugins do nothing before `plugins_loaded` is fired.
+ * So it is what Polylang intends to do.
+ * But some plugins load their textdomain as soon as loaded, thus before `plugins_loaded` is fired.
+ * This class defers textdomain loading until the language is defined either in a `plugins_loaded` action
+ * or in a `wp` action (when the language is set from content on frontend).
  *
  * @since 1.2
+ * @since 3.6 Singleton removed, instantiate at your own risk!
```
Chouby, line 14: In fact it has never been a singleton. I just checked and the constructor has always been public.
```suggestion
 ```

File: include/olt-manager.php [hunk 5]
```diff
@@ -85,155 +60,43 @@ public static function instance() {
...
-						load_textdomain( $textdomain['domain'], WP_LANG_DIR . "/plugins/{$textdomain['domain']}-$new_locale.mo" );
-					}
-				}
-			}
-		}
-
-		// First remove taxonomies and post_types labels that we don't need to translate
-		$taxonomies = get_taxonomies( array( '_pll' => true ) );
-		$post_types = get_post_types( array( '_pll' => true ) );
+		// Our load_textdomain_mofile filter has done its job. let's remove it before calling load_textdomain.
```
Chouby, line 70: We don't call `load_text_domain()` anymore.
```suggestion
		// Our load_textdomain_mofile filter has done its job. let's remove it to enable translation.
```

include/olt-manager.php
```diff
@@ -85,155 +60,43 @@ public static function instance() {
...
-		/**
-		 * Fires after the post types and taxonomies labels have been translated
-		 * This allows plugins to translate text the same way we do for post types and taxonomies labels
-		 *
-		 * @since 1.2
-		 *
-		 * @param array $labels list of strings to trnaslate
-		 */
-		do_action_ref_array( 'pll_translate_labels', array( &$this->labels ) );
+		do_action( 'change_locale', $new_locale );
```
Chouby, line 85:
```suggestion
		/** This action is documented in wp-includes/class-wp-locale-switcher.php */
		do_action( 'change_locale', $new_locale );
```

include/olt-manager.php
```diff
@@ -85,155 +60,43 @@ public static function instance() {
...
 
-		// Free memory.
-		$this->default_locale   = null;
-		$this->list_textdomains = array();
-		$this->labels           = array();
+		do_action_deprecated( 'pll_translate_labels', array(), '3.6', 'change_locale' );
```
Chouby, line 87:
```suggestion
		do_action_deprecated( 'pll_translate_labels', array(), '3.7', 'change_locale' );
```

include/olt-manager.php
```diff
@@ -56,16 +36,11 @@ public function __construct() {
...
 
-		// Saves the default locale before we start any language manipulation
-		$this->default_locale = get_locale();
-
-		// Filters for text domain management
-		add_filter( 'load_textdomain_mofile', array( $this, 'load_textdomain_mofile' ), 10, 2 );
-		add_filter( 'gettext', array( $this, 'gettext' ), 10, 3 );
-		add_filter( 'gettext_with_context', array( $this, 'gettext_with_context' ), 10, 4 );
+		// Filters for text domain management.
+		add_filter( 'load_textdomain_mofile', array( $this, 'bypass_load_textdomain_mofile' ) );
```
Chouby, line 40:
```suggestion
		add_filter( 'load_textdomain_mofile', '__return_empty_string' );
```

include/olt-manager.php
```diff
@@ -85,155 +60,43 @@ public static function instance() {
...
-					}
-				}
-			}
-		}
-
-		// First remove taxonomies and post_types labels that we don't need to translate
-		$taxonomies = get_taxonomies( array( '_pll' => true ) );
-		$post_types = get_post_types( array( '_pll' => true ) );
+		// Our load_textdomain_mofile filter has done its job. let's remove it before calling load_textdomain.
+		remove_filter( 'load_textdomain_mofile', array( $this, 'bypass_load_textdomain_mofile' ) );
```
Chouby, line 71:
```suggestion
		remove_filter( 'load_textdomain_mofile', '__return_empty_string' );
```

include/olt-manager.php
```diff
@@ -85,155 +60,43 @@ public static function instance() {
...
-						__( $label, $this->labels[ $label ]['domain'] );
-					// PHPCS:enable
-				}
-				else {
-					$type->labels->$key = $translated[ $label ];
-				}
-			}
-		}
+	public function bypass_load_textdomain_mofile() {
+		return '';
```
Chouby, line 99: I guess we can remove this method and use `__return_empty_string()` instead.

tests/phpunit/tests/test-olt-manager.php
```diff
@@ -21,4 +30,45 @@ public function test_polylang_first() {
...
+		$options['force_lang']   = 0;
+		$options['default_lang'] = 'en';
+		$model                   = new PLL_Model( $options );
+		$links_model             = $model->get_links_model();
+		$frontend                = new PLL_Frontend( $links_model );
+
+		/*
+		 *  Calls `_load_textdomain_just_in_time()` *before* the current language is defined!
+		 */
+		__( 'Dashboard', 'foo' ); // phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
```
Chouby, line 62: I suggest to perform an assertion here to check that no translation is loaded.
```suggestion
		$this->assertSame( 'Dashboard', __( 'Dashboard', 'foo' ) ); // phpcs:ignore WordPress.WP.I18n.TextDomainMismatch
```

In fact to be even more complete we could test with a WPLANG option set to a 3rd language (not en_US or fr_FR), let say de_DE, to make sure that foo-de_DE.mo is not loaded. What do you think?

tests/phpunit/tests/test-olt-manager.php
Hug0-Drelon: What?! They commit directly in `wordpress-develop` instead of the dedicated repo?
I learned something, thanks!

tests/phpunit/tests/test-olt-manager.php
Hug0-Drelon: Good idea, done in 6cb66abb.

include/olt-manager.php
Hug0-Drelon: Done in beedce96.

PR #1397, opened 2023-12-12
Avoid a notice when using system CRON

### LINKED ISSUE #1379: Notice about missing variables $_SERVER[HTTP_HOST] and $_SERVER[REQUEST_URI].
## Prerequisites

## Introduction

After disabling the default cron task and enabling the call in task schedule, the plugin returns a message in the logs.
The message is generated by the `pll_get_requested_url` function in the function.php file.

```
PHP Notice: $_SERVER['HTTP_HOST'] or $_SERVER['REQUEST_URI'] are required but not set. in /wp-content/plugins/polylang/include/functions.php on line 98
```

1. Set the `WP_DEBUG` variable to true in wp-config.php.
2. Set the `DISABLE_WP_CRON` variable to true in wp-config.php.
3. Add running wp-cron.php file in task schedule.

## Possible solution

I suggest adding an additional if to check if a cron task is not running, I checked and it works.

```php
if ( defined( 'DOING_CRON' ) && DOING_CRON ) {
	return get_option( 'home' );
}
```

Full function code:

```php
function pll_get_requested_url() {
	if ( isset( $_SERVER['HTTP_HOST'], $_SERVER['REQUEST_URI'] ) ) {
		return set_url_scheme( esc_url_raw( wp_unslash( 'http://' . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'] ) ) );
	}

	/*
	 * In WP CLI context, few developers define superglobals in wp-config.php
	 * as proposed in https://make.wordpress.org/cli/handbook/common-issues/#php-notice-undefined-index-on-_server-superglobal
	 * So let's return the unfiltered home url to avoid a bunch of notices.
	 */
	if ( defined( 'WP_CLI' ) && WP_CLI ) {
		return get_option( 'home' );
	}

	if ( defined( 'DOING_CRON' ) && DOING_CRON ) {
		return get_option( 'home' );
	}

	if ( WP_DEBUG ) {
		// phpcs:ignore WordPress.PHP.DevelopmentFunctions
		trigger_error( '$_SERVER[\'HTTP_HOST\'] or $_SERVER[\'REQUEST_URI\'] are required but not set.' );
	}

	return '';
}
```

Chouby:
Fix #1379 
Props arielruminski

### DIFF
--- a/include/functions.php
+++ b/include/functions.php
@@ -93,6 +93,13 @@ function pll_get_requested_url() {
 		return get_option( 'home' );
 	}
 
+	/*
+	 * When using system CRON instead of WP_CRON, the superglobals are likely undefined.
+	 */
+	if ( defined( 'DOING_CRON' ) && DOING_CRON ) {
+		return get_option( 'home' );
+	}
+
 	if ( WP_DEBUG ) {
 		// phpcs:ignore WordPress.PHP.DevelopmentFunctions
 		trigger_error( '$_SERVER[\'HTTP_HOST\'] or $_SERVER[\'REQUEST_URI\'] are required but not set.' );


Screenfeed (commented):
Shouldn't we make sure that `get_option( 'home' )` is a non-empty string instead of doing `@var string`?
I suppose it's part of the things that we must assume to be right.

Screenfeed (approved):

PR #1399, opened 2023-12-18
Fix `wp_tag_cloud()` return type for PHPStan

Screenfeed:
This PR fixes the following issue reported by PHPStan:
```
 ------ ---------------------------------------------------------------------- 
  Line   admin/admin-filters-term.php                                          
 ------ ---------------------------------------------------------------------- 
         Ignored error pattern #^Parameter \#3 \.\.\.\$values of function      
         sprintf expects bool\|float\|int\|string\|null,                       
         array<string>\|string\|void given\.$# in path                         
         /home/runner/work/polylang/polylang/admin/admin-filters-term.php was  
         not matched in reported errors.                                       
  482    Parameter #3 ...$values of function sprintf expects                   
         bool|float|int|string|null, array<string>|string given.               
 ------ ---------------------------------------------------------------------- 
```

### DIFF
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -478,9 +478,10 @@ class PLL_Admin_Filters_Term {
 					$args = array_merge( $args, array( 'link' => 'edit' ) );
 				}
 
+				/** @var string */
 				$tag_cloud = wp_tag_cloud( $args );
 
-				if ( ! empty( $tag_cloud ) && is_string( $tag_cloud ) ) {
+				if ( ! empty( $tag_cloud ) ) {
 					$html = sprintf( '<div class="tagcloud"><h2>%1$s</h2>%2$s</div>', esc_html( $tax->labels->popular_items ), $tag_cloud );
 					$x->Add( array( 'what' => 'tag_cloud', 'data' => $html ) );
 				}
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -478,7 +478,7 @@ class PLL_Admin_Filters_Term {
 					$args = array_merge( $args, array( 'link' => 'edit' ) );
 				}
 
-				/** @var string */
+				/** @var string|void */
 				$tag_cloud = wp_tag_cloud( $args );
 
 				if ( ! empty( $tag_cloud ) ) {
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -478,10 +478,10 @@ class PLL_Admin_Filters_Term {
 					$args = array_merge( $args, array( 'link' => 'edit' ) );
 				}
 
-				/** @var string|void */
 				$tag_cloud = wp_tag_cloud( $args );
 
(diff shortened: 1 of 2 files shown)


admin/admin-filters-term.php
```diff
@@ -478,7 +478,9 @@ public function term_lang_choice() {
 					$args = array_merge( $args, array( 'link' => 'edit' ) );
 				}
 
-				if ( $tag_cloud = wp_tag_cloud( $args ) ) {
+				$tag_cloud = wp_tag_cloud( $args );
+
+				if ( ! empty( $tag_cloud ) && is_string( $tag_cloud ) ) {
```
Hug0-Drelon, line 483: By default `format` argument is set to `flat` so if return isn't void, it must be a string.
PHPStan didn't catch it?

Screenfeed: I don't know what changed but the return type is simply `@return void|string|string[]` in the stubs, while it should be a conditional return type based on `format` and `echo`.

Hug0-Drelon: Well, maybe some day phpstan-wordpress will support the conditional return type for `wp_tag_cloud()`.

Hug0-Drelon (approved):

admin/admin-filters-term.php
Chouby: I guess that the dynamic return type of `wp_tag_cloud()` is not handled. If you are sure that we can get only a string then I suggest to document it as 
```php
/** @var string **/
tag_cloud = wp_tag_cloud( $args );
```
You can also write a PR for PHPStan WordPress.

Screenfeed: You're right, I'll do that.
The dynamic return type seems to be [a work in progress](https://github.com/szepeviktor/phpstan-wordpress/pull/147).

Screenfeed: Done in [4169e15](https://github.com/polylang/polylang/pull/1399/commits/4169e15958c28d37e2c95667b8f76fdc55d9f76d).

admin/admin-filters-term.php
```diff
@@ -478,7 +478,10 @@ public function term_lang_choice() {
 					$args = array_merge( $args, array( 'link' => 'edit' ) );
 				}
 
-				if ( $tag_cloud = wp_tag_cloud( $args ) ) {
+				/** @var string */
```
Hug0-Drelon, line 481:
```suggestion
				/** @var string|void */
```
Seems more accurate to me.

Screenfeed: It can't be `void` because of `'echo' => false`.

Hug0-Drelon: It also returns `void` in case of failure according to the [documentation](https://developer.wordpress.org/reference/functions/wp_tag_cloud/#return).

Screenfeed: AH !

Screenfeed:
WTF PHPStan, how?!

PR #1400, opened 2023-12-21
Move create_media_translation() to PLL_Translated_Posts

Chouby:
Closes https://github.com/polylang/polylang-pro/issues/1360
See also https://github.com/polylang/polylang-pro/issues/1108

`PLL_CRUD_Posts` should only include methods hooked to actions fired when posts are created, updated or deleted and should not include public method used outside these contexts.

This PR proposes to move the method `create_media_translation()` to `PLL_Translated_Posts`

### DIFF
--- a/include/translated-post.php
+++ b/include/translated-post.php
@@ -304,6 +304,78 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 	}
 
 	/**
+	 * Creates a media translation
+	 *
+	 * @since 1.8
+	 * @since 3.6 Moved from PLL_CRUD_Posts.
+	 *
+	 * @param int           $post_id Original attachment id.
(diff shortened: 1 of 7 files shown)


include/translated-post.php
Hug0-Drelon, line 1: I'm not agains moving `create_media_translation()` into `PLL_Translated_Posts`.
Except that it break symmetry between `PLL_Translated_posts` and `PLL_Translated_Terms`.
Another solution could be a new class dedicated to medias (and loaded only if the corresponding option allows it).
It could gather all other thins related to media.

Hug0-Drelon: Using `create_media_translation()` is a PITA when it's from `PLL_CRUD_Posts`, so no need to do the same by moving it to a new class.

Hug0-Drelon (approved):

PR #1402, opened 2023-12-21
Support JSON encoding for metas in `wpml-config.xml`

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/1936.

This PR is about 2 things:
- Grad the encoding for metas from the `wpml-config.xml` files, and filter `pll_post_meta_encodings` + `pll_term_meta_encodings` accordingly.
- Reduce duplicated code in metas-related methods in `PLL_WPML_Config`.

The method `PLL_WPML_Config::parse_xml_metas()` will fetch the data from the files and it to cache, to prevent re-doing `SimpleXMLElement::xpath()` + fetching attributes for the same fields. Benefits from this cache are probably low though.

### DIFF
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -10,11 +10,13 @@
  *
  * @since 1.0
  *
- * @phpstan-type ParsedMeta array{
- *     name: non-falsy-string,
- *     action: string,
- *     encoding: 'json'|''
- * }
+ * @phpstan-type ParsedMetas array<
+ *     non-falsy-string,
+ *     array{
+ *         action: string,
+ *         encoding: 'json'|''
+ *     }
+ * >
  */
 class PLL_WPML_Config {
 	/**
@@ -59,7 +61,7 @@ class PLL_WPML_Config {
 	 *
 	 * @var array
 	 *
-	 * @phpstan-var array<non-falsy-string, list<ParsedMeta>>
+	 * @phpstan-var array<non-falsy-string, ParsedMetas>
 	 */
 	private $parsed_metas = array();
 
@@ -844,11 +846,11 @@ class PLL_WPML_Config {
 		$parsed_metas    = $this->parse_xml_metas( $xpath );
 		$metas_to_remove = array();
 
-		foreach ( $parsed_metas as $parsed_meta ) {
+		foreach ( $parsed_metas as $name => $parsed_meta ) {
 			if ( 'copy' === $parsed_meta['action'] || ( ! $sync && in_array( $parsed_meta['action'], array( 'translate', 'copy-once' ), true ) ) ) {
-				$metas[] = $parsed_meta['name'];
+				$metas[] = $name;
 			} else {
-				$metas_to_remove[] = $parsed_meta['name'];
+				$metas_to_remove[] = $name;
 			}
 		}
 
@@ -878,9 +880,9 @@ class PLL_WPML_Config {
 	private function add_metas_to_export( array $metas, string $xpath ) {
 		$fields = $this->parse_xml_metas( $xpath );
 
-		foreach ( $fields as $field ) {
+		foreach ( $fields as $name => $field ) {
 			if ( 'translate' === $field['action'] ) {
-				$metas[ $field['name'] ] = 1;
+				$metas[ $name ] = 1;
 			}
 		}
 
@@ -902,9 +904,9 @@ class PLL_WPML_Config {
 	private function add_metas_encodings( array $metas, string $xpath ): array {
 		$parsed_metas = $this->parse_xml_metas( $xpath );
 
-		foreach ( $parsed_metas as $parsed_meta ) {
+		foreach ( $parsed_metas as $name => $parsed_meta ) {
 			if ( ! empty( $parsed_meta['encoding'] ) ) {
-				$metas[ $parsed_meta['name'] ] = $parsed_meta['encoding'];
+				$metas[ $name ] = $parsed_meta['encoding'];
 			}
 		}
 
@@ -921,7 +923,7 @@ class PLL_WPML_Config {
 	 * @return array
 	 *
 	 * @phpstan-param non-falsy-string $xpath
-	 * @phpstan-return list<ParsedMeta>
+	 * @phpstan-return ParsedMetas
 	 */
 	private function parse_xml_metas( string $xpath ): array {
 		if ( isset( $this->parsed_metas[ $xpath ] ) ) {
@@ -945,14 +947,13 @@ class PLL_WPML_Config {
 				}
 
 				$data = array(
-					'name'     => $name,
 					'action'   => $this->get_field_attribute( $custom_field, 'action' ),
 					'encoding' => $this->get_field_attribute( $custom_field, 'encoding' ),
 				);
 
 				$data['encoding'] = 'json' === $data['encoding'] ? 'json' : ''; // Only JSON is supported for now.
 
-				$this->parsed_metas[ $xpath ][] = $data;
+				$this->parsed_metas[ $xpath ][ $name ] = $data;
 			}
 		}
 
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -931,14 +931,14 @@ class PLL_WPML_Config {
 		$this->parsed_metas[ $xpath ] = array();
 
 		foreach ( $this->xmls as $xml ) {
-			$cfs = $xml->xpath( $xpath );
+			$custom_fields = $xml->xpath( $xpath );
 
-			if ( ! is_array( $cfs ) ) {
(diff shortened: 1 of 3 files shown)


modules/wpml/wpml-config.php
```diff
@@ -849,4 +826,136 @@ private function normalize_path( string $path ): string {
...
+	 *
+	 * @phpstan-param array<non-falsy-string> $metas
+	 * @phpstan-param non-falsy-string $xpath
+	 */
+	private function filter_metas_to_copy( array $metas, string $xpath, bool $sync ): array {
+		$parsed_metas    = $this->parse_xml_metas( $xpath );
+		$metas_to_remove = array();
+
+		foreach ( $parsed_metas as $parsed_meta ) {
+			if ( 'copy' === $parsed_meta['action'] || ( ! $sync && in_array( $parsed_meta['action'], array( 'translate', 'copy-once' ), true ) ) ) {
```
Hug0-Drelon, line 848:
```suggestion
                         $do_sync =  'copy' === $parsed_meta['action'];
                         $do_copy = ! $sync && in_array( $parsed_meta['action'], array( 'translate', 'copy-once' ), true ) );
			if ( $do_sync || $do_copy ) {
```
Do you like it?

modules/wpml/wpml-config.php
```diff
@@ -849,4 +826,136 @@ private function normalize_path( string $path ): string {
...
+	 */
+	private function parse_xml_metas( string $xpath ): array {
+		if ( isset( $this->parsed_metas[ $xpath ] ) ) {
+			return $this->parsed_metas[ $xpath ];
+		}
+
+		$this->parsed_metas[ $xpath ] = array();
+
+		foreach ( $this->xmls as $xml ) {
+			$cfs = $xml->xpath( $xpath );
```
Hug0-Drelon, line 934:
```suggestion
			$custom_fields = $xml->xpath( $xpath );
```
More readable IMHO, but I let you see.

Hug0-Drelon (changes requested):

modules/wpml/wpml-config.php
```diff
@@ -849,4 +826,136 @@ private function normalize_path( string $path ): string {
...
+	 */
+	private function filter_metas_to_copy( array $metas, string $xpath, bool $sync ): array {
+		$parsed_metas    = $this->parse_xml_metas( $xpath );
+		$metas_to_remove = array();
+
+		foreach ( $parsed_metas as $parsed_meta ) {
+			if ( 'copy' === $parsed_meta['action'] || ( ! $sync && in_array( $parsed_meta['action'], array( 'translate', 'copy-once' ), true ) ) ) {
+				$metas[] = $parsed_meta['name'];
+			} else {
+				$metas_to_remove[] = $parsed_meta['name'];
```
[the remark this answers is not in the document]
Screenfeed: Yes, only one `array_diff()` at the end instead of one in each loop.

modules/wpml/wpml-config.php
```diff
@@ -48,6 +54,15 @@ class PLL_WPML_Config {
...
 	private $open_basedir_paths;
 
+	/**
+	 * Cache for parsed metas.
+	 *
+	 * @var array
+	 *
+	 * @phpstan-var array<non-falsy-string, list<ParsedMeta>>
+	 */
+	private $parsed_metas = array();
```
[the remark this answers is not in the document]
Screenfeed: We'd probably lose PHPStan's typing.

modules/wpml/wpml-config.php
Screenfeed: More readable indeed but we slightly lose in perfs.
- Current case: if action is `'copy'`, we go directly into the `if`.
- Suggested case: if action is `'copy'`, we still need to evaluate `$do_copy`.

So I don't know ^^'

modules/wpml/wpml-config.php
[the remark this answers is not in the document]
Screenfeed: Agree. Done in [1d7cfa2](https://github.com/polylang/polylang/pull/1402/commits/1d7cfa20046bd9d041145a564ce437274277601a).

modules/wpml/wpml-config.php
Screenfeed: Yep. Done in [35fbd57](https://github.com/polylang/polylang/pull/1402/commits/35fbd57e3f9419f7dc04512919be4b02ab4694d5).

modules/wpml/wpml-config.php
Hug0-Drelon: I'll let @Chouby make a decision here 😅

Hug0-Drelon (approved):

modules/wpml/wpml-config.php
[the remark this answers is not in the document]
Screenfeed: OK I see one benefit: `PLL_Cache` handles multisite.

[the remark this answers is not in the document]
Screenfeed: So... there is a way to use `PLL_Cache` without losing PHPStan's typing.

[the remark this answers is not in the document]
Screenfeed: Ah, `PLL_Cache` would be useless here I think, because the paths to the files are also stored in a class property (`$files`).

modules/wpml/wpml-config.php
```diff
@@ -173,34 +190,18 @@ public function get_files() {
...
-				if ( 'copy' === $action || ( ! $sync && in_array( $action, array( 'translate', 'copy-once' ), true ) ) ) {
-					$metas[] = (string) $cf;
-				} else {
-					$metas = array_diff( $metas, array( (string) $cf ) );
-				}
-			}
-		}
-
-		return $metas;
+		return $this->filter_metas_to_copy( (array) $metas, 'custom-fields/custom-field', (bool) $sync );
```
Chouby, line 204: Smart move. 
Do you think it's worth to merge the 2 methods `copy_post_metas()` and copy_term_metas()` in `filter_metas_to_copy()` removing the 2nd param and replacing it by something close to:
`$xpath = current_filter() ===  'pll_copy_post_metas' ? 'custom-fields/custom-field' : 'custom-term-fields/custom-term-field'

tests/phpunit/tests/test-wpml-config.php
```diff
@@ -107,52 +107,93 @@ protected function translate_options( $slug ) {
...
+		add_post_meta( $from, 'date-added', 2007 ); // `ignore`
+		add_post_meta( $from, 'a_json_meta', $json_en ); // `translate` + encoding.
 
 		$fr = $to = self::factory()->post->create();
 		self::$model->post->set_language( $to, 'fr' );
 		self::$model->post->save_translations( $en, compact( 'en', 'fr' ) );
 
-		// copy
+		// Enable translation for the JSON meta.
+		add_filter(
```
Chouby, line 131: I'm not sure to understand the purpose of this filter in the test.

modules/wpml/wpml-config.php
Screenfeed: It can be done but I'm not a big fan of it, I prefer to inject the path as a parameter instead of sniffing the current filter.

modules/wpml/wpml-config.php
```diff
@@ -849,4 +826,136 @@ private function normalize_path( string $path ): string {
...
+
+				if ( empty( $name ) ) {
+					continue;
+				}
+
+				$data = array(
+					'name'     => $name,
+					'action'   => $this->get_field_attribute( $custom_field, 'action' ),
+					'encoding' => $this->get_field_attribute( $custom_field, 'encoding' ),
+				);
```
Chouby, line 951: Wouldn't it be easier to manipulate?
```suggestion
				$data[ $name ] = array(
					'action'   => $this->get_field_attribute( $custom_field, 'action' ),
					'encoding' => $this->get_field_attribute( $custom_field, 'encoding' ),
				);
```

tests/phpunit/tests/test-wpml-config.php
Screenfeed: It's only to make sure that the whole process is working. But I think this test has little value for this PR afterall 🤔

modules/wpml/wpml-config.php
Screenfeed: Why not.

Screenfeed: Done in [43c8b85](https://github.com/polylang/polylang/pull/1402/commits/43c8b851293473044f894f9d36e115e3cf7669fd).

tests/phpunit/tests/test-wpml-config.php
Screenfeed: Fixed in [f1796c7](https://github.com/polylang/polylang/pull/1402/commits/f1796c7eb2a973e28afa2a73b213e6119e01b396).

PR #1403, opened 2023-12-22
Use a template and generics to document `PLL_Cache`

Screenfeed:
With this PR, `PLL_Cache` is not stuck with `mixed` for the cached data anymore.

This allows to be more precise about the data stored by `->set()` and returned by `->get()`.

##Impact in Polylang

Minimal. For now on, we need to specify the type of data stored by `PLL_Cache`.
Example:
```php
/**
 * @var PLL_Cache<string>
 */
private $cache;
```
Cache keys are already expected to be `non-empty-string`, and with this we specify to PHPStan that the data stored is expected to be strings. Of course we can still use `PLL_Cache<mixed>` when we expect to cache any kind of data.

##Also in this PR
- `set()` now returns the cached value. This only allows a shorthand like `return $this->cache->set( 'foobar', $data );` instead of setting the cache then returning the value.
- In `clean()`, the value `'0'` is allowed as a valid key.
- Several entries removed from the baseline.
- `$key` is type-hinted as a string. Debatable, maybe we'll want to use numbers someday.
- New method `PLL_Cache::get_unique_key()`.

### DIFF
--- a/frontend/frontend-filters-widgets.php
+++ b/frontend/frontend-filters-widgets.php
@@ -63,9 +63,7 @@ class PLL_Frontend_Filters_Widgets {
 
 		$sidebars_widgets = $this->filter_widgets_sidebars( $sidebars_widgets, $wp_registered_widgets );
 
-		$this->cache->set( "sidebars_widgets_{$cache_key}", $sidebars_widgets );
-
-		return $sidebars_widgets;
+		return $this->cache->set( "sidebars_widgets_{$cache_key}", $sidebars_widgets );
 	}
 
 	/**
--- a/frontend/frontend-filters-widgets.php
+++ b/frontend/frontend-filters-widgets.php
@@ -12,7 +12,7 @@ class PLL_Frontend_Filters_Widgets {
 	/**
 	 * Internal non persistent cache object.
 	 *
-	 * @var PLL_Cache
+	 * @var PLL_Cache<array>
 	 */
 	public $cache;
 
--- a/frontend/frontend-filters-widgets.php
+++ b/frontend/frontend-filters-widgets.php
@@ -54,8 +54,8 @@ class PLL_Frontend_Filters_Widgets {
 			return $sidebars_widgets;
 		}
 
-		$cache_key         = md5( maybe_serialize( $sidebars_widgets ) );
-		$_sidebars_widgets = $this->cache->get( "sidebars_widgets_{$cache_key}" );
+		$cache_key         = $this->cache->get_unique_key( 'sidebars_widgets_', $sidebars_widgets );
+		$_sidebars_widgets = $this->cache->get( $cache_key );
 
 		if ( false !== $_sidebars_widgets ) {
 			return $_sidebars_widgets;
@@ -63,7 +63,7 @@ class PLL_Frontend_Filters_Widgets {
 
 		$sidebars_widgets = $this->filter_widgets_sidebars( $sidebars_widgets, $wp_registered_widgets );
 
-		return $this->cache->set( "sidebars_widgets_{$cache_key}", $sidebars_widgets );
+		return $this->cache->set( $cache_key, $sidebars_widgets );
 	}
 
 	/**
--- a/include/cache.php
+++ b/include/cache.php
@@ -63,7 +63,7 @@ class PLL_Cache {
 	 * @phpstan-param TCacheData $data
 	 * @phpstan-return TCacheData
 	 */
-	public function set( string $key, $data ) {
+	public function set( $key, $data ) {
 		$this->cache[ $this->blog_id ][ $key ] = $data;
 		return $data;
 	}
@@ -79,7 +79,7 @@ class PLL_Cache {
 	 * @phpstan-param non-empty-string $key
 	 * @phpstan-return TCacheData|false
 	 */
-	public function get( string $key ) {
+	public function get( $key ) {
 		return isset( $this->cache[ $this->blog_id ][ $key ] ) ? $this->cache[ $this->blog_id ][ $key ] : false;
 	}
 
@@ -92,7 +92,7 @@ class PLL_Cache {
 	 *                    Default is an empty string.
 	 * @return void
 	 */
-	public function clean( string $key = '' ) {
+	public function clean( $key = '' ) {
 		if ( '' === $key ) {
 			unset( $this->cache[ $this->blog_id ] );
 		} else {
--- a/include/cache.php
+++ b/include/cache.php
@@ -109,7 +109,8 @@ class PLL_Cache {
 	 * @param string|array|object $data   Data.
 	 * @return string
 	 *
-	 * @phpstan-return ( $prefix is non-empty-string ? non-empty-string : string )
+	 * @phpstan-param non-empty-string $prefix
+	 * @phpstan-return non-empty-string
 	 */
 	public function get_unique_key( string $prefix, $data ): string {
 		$data = maybe_serialize( $data );
--- a/include/cache.php
+++ b/include/cache.php
@@ -99,4 +99,21 @@ class PLL_Cache {
 			unset( $this->cache[ $this->blog_id ][ $key ] );
 		}
 	}
+
+	/**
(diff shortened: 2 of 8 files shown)


frontend/frontend-filters-widgets.php
```diff
@@ -63,9 +63,7 @@ public function sidebars_widgets( $sidebars_widgets ) {
 
 		$sidebars_widgets = $this->filter_widgets_sidebars( $sidebars_widgets, $wp_registered_widgets );
 
-		$this->cache->set( "sidebars_widgets_{$cache_key}", $sidebars_widgets );
-
-		return $sidebars_widgets;
+		return $this->cache->set( "sidebars_widgets_{$cache_key}", $sidebars_widgets );
```
Hug0-Drelon, line 66: Do you think it's a good enhancement to have a static method like `PLL_Cache::get_unique_key()`?

include/model.php
```diff
@@ -12,7 +12,7 @@ class PLL_Model {
 	/**
 	 * Internal non persistent cache object.
 	 *
-	 * @var PLL_Cache
+	 * @var PLL_Cache<mixed>
```
Hug0-Drelon, line 15:
```suggestion
	 * @var PLL_Cache<PLL_Language|array>
```
As of today, `PLL_Model` caches only this two kind of data IMHO.

include/translate-option.php
```diff
@@ -40,7 +40,7 @@ class PLL_Translate_Option {
 	/**
 	 * Cache for the translated values.
 	 *
-	 * @var PLL_Cache
+	 * @var PLL_Cache<mixed>
```
Hug0-Drelon, line 43:
```suggestion
	 * @var PLL_Cache<array|string>
```
According to the result of `self::translate_string_recursive()`.

Hug0-Drelon (changes requested):

include/model.php
Screenfeed: It can also be `array<string>` because of [this](https://github.com/polylang/polylang/blob/46e1a5be8c95ddf7043b964de29ffa9b13fc4a1e/include/translated-post.php#L150) and [this](https://github.com/polylang/polylang/blob/46e1a5be8c95ddf7043b964de29ffa9b13fc4a1e/include/translated-term.php#L218).
The full type would be `PLL_Cache<PLL_Language|list<PLL_Language>|array<string>>` then. But then it triggers new issues like `Cannot call method get_tax_prop() on PLL_Language|string.` and `Cannot access property $is_default on PLL_Language|string.` and I don't see a way to fix this because the type returned by `PLL_Cache::get()` depends on the 1st argument passed to it (BTW, it might come back to bite our ass someday if we store all kinds of things in a same instance of `PLL_Cache` like in this case 🤔).

frontend/frontend-filters-widgets.php
Screenfeed: Yes, I think so. I see the same thing is also used in `PLL_Cache::count_posts()`.
It's out of scope but let's see if this passes the final review 😁

include/model.php
Hug0-Drelon: Yeah you're right, it becomes too hard to maintain that way.

frontend/frontend-filters-widgets.php
Screenfeed: Done in [cfd7677](https://github.com/polylang/polylang/pull/1403/commits/cfd7677bbcc1147c410dc3d4330e79e23cead2d3).

include/cache.php
```diff
@@ -46,43 +50,70 @@ public function switch_blog( $new_blog_id ) {
...
 			unset( $this->cache[ $this->blog_id ][ $key ] );
 		}
 	}
+
+	/**
+	 * Generates and returns a "unique" cache key, depending on `$data` and prefixed by `$prefix`.
+	 *
+	 * @since 3.6
+	 *
+	 * @param string              $prefix String to prefix the cache key.
```
Hug0-Drelon, line 108: I would require `$prefix` to be non-empty, otherwise we could have an empty cache key, couldn't we?

Screenfeed: No but you highlight something incorrect with what I've done.
I did this:
```
@phpstan-return ( $prefix is non-empty-string ? non-empty-string : string )
```
But this is incorrect, because even if `$prefix` is an empty string, the returned cache key will not be an empty string, thanks to `md5()`.

I would keep the possibility to have `$prefix` as an empty string because we may want to generate several keys with the same md5. Ex:
```php
$key            = $this->cache->get_unique_key( '', $data );
$foobar_key     = 'foobar_' . $key;
$foobar_baz_key = 'foobar_baz_' . $key;
```

But maybe we may prefer to enforce `$prefix` as a non empty string to prevent bugs where `$prefix` is empty while we expect it not to be?
What do you think?

Hug0-Drelon: Forcing `$prefix` to non-empty is safer IMHO. Also, you can still do:
```php
$foobar_key = $this->cache->get_unique_key( 'foobar_', $data );
$foobar_baz_key = $this->cache->get_unique_key( 'foobar_baz_', $data );
```
And you get the same result, right? Or am I missing something 🤔

Screenfeed: Done in [150d298](https://github.com/polylang/polylang/pull/1403/commits/150d298b78a398c7e390d7be93e513aa44028f21).

include/cache.php
```diff
@@ -46,43 +50,71 @@ public function switch_blog( $new_blog_id ) {
...
+	 * @param string|array|object $data   Data.
+	 * @return string
+	 *
+	 * @phpstan-param non-empty-string $prefix
+	 * @phpstan-return non-empty-string
+	 */
+	public function get_unique_key( string $prefix, $data ): string {
```
Chouby, line 115: I suggest to make the prefix optional.

include/cache.php
```diff
@@ -46,43 +50,71 @@ public function switch_blog( $new_blog_id ) {
...
 	 * @since 1.7
 	 *
-	 * @param string $key Cache key.
+	 * @param string $key Optional. Cache key. An empty string to clean the whole cache for the current blog.
+	 *                    Default is an empty string.
 	 * @return void
 	 */
-	public function clean( $key = '' ) {
-		if ( empty( $key ) ) {
+	public function clean( string $key = '' ) {
```
Chouby, line 95: I propose not to typehint the existing methods `set()`, `get()`, `clean()`. Although documented as string, nothing prevented to use any possible array key.

Chouby (changes requested):

include/cache.php
Screenfeed: @Chouby See [the previous conversation](https://github.com/polylang/polylang/pull/1403#discussion_r1441632921).

include/cache.php
Screenfeed: Done in [09ae96e](https://github.com/polylang/polylang/pull/1403/commits/09ae96eff2f7141c85d9852935db7bc462ddc9e7).

include/cache.php
```diff
@@ -46,43 +50,71 @@ public function switch_blog( $new_blog_id ) {
...
+	 * @param string              $prefix String to prefix the cache key.
+	 * @param string|array|object $data   Data.
+	 * @return string
+	 *
+	 * @phpstan-param non-empty-string $prefix
+	 * @phpstan-return non-empty-string
+	 */
+	public function get_unique_key( string $prefix, $data ): string {
+		$data = maybe_serialize( $data );
+		$data = is_scalar( $data ) ? (string) $data : '';
```
Chouby, line 117: The check is useless as `maybe_serialize()` always returns a scalar.

include/cache.php
```diff
@@ -46,43 +50,71 @@ public function switch_blog( $new_blog_id ) {
...
+		return $prefix . md5( $data );
```
Chouby, line 118: The check is useless as maybe_serialize() always returns a scalar.
```suggestion
		/** @var scalar **/
		$serialized = maybe_serialize( $data );
		return $prefix . md5( (string) $serialized );
```

PR #1404, opened 2023-12-22
Wilcards in `PLL_WPML_Config` and `PLL_Translate_Option`

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/1937.
Related: https://github.com/polylang/polylang-pro/pull/2073.

This PR introduces the class `PLL_Format_Util`. This PR doesn't add wildcard support to anything, it only uses the new class where wildcards were already in use.

Things to note:
1. There are some differences from the version from Polylang Pro:
    - `filter_list()` supports arrays only (we don't use anything else anymore).
    - New method `is_format()` to tell if a given string has a wildcard.
2. Thanks to the new class, `PLL_WPML_Config` and `PLL_Translate_Option` now support keys with special characters.

`PLL_Format_Util` must then be removed from Polylang Pro.

### DIFF
--- a/include/format-util.php
+++ b/include/format-util.php
@@ -1,0 +1,115 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * A class to match values against a given format.
+ *
+ * @since 3.6
+ */
+class PLL_Format_Util {
+	/**
+	 * Cache for regex patterns.
+	 * Useful when using `filter_list()` for example.
+	 *
+	 * @var string[] Formats as array keys, patterns as array values.
+	 *
+	 * @phpstan-var array<non-empty-string, non-empty-string>
+	 */
+	private $patterns = array();
+
+	/**
+	 * Filters the given list to return only the values whose the key or value matches the given format.
+	 *
+	 * @since 3.6
+	 *
+	 * @param array|Traversable|stdClass $list   An list with keys or values to match against `$format`.
+	 * @param string                     $format A format, where `*` means "any characters" (`.*`), unless escaped.
+	 * @param string                     $mode   Optional. Tell if we should filter the keys or values from `$list`.
+	 *                                           Possible values are `'use_keys'` and `'use_values'`. Default is `'use_keys'`.
+	 * @return array
+	 *
+	 * @template TArrayValue
+	 * @phpstan-param (
+	 *     $mode is 'use_keys' ?
+	 *         array<string, TArrayValue>|Traversable<string, TArrayValue>|stdClass :
+	 *         array<string>|Traversable<string>|stdClass
+	 * ) $list
+	 * @phpstan-param 'use_keys'|'use_values' $mode
+	 * @phpstan-return ($mode is 'use_keys' ? array<string, TArrayValue> : array<string>)
+	 */
+	public function filter_list( $list, string $format, string $mode = 'use_keys' ): array {
+		$filter = function ( $key ) use ( $format ) {
+			return $this->matches( (string) $key, $format );
+		};
+
+		if ( $list instanceof Traversable ) {
+			$list = iterator_to_array( $list );
+		} elseif ( ! is_array( $list ) ) {
+			$list = (array) $list;
+		}
+
+		if ( 'use_values' === $mode ) {
+			return array_filter( $list, $filter );
+		}
+
+		return array_filter( $list, $filter, ARRAY_FILTER_USE_KEY );
+	}
+
+	/**
+	 * Tells if the given string matches the given format.
+	 *
+	 * @since 3.6
+	 *
+	 * @param string $key    A string to test.
+	 * @param string $format A format, where `*` means "any characters" (`.*`), unless escaped.
+	 * @return bool
+	 */
+	public function matches( string $key, string $format ): bool {
+		if ( ! $this->is_format( $format ) ) {
+			return $key === $format;
+		}
+
+		if ( '*' === $format ) {
+			return true;
+		}
+
+		if ( empty( $this->patterns[ $format ] ) ) {
+			$pattern = addcslashes( $format, '.+?[^]$(){}=!<>|:-#/' ); // Escape regular expression characters (list from `preg_quote()` but `*` and `\` are ignored).
+			$pattern = preg_replace(
+				array(
+					'/\\\(?!\*)/', // Escape `\` characters except if followed by `*`.
+					'/(?<!\\\)\*/', // Replace `*` characters by `.*` except if preceded by `\`.
+				),
+				array( '\\', '.*' ),
+				$pattern
+			);
+
+			if ( empty( $pattern ) ) {
+				// Error.
+				return false;
+			}
+
+			$this->patterns[ $format ] = $pattern;
+		} else {
+			$pattern = $this->patterns[ $format ];
+		}
+
+		return (bool) preg_match( "/^{$pattern}$/", $key );
+	}
+
+	/**
+	 * Tells if the given string is a format (that includes a `*`).
+	 *
+	 * @since 3.6
+	 *
+	 * @param string $format Format to test.
+	 * @return bool
+	 *
+	 * @phpstan-assert-if-true non-empty-string $format
+	 */
+	public function is_format( string $format ): bool {
+		return (bool) preg_match( '/(?<!\\\)\*/', $format );
+	}
+}
--- a/include/format-util.php
+++ b/include/format-util.php
@@ -6,7 +6,7 @@
 /**
  * A class to match values against a given format.
  *
- * @since 3.6
+ * @since 3.7
  */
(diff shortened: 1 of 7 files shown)


include/translate-option.php
```diff
@@ -130,13 +130,16 @@ public function translate( $value ) {
...
 	 * @param mixed      $values Either a string to translate or a list of strings to translate.
 	 * @param array|bool $key    Array of option keys to translate.
-	 * @return array|string Translated string(s)
+	 * @return array|string Translated string(s).
 	 */
 	protected function translate_string_recursive( $values, $key ) {
 		$children = is_array( $key ) ? $key : array();
 
 		if ( is_array( $values ) || is_object( $values ) ) {
+			/** @var array|Traversable $values */
```
Hug0-Drelon, line 139: Could we pass only `array` to the matcher using `iterator_to_array()`? And thus avoid to manage several type in `PLL_Format_Utils`?

Hug0-Drelon (changes requested):

include/translate-option.php
Screenfeed: Weird, I can't find any use of `filter_list()` anymore where the 1st argument is not an array, neither in PLL and PLL Pro, while I remember we had a case with `stdClass`. I guess it disappeared in a PR.
Anyway, done in 77a0d42.

Now I wonder if we should tag this class and methods `@since 3.7` or `@since 3.6` + a `@since 3.7 Changed signature` for `filter_list()` 🤔

Hug0-Drelon: Could it me some third party plugin that stores its options to translate in an object?

Screenfeed: Each time we use `filter_list()`, we make sure to give it an array: `is_array()` x3 in PLL Pro, `(array)` cast x1 in PLL (master).

Hug0-Drelon: Sorry, I missed your last question.
For the tags I'd say; `@since3.6` + `@since 3.7 Moved from Pro to Free`. What do you think?

Screenfeed: Done in [080af71](https://github.com/polylang/polylang/pull/1404/commits/080af7116de5081d9a995ac1e1e1985d4151cfea).

modules/wpml/wpml-config.php
```diff
@@ -144,19 +146,14 @@ public function init() {
 			foreach ( $keys as $key ) {
 				$name = $this->get_field_attribute( $key, 'name' );
 
-				if ( false === strpos( $name, '*' ) ) {
+				if ( ! $matcher->is_format( $name ) ) {
```
Hug0-Drelon, line 149: I know you kept the same behavior as before. But I'm not very fond of this pattern.
IMHO this class shouldn't be aware if an option has format or not and only pass options and matchers to `PLL_Format_Util()` and get the result.
What do you think to move `PLL_Format_Util::filter_list()` inside `register_or_translate_option()` for instance?

Hug0-Drelon (changes requested):

modules/wpml/wpml-config.php
Screenfeed: I don't know how to do another way, because if it's not a format we don't use `wp_load_alloptions()`.
Also, you say that this class shouldn't be aware of that but you suggest to move this to another method of the same class 😕

Hug0-Drelon: Resolved after a chat.
There is still a performance issue with `wp_load_alloptions()` IMHO, but since it's the current behavior we may not change it.

Hug0-Drelon (approved):

include/format-util.php
```diff
@@ -0,0 +1,107 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $format Format to test.
+	 * @return bool
+	 *
+	 * @phpstan-assert-if-true non-empty-string $format
+	 */
+	public function is_format( string $format ): bool {
+		return (bool) preg_match( '/(?<!\\\)\*/', $format );
```
Chouby, line 105: You changed from `strpos( $format, '*' )` to this. Could you document why or at least document what the regex does exactly?

include/format-util.php
```diff
@@ -0,0 +1,107 @@
...
+	 * Tells if the given string is a format (that includes a `*`).
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $format Format to test.
+	 * @return bool
+	 *
+	 * @phpstan-assert-if-true non-empty-string $format
+	 */
+	public function is_format( string $format ): bool {
```
Chouby, line 104: I suggest to rename to `has_wildcard()`. What do you think?

include/translate-option.php
```diff
@@ -130,13 +130,16 @@ public function translate( $value ) {
 	 *
 	 * @param mixed      $values Either a string to translate or a list of strings to translate.
 	 * @param array|bool $key    Array of option keys to translate.
-	 * @return array|string Translated string(s)
+	 * @return array|string Translated string(s).
```
Chouby, line 133: This should avoid an ignored PHPStan error
```suggestion
	 * @return string|array|Traversable Translated string(s).
```

include/translate-option.php
```diff
@@ -130,13 +130,16 @@ public function translate( $value ) {
 	 *
 	 * @param mixed      $values Either a string to translate or a list of strings to translate.
```
Chouby, line 131: This should avoid to add `@var` inside the method.
```suggestion
	 * @param string|array|Traversable $values Either a string to translate or a list of strings to translate.
```

Chouby (changes requested):

modules/wpml/wpml-config.php
```diff
@@ -144,19 +146,14 @@ public function init() {
...
-				$pattern = '#^' . str_replace( '*', '(?:.+)', $name ) . '$#';
-				$names = preg_grep( $pattern, array_keys( wp_load_alloptions() ) );
-
-				if ( ! is_array( $names ) ) {
-					continue;
-				}
+				$names = $matcher->filter_list( (array) wp_load_alloptions(), $name );
 
-				foreach ( $names as $_name ) {
+				foreach ( $names as $_name => $_val ) {
```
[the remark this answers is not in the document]
Screenfeed: Why would it? We need to create `$_val` to access the array key. it's like:
```php
function foo( $a, $b ) {
    return $b;
}
```

IMHO, `array_keys()` isn't necessary.

include/translate-option.php
Screenfeed: But it creates a new one:
```
L120 - Parameter #2 $data of method PLL_Cache<array|string>::set() expects array|string, array|string|Traversable given.
```

include/translate-option.php
```diff
@@ -191,17 +192,17 @@ protected function register_string_recursive( $context, $option, $values, $key )
...
 				foreach ( $children as $name => $child ) {
 					if ( isset( $values[ $name ] ) ) {
 						$this->register_string_recursive( $context, $name, $values[ $name ], $child );
 						continue;
 					}
 
-					$pattern = '#^' . str_replace( '*', '(?:.+)', $name ) . '$#';
-
 					foreach ( $values as $n => $value ) {
 						// The first case could be handled by the next one, but we avoid calls to preg_match here.
```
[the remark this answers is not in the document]
Screenfeed: Removed in [1c322e5](https://github.com/polylang/polylang/pull/1404/commits/1c322e5a02d20933e2418cce1142ac6104f597ce).

Also, I added [7f7c672](https://github.com/polylang/polylang/pull/1404/commits/7f7c6725bf85d2157a308d8aee55321f7f260d82) to prevent doing `forech` if `$name` is not a format.

modules/wpml/wpml-config.php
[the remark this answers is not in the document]
Screenfeed: As we discussed, no really need to use `array_keys()` since PHPCS doesn't care.

include/format-util.php
```diff
@@ -0,0 +1,107 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * A class to match values against a given format.
+ *
+ * @since 3.6
+ * @since 3.7 Moved from Polylang Pro to Polylang.
```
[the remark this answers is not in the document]
Screenfeed: As discussed, we'll keep it that way.

include/format-util.php
Screenfeed: As discussed, we'll keep it that way, especially since we're not sure how we're going to implement "regex pattern formats" in the future.

include/translate-option.php
Screenfeed: If we change things, that will be in a separate PR.

include/translate-option.php
Screenfeed: If we change things, that will be in a separate PR.

PR #1406, opened 2023-12-28
Allow setcookie httponly args to be filtered

sebastienserre:
this PR should fix https://github.com/polylang/polylang-pro/issues/1003

### DIFF
--- a/include/cookie.php
+++ b/include/cookie.php
@@ -31,12 +31,21 @@ class PLL_Cookie {
 		 */
 		$expiration = (int) apply_filters( 'pll_cookie_expiration', YEAR_IN_SECONDS );
 
+		/**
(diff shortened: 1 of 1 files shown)


Screenfeed (approved):

include/cookie.php
```diff
@@ -34,13 +34,34 @@ protected static function parse_args( $args ) {
...
 		);
 
-		return wp_parse_args( $args, $defaults );
+		$args = wp_parse_args( $args, $defaults );
+
+		/**
+		 * Filters the Polylang cookie arguments.
+		 * /!\ This filter may be fired *before* the theme is loaded.
+		 *
+		 * @since 3.7
```
Chouby, line 49: I believe that we can push this now:
```suggestion
		 * @since 3.6
```

PR #1415, opened 2024-01-24
Fix wrong post types and taxonomies cached during a switch_blog()

Chouby:
This fixes an issue reported in https://secure.helpscout.net/conversation/2485021872/27260

Since #1345, during a `switch_blog()` we [remove rewrite rules filters](https://github.com/polylang/polylang/blob/3.5.3/include/base.php#L150) associated to the previous blog. This is intentionally done before switching the options to the new blog in `PLL()->options`, in order to have the correct list of post types and taxonomies in [`PLL_Links_Permalinks::get_rewrite_rules_filters()`](https://github.com/polylang/polylang/blob/3.5.3/include/links-permalinks.php#L192-L205).
On the other hand, `PLL_Model::get_translated_post_types()` and `PLL_Model::get_translated_taxonomies()` have a blog dependent cache which aims to store the translated post types and taxonomies of the current blog.
Since `get_translated_post_types() ` and `get_translated_taxonomies()` are called before the options are switched we cache the list of post types and taxonomies of the previous blog instead of the new blog.

### DIFF
--- a/tests/phpunit/tests/test-switch-blog-translatable-objects.php
+++ b/tests/phpunit/tests/test-switch-blog-translatable-objects.php
@@ -20,12 +20,12 @@ if ( is_multisite() ) :
 		protected function set_up_blog_with_pll( WP_Site $blog, array $languages, array $options, string $structure ) {
 			$cpt = uniqid( 'cpt_', true );
 			$tax = uniqid( 'tax_', true );
-			$this->translatable_objects[ (int) $blog->blog_id ] = array(
+			$this->translatable_objects[ $blog->blog_id ] = array(
 				'post_types' => array( $cpt => $cpt ),
 				'taxonomies' => array( $tax => $tax ),
 			);
 
-			$options = array_merge( $options, $this->translatable_objects[ (int) $blog->blog_id ] );
+			$options = array_merge( $options, $this->translatable_objects[ $blog->blog_id ] );
 
 			parent::set_up_blog_with_pll( $blog, $languages, $options, $structure );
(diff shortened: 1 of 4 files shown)


szepeviktor (approved):

tests/phpunit/tests/test-switch-blog-translatable-objects.php
```diff
@@ -0,0 +1,142 @@
...
+		 * @param WP_Site $blog      Blog to set up.
+		 * @param array   $languages Languages to create.
+		 * @param array   $options   Polylang options to use.
+		 * @param string  $structure Permalink structure to use.
+		 * @return void
+		 */
+		protected function set_up_blog_with_pll( WP_Site $blog, array $languages, array $options, string $structure ) {
+			$cpt = uniqid( 'cpt_', true );
+			$tax = uniqid( 'tax_', true );
+			$this->translatable_objects[ (int) $blog->blog_id ] = array(
```
Screenfeed, line 23: Casting array keys as integers is useless: php does it by itself.
```php
$arr = [ '7' => 'foo' ];
foreach ( $arr as $k => $v ) {
	var_dump( $k === "$k" ); // bool(false)
	var_dump( $k === (int) $k ); // bool(true)
}
```
So we can simplify the code of class a bit.

Hug0-Drelon: 👍 
Done in f2331030.

Screenfeed (approved):

PR #1418, opened 2024-02-04
Don't set post tags with names. Use term ids instead

### LINKED ISSUE #1385: All tags are attached to a post on save, if tags are set hierarchical
## Prerequisites

1. WP 6.4.1, Polylang 3.6-dev / 3.5.2, Twenty Nineteen 2.7 
2. Add the following code to `functions.php`
3. Create n amount of tags
4. Add 1 tag to a new post or to an existing post

```
add_filter( 'register_taxonomy_args', 'provide_hierarchical_tags', 10, 2 );
function provide_hierarchical_tags(array $args, string $taxonomy): array {
	if ('post_tag' !== $taxonomy) {
		return $args;
	}

	global $wp_rewrite;

	$args['hierarchical'] = true;
	$args['rewrite'] = [
		'hierarchical' => false,
		'slug' => get_option('tag_base') ? get_option('tag_base') : 'tag',
		'with_front' => ! get_option('tag_base') || $wp_rewrite->using_index_permalinks(),
		'ep_mask' => EP_TAGS,
	];

	return $args;
}
```

When I follow those steps, all available tags are attached to the post.

I was expecting that only the selected tag would have been attached to the post.

### LINKED ISSUE #1401: Polylang bug with tags (inserted wrong tag)!
## Prerequisites

1. Install fresh new WordPress with permalink custom structure `/%category%/%postname%/` (but I think it is not important).
2. Delete all plugin but install and active only Polylang. Primary language: Thai, secondary: English-US.
3. Delete all posts, pages, tags. Leave uncategorized and uncategorized-en untouch (I think category is not important).
4. Use any theme. In my testing, I use **Twenty Twenty-Four** but any theme is same.
5. Create 2 new tags with these details. 
    First tag name: **`ชื่อเต็ม กทม`** and slug is **`กรุงเทพมหานคร-อมรรัตนโก`**. 
    Second tag name: **`กรุงเทพมหานคร อมรรัตนโกสินทร์`** and slug is **`บางกอก`**.
6. Create new post or edit existing post.
7. Add tag **`กรุงเทพมหานคร อมรรัตนโกสินทร์`** or you may enter **กรุง** to search and it will be found only one (only one which is correct, the tag name **กรุงเทพ...** has only one, the other one is **ชื่อเต็ม กทม**).
8. Update or submit or save that post.
9. View your post on front page you will see that tag name is **INCORRECT**! It will be displayed **ชื่อเต็ม กทม** instead of **กรุงเทพมหานคร อมรรัตนโกสินทร์**.

The expected behavior is Polylang should not change my insert tag which is **กรุงเทพมหานคร อมรรัตนโกสินทร์**. 

The actual behavior is Polylang change my tag to be **ชื่อเต็ม กทม**.

However, if I deactivated Polylang, the only WordPress itself is working correctly.

The problem maybe because the slug name is look like other tag name and Polylang catch from that.  
You may say that I have to change it to be completely different but in real project I have tag names **หนังสือaaaa Aaaaa**, **หนังสือaaaa AaaBbbb** (aa, bb is censored) which is not possible to change because it will be unreadable (in my language) and can be duplicated with **เสียงaaa Aaaa**, etc if I remove some part of it.

Tested with Polylang 3.6-dev. WordPress 6.4.2, or 6.5 (dev version)

Chouby:
Fixes #1401 
May fix #1385 too.

WordPress may mix 2 post different post tags if the if the slug of the first tag is the same as the name of the second tag.
This is due to `wp_set_object_terms()` calling [`term_exists()`](https://github.com/WordPress/wordpress-develop/blob/6.4.3/src/wp-includes/taxonomy.php#L2806) which [searches either among slugs or names](https://github.com/WordPress/wordpress-develop/blob/6.4.3/src/wp-includes/taxonomy.php#L1606-L1611).
As a consequence, we should **never** use term names or slugs when setting post terms.

This PR replaces a call to `wp_set_post_tags()` with tag names by a call to `wp_set_post_terms()` with corresponding term_ids.

It adds 2 tests which are supposed to mimic the behavior of the classic editor and the block editor when we assign a post tag to a post.

### DIFF
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -355,12 +355,20 @@ class PLL_CRUD_Posts {
 	 * @return void
 	 */
 	public function force_tags_translation( $post_id, $post_after, $post_before ) {
-		$post_before = $post_before->to_array();
-
-		if ( ! empty( $post_before['tags_input'] ) ) {
-			// Let's ensure that `PLL_CRUD_Posts::set_object_terms()` will do its job.
-			wp_set_post_tags( $post_id, $post_before['tags_input'] );
+		if ( ! is_object_in_taxonomy( $post_before, 'post_tag' ) ) {
+			return;
 		}
+
+		$terms = get_the_terms( $post_before, 'post_tag' );
+
+		if ( empty( $terms ) ) {
+			return;
+		}
+
+		$term_ids = wp_list_pluck( $terms, 'term_id' );
+
+		// Let's ensure that `PLL_CRUD_Posts::set_object_terms()` will do its job.
+		wp_set_post_terms( $post_id, $term_ids, 'post_tag' );
 	}
 
 	/**
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -355,13 +355,13 @@ class PLL_CRUD_Posts {
 	 * @return void
 	 */
 	public function force_tags_translation( $post_id, $post_after, $post_before ) {
-		if ( ! is_object_in_taxonomy( $post_before, 'post_tag' ) ) {
+		if ( ! is_object_in_taxonomy( $post_before->post_type, 'post_tag' ) ) {
 			return;
 		}
 
 		$terms = get_the_terms( $post_before, 'post_tag' );
 
-		if ( empty( $terms ) ) {
+		if ( empty( $terms ) || ! is_array( $terms ) ) {
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


ve3:
Confirmed fixed #1401 .

Hug0-Drelon (approved):

Screenfeed (approved):

PR #1420, opened 2024-02-07
Machine Translation module preview

Screenfeed:
This PR adds a "Machine Translation" preview module.
This also brings changes in `PLL_Settings_Module` and its property `$active_option`, to prevent duplicating code in modules.

The modification in CSS is for the `vertical-align` to be reused in the news settings fields in PLL Pro.


Chouby:
image
I expected that the module appears as deactivated as the others.

js/src/admin.js
```diff
@@ -2,6 +2,67 @@
...
+			const action    = button.getAttribute( 'data-action' );
+			const nonce     = button.getAttribute( 'data-nonce' );
+			const row       = button.closest( 'tr' );
+			const errorElms = row.querySelectorAll( '.pll-error-message-text' );
+
+			if ( ! action || ! nonce || ! row || ! errorElms.length ) {
+				return;
+			}
+
+			const sep = ( ajaxurl.indexOf( '?' ) !== -1 ) ? '&' : '?';
```
manooweb, line 36: See if we couldn't also depend on `wp.url` package? 🤔
There are also functions available in this package like `addQueryArgs` which already manage the query string separator.

manooweb (changes requested):

Screenfeed:
> I expected that the module appears as deactivated as the others.

Fixed in [369842c](https://github.com/polylang/polylang/pull/1420/commits/369842c57df04ac79cb26b69b6c1c2b908f9604b).
I introduced some improvements to `PLL_Settings_Module`, to prevent duplicating code in each "preview" class.
For example this avoids to overwrite `is_active()` in `PLL_Settings_Preview_Machine_Translation` (like we did in `PLL_Settings_Preview_Share_Slug`) to make it always return `false`, and then re-overwrite it again in PLL Pro to re-enable the original method.

js/src/admin.js
Screenfeed: Done in [6ceaf75](https://github.com/polylang/polylang/pull/1420/commits/6ceaf753e97aba3322e52be876406ff58e3d5424).

js/src/admin.js
```diff
@@ -33,16 +33,16 @@ window.pll.settings_ajax_button = () => {
 				return;
 			}
 
-			const sep = ( ajaxurl.indexOf( '?' ) !== -1 ) ? '&' : '?';
-			let   url = ajaxurl + sep + 'action=' + action + '&_pll_nonce=' + nonce + '&pll_ajax_settings=1';
-
+			let urlParams = { 'action': action, '_pll_nonce': nonce, 'pll_ajax_settings': 1 };
```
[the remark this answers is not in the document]
Screenfeed: Fixed in [3314243](https://github.com/polylang/polylang/pull/1420/commits/33142432e8760b53abf6b1ea3b09b0b2efa684f2).

js/src/admin.js
```diff
@@ -2,6 +2,67 @@
  * @package Polylang
  */
 
+window.pll = window.pll || {};
```
Hug0-Drelon, line 5: This is an old fashion way of accessing `pll` from other script sheets. This is not what we want here.
Also it would break without error if some third party do the same.
Please use `const pllMachineTranslationButton` or something instead.

js/src/admin.js
```diff
@@ -2,6 +2,67 @@
...
+		fn();
+	} else {
+		document.addEventListener( 'DOMContentLoaded', fn );
+	}
+};
+
+/**
+ * Settings: manages AJAX calls attached to `.pll-ajax-button` buttons.
+ */
+window.pll.settings_ajax_button = () => {
```
Hug0-Drelon, line 23: I see you have a modular approach so I suggest to go all in.
- Move the code in a dedicated file then import it here.
- Have a `init()` function so you elements are prepared before the event listener callback is called.

Bonus, camelCase please ❤️

js/src/admin.js
```diff
@@ -2,6 +2,67 @@
...
+			} ).then( ( json ) => {
+				if ( json.success ) {
+					row.classList.add( 'notice-success', 'notice-alt' );
+				} else {
+					errorElms[0].textContent = json.data;
+					row.classList.add( 'notice-error', 'notice-alt' );
+				}
+			} ).catch( () => {
+				button.removeAttribute( 'disabled' );
+				errorElms[0].textContent = 'Unknown error.';
```
Hug0-Drelon, line 57: Maybe we want to translate that.

js/src/admin.js
```diff
@@ -2,6 +2,67 @@
...
+
+			button.setAttribute( 'disabled', 'disabled' );
+			row.classList.remove( 'notice-success', 'notice-error', 'notice-alt' );
+
+			fetch( url ).then( ( response ) => {
+				button.removeAttribute( 'disabled' );
+				return response.json();
+			} ).then( ( json ) => {
+				if ( json.success ) {
+					row.classList.add( 'notice-success', 'notice-alt' );
```
Hug0-Drelon, line 50: This kind of logic could benefit from a `toggle` function along `button.removeAttribute( 'disabled' );`.

Hug0-Drelon (changes requested):

js/src/admin.js
Screenfeed: It wouldn't break unless the plugin does this directly:
```JS
window.pll = { ... };
```
In this case we can say it's intentional.

`const` added in [411277a](https://github.com/polylang/polylang/pull/1420/commits/411277a5b02769446f93f41ed8cbed2522dddc41).

js/src/admin.js
Screenfeed: Done in [411277a](https://github.com/polylang/polylang/pull/1420/commits/411277a5b02769446f93f41ed8cbed2522dddc41).

js/src/admin.js
Screenfeed: I didn't went through, it adds code instead of removing some.

js/src/admin.js
Screenfeed: Done in [411277a](https://github.com/polylang/polylang/pull/1420/commits/411277a5b02769446f93f41ed8cbed2522dddc41) and [9ce6f69](https://github.com/polylang/polylang/pull/1420/commits/9ce6f69003c2b3d8c8f7a5f4d9a3d8a881be5249).

js/src/admin.js
```diff
@@ -2,61 +2,7 @@
...
-					errorElms[0].textContent = pll_admin.unknown_error;
-					row.classList.add( 'notice-error', 'notice-alt' );
-				} )
-			} );
-		} );
-	}
-};
-
-pllMachineTranslationAjaxButton.init();
+import {} from './settings';
```
Hug0-Drelon, line 5:
```suggestion
import { pllMachineTranslationAjaxButton } from './settings';

pllMachineTranslationAjaxButton.init();
```
So it describe what it does, and run the code.

js/src/admin.js
manooweb: 🤔 we can use `toggleAttribute( 'disabled')` everywhere instead of `setAttribute` or `removeAttribute`
See https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute

Screenfeed: Nope, I'm against any `toggle*()`, it's too unpredictable: when you use it don't know if it will add or remove the attribute.

js/src/admin.js
manooweb: Please put `settings.js` in lib sub folder not to be built itself.

Screenfeed: > Please put `settings.js` in `lib` sub folder not to be built itself.

Done in [f916e62](https://github.com/polylang/polylang/pull/1420/commits/f916e6260bd5b55597307fa82b2d20bb6d66038a).

Hug0-Drelon (approved):

manooweb (approved):

modules/machine-translation/settings-preview-machine-translation.php
```diff
@@ -0,0 +1,39 @@
...
+	 *
+	 * @since 3.6
+	 *
+	 * @param PLL_Settings $polylang Polylang object.
+	 */
+	public function __construct( &$polylang ) {
+		parent::__construct(
+			$polylang,
+			array(
+				'module'        => 'machine-translation',
```
manooweb, line 32: Settings actions are based on module name.
For example https://github.com/polylang/polylang/blob/3.5.4/settings/settings.php#L260-L266
When Polylang Pro is activated the module name is the full namespaced qualified class name
https://github.com/polylang/polylang-pro/blob/ef45ad2f04073a66397fa52011d3bd78e87aa85f/modules/machine-translation/load.php#L27-L31 
This is this name which is used as key in the settings `modules` array because the class name doesn't starts with `PLL_Settings_` 
See https://github.com/polylang/polylang/blob/3.5.4/settings/settings.php#L89
It is also used as module property in the instance itself 
https://github.com/polylang/polylang/blob/3.5.4/settings/settings-module.php#L116

So sanitization make this module name unknown 
https://github.com/polylang/polylang/blob/3.5.4/settings/settings.php#L262 because `wp_syntex\\polylang_pro\\modules\\machine_translation\\module_settings` is sanitized to `wp_syntexpolylang_promodulesmachine_translationmodule_settings`

manooweb (changes requested):

modules/machine-translation/settings-preview-machine-translation.php
Screenfeed: Fixed in [ed560d7](https://github.com/polylang/polylang/pull/1420/commits/ed560d77b66a28f77553259469920d5aaeb16079).
@manooweb You were not looking at the right git branch, this system [has changed a bit](https://github.com/polylang/polylang-pro/blob/f5251bbd3e47ed4d396c8475af4c52cc045a8cf4/modules/admin-loader/admin-loader.php#L91-L92).

Screenfeed: @manooweb Ah! You're right, I missed it.

Screenfeed: Fixed in [c284291](https://github.com/polylang/polylang-pro/pull/1974/commits/c284291b6ac6e992b7b36c86aa0a807bc4ab2924).

manooweb (approved):

modules/machine-translation/settings-preview-machine-translation.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @since 3.6
+ */
+class PLL_Settings_Preview_Machine_Translation extends PLL_Settings_Module {
+	/**
+	 * Stores the display order priority.
+	 *
+	 * @var int
+	 */
+	public $priority = 85;
```
Chouby, line 19: I think it's free and still < License key prio:
```suggestion
	public $priority = 90;
```

modules/machine-translation/settings-preview-machine-translation.php
```diff
@@ -0,0 +1,39 @@
...
+		parent::__construct(
+			$polylang,
+			array(
+				'module'        => 'machine_translation',
+				'title'         => __( 'Machine Translation', 'polylang' ),
+				'description'   => __( 'Allows linkage to an external translation solution.', 'polylang' ),
+				'active_option' => '',
+			)
+		);
+	}
```
Chouby, line 38: For previews like this one (for share slugs and translated slugs too), I suggest to merge current arguments with arguments coming from the child class:
```suggestion
	public function __construct( &$polylang, $args = array() ) {
		$defaults = array(
			'module'        => 'machine_translation',
			'title'         => __( 'Machine Translation', 'polylang' ),
			'description'   => __( 'Allows linkage to an external translation solution.', 'polylang' ),
			'active_option' => '',
		);

		parent::__construct( $polylang, array_merge( $default, $args ) );
	}
```
What do you think

modules/machine-translation/settings-preview-machine-translation.php
Screenfeed: Done in [72fb3a5](https://github.com/polylang/polylang/pull/1420/commits/72fb3a59c4b05435a120e5467ff525323cf5fe15).

modules/machine-translation/settings-preview-machine-translation.php
Screenfeed: Done in [9732372](https://github.com/polylang/polylang/pull/1420/commits/97323722ce8c9f85ede58ff4d9bdc3bc31584b72).

Screenfeed:
Following the discussion with @Chouby, `PLL_Settings_Module::$active_option` can now have 3 values: `'none'` (is always active), `'preview'` (the module is a preview in PLL, so, always, inactive), and an option key (can be activated/deactivated). In PLL Pro, a `'preview'` module must be overwritten by one of the other 2 values.

settings/settings-module.php
```diff
@@ -145,62 +170,62 @@ public function __construct( &$polylang, $args ) {
...
-	 * Tells if the module is active
+	 * Tells if the module is active.
 	 *
 	 * @since 1.8
 	 *
 	 * @return bool
 	 */
 	public function is_active() {
-		return empty( $this->active_option ) || ! empty( $this->options[ $this->active_option ] );
+		return 'none' === $this->active_option || 'preview' !== $this->active_option && ! empty( $this->options[ $this->active_option ] );
```
Chouby, line 185: Could you add parentheses? I don't like relying on operator precedence.

settings/settings-module.php
```diff
@@ -29,9 +29,15 @@ class PLL_Settings_Module {
 	public $links_model;
 
 	/**
-	 * Stores if the module is active.
+	 * Key to use to manage the module activation state.
+	 * Possible values:
+	 * - An option key for a module that can be activated/deactivated.
+	 * - 'none' for a module that is always active.
```
Chouby, line 35: I know that I wrote that some days ago but it seems that's incorrect. Instead it looks like 'none' corresponds to "no settings available". Whether it is active or not is programmatically decided. See `PLL_Settings_Browser` for ex.

Chouby (changes requested):

settings/settings-module.php
Screenfeed: Done in [daeb7eb](https://github.com/polylang/polylang/pull/1420/commits/daeb7eb86853195c76675b76bd5141c96be20faa).

settings/settings-module.php
Screenfeed: You're right, I forgot this case. Done in [01e6886](https://github.com/polylang/polylang/pull/1420/commits/01e6886688f096099ab78716d813791289a8daf1).

PR #1421, opened 2024-02-09
Add DeepL language codes in our list.

Hug0-Drelon:
## What?
All in the title.

## How?
Add a key `deepl` to language list in `languages.php`.
This key is for target language code.
Source code is the same as the one already here (key `code`).

### DIFF
--- a/settings/languages.php
+++ b/settings/languages.php
@@ -80,6 +80,7 @@ return array(
 		'dir'      => 'rtl',
 		'flag'     => 'arab',
 		'facebook' => 'ar_AR',
+		'deepl'    => 'AR',
 	),
 	'arg' => array(
 		'code'     => 'an',
@@ -98,6 +99,7 @@ return array(
 		'dir'      => 'rtl',
 		'flag'     => 'ma',
 		'facebook' => 'ar_AR',
+		'deepl'    => 'AR',
 	),
 	'as' => array(
 		'code'     => 'as',
--- a/settings/languages.php
+++ b/settings/languages.php
@@ -138,6 +138,7 @@ return array(
 		'dir'      => 'ltr',
 		'flag'     => 'bg',
 		'facebook' => 'bg_BG',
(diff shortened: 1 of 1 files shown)


Marianne380 (approved):
LGTM
Just to be sure, we're not adding **Arabic** because it hasn't yet been added to the `/languages` endpoint?

Hug0-Drelon:
> LGTM Just to be sure, we're not adding **Arabic** because it hasn't yet been added to the `/languages` endpoint?

@Marianne380 To my knowledge, Arabic isn't available yet in DeepL. At least through their API. Do you have contradictory infromation?

Chouby (changes requested):
From [the doc](https://www.deepl.com/docs-api/translate-text):
>[1] Please note that Arabic has not yet been added to the /languages endpoint because it does not yet support document translation; only text translation is supported for Arabic at this time. When document translation support is added for Arabic, we will a) remove this note and b) add Arabic to the /languages endpoint.

So we must add `AR`.

I am not sure that we can use simplified Chinese for `zh_HK` and `zh_TW`, but let's keep it like that. If a Chinese speaking user tells us its nonsense we can easily remove that in the future.

Hug0-Drelon:
> From [the doc](https://www.deepl.com/docs-api/translate-text):
> 
> > [1] Please note that Arabic has not yet been added to the /languages endpoint because it does not yet support document translation; only text translation is supported for Arabic at this time. When document translation support is added for Arabic, we will a) remove this note and b) add Arabic to the /languages endpoint.
> 
> So we must add `AR`.
> 
> I am not sure that we can use simplified Chinese for `zh_HK` and `zh_TW`, but let's keep it like that. If a Chinese speaking user tells us its nonsense we can easily remove that in the future.

@Chouby My bad, I didn't read that part of the doc. I based my knowledge regarding language availability only through their API.
I added Arabic in 0b9dc4e8.

PR #1424, opened 2024-02-15
Replace settings error slug with polylang

manooweb:
Fixes partially https://github.com/polylang/polylang-pro/issues/2033 Polylang part.

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -181,17 +181,17 @@ class PLL_Settings extends PLL_Admin_Base {
 
 				if ( is_wp_error( $errors ) ) {
 					foreach ( $errors->get_error_codes() as $code ) {
-						add_settings_error( 'general', $code, $errors->get_error_message( $code ) );
+						add_settings_error( 'polylang', $code, $errors->get_error_message( $code ) );
 					}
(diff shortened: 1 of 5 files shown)


settings/settings.php
Hug0-Drelon, line 1: Should we take the opportunity to use `esc_html__()` instead of `__()` in this file?

manooweb: It's another task of #2023

Chouby (commented):
Shouldn't this test fail ? https://github.com/polylang/polylang/blob/3.5.4/tests/phpunit/tests/test-settings.php#L109-L119

manooweb:
> Shouldn't this test fail ? https://github.com/polylang/polylang/blob/3.5.4/tests/phpunit/tests/test-settings.php#L109-L119

No! Because notices in this case are displayed by WordPress without using any notice slug.
So we can set any notice slug we want here.

See https://github.com/WordPress/WordPress/blob/6.4.3/wp-admin/options-head.php#L18 used [here](https://github.com/polylang/polylang/blob/3.5.4/settings/view-languages.php#L12)

PR #1425, opened 2024-02-16
Fixes PHPStan WP_Query::$query undefined error

manooweb:
This [error reported](https://github.com/polylang/polylang/actions/runs/7905456130/job/21578129342?pr=1423)

```log
Error: Access to an undefined property object::$query.
 ------ -------------------------------------------------------------------------- 
  Line   frontend/canonical.php                                                    
 ------ -------------------------------------------------------------------------- 
  257    Access to an undefined property object::$query.                           
         💡 Learn more:                                                             
            https://phpstan.org/blog/solving-phpstan-access-to-undefined-property  
 ------ -------------------------------------------------------------------------- 

```
seems due to PHPStan [1.10.58](https://github.com/phpstan/phpstan/releases/tag/1.10.58) release.

If seems that PHPdoc `@global WP_Query $wp_query` annotation has no effect because when add a `@var WP_Query` annotation 

```
/**
 * @var WP_Query
 */
global $wp_query;

```
we don't have the same error

```
  Line   canonical.php
 ------ --------------------------------------------------------------------------
  :258   Property WP_Query::$tax_query (WP_Tax_Query) in isset() is not nullable.
 ------ --------------------------------------------------------------------------

```

exactly like in this [PHPstan playground](https://phpstan.org/r/9e08cc6d-b602-46d4-995f-830fe7f59d47).

Uses `! is_null()` instead of `isset()` seems to solve the error.

NB: uses `! empty()` has the same effect as `isset()` ➡ the initial PHPStan error reported.

### DIFF
--- a/frontend/canonical.php
+++ b/frontend/canonical.php
@@ -252,7 +252,7 @@ class PLL_Canonical {
 
 		$backup_wp_query = $wp_query;
 
-		if ( isset( $wp_query->tax_query ) ) {
+		if ( ! is_null( $wp_query->tax_query ) ) {
 			unset( $wp_query->tax_query->queried_terms['language'] );
 			unset( $wp_query->query['lang'] );
 		}
--- a/frontend/canonical.php
+++ b/frontend/canonical.php
@@ -246,13 +246,16 @@ class PLL_Canonical {
 	 * @return string
 	 */
 	protected function redirect_canonical( $url, $language ) {
+		/**
+		 * @var WP_Query
+		 */
 		global $wp_query;
 
 		$this->curlang = $language; // Hack to filter the `page_for_posts` option in the correct language.
 
 		$backup_wp_query = $wp_query;
 
-		if ( ! is_null( $wp_query->tax_query ) ) {
+		if ( isset( $wp_query->tax_query ) ) {
 			unset( $wp_query->tax_query->queried_terms['language'] );
 			unset( $wp_query->query['lang'] );
 		}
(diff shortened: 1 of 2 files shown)


Screenfeed (approved):
I don't know if we should fix this here, or if something should be done on PHPStan's side.

From what I understand (and I may be wrong), this property is tagged `@var WP_Tax_Query` but can be empty, which means it should be tagged `@var WP_Tax_Query|null`. But doing this change will probably trigger a lot of "you're using this property but it may be empty" errors (I don't mean for us specifically).

So your solution may be the best one for us since it is trivial and has zero impact.

frontend/canonical.php
```diff
@@ -252,7 +252,7 @@ protected function redirect_canonical( $url, $language ) {
 
 		$backup_wp_query = $wp_query;
 
-		if ( isset( $wp_query->tax_query ) ) {
+		if ( ! is_null( $wp_query->tax_query ) ) {
```
Chouby, line 255: This rather looks like a trick reliying on a wrong PHPStan detection. The issue is not on `tax_query` but on `query`
1. Currently `$wp_query` is an anonymous object for PHPStan, so we should check that the property exists before accessing it.
```php
if ( isset( $wp_query->query ) ) {
	unset( $wp_query->query['lang'] );
}
```
This is one solution.
2. A second solution would be to document the variable as `WP_Query` as you mentioned. The new reported error is due to a doc error in WP. You could open a ticket and meanwhile we could temporarily ignore this error. 
3. Side note: Don't use `is_null()` but `null ===` instead which is faster (speed is not important here but it's a good habit to have).

Chouby (changes requested):

frontend/canonical.php
manooweb: I chose second solution in 7529b53faa2c646b1912fb09f55c3728ba2a0c6e 
See https://core.trac.wordpress.org/ticket/60563

PR #1430, opened 2024-02-28
Adds an add_settings_error() wrapper and uses it

manooweb:
Fixes partially https://github.com/polylang/polylang-pro/issues/2060 (Polylang parts)

See [#2060](https://github.com/polylang/polylang-pro/issues/2060) for explanations of what this PR propose.
First step required before applying to Polylang Pro.

Finally, this wrapper function name is `pll_add_notice()`.

We will need to take in account the [Polylang Pro special case](https://github.com/polylang/polylang-pro/blob/3.6-beta1/modules/machine-translation/Module_Settings.php#L129-L144).  ➡ In https://github.com/polylang/polylang-pro/pull/2067 we decided to pass the whole data as a `WP_Error()` extra data string. So the string is passed to `add_settings_error()` $type parameter as before.

### DIFF
--- a/include/functions.php
+++ b/include/functions.php
@@ -201,3 +201,37 @@ function pll_is_plugin_active( string $plugin_name ) {
 
 	return in_array( $plugin_name, $plugins );
 }
+
+/**
+ * Prepares and registers errors
+ *
+ * Wraps `add_settings_error()` to make its use more consistent.
+ *
+ * @since 3.6
+ *
+ * @param WP_Error $error Error object.
+ * @return void
+ */
+function pll_add_settings_error( WP_Error $error ) {
+	if ( ! $error->has_errors() ) {
+		return;
+	}
+
+	foreach ( $error->get_error_codes() as $error_code ) {
+		// Extract the "error" type.
+		$data = $error->get_error_data( $error_code );
+		$type = empty( $data ) || ! is_string( $data ) ? 'error' : $data;
+
+		$message = wp_kses(
+			$error->get_error_message( $error_code ),
+			array(
+				'a' => array( 'href' ),
+				'br',
+				'code',
+				'em',
+			)
+		);
+
+		add_settings_error( 'polylang', $error_code, $message, $type );
+	}
+}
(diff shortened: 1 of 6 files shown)


Screenfeed (approved):

include/functions.php
```diff
@@ -201,3 +201,37 @@ function pll_is_plugin_active( string $plugin_name ) {
...
+ * Prepares and registers errors
+ *
+ * Wraps `add_settings_error()` to make its use more consistent.
+ *
+ * @since 3.6
+ *
+ * @param WP_Error $error Error object.
+ * @return void
+ */
+function pll_add_settings_error( WP_Error $error ) {
```
Chouby, line 215: As we don't use them only for settings, what do you think to rename the function? Maybe `pll_add_error_notice()` ?

include/functions.php
```diff
@@ -201,3 +201,37 @@ function pll_is_plugin_active( string $plugin_name ) {
...
+		return;
+	}
+
+	foreach ( $error->get_error_codes() as $error_code ) {
+		// Extract the "error" type.
+		$data = $error->get_error_data( $error_code );
+		$type = empty( $data ) || ! is_string( $data ) ? 'error' : $data;
+
+		$message = wp_kses(
+			$error->get_error_message( $error_code ),
```
Screenfeed, line 226: What about this?
```suggestion
			implode( '<br>', $error->get_error_messages( $error_code ) ),
```
This is done [here](https://github.com/polylang/polylang-pro/blob/fa6667c34f6f0afca29027c0c6a12e80b9f4687c/modules/import-export/export/export-strings-action.php#L161-L162) and [here](https://github.com/polylang/polylang-pro/blob/fa6667c34f6f0afca29027c0c6a12e80b9f4687c/modules/import-export/import/import-action.php#L320-L321).

Screenfeed: Yes, that's because I was reviewing this other PR that I suggested this ;)

include/functions.php
manooweb: Renamed to `pll_add_notice()` in bee5b32d5d80cfa3233f596a38193a992be6429c

PR #1431, opened 2024-02-29
Hide the "The language is set from content" option

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/1921.

This PR completely hides the "The language is set from content" option in two cases:
- on new installations,
- on upgrade, if it isn't the option selected.

## Two ways to keep it

1- Define a constant in the `wp-config.php` file, this works only if added BEFORE the first activation or upgrade to 3.7:
```php
define( 'PLL_DISPLAY_LANGUAGE_FROM_CONTENT_OPTION', true );
```

2- Or add a filter to a Must-Use plugin, this works only if added BEFORE the first activation or upgrade to 3.7:
```php
add_filter( 'pll_display_language_from_content_option', '__return_true' );
```

Both can be removed after Polylang's first activation/upgrade.

## Dev note

```php
abstract class Abstract_Option {
    public function __construct( $value = null, Options $options ) {
```
>Optional `$value` before `$options`: deprecated in php 8.0.

```php
abstract class Abstract_Option {
    public function __construct( Options $options, $value ) {
```
>PHPStan: `$options` is not used.

```php
abstract class Abstract_Option {
    public function __construct( $value, Options $options ) {
```
>PHPStan: `$options` is not used.

So the solution I chose is:

```php
abstract class Abstract_Option {
    public function __construct( $value ) {} // Default value `null` removed so it behaves the same for each class.
}

class Force_Lang extends Abstract_Option {
    public function __construct( $value, Options $options ) {}
}
```

### DIFF
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -52,9 +52,9 @@ abstract class Abstract_Option {
 	 *
 	 * @since 3.7
 	 *
-	 * @param mixed $value Optional. Option value.
+	 * @param mixed $value Option value. Use `null` to set the default value.
 	 */
-	public function __construct( $value = null ) {
+	public function __construct( $value ) {
 		$this->errors = new WP_Error();
 
 		if ( ! isset( $value ) ) {
--- a/include/Options/Business/Force_Lang.php
+++ b/include/Options/Business/Force_Lang.php
@@ -6,15 +6,38 @@
 namespace WP_Syntex\Polylang\Options\Business;
 
 use WP_Syntex\Polylang\Options\Abstract_Option;
+use WP_Syntex\Polylang\Options\Options;
 
 defined( 'ABSPATH' ) || exit;
 
 /**
  * Class defining the "Determine how the current language is defined" option.
+ * /!\ Constructor depends on `hide_language_from_content_option`: this option must be set AFTER `hide_language_from_content_option`.
  *
  * @since 3.7
  */
 class Force_Lang extends Abstract_Option {
+	/**
+	 * @var array
+	 */
+	private $enum = array( 0, 1, 2, 3 );
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.7
+	 *
+	 * @param mixed   $value   Option value. Use `null` to set the default value.
+	 * @param Options $options All options.
+	 */
+	public function __construct( $value, Options $options ) {
+		if ( $options->get( 'hide_language_from_content_option' ) ) {
+			$this->enum = array( 1, 2, 3 );
+		}
+
+		parent::__construct( $value );
+	}
+
 	/**
 	 * Returns option key.
 	 *
@@ -51,7 +74,7 @@ class Force_Lang extends Abstract_Option {
 	protected function get_data_structure(): array {
 		return array(
 			'type' => 'integer',
-			'enum' => array( 0, 1, 2, 3 ),
+			'enum' => $this->enum,
 		);
 	}
 
--- a/include/Options/Business/Force_Lang.php
+++ b/include/Options/Business/Force_Lang.php
@@ -20,7 +20,7 @@ class Force_Lang extends Abstract_Option {
 	/**
 	 * @var array
 	 */
-	private $enum = array( 0, 1, 2, 3 );
+	private $enum = array( 1, 2, 3 );
 
 	/**
 	 * Constructor.
@@ -31,8 +31,8 @@ class Force_Lang extends Abstract_Option {
 	 * @param Options $options All options.
 	 */
 	public function __construct( $value, Options $options ) {
-		if ( $options->get( 'hide_language_from_content_option' ) ) {
-			$this->enum = array( 1, 2, 3 );
+		if ( ! $options->get( 'hide_language_from_content_option' ) ) {
+			$this->enum = array( 0, 1, 2, 3 );
 		}
 
 		parent::__construct( $value );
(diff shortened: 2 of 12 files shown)


manooweb (changes requested):

settings/settings-url.php
```diff
@@ -51,18 +51,28 @@ public function __construct( &$polylang ) {
 	 * @return void
 	 */
 	protected function force_lang() {
-		?>
-		<p class="description"><?php esc_html_e( 'Some themes or plugins may not be fully compatible with the language defined by the content or by domains.', 'polylang' ); ?></p>
-		<label>
+		if ( ! empty( $this->options['hide_language_from_content_option'] ) ) {
+			// Hide the "Language is set from content" option.
+			?>
+			<p class="description"><?php esc_html_e( 'Some themes or plugins may not be fully compatible with the language defined by domains.', 'polylang' ); ?></p>
```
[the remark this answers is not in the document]
Screenfeed: Good thing it is not then :p
>with the language defined by domains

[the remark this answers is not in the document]
manooweb: 😴

install/install.php
```diff
@@ -87,20 +87,32 @@ public function wp_version_notice() {
 	 * @return array
 	 */
 	public static function get_default_options() {
+		$show_option = defined( 'PLL_SHOW_LANGUAGE_FROM_CONTENT_OPTION' ) && PLL_SHOW_LANGUAGE_FROM_CONTENT_OPTION;
+		/**
+		 * Filters whether the "The language is set from content" option should be shown on new installations.
+		 * This is only available during Polylang's first activation.
+		 *
+		 * @since 3.7
```
manooweb, line 95: FYI, as there is no milestone 3.7 now and the issue is in 3.6 backlog, I moved the PR in 3.6 project

manooweb (approved):

include/Options/Business/Hide_Language_From_Content_Option.php
Hug0-Drelon, line 1: Why an option to store how the UI should behave instead of a transient for instance? Or even unregister the options value related to "set from content"?
In my mind, Polylang options are here to tell how the plugin should behave regarding the business logic.

[EDIT]: Should be discussed in depth with the team.

Screenfeed: Transients disappear, options stay :)

Screenfeed: [606d6f6](https://github.com/polylang/polylang/pull/1431/commits/606d6f68055ee65d42ac5f8a21cf7708390365da) removed the choice `0` from `force_lang`'s enum.

include/Options/Business/Force_Lang.php
```diff
@@ -6,15 +6,38 @@
...
  * Class defining the "Determine how the current language is defined" option.
+ * /!\ Constructor depends on `hide_language_from_content_option`: this option must be set AFTER `hide_language_from_content_option`.
  *
  * @since 3.7
  */
 class Force_Lang extends Abstract_Option {
+	/**
+	 * @var array
+	 */
+	private $enum = array( 0, 1, 2, 3 );
```
Hug0-Drelon, line 23: i'd do the opposite and use `array( 1, 2, 3 )` as default.
Since it'll be the default behavior of Polylang in 3.7.

include/Options/Abstract_Option.php
```diff
@@ -52,9 +52,9 @@ abstract class Abstract_Option {
...
 	 * @since 3.7
 	 *
-	 * @param mixed $value Optional. Option value.
+	 * @param mixed $value Option value. Use `null` to set the default value.
 	 */
-	public function __construct( $value = null ) {
+	public function __construct( $value ) {
 		$this->errors = new WP_Error();
 
 		if ( ! isset( $value ) ) {
```
Hug0-Drelon, line 60:
```suggestion
		if ( is_null( $value ) ) {
```
More accurate and descriptive according to your change, don't you agree?

include/Options/Business/Force_Lang.php
```diff
@@ -6,15 +6,38 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.7
+	 *
+	 * @param mixed   $value   Option value. Use `null` to set the default value.
+	 * @param Options $options All options.
+	 */
+	public function __construct( $value, Options $options ) {
+		if ( $options->get( 'hide_language_from_content_option' ) ) {
```
Hug0-Drelon, line 34: Could the condition be something like:
```php
if ( PLL_Install::should_hide_language_from_content_option() && 0 !== (int) $value ) {
    // Then remove "set from content" value in force lang.
}
``` 
```

Hug0-Drelon (changes requested):

include/Options/Business/Force_Lang.php
Screenfeed: Done in [beb70ec](https://github.com/polylang/polylang/pull/1431/commits/beb70ecf2d0c380bf2028bbb8210e97f59903ce5).

Chouby (changes requested):
I would simplify the whole thing by using a totally separate option, let say `pll_language_from_content_enabled`, thus:
- We don't expose the option together with the others as it's not supposed to
- We don't provide a constant or additional filter as there are already existing, ex: `pre_option_pll_language_from_content_enabled`.
- We don't need to modify `Abstract_Option`

Inspiration source: WordPress option `link_manager_enabled`

Chouby:
Closed in favor of #1517

PR #1434, opened 2024-03-01
Adds sub-levels attribute keys support in wpml config

Marianne380:
See https://github.com/polylang/polylang-pro/issues/1933

Adds a `get_field_attributes()` method to handle block sub-attributes in the **wpml-config.xml** file.

### DIFF
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -597,13 +597,11 @@ class PLL_WPML_Config {
 	 * @since 3.6
 	 *
 	 * @param  SimpleXMLElement $field A XML node.
-	 * @param  array            $attrs The attributes.
 	 * @return array An array of attributes.
 	 *
-	 * @phpstan-param array<non-empty-string, array|true> $attrs
 	 * @phpstan-return array<non-empty-string, array|true>
 	 */
-	private function get_field_attributes( SimpleXMLElement $field, array $attrs = array() ): array {
+	private function get_field_attributes( SimpleXMLElement $field ): array {
 		$name = $this->get_field_attribute( $field, 'name' );
 
 		if ( '' === $name ) {
@@ -616,25 +614,19 @@ class PLL_WPML_Config {
 			return array( $name => true );
 		}
 
-		if ( ! isset( $attrs[ $name ] ) || ! is_array( $attrs[ $name ] ) ) {
-			$attrs[ $name ] = array();
-		}
+		$sub_attributes = array();
 
 		foreach ( $children as $child ) {
-			$sub = $this->get_field_attributes( $child, $attrs[ $name ] );
+			$sub = $this->get_field_attributes( $child );
 
 			if ( empty( $sub ) ) {
 				continue;
 			}
 
-			$attrs[ $name ] = array_merge( $attrs[ $name ], $sub );
+			$sub_attributes[ $name ] = array_merge( $sub_attributes[ $name ] ?? array(), $sub );
 		}
 
-		if ( empty( $attrs[ $name ] ) ) {
-			unset( $attrs[ $name ] );
-		}
-
-		return $attrs;
+		return $sub_attributes;
 	}
 
 	/**
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -576,7 +576,7 @@ class PLL_WPML_Config {
 	}
 
 	/**
-	 * Gets attributes values.
+	 * Gets attributes values recursively.
 	 *
 	 * @since 3.6
 	 *
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -484,17 +484,18 @@ class PLL_WPML_Config {
 							break;
 					}
 
-					if ( '' === $rule || ( is_array( $rule ) && empty( array_filter( $rule ) ) ) ) {
-						continue;
-					}
-
 					if ( is_array( $rule ) ) {
+						$rule = array_filter( $rule );
+						if ( empty( $rule ) ) {
+							continue;
+						}
+						
 						if ( isset( $parsing_rules[ $child_tag ][ $block_name ] ) ) {
 							$parsing_rules[ $child_tag ][ $block_name ] = array_merge( $parsing_rules[ $child_tag ][ $block_name ], $rule );
 						} else {
 							$parsing_rules[ $child_tag ][ $block_name ] = $rule;
 						}
-					} else {
+					} elseif ( '' !== $rule ) {
 						$parsing_rules[ $child_tag ][ $block_name ][] = $rule;
 					}
 				}
--- a/modules/wpml/wpml-config.php
+++ b/modules/wpml/wpml-config.php
@@ -583,7 +583,7 @@ class PLL_WPML_Config {
 	 * @param  SimpleXMLElement $field A XML node.
 	 * @param  array            $attrs The attributes.
 	 * @param  bool             $first Tells if this is the first time in the function..
-	 * @return array|string
+	 * @return array|string An array if there are sub-attributes, otherwise a string.
 	 */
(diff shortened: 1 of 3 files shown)


modules/wpml/wpml-config.php
```diff
@@ -565,6 +575,40 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
 		return trim( (string) $attributes[ $attribute_name ] );
 	}
 
+	/**
+	 * Gets attributes values.
```
Hug0-Drelon, line 579:
```suggestion
	 * Gets attributes values recursively.
```

modules/wpml/wpml-config.php
```diff
@@ -565,6 +575,40 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
...
+				if ( ! isset( $attrs[ $name ] ) || ! is_array( $attrs[ $name ] ) ) {
+					$attrs[ $name ] = array();
+				}
+				$attrs[ $name ] = $this->get_field_attributes( $child, $attrs[ $name ], false );
+			}
+		} else {
+			$attrs[ $name ] = true;
+		}
+
+		return $attrs;
```
Hug0-Drelon, line 609:
```suggestion
                 if ( 1 === count( $attrs ) ) {
                     return reset( $attrs );
                 }
                 
		return $attrs;
```
Does it work? If so it will make `$first` parameter useless.
My point is to simplify the method and make it truly recursive.

modules/wpml/wpml-config.php
```diff
@@ -565,6 +575,40 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
...
 
+	/**
+	 * Gets attributes values.
+	 *
+	 * @since 3.6
+	 *
+	 * @param  SimpleXMLElement $field A XML node.
+	 * @param  array            $attrs The attributes.
+	 * @param  bool             $first Tells if this is the first time in the function..
+	 * @return array|string
```
Hug0-Drelon, line 586: Maybe explain when we get a string and when we get an array.

Hug0-Drelon (changes requested):

modules/wpml/wpml-config.php
Marianne380: No, because it won't return the right values for the sub-attributes.

Hug0-Drelon: My bad, this should work:
```php
/**
	 * Gets attributes values recursively.
	 *
	 * @since 3.6
	 *
	 * @param  SimpleXMLElement $field A XML node.
	 * @param  array            $attrs The attributes.
	 * @return array|string An array if there are sub-attributes, otherwise a string.
	 */
	private function get_field_attributes( SimpleXMLElement $field, array $attrs = array() ) {
		$name = $this->get_field_attribute( $field, 'name' );

		$children = $field->children();

		if ( 0 === count( $children ) ) {
			// No sub-attributes, don't go further and just return the attribute's value.
			return $name;
		}

		if ( count( $children ) ) {
			foreach ( $children as $child ) {
				if ( ! isset( $attrs[ $name ] ) || ! is_array( $attrs[ $name ] ) ) {
					$attrs[ $name ] = array();
				}

				$sub = $this->get_field_attributes( $child, $attrs[ $name ] );

				if ( is_string( $sub ) ) {
					$sub = array( $sub => true );
				}

				$attrs[ $name ] = array_merge( $attrs[ $name ], $sub );
			}
		} else {
			$attrs[ $name ] = true;
		}

		return $attrs;
	}
```
What do you think?

modules/wpml/wpml-config.php
Hug0-Drelon, line 1: Sorry my suggestion was not perfect, it remains one last PHPStan error.

Marianne380: Yes I already fix it.

Hug0-Drelon (approved):

modules/wpml/wpml-config.php
```diff
@@ -480,11 +480,21 @@ protected function extract_blocks_parsing_rules() {
...
+						continue;
+					}
+
+					if ( is_array( $rule ) ) {
+						if ( isset( $parsing_rules[ $child_tag ][ $block_name ] ) ) {
+							$parsing_rules[ $child_tag ][ $block_name ] = array_merge( $parsing_rules[ $child_tag ][ $block_name ], $rule );
+						} else {
+							$parsing_rules[ $child_tag ][ $block_name ] = $rule;
+						}
+					} else {
```
Chouby, line 497: To avoid the same test 2 times:
```suggestion
					if ( is_array( $rule ) ) {
						$rule = array_filter( $rule );
						if ( empty( $rule ) ) {
							continue;
						}
						
						if ( isset( $parsing_rules[ $child_tag ][ $block_name ] ) ) {
							$parsing_rules[ $child_tag ][ $block_name ] = array_merge( $parsing_rules[ $child_tag ][ $block_name ], $rule );
						} else {
							$parsing_rules[ $child_tag ][ $block_name ] = $rule;
						}
					} elseif ( '' !== $rule ) {
```

Chouby (changes requested):

modules/wpml/wpml-config.php
```diff
@@ -582,32 +598,42 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
...
+			return array( $name => true );
 		}
 
-		foreach ( $children as $child ) {
-			if ( ! isset( $attrs[ $name ] ) || ! is_array( $attrs[ $name ] ) ) {
-				$attrs[ $name ] = array();
-			}
+		if ( ! isset( $attrs[ $name ] ) || ! is_array( $attrs[ $name ] ) ) {
+			$attrs[ $name ] = array();
+		}
```
Hug0-Drelon, line 621:
```suggestion
		$sub_attributes = array();
		if ( isset( $attrs[ $name ] ) && is_array( $attrs[ $name ] ) ) {
			$sub_attributes = $attrs[ $name ];
		}
```
So after there is no need to unset `$attrs[ $name ]` at the end.

Hug0-Drelon (changes requested):

modules/wpml/wpml-config.php
Hug0-Drelon: You could check `isset( $attrs[ $name ] )` and then do the `array_merge()` or create it otherwise.
But I let you see what you prefer.

Screenfeed: I'll keep it as it is, what you suggest complicates things actually because we have to do the `if ( isset() && is_array() ) {` twice.

Screenfeed: @Hug0-Drelon I think I find a way to simplify it with [73298ee](https://github.com/polylang/polylang/pull/1434/commits/73298eebb39035a47d1a3a1d4afe334830f136d2).

modules/wpml/wpml-config.php
```diff
@@ -477,15 +493,25 @@ protected function extract_blocks_parsing_rules() {
...
+							$rule = $this->get_field_attributes( $child );
+
+							if ( empty( $rule ) ) {
+								break;
+							}
 
-					if ( '' !== $rule ) {
-						$parsing_rules[ $child_tag ][ $block_name ][] = $rule;
+							if ( isset( $parsing_rules['key'][ $block_name ] ) ) {
+								$parsing_rules['key'][ $block_name ] = array_merge( $parsing_rules['key'][ $block_name ], $rule );
```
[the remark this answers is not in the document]
Screenfeed: @Chouby Here we're looping through `$block->children()`. So, for example, through the 1st level of `key` tags here:
```xml
<gutenberg-block type="my-plugin/my-block" translate="1">
	<key name="headingTitle" />
	<key name="first-level">
		<key name="second-level" />
	</key>
</gutenberg-block>
```

1. We don't need to merge recursively: 1st lap we add `headingTitle`, 2nd lap we merge `first-level` and its children.
2. Yes we want to merge. If we overwrite, we will remove `headingTitle` when adding `first-level`.

PR #1436, opened 2024-03-01
Update post.js – fix #1435

### LINKED ISSUE #1435: JS-Error after QuickEdit: Uncaught TypeError: form.classList is undefined
## Prerequisites

1. Quick edit a post or page.
2. Hit "Update"
3. An `Uncaught TypeError` occurs in the js console

grafik

When I follow those steps, I see an `Uncaught TypeError` in the JS console.

I was able to track down the issue to these lines:
https://github.com/polylang/polylang/blob/17fe4aabca9b4d91f544038ea8c8ed601b3a4958/js/src/post.js#L27-L28

When I `console.log(mutation.addedNodes)` the first node is a text node containing two tabs.
grafik

mcguffin:
Fixes #1435

Make sure only actual Element Nodes are processed.

### DIFF
--- a/js/src/post.js
+++ b/js/src/post.js
@@ -24,7 +24,8 @@ jQuery(
 	function ( $ ) {
 		const handleQuickEditInsertion = ( mutationsList ) => {
 			for ( const mutation of mutationsList ) {
-				const form = mutation.addedNodes[0];
+				const addedNodes = Array.from( mutation.addedNodes ).filter( el => el.nodeType === Node.ELEMENT_NODE )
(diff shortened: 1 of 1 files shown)


manooweb (approved):
Good catch and thanks again @mcguffin.

FYI, this script only necessary at the quick edit initialization. When we update the quick edit other DOM mutations are done by WordPress and we don't need to do anything.
You're right as `classList` is an Element property we need to ensure that the node we want to check is a type of DOM element and therefore avoid the javascript error you reported.

PR #1439, opened 2024-03-05
Fix possible warning in view-translations-post.php

Chouby:
Reported by https://wordpress.org/support/topic/polylang-warning-home-page-issue-2/

```
Warning: Attempt to read property “ID” on null in /home/6/b/businessaccessin/www/wp-content/plugins/polylang/admin/view-translations-post.php on line 59
```

My guess is that it could occur if we get a translation post id but the post doesn't exist in the DB. 
While on it, I took the opportunity to reformat the old code in the file (renamed variables with more meaningful names, strict comparisons, type consistency).

### DIFF
--- a/admin/view-translations-post.php
+++ b/admin/view-translations-post.php
@@ -6,31 +6,32 @@
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly
+	exit; // Don't access directly.
 }
 ?>
 <p><strong><?php esc_html_e( 'Translations', 'polylang' ); ?></strong></p>
 <table>
 	<?php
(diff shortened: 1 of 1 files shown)


admin/view-translations-post.php
```diff
@@ -6,31 +6,32 @@
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly
+	exit; // Don't access directly.
```
manooweb, line 9: As you took the opportunity to rewrite some other things, you can also rewrite this condition as we decided `defined( 'ABSPATH' ) || exit;`.

manooweb (changes requested):

Screenfeed (approved):

PR #1443, opened 2024-03-08
Activates PHPstan tool on PHP views

manooweb:
Same as https://github.com/polylang/polylang-pro/pull/2077 initiated on Polylang Pro

We can document `$this` outside a class since PHPStan [0.12.30](https://github.com/phpstan/phpstan/releases/tag/0.12.30) and completed in PHPStan [0.12.67](https://github.com/phpstan/phpstan/releases/tag/0.12.67). See complete discussion in https://github.com/phpstan/phpstan/issues/351

However it seems there is still an issue if this class has private or protected properties See However it seems there still has an issue if this class has private or protected properties.

It explains why I chose to assign these properties as local variable and document them in views instead of `$this` espacially in the wizard module.

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+defined( 'ABSPATH' ) || exit;
+
 /**
  * Manages filters and actions related to the classic editor
  *
@@ -15,7 +17,7 @@ class PLL_Admin_Classic_Editor {
 	public $model;
 
 	/**
-	 * @var PLL_Admin_Links|null
+	 * @var PLL_Admin_Links
 	 */
 	public $links;
 
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+defined( 'ABSPATH' ) || exit;
+
 /**
  * Manages filters and actions related to terms on admin side
  *
@@ -15,7 +17,7 @@ class PLL_Admin_Filters_Term {
 	public $model;
 
 	/**
-	 * @var PLL_Admin_Links|null
+	 * @var PLL_Admin_Links
 	 */
 	public $links;
 
--- a/admin/view-translations-media.php
+++ b/admin/view-translations-media.php
@@ -4,17 +4,19 @@
  * Needs WP 3.5+
  *
  * @package Polylang
+ *
+ * @var PLL_Admin_Classic_Editor $this
+ * @var PLL_Language             $lang
+ * @var int                      $post_ID
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly
-}
+defined( 'ABSPATH' ) || exit; // Don't access directly
 ?>
 <p><strong><?php esc_html_e( 'Translations', 'polylang' ); ?></strong></p>
 <table>
 	<?php
 	foreach ( $this->model->get_languages_list() as $language ) {
-		if ( $language->term_id == $lang->term_id ) {
+		if ( $language->term_id === $lang->term_id ) {
 			continue;
 		}
 		?>
@@ -22,12 +24,13 @@ if ( ! defined( 'ABSPATH' ) ) {
 			<td class = "pll-media-language-column"><span class = "pll-translation-flag"><?php echo $language->flag; // phpcs:ignore WordPress.Security.EscapeOutput ?></span><?php echo esc_html( $language->name ); ?></td>
 			<td class = "pll-media-edit-column">
 				<?php
-				if ( ( $translation_id = $this->model->post->get_translation( $post_ID, $language ) ) && $translation_id !== $post_ID ) {
+				$translation_id = $this->model->post->get_translation( $post_ID, $language );
+				if ( ! empty( $translation_id ) && $translation_id !== $post_ID ) {
 					// The translation exists
 					printf(
 						'<input type="hidden" name="media_tr_lang[%s]" value="%d" />',
 						esc_attr( $language->slug ),
-						esc_attr( $translation_id )
+						esc_attr( (string) $translation_id )
 					);
 					echo $this->links->edit_post_translation_link( $translation_id ); // phpcs:ignore WordPress.Security.EscapeOutput
 				} else {
--- a/admin/view-translations-term.php
+++ b/admin/view-translations-term.php
@@ -3,11 +3,14 @@
  * Displays the translations fields for terms
  *
  * @package Polylang
+ *
+ * @var PLL_Admin_Filters_Term $this
+ * @var PLL_Language           $lang
+ * @var string                 $taxonomy
+ * @var string                 $post_type
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly
-}
+defined( 'ABSPATH' ) || exit; // Don't access directly
 
 if ( isset( $term_id ) ) {
 	// Edit term form ?>
@@ -31,7 +34,7 @@ else {
 
 		// Look for any existing translation in this language
 		// Take care not to propose a self link
-		$translation = 0;
+		$translation = null;
 		if ( isset( $term_id ) && ( $translation_id = $this->model->term->get_translation( $term_id, $language ) ) && $translation_id != $term_id ) {
 			$translation = get_term( $translation_id, $taxonomy );
 		}
@@ -39,11 +42,14 @@ else {
 			$translation = get_term( $translation_id, $taxonomy );
 		}
 
+		$add_link = '';
+		$link     = '';
 		if ( isset( $term_id ) ) { // Do not display the add new link in add term form ( $term_id not set !!! )
-			$link = $add_link = $this->links->new_term_translation_link( $term_id, $taxonomy, $post_type, $language );
+			$link = $this->links->new_term_translation_link( $term_id, $taxonomy, $post_type, $language );
+			$add_link = $link;
 		}
 
-		if ( $translation ) {
+		if ( $translation instanceof WP_Term ) {
 			$link = $this->links->edit_term_translation_link( $translation->term_id, $taxonomy, $post_type );
 		}
 		?>
@@ -75,8 +81,8 @@ else {
 					esc_attr( $language->slug ),
 					/* translators: accessibility text */
 					esc_html__( 'Translation', 'polylang' ),
-					( empty( $translation ) ? 0 : esc_attr( $translation->term_id ) ),
-					( empty( $translation ) ? '' : esc_attr( $translation->name ) ),
+					( ! $translation instanceof WP_Term ? 0 : esc_attr( (string) $translation->term_id ) ),
+					( ! $translation instanceof WP_Term ? '' : esc_attr( $translation->name ) ),
 					disabled( empty( $disabled ), false, false ),
 					esc_attr( $language->get_locale( 'display' ) ),
 					( $language->is_rtl ? 'rtl' : 'ltr' )
--- a/modules/wizard/view-wizard-page.php
+++ b/modules/wizard/view-wizard-page.php
@@ -5,11 +5,22 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var array[]    $steps {
+ *   List of steps.
+ *
+ *     @type string   $name    I18n string which names the step.
+ *     @type callable $view    The callback function use to display the step content.
+ *     @type callable $handler The callback function use to process the step after it is submitted.
+ *     @type array    $scripts List of scripts handle needed by the step.
+ *     @type array    $styles  The list of styles handle needed by the step.
+ * }
+ * @var string   $current_step Current step.
+ * @var string[] $styles       List of wizard page styles.
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly.
-}
+defined( 'ABSPATH' ) || exit; // Don't access directly.
+
 $admin_body_class = array( 'pll-wizard', 'wp-core-ui' );
 if ( is_rtl() ) {
 	$admin_body_class[] = 'rtl';
@@ -32,9 +43,8 @@ if ( is_rtl() ) {
 		<script>
 			var ajaxurl = '<?php echo esc_url( admin_url( 'admin-ajax.php', 'relative' ) ); ?>';
 		</script>
-		<?php do_action( 'admin_enqueue_scripts' ); ?>
-		<?php wp_print_scripts( $this->steps[ $this->step ]['scripts'] ); ?>
-		<?php wp_print_styles( array_merge( $this->styles, $this->steps[ $this->step ]['styles'] ) ); ?>
+		<?php wp_print_scripts( $steps[ $current_step ]['scripts'] ); ?>
+		<?php wp_print_styles( array_merge( $styles, $steps[ $current_step ]['styles'] ) ); ?>
 		<?php do_action( 'admin_head' ); ?>
 	</head>
 	<body class="<?php echo join( ' ', array_map( 'sanitize_key', $admin_body_class ) ); ?>">
@@ -46,10 +56,10 @@ if ( is_rtl() ) {
 		</h1>
 		<ol class="pll-wizard-steps">
 			<?php
-			foreach ( $this->steps as $step_key => $step ) {
-				$is_completed = array_search( $this->step, array_keys( $this->steps ), true ) > array_search( $step_key, array_keys( $this->steps ), true );
+			foreach ( $steps as $step_key => $step ) {
+				$is_completed = array_search( $step, array_keys( $steps ), true ) > array_search( $step_key, array_keys( $steps ), true );
 
-				if ( $step_key === $this->step ) {
+				if ( $step_key === $current_step ) {
 					?>
 					<li class="active"><?php echo esc_html( $step['name'] ); ?></li>
 					<?php
@@ -72,15 +82,15 @@ if ( is_rtl() ) {
 			?>
 		</ol>
 		<div class="pll-wizard-content">
-			<form method="post" class="<?php echo esc_attr( "{$this->step}-step" ); ?>">
+			<form method="post" class="<?php echo esc_attr( "{$current_step}-step" ); ?>">
 				<?php
 				wp_nonce_field( 'pll-wizard', '_pll_nonce' );
 
-				if ( ! empty( $this->steps[ $this->step ]['view'] ) ) {
-					call_user_func( $this->steps[ $this->step ]['view'], $this );
+				if ( ! empty( $steps[ $current_step ]['view'] ) ) {
+					call_user_func( $steps[ $current_step ]['view'] );
 				}
 				?>
-				<?php if ( 'last' !== $this->step ) : ?>
+				<?php if ( 'last' !== $current_step ) : ?>
 				<p class="pll-wizard-actions step">
 					<button
 						type="submit"
@@ -95,7 +105,7 @@ if ( is_rtl() ) {
 			</form>
 		</div>
 		<div class="pll-wizard-footer">
-			<?php if ( 'last' !== $this->step ) : ?>
+			<?php if ( 'last' !== $current_step ) : ?>
 				<a
 					class="pll-wizard-footer-links"
 					href="<?php echo esc_url( admin_url() ); ?>"
--- a/modules/wizard/view-wizard-step-home-page.php
+++ b/modules/wizard/view-wizard-step-home-page.php
@@ -5,54 +5,60 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var PLL_Admin_Model $model
+ * @var array           $options
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly.
-}
+defined( 'ABSPATH' ) || exit; // Don't access directly.
 
-$languages = $this->model->get_languages_list();
-$default_language = count( $languages ) > 0 ? $this->options['default_lang'] : null;
+$languages = $model->get_languages_list();
+$default_language = count( $languages ) > 0 ? $options['default_lang'] : null;
 $home_page_id = get_option( 'page_on_front' );
-$translations = $this->model->post->get_translations( $home_page_id );
+$home_page_id = is_scalar( $home_page_id ) ? (int) $home_page_id : 0;
+$translations = $model->post->get_translations( $home_page_id );
 $untranslated_languages = array();
 $home_page = $home_page_id > 0 ? get_post( $home_page_id ) : null;
-$home_page_language = $this->model->post->get_language( $home_page_id );
+$home_page_language = $model->post->get_language( $home_page_id );
+
+if ( empty( $home_page ) ) {
+	return;
+}
 
 foreach ( $languages as $language ) {
-	if ( ! $this->model->post->get( $home_page_id, $language ) ) {
+	if ( ! $model->post->get( $home_page_id, $language ) ) {
 		$untranslated_languages[] = $language;
 	}
 }
 ?>
-<input type="hidden" name="home_page" value="<?php echo esc_attr( $home_page->ID ); ?>" />
+<input type="hidden" name="home_page" value="<?php echo esc_attr( (string) $home_page->ID ); ?>" />
 <input type="hidden" name="home_page_title" value="<?php echo esc_attr( $home_page->post_title ); ?>" />
-<?php if ( false !== $home_page_language ) : ?>
+<?php if ( ! empty( $home_page_language ) ) : ?>
 	<input type="hidden" name="home_page_language" value="<?php echo esc_attr( $home_page_language->slug ); ?>" />
+	<h2><?php esc_html_e( 'Homepage', 'polylang' ); ?></h2>
+	<p>
+		<?php
+			printf(
+				/* translators: %s is the post title of the front page */
+				esc_html__( 'You defined this page as your static homepage: %s.', 'polylang' ),
+				'<strong>' . esc_html( $home_page->post_title ) . '</strong>'
+			);
+		?>
+		<br />
+		<?php
+			printf(
+				/* translators: %s is the language of the front page ( flag, native name and locale ) */
+				esc_html__( 'Its language is : %s.', 'polylang' ),
+				$home_page_language->flag . ' <strong>' . esc_html( $home_page_language->name ) . ' ' . esc_html( $home_page_language->locale ) . '</strong>' //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			)
+		?>
+		<br />
+		<?php esc_html_e( 'For your site to work correctly, this page must be translated in all available languages.', 'polylang' ); ?>
+	</p>
+	<p>
+		<?php esc_html_e( 'After the pages is created, it is up to you to put the translated content in each page linked to each language.', 'polylang' ); ?>
+	</p>
 <?php endif; ?>
-<h2><?php esc_html_e( 'Homepage', 'polylang' ); ?></h2>
-<p>
-	<?php
-		printf(
-			/* translators: %s is the post title of the front page */
-			esc_html__( 'You defined this page as your static homepage: %s.', 'polylang' ),
-			'<strong>' . esc_html( $home_page->post_title ) . '</strong>'
-		);
-		?>
-	<br />
-	<?php
-		printf(
-			/* translators: %s is the language of the front page ( flag, native name and locale ) */
-			esc_html__( 'Its language is : %s.', 'polylang' ),
-			$home_page_language->flag . ' <strong>' . esc_html( $home_page_language->name ) . ' ' . esc_html( $home_page_language->locale ) . '</strong>' //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		)
-		?>
-	<br />
-	<?php esc_html_e( 'For your site to work correctly, this page must be translated in all available languages.', 'polylang' ); ?>
-</p>
-<p>
-	<?php esc_html_e( 'After the pages is created, it is up to you to put the translated content in each page linked to each language.', 'polylang' ); ?>
-</p>
 <?php if ( $translations ) : ?>
 <table id="translated-languages" class="striped">
 	<thead>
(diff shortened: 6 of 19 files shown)


modules/wizard/view-wizard-page.php
```diff
@@ -5,11 +5,22 @@
...
  *
  * @since 2.7
+ *
+ * @var array[]    $steps {
+ *   List of steps.
+ *
+ *     @type string   $name    I18n string which names the step.
+ *     @type callable $view    The callback function use to display the step content.
+ *     @type callable $handler The callback function use to process the step after it is submitted.
+ *     @type array    $scripts List of scripts handle needed by the step.
```
Hug0-Drelon, line 15:
```suggestion
 *     @type array    $scripts List of script handles needed by the step.
```

modules/wizard/view-wizard-page.php
```diff
@@ -5,11 +5,22 @@
...
+ *     @type array    $styles  The list of styles handle needed by the step.
```
Hug0-Drelon, line 16:
```suggestion
 *     @type array    $styles  The list of style handles needed by the step.
```

admin/view-translations-media.php
```diff
@@ -4,30 +4,33 @@
...
+ *
+ * @var PLL_Admin_Classic_Editor $this
+ * @var PLL_Language             $lang
+ * @var int                      $post_ID
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly
-}
+defined( 'ABSPATH' ) || exit; // Don't access directly
```
Marianne380, line 13:
```suggestion
defined( 'ABSPATH' ) || exit; // Don't access directly.
```

Do we add this comment every time or not? 
If so, add it where it's missing.

modules/wizard/view-wizard-step-media.php
```diff
@@ -7,12 +7,10 @@
...
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly.
-}
+defined( 'ABSPATH' ) || exit; // Don't access directly.
 
 $default_options = PLL_Install::get_default_options();
-$options = wp_parse_args( get_option( 'polylang' ), $default_options );
+$options = wp_parse_args( (array) get_option( 'polylang' ), $default_options );
```
Hug0-Drelon, line 13: Shouldn't we pass Polylang options to this view (like the other Wizard views)  instead of using `get_option()`?

modules/wizard/wizard.php
```diff
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+defined( 'ABSPATH' ) || exit;
```
Hug0-Drelon, line 6: Not needed if we stick to WordPress recommendations.

Hug0-Drelon (changes requested):

admin/view-translations-media.php
szepeviktor: This is a protection against direct execution.
But files without side effects do not need it. A file with only `class Name { ... }` is parsed but execution only involves defining a class, no instantiation.

admin/admin-filters-term.php
```diff
@@ -15,7 +17,7 @@ class PLL_Admin_Filters_Term {
 	public $model;
 
 	/**
-	 * @var PLL_Admin_Links|null
+	 * @var PLL_Admin_Links
```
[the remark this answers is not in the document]
manooweb: 👍
I just fixed what PHPStan reported.

It seems we can also remove `|null` on `$filter_lang` and `$pref_lang` properties and therefore remove one ignored error in phpstan-baseline.neon file.

Done in 2170aece36c83653575d7d2050389d2f6a80ba95

admin/view-translations-media.php
manooweb: As @screenfeed said to me: we also put it when a class extends another class, implements a interface or use a trait. If there is a fatal error it could have a full path disclosure which display the webroot path.

For consistency we decided to put it everywhere not to forgot it and if a class definition change in the future.

In this PR, I put it only in file I needed to modify. The goal isn't to modify all the polylang files.

admin/admin-classic-editor.php
```diff
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+defined( 'ABSPATH' ) || exit;
```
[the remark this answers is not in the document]
manooweb: See https://github.com/polylang/polylang/pull/1443#discussion_r1559022178

admin/admin-filters-term.php
```diff
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+defined( 'ABSPATH' ) || exit;
```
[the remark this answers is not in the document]
manooweb: See https://github.com/polylang/polylang/pull/1443#discussion_r1559022178

modules/wizard/wizard.php
manooweb: See https://github.com/polylang/polylang/pull/1443#discussion_r1559022178

admin/view-translations-media.php
manooweb: > Do we add this comment every time or not? If so, add it where it's missing.

Concerning the comment. I modified this line manually and the comment was here before the modification and then I kept it like this. The comment is useless. We can remove it everywhere.

Done for this file in 7eb6ea6e27ab6daf91e444750528577a1b7afa4e

modules/wizard/view-wizard-step-media.php
manooweb: Done in 7eb6ea6e27ab6daf91e444750528577a1b7afa4e

szepeviktor (approved):

manooweb (commented):
Leave an reply in a review comment

szepeviktor:
Very nice job 🎉

modules/wizard/wizard.php
```diff
@@ -274,6 +276,10 @@ public function wizard_notice() {
 	 */
 	public function display_wizard_page() {
 		set_current_screen( 'pll-wizard' );
+		do_action( 'admin_enqueue_scripts' );
```
Hug0-Drelon, line 279: Where does this `do_action()` come from?

manooweb: Keep in mind that the wizard is outside a WordPress admin context. So we need to enqueue our scripts ourselves.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -5,54 +5,60 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var PLL_Admin_Model $model   A reference to `PLL_Model` instance.
```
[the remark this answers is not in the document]
manooweb: 🤔 See 
- https://github.com/polylang/polylang/blob/also-run-phpstan-on-views/modules/wizard/wizard.php#L16-L20
- https://github.com/polylang/polylang/blob/also-run-phpstan-on-views/modules/wizard/wizard.php#L65
- https://github.com/polylang/polylang/blob/also-run-phpstan-on-views/modules/wizard/wizard.php#L757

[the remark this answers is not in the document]
Hug0-Drelon: There is no reason to do that, moreover if the options are referenced separately.
We should remove `&`.

Also, you used `$model = $this->model;` everywhere before including corresponding views. Initializing a new variable and assigning it a referenced value dereferences it.
See https://onlinephp.io/c/3dab8

modules/wizard/wizard.php
Hug0-Drelon: I understand that, but I can find where it's moved from.

modules/wizard/view-wizard-step-home-page.php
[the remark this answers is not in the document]
manooweb: As you want 😁 `Reference` word is confusing and there is no comment about it here https://github.com/polylang/polylang-pro/pull/2077/files#diff-a5a75717f08504f889c9108147b77aa89721403362da29e1f9196376fca4561a and here https://github.com/polylang/polylang-pro/pull/2077/files#diff-01c90a42bb33395dca5108c73e378f0a5db786cc602fc3878355458613d04cba

modules/wizard/wizard.php
manooweb: It was in `view-wizard-page.php` https://github.com/polylang/polylang/pull/1443/files#diff-3a4ba45b721c33af7e4372a25537b6a042cd2a5aa3c5694a0e2ec26a84b3b623

https://github.com/polylang/polylang/blob/master/modules/wizard/view-wizard-page.php#L35

Hug0-Drelon: Thanks!
Indeed, way better your way.

modules/wizard/view-wizard-step-home-page.php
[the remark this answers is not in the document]
manooweb: Done in c6f7acde07f61a9cd5c3e982008f29cdd2ba36cf

modules/wizard/view-wizard-step-untranslated-contents.php
```diff
@@ -5,13 +5,13 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var PLL_Admin_Model $model A reference to `PLL_Model` instance.
```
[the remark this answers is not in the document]
manooweb: Done in c6f7acde07f61a9cd5c3e982008f29cdd2ba36cf

modules/wizard/view-wizard-step-media.php
```diff
@@ -5,11 +5,11 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var array $options
```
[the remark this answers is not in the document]
manooweb: Done in e8a431a8c1f97bcb2e18ac9108b9397d944b44de

admin/view-translations-media.php
```diff
@@ -4,30 +4,33 @@
...
-	exit; // Don't access directly
-}
+defined( 'ABSPATH' ) || exit;
 ?>
 <p><strong><?php esc_html_e( 'Translations', 'polylang' ); ?></strong></p>
 <table>
 	<?php
 	foreach ( $this->model->get_languages_list() as $language ) {
-		if ( $language->term_id == $lang->term_id ) {
+		if ( $language->term_id === $lang->term_id ) {
```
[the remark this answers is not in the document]
Hug0-Drelon: I see, nevermind.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -5,54 +5,60 @@
...
-			/* translators: %s is the language of the front page ( flag, native name and locale ) */
-			esc_html__( 'Its language is : %s.', 'polylang' ),
-			$home_page_language->flag . ' <strong>' . esc_html( $home_page_language->name ) . ' ' . esc_html( $home_page_language->locale ) . '</strong>' //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		)
+		<br />
+		<?php
+			printf(
+				/* translators: %s is the language of the front page ( flag, native name and locale ) */
+				esc_html__( 'Its language is : %s.', 'polylang' ),
+				$home_page_language->flag . ' <strong>' . esc_html( $home_page_language->name ) . ' ' . esc_html( $home_page_language->locale ) . '</strong>' //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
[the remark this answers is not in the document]
Hug0-Drelon: Indeed, `PLL_Language::$flag` stores a escaped string.

In another hand, since its a public property and not a getter that we're using, nothing ensures it's still safe here.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -5,54 +5,60 @@
...
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly.
-}
+defined( 'ABSPATH' ) || exit;
 
-$languages = $this->model->get_languages_list();
-$default_language = count( $languages ) > 0 ? $this->options['default_lang'] : null;
+$languages = $model->get_languages_list();
```
Hug0-Drelon, line 15:
```suggestion
$languages = $model->languages->get_list();
```

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -63,7 +69,10 @@
 	<tbody>
 	<?php
 	foreach ( array_keys( $translations ) as $lang ) {
-		$language = $this->model->get_language( $lang );
+		$language = $model->get_language( $lang );
```
Hug0-Drelon, line 72:
```suggestion
		$language = $model->languages->get( $lang );
```

modules/wizard/view-wizard-step-languages.php
```diff
@@ -5,14 +5,15 @@
...
  */
 
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly.
-}
+defined( 'ABSPATH' ) || exit;
 
-$existing_languages = $this->model->get_languages_list();
-$default_language   = count( $existing_languages ) > 0 ? $this->options['default_lang'] : null;
+$existing_languages = $model->get_languages_list();
```
Hug0-Drelon, line 15:
```suggestion
$existing_languages = $model->languages->get_list();
```

Hug0-Drelon (changes requested):

modules/wizard/view-wizard-step-home-page.php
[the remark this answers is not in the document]
manooweb: Yes! But this property contains HTML and we can't escape it here otherwise the image won't be displayed correctly, won't it?

Hug0-Drelon (approved):

admin/view-translations-media.php
```diff
@@ -4,30 +4,33 @@
...
 				<?php
-				if ( ( $translation_id = $this->model->post->get_translation( $post_ID, $language ) ) && $translation_id !== $post_ID ) {
+				$translation_id = $this->model->post->get_translation( $post_ID, $language );
+				if ( ! empty( $translation_id ) && $translation_id !== $post_ID ) {
 					// The translation exists
 					printf(
 						'<input type="hidden" name="media_tr_lang[%s]" value="%d" />',
 						esc_attr( $language->slug ),
-						esc_attr( $translation_id )
+						esc_attr( (string) $translation_id )
```
Chouby, line 33: To be consistent with `%d`.
```suggestion
						(int) $translation_id
```

admin/view-translations-term.php
```diff
@@ -31,19 +34,23 @@
...
 			$translation = get_term( $translation_id, $taxonomy );
 		}
 
+		$add_link = '';
+		$link     = '';
 		if ( isset( $term_id ) ) { // Do not display the add new link in add term form ( $term_id not set !!! )
-			$link = $add_link = $this->links->new_term_translation_link( $term_id, $taxonomy, $post_type, $language );
+			$link = $this->links->new_term_translation_link( $term_id, $taxonomy, $post_type, $language );
+			$add_link = $link;
 		}
```
Chouby, line 50: Let's simplify:
```php
		$add_link = isset( $term_id ) ? $this->links->new_term_translation_link( $term_id, $taxonomy, $post_type, $language ); // Do not display the add new link in add term form ($term_id not set).			
		$link = $add_link;

```

admin/view-translations-term.php
```diff
@@ -31,19 +34,23 @@
...
 
-		if ( $translation ) {
+		$is_existing_translation = $translation instanceof WP_Term;
```
Chouby, line 52: OK to introduce this new variable, I would however assign it a bit sooner just after https://github.com/polylang/polylang/pull/1443/files#diff-52adbff083959ccc3e35f99a6bc985248c93f6721de8577afafb12788baf697dR38-R43 and rename it to `$translation_exists`.

admin/view-translations-term.php
```diff
@@ -75,8 +82,8 @@
 					esc_attr( $language->slug ),
 					/* translators: accessibility text */
 					esc_html__( 'Translation', 'polylang' ),
-					( empty( $translation ) ? 0 : esc_attr( $translation->term_id ) ),
-					( empty( $translation ) ? '' : esc_attr( $translation->name ) ),
+					( ! $is_existing_translation ? '0' : esc_attr( (string) $translation->term_id ) ),
+					( ! $is_existing_translation ? '' : esc_attr( $translation->name ) ),
```
Chouby, line 86: This order is more convenient to read:
```suggestion
					$translation_exists ? (int) $translation->term_id : 0,
					$translation_exists ? esc_attr( $translation->name ) : '',
```
NB: The first line supposes that we use `%3$d` for the value.

modules/wizard/view-wizard-page.php
```diff
@@ -32,9 +43,8 @@
 		<script>
 			var ajaxurl = '<?php echo esc_url( admin_url( 'admin-ajax.php', 'relative' ) ); ?>';
 		</script>
-		<?php do_action( 'admin_enqueue_scripts' ); ?>
```
Chouby, line 35: Why removing this line?

modules/wizard/wizard.php
```diff
@@ -28,7 +30,7 @@ class PLL_Wizard {
 	/**
 	 * List of steps.
 	 *
-	 * @var array $steps {
+	 * @var array[] $steps {
```
Chouby, line 33: Not true according to the doc below:
```suggestion
	 * @var array $steps {
```

modules/wizard/view-wizard-page.php
```diff
@@ -46,10 +56,10 @@
 		</h1>
 		<ol class="pll-wizard-steps">
 			<?php
-			foreach ( $this->steps as $step_key => $step ) {
-				$is_completed = array_search( $this->step, array_keys( $this->steps ), true ) > array_search( $step_key, array_keys( $this->steps ), true );
+			foreach ( $steps as $step_key => $step ) {
+				$is_completed = array_search( $step, array_keys( $steps ), true ) > array_search( $step_key, array_keys( $steps ), true );
```
Chouby, line 60: `$this->step` is not `$step`:
```suggestion
				$is_completed = array_search( $current_step, array_keys( $steps ), true ) > array_search( $step_key, array_keys( $steps ), true );
```

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -5,54 +5,60 @@
...
+if ( empty( $home_page ) ) {
+	return;
+}
 
 foreach ( $languages as $language ) {
-	if ( ! $this->model->post->get( $home_page_id, $language ) ) {
+	if ( ! $model->post->get( $home_page_id, $language ) ) {
 		$untranslated_languages[] = $language;
 	}
 }
```
Chouby, line 32: Part of the code here should be in `display_step_home_page()` as several variables are not used in the html part.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -5,54 +5,60 @@
...
 ?>
-<input type="hidden" name="home_page" value="<?php echo esc_attr( $home_page->ID ); ?>" />
+<input type="hidden" name="home_page" value="<?php echo esc_attr( (string) $home_page->ID ); ?>" />
 <input type="hidden" name="home_page_title" value="<?php echo esc_attr( $home_page->post_title ); ?>" />
-<?php if ( false !== $home_page_language ) : ?>
+<?php if ( ! empty( $home_page_language ) ) : ?>
 	<input type="hidden" name="home_page_language" value="<?php echo esc_attr( $home_page_language->slug ); ?>" />
-<?php endif; ?>
```
Chouby, line 32: The code is changing here. Why?

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -63,7 +69,10 @@
 	<tbody>
 	<?php
 	foreach ( array_keys( $translations ) as $lang ) {
-		$language = $this->model->get_language( $lang );
+		$language = $model->languages->get( $lang );
+		if ( empty( $language ) ) {
+			continue;
+		}
```
Chouby, line 75: Is it possible to have an empty language here? If not just document:
```suggestion
		/** @var PLL_language */
		$language = $model->languages->get( $lang );
```

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -88,15 +97,15 @@
...
+	<?php if ( ! empty( $default_language ) ) : ?>
 		<caption><span class="icon-default-lang"></span> <?php esc_html_e( 'Default language', 'polylang' ); ?></caption>
 	<?php endif; ?>
 	<thead>
 		<?php if ( count( $untranslated_languages ) >= 1 ) : ?>
 			<tr>
 				<th><?php esc_html_e( 'We are going to prepare this page in', 'polylang' ); ?></th>
 			</tr>
-		<?php elseif ( false !== $home_page_language && count( $untranslated_languages ) <= 0 ) : ?>
+		<?php elseif ( ! empty( $home_page_language ) ) : ?>
```
Chouby, line 108: The code is changing here. Why?

modules/wizard/view-wizard-step-languages.php
```diff
@@ -5,14 +5,15 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var PLL_Admin_Model $model   `PLL_Admin_Model` instance.
+ * @var array           $options List of Polylang options.
```
Chouby, line 10: Shouldn't our new `Options` class instead or is it too soon and require many changes elsewhere?

modules/wizard/view-wizard-step-licenses.php
```diff
@@ -8,7 +8,7 @@
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Don't access directly.
+	exit;
 }
```
Chouby, line 12: Since you did it everywhere else:
```php
defined( 'ABSPATH' ) || exit;
```

Chouby (changes requested):

settings/settings.php
```diff
@@ -368,7 +372,7 @@ public static function redirect( $args = array() ) {
 	 *
 	 * @since 2.3
 	 *
-	 * @return string[] {
+	 * @return string[][] {
```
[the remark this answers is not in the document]
manooweb: The doc below is wrong. The returned value is really an array of array of string: first level array of code language and second level array of language properties.

modules/wizard/view-wizard-page.php
manooweb: This line has been moved to `PLL_Wizard::display_wizard_page()` to be able to document the local view variable `$styles` and because $this->styles is assigned on `PLL_Wizard::enqueue_scripts()`

See 
- https://github.com/polylang/polylang/blob/36ff6ffbec62dedc2e4dcfa6ed4ce96554c9927c/modules/wizard/wizard.php#L279
- https://github.com/polylang/polylang/blob/36ff6ffbec62dedc2e4dcfa6ed4ce96554c9927c/modules/wizard/wizard.php#L282
- https://github.com/polylang/polylang/blob/36ff6ffbec62dedc2e4dcfa6ed4ce96554c9927c/modules/wizard/view-wizard-page.php#L19

modules/wizard/view-wizard-step-home-page.php
manooweb: The `endif` has been moved down to cover the use of `$home_page_language`.
However it's a mistake only the corresponding printf function should be concerned by this condition.

modules/wizard/view-wizard-step-home-page.php
manooweb: The second condition is useless.
As `PLL_Translatable_Object::get_language()` returns `PLL_Language` or `false` do you prefer to keep the first condition as before?

modules/wizard/view-wizard-step-languages.php
manooweb: `$options` is now useless since using `WP_Syntex\Polylang\Model\Languages::get_default()` method.

modules/wizard/view-wizard-step-home-page.php
manooweb: `WP_Syntex\Polylang\Model\Languages::get()` could returned `PLL_Language` or `false` so PHPStan grumbles

```
  Line   modules\wizard\view-wizard-step-home-page.php
 ------ ------------------------------------------------------------------
  :75    Class PLL_Language referenced with incorrect case: PLL_language.
 ------ ------------------------------------------------------------------
```
even if this line is documented.

settings/settings.php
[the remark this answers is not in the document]
manooweb: Done in e5a048f54b8e5a44ca569d7564a1aa613e7c5354

admin/view-translations-term.php
manooweb: Done in d5f169185a4ef482de76b4ae8c4de6a944c5d879

admin/view-translations-term.php
manooweb: Done in e6e41a36cf9ff7fd0a30ec02e2b1c40b0e44eeb7 for `printf` pattern.

modules/wizard/view-wizard-step-licenses.php
manooweb: Done in e6e41a36cf9ff7fd0a30ec02e2b1c40b0e44eeb7 and on other missing views

admin/view-translations-term.php
manooweb: Done in ace25b6b7e487239403ab9f0369325565fec4153

modules/wizard/view-wizard-step-home-page.php
manooweb: Reverted in ace25b6b7e487239403ab9f0369325565fec4153 and adds another condition to protect the use of $home_page_language properties.

modules/wizard/view-wizard-step-languages.php
```diff
@@ -5,14 +5,15 @@
  * @package Polylang
  *
  * @since 2.7
+ *
+ * @var PLL_Admin_Model $model   `PLL_Admin_Model` instance.
```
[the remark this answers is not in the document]
manooweb: Done in 6d78833ed89cba3494137090c082361ed623b37a and for other views.

modules/wizard/view-wizard-step-languages.php
manooweb: Removed in 60f12a0f0250bbbaa4b1c5972142a572ee14ff18

modules/wizard/view-wizard-step-home-page.php
manooweb: Reverted in 0803da4d422f7b8b9ac56b0a1d9d14473bddf434

modules/wizard/wizard.php
```diff
@@ -274,6 +276,10 @@ public function wizard_notice() {
 	 */
 	public function display_wizard_page() {
 		set_current_screen( 'pll-wizard' );
+		do_action( 'admin_enqueue_scripts' );
+		$steps          = $this->steps;
+		$current_step   = $this->step;
```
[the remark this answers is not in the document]
manooweb: Done in 2304eb8953f50a3151834056d2dd4071c758e7a1

modules/wizard/view-wizard-step-home-page.php
manooweb: Done in 77e9faafdec597c021e8846ecbde2dd9cb6e21a7

modules/wizard/wizard.php
Chouby: Fixed by 3edbba947c867926dba47b49dc85b08041b1a019

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -5,29 +5,21 @@
...
-	if ( ! $this->model->post->get( $home_page_id, $language ) ) {
-		$untranslated_languages[] = $language;
-	}
-}
 ?>
-<input type="hidden" name="home_page" value="<?php echo esc_attr( $home_page->ID ); ?>" />
+<input type="hidden" name="home_page" value="<?php echo esc_attr( (string) $home_page->ID ); ?>" />
 <input type="hidden" name="home_page_title" value="<?php echo esc_attr( $home_page->post_title ); ?>" />
-<?php if ( false !== $home_page_language ) : ?>
+<?php if ( ! empty( $home_page_language ) ) : ?>
```
Chouby, line 22: Note sure that we should test this within the view. If everything went smoothly, this is defined and not empty. If something went wrong, displaying half of the view may just make things worse. 

The same for the default language. 

Maybe we could make tests before we add the  step.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -41,12 +33,14 @@
 		?>
 	<br />
 	<?php
+	if ( ! empty( $home_page_language ) ) {
```
Chouby, line 36: See above.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -88,15 +85,15 @@
 </table>
 <?php endif; ?>
 <table id="untranslated-languages" class="striped">
-	<?php if ( ! is_null( $default_language ) ) : ?>
+	<?php if ( ! empty( $default_language ) ) : ?>
```
Chouby, line 88: See above.

modules/wizard/view-wizard-step-home-page.php
```diff
@@ -88,15 +85,15 @@
...
 		<caption><span class="icon-default-lang"></span> <?php esc_html_e( 'Default language', 'polylang' ); ?></caption>
 	<?php endif; ?>
 	<thead>
 		<?php if ( count( $untranslated_languages ) >= 1 ) : ?>
 			<tr>
 				<th><?php esc_html_e( 'We are going to prepare this page in', 'polylang' ); ?></th>
 			</tr>
-		<?php elseif ( false !== $home_page_language && count( $untranslated_languages ) <= 0 ) : ?>
+		<?php elseif ( false !== $home_page_language ) : ?>
```
Chouby, line 96: See above

Chouby (changes requested):

modules/wizard/view-wizard-step-home-page.php
manooweb: Normally, everything goes smoothly because the step before this one is the assignment of a language to the whole content including the home page.

I'm going to see the PHPStan behaviour when I code an early return in the method as I did for the `$home_page`
https://github.com/polylang/polylang/blob/da9ac88efbdd7cd942872833cb3ba9d74b49fd36/modules/wizard/wizard.php#L770-L772

I think it's a edge case if a user remind the URL of this wizard step and start at this point when Polylang isn't yet properly set up.

manooweb: Done in a2a3184aa0d7e724d30c60e41f6c799aaafea106

PR #1446, opened 2024-03-15
is Polylang network activated or not ?

sebastienserre:
In a Multisite context, add a test if Polylang is network activated or not


modules/site-health/admin-site-health.php
```diff
@@ -246,6 +246,17 @@ public function info_options( $debug_info ) {
 			}
 		}
 
+		//Multisite
+		if ( is_multisite() ){
+			$plugin = POLYLANG_ROOT_FILE;
+			$network_activated = __( 'No', 'polylang' );
+			if ( is_plugin_active_for_network( 'polylang/polylang.php' ) || is_plugin_active_for_network( 'polylang-pro/polylang.php' )){
```
Chouby, line 253: Use `POLYLANG_BASENAME` instead

modules/site-health/admin-site-health.php
```diff
@@ -246,6 +246,17 @@ public function info_options( $debug_info ) {
...
+				$network_activated = __( 'Yes', 'polylang' );
```
Chouby, line 254: I know translations are more performant now however an `if() ... else` would avoid you to translate 2 stings when only one is necessary.

Chouby (changes requested):
I would rather move this new section to `info()` as this is not related to Polylang options.

sebastienserre:
Replaced by #1712

PR #1449, opened 2024-03-25
Fix false returned with pll_get_post()

loic-parent:
When I try to get a post_id of a page defined in another site using switch_to_blog($blog_id) and pll_get_post($post_id), the value returned was false instead of the existing id because the id was not defined in the cache object.

### DIFF
--- a/include/cache.php
+++ b/include/cache.php
@@ -64,9 +64,7 @@ class PLL_Cache {
 	 * @phpstan-return TCacheData
 	 */
 	public function set( $key, $data ) {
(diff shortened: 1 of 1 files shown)


loic-parent (commented):
Remove the return.

Note: a call to this method stored in a variable is called in the function "sidebars_widgets" (`vendor/wpsyntex/polylang/frontend/frontend-filters-widgets.php:66`).

loic-parent:
(Related to [this issue](https://github.com/polylang/polylang/issues/1452))

PR #1450, opened 2024-03-25
WPCS 3.1

Chouby:
- Remove `sanitize_locale_name()` from custom properties now that's correctly handled by WPCS.
- Fix new errors reported
`Mixing different binary boolean operators within an expression without using parentheses to clarify precedence is not allowed.`

I also added `wp_verify_nonce()` to the `customUnslashingSanitizingFunctions`.

### DIFF
--- a/include/translated-post.php
+++ b/include/translated-post.php
@@ -167,7 +167,7 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 	 */
 	public function is_translated_object_type( $object_type ) {
 		$post_types = $this->get_translated_object_types( false );
-		return ( is_array( $object_type ) && array_intersect( $object_type, $post_types ) || in_array( $object_type, $post_types ) || 'any' === $object_type && ! empty( $post_types ) );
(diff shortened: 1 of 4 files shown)


Screenfeed (commented):
All these parenthesis are not helping readability 😵‍💫

include/translated-post.php
```diff
@@ -167,7 +167,7 @@ public function get_translated_object_types( $filter = true ) {
 	 */
 	public function is_translated_object_type( $object_type ) {
 		$post_types = $this->get_translated_object_types( false );
-		return ( is_array( $object_type ) && array_intersect( $object_type, $post_types ) || in_array( $object_type, $post_types ) || 'any' === $object_type && ! empty( $post_types ) );
+		return ( ( is_array( $object_type ) && array_intersect( $object_type, $post_types ) ) || in_array( $object_type, $post_types ) || ( 'any' === $object_type && ! empty( $post_types ) ) );
```
Screenfeed, line 170:
```suggestion
		return ( is_array( $object_type ) && array_intersect( $object_type, $post_types ) ) || in_array( $object_type, $post_types ) || ( 'any' === $object_type && ! empty( $post_types ) );
```

Screenfeed (approved):

PR #1459, opened 2024-04-03
Fix search block filter

manooweb:
Reported in HS [#28291](https://secure.helpscout.net/conversation/2556149296/28291?folderId=6991790) ticket

## Why?
Because `<form>` tag generated for the search block only rendered as a button has some carriage return between `<form` and `>`.
So the regular expression `#<form.+?>#` doesn't match anything because the `>` character is never reached.
See https://github.com/polylang/polylang/blob/3.6/frontend/frontend-filters-search.php#L72

## How?
Use `#<form.+?>#s` regular expression instead to be able to match any character including carriage return character by considering the string as a single line.

### DIFF
--- a/frontend/frontend-filters-search.php
+++ b/frontend/frontend-filters-search.php
@@ -69,7 +69,7 @@ class PLL_Frontend_Filters_Search {
 
 		if ( $this->links_model->using_permalinks ) {
 			// Take care to modify only the url in the <form> tag.
-			preg_match( '#<form[\s\S]+?>#', $form, $matches );
+			preg_match( '#<form.+?>#s', $form, $matches );
 			$old = reset( $matches );
 			if ( empty( $old ) ) {
 				return $form;
--- a/frontend/frontend-filters-search.php
+++ b/frontend/frontend-filters-search.php
@@ -69,7 +69,7 @@ class PLL_Frontend_Filters_Search {
 
 		if ( $this->links_model->using_permalinks ) {
 			// Take care to modify only the url in the <form> tag.
-			preg_match( '#<form.+?>#', $form, $matches );
+			preg_match( '#<form[\s\S]+?>#', $form, $matches );
 			$old = reset( $matches );
 			if ( empty( $old ) ) {
 				return $form;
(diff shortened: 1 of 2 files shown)


frontend/frontend-filters-search.php
```diff
@@ -69,7 +69,7 @@ public function get_search_form( $form ) {
 
 		if ( $this->links_model->using_permalinks ) {
 			// Take care to modify only the url in the <form> tag.
-			preg_match( '#<form.+?>#', $form, $matches );
+			preg_match( '#<form[\s\S]+?>#', $form, $matches );
```
Screenfeed, line 72: FYI, it could also be fixed like this:
```suggestion
			preg_match( '#<form.+?>#s', $form, $matches );
```
Or this:
```suggestion
			preg_match( '#<form.+>#sU', $form, $matches );
```

tests/phpunit/tests/test-search-form.php
```diff
@@ -70,9 +70,19 @@ public function test_search_block() {
...
 
 		$this->frontend->curlang = self::$model->get_language( 'fr' );
-		$form = do_blocks( '<!-- wp:search /-->' );
+		$form                = do_blocks( '<!-- wp:search /-->' );
+		$form_button_outside = do_blocks( '<!-- wp:search {"showLabel":false,"buttonUseIcon":true,"align":"left"} /-->' );
+		$form_button_inside  = do_blocks( '<!-- wp:search {"showLabel":false,"buttonPosition":"button-inside","buttonUseIcon":true,"align":"left"} /-->' );
+		$form_no_button      = do_blocks( '<!-- wp:search {"showLabel":false,"buttonPosition":"no-button","buttonUseIcon":true,"align":"left"} /-->' );
+		$form_button_only    = do_blocks( '<!-- wp:search {"showLabel":false,"buttonPosition":"button-only","buttonUseIcon":true,"isSearchFieldHidden":true,"align":"left"} /-->' );
+
+		$form = do_blocks( '<!-- wp:search {"showLabel":false,"buttonPosition":"button-only","buttonUseIcon":true,"isSearchFieldHidden":true} /-->' );
```
Screenfeed, line 79: Overwriting `$form`?

Screenfeed (changes requested):

tests/phpunit/tests/test-search-form.php
manooweb: 😮 Oops! copy/paste forgotten

frontend/frontend-filters-search.php
manooweb: `U`ngreedy not sure is what you want because it matches the whole line until the `</form>`.
`s`ingle line seems sufficient not to process carriage return character. So I chose this one.

tests/phpunit/tests/test-search-form.php
manooweb: Done in 64244df1412f71a0a1e3d56bd582207b5da0f61f

frontend/frontend-filters-search.php
manooweb: Done in 64244df1412f71a0a1e3d56bd582207b5da0f61f

PR #1468, opened 2024-04-12
Filter locale on REST context

Marianne380:
Fix https://github.com/polylang/polylang-wc/issues/787.

Filter locale on **REST** context to set the right current language user meta.

**_Needs https://github.com/polylang/polylang/pull/1395 to work because of `do_action( 'change_locale', $new_locale );` which reload the WC localisation files._**

### DIFF
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -134,10 +134,10 @@ class PLL_REST_Request extends PLL_Base {
 	 * @return string
 	 */
 	public function get_locale( $locale ) {
-		if ( ! empty( $this->curlang ) ) {
-			return $this->curlang->locale;
+		if ( empty( $this->curlang ) ) {
+			return $locale;
 		}
 
-		return $locale;
+		return $this->curlang->locale;
 	}
 }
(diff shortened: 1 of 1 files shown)


include/rest-request.php
```diff
@@ -122,4 +124,20 @@ public function set_language( $result, $server, $request ) {
...
+	 *
+	 * @param string $locale The current ID.
+	 * @return string
+	 */
+	public function get_locale( $locale ) {
+		if ( ! empty( $this->curlang ) ) {
+			return $this->curlang->locale;
+		}
+
+		return $locale;
```
szepeviktor, line 141:
```suggestion
		if ( empty( $this->curlang ) ) {
			return $locale;
		}

		return $this->curlang->locale;
```

Please put the unhappy path in the conditional.

[_source_](https://github.com/szepeviktor/very-clean-code#-happy-and-unhappy-path)

szepeviktor:
@Chouby PHPStan says `$this` is of type mixed because `PLL_Static_Pages` has constructor with a pass-by-ref (`&$polylang`) parameter. So `PLL_Static_Pages` may modify `$this`.

Hug0-Drelon (approved):
LGTM, but I think it'd be worthy to add some tests. What do you think?

Marianne380:
Replaced by https://github.com/polylang/polylang-wc/pull/791.

PR #1469, opened 2024-04-15
Fix new issues reported by phpstan

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2088.

1. Fix baseline for `add_submenu_page()`: https://github.com/php-stubs/wordpress-stubs/commit/794559556de631c52b98349890956d306ac38884
2. Remove exclusion about the property `$tax_query`: https://core.trac.wordpress.org/changeset/57761
3. Fix use of `sanitize_option()`: https://github.com/WordPress/WordPress/commit/b9638e6b00ed2cc03dc3ee8cc0c5cc987bc2db23 and https://github.com/php-stubs/wordpress-stubs/commit/379f17a90c01498d4c99a0d15aab6e7aa6a2c840#diff-4dc94858ed12d457b90cbe75a4151de2b69d60a4693fc574a0bf8d4b71923f31R109833.

### DIFF
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -404,7 +404,8 @@ class PLL_Translate_Option {
 	 * @return string Sanitized value.
 	 */
 	public function sanitize_option( $value, $name ) {
+		/** @var string $value */
 		$value = sanitize_option( $name, $value );
-		return is_scalar( $value ) ? (string) $value : '';
+		return $value;
 	}
 }
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -404,6 +404,7 @@ class PLL_Translate_Option {
 	 * @return string Sanitized value.
 	 */
 	public function sanitize_option( $value, $name ) {
-		return sanitize_option( $name, $value );
+		$value = sanitize_option( $name, $value );
+		return is_scalar( $value ) ? (string) $value : '';
 	}
 }
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26,7 +26,7 @@ parameters:
 			path: admin/admin-base.php
 
 		-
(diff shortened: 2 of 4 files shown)


manooweb (approved):

include/translate-option.php
```diff
@@ -404,6 +404,7 @@ protected function check_value_recursive( $old_values, $values, $key, $mo ) {
 	 * @return string Sanitized value.
 	 */
 	public function sanitize_option( $value, $name ) {
-		return sanitize_option( $name, $value );
+		$value = sanitize_option( $name, $value );
+		return is_scalar( $value ) ? (string) $value : '';
```
Chouby, line 408: I am in favor to keep the code as it was and handle this with doc since it's more a doc issue, as `sanitize_option()` doesn't change the type of `$value`.

phpstan-baseline.neon
```diff
@@ -26,7 +26,7 @@ parameters:
 			path: admin/admin-base.php
 
 		-
-			message: "#^Parameter \\#6 \\$callback of function add_submenu_page expects callable\\(\\)\\: mixed, array\\{\\$this\\(PLL_Admin_Base\\), 'languages_page'\\} given\\.$#"
+			message: "#^Parameter \\#6 \\$callback of function add_submenu_page expects ''\\|\\(callable\\(\\)\\: mixed\\), array\\{\\$this\\(PLL_Admin_Base\\), 'languages_page'\\} given\\.$#"
```
Chouby, line 29: Couldn't we fix the error instead by truly creating the missing method even if it does nothing? I have abused the system in the past, knowing the method cannot be called from `PLL_Admin` but only from `PLL_Settings`.

Chouby (changes requested):

phpstan-baseline.neon
Screenfeed: Done in [25115f6](https://github.com/polylang/polylang/pull/1469/commits/25115f6f9299ba39abc6b0f64149a68d07669b01).

PR #1477, opened 2024-04-26
Fix API documentation.

Hug0-Drelon:
## What?
`pll_get_post()` and `pll_get_term()` documentation is not accurate.
Indeed, the passed ID is never returned when not translated in the given language (or current one). Instead `0` is returned.
Before 3.4, we used to return `false` and thus we should keep returning falsy value to avoid breaking changes.

## How?
- Fix the doc for `pll_get_post()` and `pll_get_term()`.
- Fix the doc for `PLL_Translated_Object::get()`.
- Add some tests to ensure `pll_get_post()` and `pll_get_term()` returned types are correct.

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -114,7 +114,7 @@ function pll_default_language( $field = 'slug' ) {
  *
  * @param int                 $post_id Post ID.
  * @param PLL_Language|string $lang    Optional language (object or slug), defaults to the current language.
- * @return int|false The translation post ID if exists, otherwise the passed ID. False if the passed object has no language or if the language doesn't exist.
+ * @return int|false The translation post ID if exists, 0 if not translated. False if the passed object has no language or if the language doesn't exist.
  *
  * @phpstan-return int<0, max>|false
  */
@@ -138,7 +138,7 @@ function pll_get_post( $post_id, $lang = '' ) {
  *
  * @param int                 $term_id Term ID.
  * @param PLL_Language|string $lang    Optional language (object or slug), defaults to the current language.
- * @return int|false The translation term ID if exists, otherwise the passed ID. False if the passed object has no language or if the language doesn't exist.
+ * @return int|false The translation term ID if exists, 0 if not translated. False if the passed object has no language or if the language doesn't exist.
  *
  * @phpstan-return int<0, max>|false
  */
@@ -450,7 +450,7 @@ function pll_save_term_translations( $arr ) {
  *                      Pass `\OBJECT` constant to get the language object. A composite value can be used for language
  *                      term property values, in the form of `{language_taxonomy_name}:{property_name}` (see
  *                      {@see PLL_Language::get_tax_prop()} for the possible values). Ex: `term_language:term_taxonomy_id`.
- * @return string|int|bool|string[]|PLL_Language The requested field or object for the post language, `false` if no language is associated to that post.
+ * @return string|int|false|string[]|PLL_Language The requested field or object for the post language, `false` if no language is associated to that post.
  *
  * @phpstan-return (
  *     $field is \OBJECT ? PLL_Language : (
@@ -480,7 +480,7 @@ function pll_get_post_language( $post_id, $field = 'slug' ) {
  *                      Pass `\OBJECT` constant to get the language object. A composite value can be used for language
  *                      term property values, in the form of `{language_taxonomy_name}:{property_name}` (see
  *                      {@see PLL_Language::get_tax_prop()} for the possible values). Ex: `term_language:term_taxonomy_id`.
- * @return string|int|bool|string[]|PLL_Language The requested field or object for the post language, `false` if no language is associated to that term.
+ * @return string|int|false|string[]|PLL_Language The requested field or object for the post language, `false` if no language is associated to that term.
  *
  * @phpstan-return (
  *     $field is \OBJECT ? PLL_Language : (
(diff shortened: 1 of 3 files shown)


include/api.php
```diff
@@ -114,7 +114,7 @@ function pll_default_language( $field = 'slug' ) {
  *
  * @param int                 $post_id Post ID.
  * @param PLL_Language|string $lang    Optional language (object or slug), defaults to the current language.
- * @return int|false The translation post ID if exists, otherwise the passed ID. False if the passed object has no language or if the language doesn't exist.
+ * @return int|false The translation post ID if exists, 0 if not translated. False if the passed object has no language or if the language doesn't exist.
```
Marianne380, line 117: Shouldn't it be : 
```suggestion
 * @return int|false The translation post ID if exists, 0 if not translated or if the passed object has no language. False if the language doesn't exist.
```

Hug0-Drelon: Indeed, see d192b43a. I also improve the returned value when the language doesn't exist (according to the documentation).

include/api.php
```diff
@@ -114,12 +114,12 @@ function pll_default_language( $field = 'slug' ) {
...
  * @param int                 $post_id Post ID.
  * @param PLL_Language|string $lang    Optional language (object or slug), defaults to the current language.
- * @return int|false The translation post ID if exists, otherwise the passed ID. False if the passed object has no language or if the language doesn't exist.
+ * @return int|false The translation post ID if exists, 0 if not translated or if the post has no language. False if the language doesn't exist.
  *
  * @phpstan-return int<0, max>|false
  */
-function pll_get_post( $post_id, $lang = '' ) {
-	$lang = $lang ? $lang : pll_current_language();
+function pll_get_post( $post_id, $lang = null ) {
```
Marianne380, line 121: Should `null` be added in the `@param $lang` in doc ?

Hug0-Drelon: The doc states `defaults to the current language` which seems enough to me.
Default to `null` is not self explanatory IMHO.

Marianne380 (approved):

include/api.php
```diff
@@ -114,12 +114,12 @@ function pll_default_language( $field = 'slug' ) {
  *
  * @param int                 $post_id Post ID.
  * @param PLL_Language|string $lang    Optional language (object or slug), defaults to the current language.
- * @return int|false The translation post ID if exists, otherwise the passed ID. False if the passed object has no language or if the language doesn't exist.
+ * @return int|false The translation post ID if exists, 0 if not translated or if the post has no language. False if the language doesn't exist.
```
Chouby, line 117: In my opinion, we should stop returning `false` and return `0` instead. There is no point changing the type except messing the doc. In fact here the doc is still incorrect.
[The method returns 0 when the language doesn't exist](https://github.com/polylang/polylang/blob/3.6.1/include/translated-object.php#L334-L337) and false [only when no translation's found](https://github.com/polylang/polylang/blob/3.6.1/include/translated-object.php#L312).

include/api.php
Chouby, line 121: Historically this function was meant to accept a language code. And the default value `pll_current_language()` returns a string, so I would have kept the empty string as param,
Not very important though.

include/api.php
```diff
@@ -114,12 +114,12 @@ function pll_default_language( $field = 'slug' ) {
...
  *
  * @phpstan-return int<0, max>|false
  */
-function pll_get_post( $post_id, $lang = '' ) {
-	$lang = $lang ? $lang : pll_current_language();
+function pll_get_post( $post_id, $lang = null ) {
+	$lang = $lang ? PLL()->model->get_language( $lang ) : pll_current_language();
```
Chouby, line 122: I don't like this because you are mixing 2 types. Moreover `PLL_Translated_Object::get()` make the conversion anyway. So I propose to only check if a language was provided or not?
```suggestion
	if ( empty( $lang ) ) {
		$lang = pll_current_language();
	}
```
The second check after this is probably useless as included in `PLL_Translated_Object::get()`.

Chouby (changes requested):

include/api.php
Hug0-Drelon: Reverted in 6ff20f1d, also use empty string in `pll_get_term()` for constistency.

include/api.php
Hug0-Drelon: Well I reverted changes on this line in 6ff20f1d.
No sure what you suggest is correct since `PLL_Translated_Object::get()` doesn't accept `false` and `pll_current_language()` *can* return `false`. What do you think?

include/api.php
Hug0-Drelon: I also added tests for unexisting language in 0e58e701.

Chouby:
While we are at it, I propose that we change the returned value `false` to 0 for [`get_translation()`](https://github.com/polylang/polylang/blob/3.6.1/include/translated-object.php#L303-L313) too. What do you think?
Also we should update our changelog, be more precise for the changes made in 3.4 and add one for the changes in this PR.

Hug0-Drelon:
> While we are at it, I propose that we change the returned value `false` to 0 for [`get_translation()`](https://github.com/polylang/polylang/blob/3.6.1/include/translated-object.php#L303-L313) too. What do you think? Also we should update our changelog, be more precise for the changes made in 3.4 and add one for the changes in this PR.

Good idea, done in 4f1054e0 for `PLL_Translated_Object::get_translation()` and 855202c1 for the 3.4 changelog.
I also added some tests.
Do we agree 3.7 changelog will be updated just before the release? Not right now?

PR #1481, opened 2024-05-21
Use the new `Options` class

Screenfeed:
This PR replaces the classic `*_option( 'polylang' )` calls by the new `Options` class.

Documentation for the `$options` class properties (like `PLL_Base::$options` is not changed: the type is kept as `array` for now, this prevents a lot of changes everywhere (some being related to PHPStan). This will be done in a future PR.

> [!WARNING]
> The `composer.json` file must be kept to keep PHPStan working.


include/options/registry.php
```diff
@@ -0,0 +1,133 @@
...
+			'description' => 'Time of first activation of Polylang.',
+		),
+		'previous_version'    => array(
+			'class'       => Primitive\String_Type::class,
+			'default'     => '',
+			'description' => "Polylang's previous version.",
+		),
+		'version'             => array(
+			'class'       => Primitive\String_Type::class,
+			'default'     => POLYLANG_VERSION,
```
Hug0-Drelon, line 109: Should constant definition be checked before for extra safety? Or is it overkill?

install/install-base.php
```diff
@@ -131,4 +134,41 @@ public function new_site( $new_site ) {
...
+
+	/**
+	 * Returns the plugin's options.
+	 *
+	 * @since 3.7
+	 *
+	 * @return Options
+	 */
+	protected function get_options(): Options {
+		if ( isset( $GLOBALS['polylang'], $GLOBALS['polylang']->model, $GLOBALS['polylang']->model->options ) ) {
```
Hug0-Drelon, line 146:
```suggestion
		if ( isset( $GLOBALS['polylang']->model->options ) ) {
```
It's sufficient IMHO. See https://onlinephp.io/c/8133d

install/install-base.php
```diff
@@ -131,4 +134,41 @@ public function new_site( $new_site ) {
...
+			return $GLOBALS['polylang']->model->options;
+		}
+
+		add_action( 'pll_init_options_for_blog', array( Options_Registry::class, 'register_options' ), 80 ); // Custom priority.
```
Hug0-Drelon, line 150:
```suggestion
		add_action( 'pll_init_options_for_blog', array( Options_Registry::class, 'register_options' ), 80 ); // Custom priority to avoid conflict with the hook added in `Polylang` class.
```

install/install-base.php
```diff
@@ -131,4 +134,41 @@ public function new_site( $new_site ) {
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param Options $options The options.
+	 * @return void
+	 */
+	protected function save_options( Options $options ): void {
+		$options->save();
+
+		if ( isset( $GLOBALS['polylang'], $GLOBALS['polylang']->model, $GLOBALS['polylang']->model->options ) ) {
```
Hug0-Drelon, line 167:
```suggestion
		if ( isset( $GLOBALS['polylang']->model->options ) ) {
```
Same suggestion as above.

install/install.php
```diff
@@ -112,18 +87,15 @@ public static function get_default_options() {
...
-		else {
-			update_option( 'polylang', self::get_default_options() );
+		$options = $this->get_options();
+
+		if ( version_compare( $options->get( 'version' ), POLYLANG_VERSION, '<' ) ) {
+			// Check if we will be able to upgrade.
+			( new PLL_Upgrade( $options ) )->can_activate();
 		}
 
+		$this->save_options( $options );
```
Hug0-Drelon, line 97: Are we sure the default ones are saved here? 🤔

modules/wizard/wizard.php
```diff
@@ -86,9 +88,9 @@ public function __construct( &$polylang ) {
 	 * @return void
 	 */
 	public static function start_wizard( $network_wide ) {
-		$options = get_option( 'polylang' );
+		$options = (array) get_option( Options::OPTION_NAME, array() );
```
Hug0-Drelon, line 91: Could we use `PLL()->options` here? Any advice @Chouby?

settings/settings-cpt.php
```diff
@@ -159,19 +159,32 @@ protected function form() {
 	}
 
 	/**
-	 * Sanitizes the settings before saving.
+	 * Prepare the received data before saving.
```
Hug0-Drelon, line 162:
```suggestion
	 * Prepares the received data before saving.
```

settings/settings-module.php
```diff
@@ -273,29 +271,27 @@ public function save_options() {
 
 		if ( isset( $_POST['module'] ) && $this->module === $_POST['module'] ) {
 			// It's up to the child class to decide which options are saved, whether there are errors or not
-			$post = array_diff_key( $_POST, array_flip( array( 'action', 'module', 'pll_ajax_backend', '_pll_nonce' ) ) );
-			$options = $this->update( $post );
-			$this->options = array_merge( $this->options, $options );
-			update_option( 'polylang', $this->options );
+			$post   = array_diff_key( $_POST, array_flip( array( 'action', 'module', 'pll_ajax_backend', 'pll_ajax_settings', '_pll_nonce' ) ) );
+			$errors = $this->options->merge( $this->prepare_raw_data( $post ) );
```
Hug0-Drelon, line 275:
```suggestion
			$posted_options   = array_diff_key( $_POST, array_flip( array( 'action', 'module', 'pll_ajax_backend', 'pll_ajax_settings', '_pll_nonce' ) ) );
			$errors                    = $this->options->merge( $this->prepare_raw_data( $posted_options ) );
```
Let's take the opportunity to set a more suitable name.

settings/settings-url.php
```diff
@@ -242,91 +242,22 @@ public function form() {
 	}
 
 	/**
-	 * Sanitizes the settings before saving.
+	 * Prepare the received data before saving.
```
Hug0-Drelon, line 245:
```suggestion
	 * Prepares the received data before saving.
```

settings/settings-url.php
```diff
@@ -242,91 +242,22 @@ public function form() {
...
-			$response = function_exists( 'vip_safe_wp_remote_get' ) ? vip_safe_wp_remote_get( esc_url_raw( $url ) ) : wp_remote_get( esc_url_raw( $url ) );
-			$response_code = wp_remote_retrieve_response_code( $response );
+	protected function prepare_raw_data( array $options ): array {
+		$defaults = array(
+			'force_lang'    => 0,
+			'domains'       => array(),
+			'hide_default'  => 0,
+			'rewrite'       => 0,
+			'redirect_lang' => 0,
+		);
```
Hug0-Drelon, line 259: I must say I'm not a big fan of this, it's a bit anti-pattern with what we've done so far regarding options classes and their encapsulation. 
What do you think?

tests/phpunit/includes/options-trait.php
```diff
@@ -0,0 +1,51 @@
...
+	 * @param array|false $options Initial options. Use `false` to delete previous options.
+	 * @return Options
+	 */
+	protected static function create_options( $options = array() ): Options {
+		if ( ! is_array( $options ) ) {
+			// Reset.
+			delete_option( Options::OPTION_NAME );
+		} elseif ( ! empty( $options ) ) {
+			// Use the given options.
+			$prev_options = get_option( Options::OPTION_NAME );
```
Hug0-Drelon, line 36: Do we really need to take care of stored options in the database here?
Could it do more bad than good (regarding potential side effects between tests)?

tests/phpunit/includes/testcase-preload-paths.php
```diff
@@ -16,21 +16,12 @@ public function set_up() {
 		register_taxonomy( 'trtax', 'trcpt', array( 'show_in_rest' => true ) ); // Translated custom taxonomy.
 		register_taxonomy( 'trtax_with_no_namespace', 'trcpt', array( 'show_in_rest' => true, 'rest_namespace' => null ) ); // Translated custom taxonomy with namespace explicitly set to null as in CPTUI.
 
-		$options = array_merge(
-			PLL_Install::get_default_options(),
+		$options = self::create_reset_options();
```
Hug0-Drelon, line 19: I see you're doing such a thing in many places. Given that, could `self::create_reset_options()` accept an array of options to merge with the default ones directly?

tests/phpunit/includes/testcase-trait.php
```diff
@@ -242,9 +245,9 @@ protected function require_wp_menus( $trigger_hooks = true ) {
 	 * @return void
 	 */
 	protected static function create_model_legacy() {
-		$options                  = PLL_Install::get_default_options();
-		$options['hide_default']  = 0;
-		$options['media_support'] = 1;
-		self::$model              = new PLL_Admin_Model( $options );
+		$options = self::create_reset_options();
```
Hug0-Drelon, line 248: Same comment as before :)

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -13,7 +13,6 @@ public function set_up() {
 		require_once POLYLANG_DIR . '/include/api.php';
 
 		self::$model->options['hide_default'] = 0;
-		update_option( 'polylang', self::$model->options ); // make sure we have options in DB ( needed by PLL_WP_Import )
```
Hug0-Drelon, line 16: Does the constraint in the comment is still met?
Because `shutdown` is run waaaay before any tests are run I think. SO not sure options are already saved.

tests/phpunit/tests/test-canonical-domain.php
```diff
@@ -14,23 +14,29 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
...
+					'en' => 'http://example.org',
+					'fr' => $domain,
+				),
+			)
 		);
-
-		self::$model->options['hide_default'] = 1;
-		self::$model->options['force_lang'] = 3;
-		self::$model->options['domains'] = $domains;
+		update_option( 'polylang', $options );
```
Hug0-Drelon, line 28: Looks tricky, could `self::create_options();` be made *before* `array_merge()`?

tests/phpunit/tests/test-links-default.php
```diff
@@ -20,10 +20,10 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 	public function set_up() {
 		parent::set_up();
 
+		register_post_type( 'cpt', array( 'public' => true ) ); // translated custom post type
 		self::$model->options['post_types'] = array(
 			'cpt' => 'cpt',
 		);
```
Hug0-Drelon, line 26:
```suggestion
		self::$model->options['post_types'] = array( 'cpt' );
```
So it's the same way everywhere.

tests/phpunit/tests/test-model.php
```diff
@@ -235,6 +232,8 @@ public function test_maybe_create_language_terms() {
 		}
 
 		self::$model->clean_languages_cache();
+		$links_model = self::$model->get_links_model();
+		$GLOBALS['polylang'] = new PLL_Admin( $links_model );
```
Hug0-Drelon, line 236: This is here only to validate taxonomies right?
If not set, does the validation still occur?

include/base.php
```diff
@@ -163,7 +162,7 @@ public function switch_blog( $new_blog_id, $prev_blog_id ) {
 	 * @return bool
 	 */
 	protected function is_active_on_current_site(): bool {
-		return pll_is_plugin_active( POLYLANG_BASENAME ) && get_option( 'polylang' );
+		return pll_is_plugin_active( POLYLANG_BASENAME );
```
Hug0-Drelon, line 165: Not sure we should remove `get_option()` here, if I remember correctly it's here to really ensure there are options in DB for the current site for extra safety.

Hug0-Drelon (changes requested):

include/options/registry.php
Screenfeed: This constant is literally the first thing that Polylang defines, so I can't see any scenario where this class can be called before the constant is defined. Do you have one in mind?

Hug0-Drelon: I see, thanks.
Maybe in a CLI environment (not implemented yet) or in unit tests. But we should be able to work around that.

tests/phpunit/tests/plugins/test-wp-importer.php
Screenfeed: As far that I can tell, the comment is a bit misleading with these new options classes. The options do not have to be in the DB, they have to be available in `PLL_WP_Import`, which uses `PLL()->options`.

Then, we have this in `set_up()`:
```php
self::$model->options['hide_default'] = 0;
// ...
$pll_admin = new PLL_Admin( $links_model );
$GLOBALS['polylang'] = &$pll_admin;
```

`$GLOBALS['polylang']` is built by using `self::$model->options`.

tests/phpunit/tests/test-canonical-domain.php
Screenfeed: No, that's the whole point. The point is to have the domains set in the options BEFORE calling `self::create_options()` to avoid having to merge them, and so, calling the sanitization process, which requires `$GLOBALS['polylang']` and the languages to be ready (`_doing_it_wrong()`).

tests/phpunit/tests/test-links-default.php
Screenfeed: I didn't change it because there loooots of places to modify then. And since it accepts both...

modules/wizard/wizard.php
Screenfeed: I don't think we can, this method is called that way:
```php
register_activation_hook( POLYLANG_BASENAME, array( 'PLL_Wizard', 'start_wizard' ) );
```

`$GLOBALS['polylang']` will be created at `'plugins_loaded'`, so it is not available at the activation hook.

install/install.php
Screenfeed: Done in [ea2f1ca](https://github.com/polylang/polylang/pull/1481/commits/ea2f1caf75970963f20d82f58fb72e7d6d5f3ebb).

tests/phpunit/tests/test-model.php
Screenfeed: It used in `Options\Business\Language_Taxonomies` (with `PLL()`).

tests/phpunit/includes/options-trait.php
Screenfeed: Removed in [37ee8c9](https://github.com/polylang/polylang/pull/1481/commits/37ee8c9e8a61d2efd62e24dd25c76c47e1f3333c).

tests/phpunit/includes/testcase-preload-paths.php
Screenfeed: Done in [37ee8c9](https://github.com/polylang/polylang/pull/1481/commits/37ee8c9e8a61d2efd62e24dd25c76c47e1f3333c).

tests/phpunit/includes/testcase-trait.php
Screenfeed: Done in [37ee8c9](https://github.com/polylang/polylang/pull/1481/commits/37ee8c9e8a61d2efd62e24dd25c76c47e1f3333c).

tests/phpunit/tests/test-canonical-domain.php
Screenfeed: Improved in [37ee8c9](https://github.com/polylang/polylang/pull/1481/commits/37ee8c9e8a61d2efd62e24dd25c76c47e1f3333c).

include/base.php
Screenfeed: Fixed in [37ee8c9](https://github.com/polylang/polylang/pull/1481/commits/37ee8c9e8a61d2efd62e24dd25c76c47e1f3333c).

modules/wizard/wizard.php
Hug0-Drelon: I see, thanks for the explanation 👍

tests/phpunit/tests/plugins/test-wp-importer.php
Hug0-Drelon: Okay 👍

Hug0-Drelon (approved):

settings/settings-cpt.php
```diff
@@ -159,19 +159,32 @@ protected function form() {
...
+			$newoptions['post_types'],
+			get_post_types( array( 'public' => true, '_builtin' => false ) )
+		);
+
+		// Limit to public taxonomies.
+		$newoptions['taxonomies'] = array_intersect(
+			$newoptions['taxonomies'],
+			get_taxonomies( array( 'public' => true, '_builtin' => false ) )
+		);
+
```
Chouby, line 187: I don't believe it's necessary. It's a choice to display only public types but the options can accept any types.

settings/settings-module.php
```diff
@@ -247,14 +245,14 @@ public function get_form() {
...
+	 * @since 3.7
 	 *
-	 * @param array $options Unsanitized options to save.
-	 * @return array Options
+	 * @param array $options Raw values to save.
+	 * @return array
 	 */
-	protected function update( $options ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
+	protected function prepare_raw_data( array $options ): array { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
 		return array(); // It's responsibility of the child class to decide what is saved.
```
Chouby, line 256: `update()` was responsible for the validation. It's not the case anymore for the new method, so I would expect that it returns unchanged options. Only options that need preparation to change the format would need to override this method. Am I wrong?

settings/settings-url.php
```diff
@@ -242,91 +242,22 @@ public function form() {
...
```
Chouby, line 259: Isn't this already handled by the Option objects?

include/class-polylang.php
```diff
@@ -165,19 +168,22 @@ public static function define_constants() {
 	 */
 	public function init() {
 		self::define_constants();
-		$options = get_option( 'polylang' );
+
+		// Plugin options.
+		add_action( 'pll_init_options_for_blog', array( Options_Registry::class, 'register_options' ) );
```
Chouby, line 173: Couldn't this be moved to `Options`?

Chouby (changes requested):

include/translated-post.php
```diff
@@ -122,7 +122,7 @@ public function get_translated_object_types( $filter = true ) {
 		if ( false === $post_types ) {
 			$post_types = array( 'post' => 'post', 'page' => 'page', 'wp_block' => 'wp_block' );
 
-			if ( ! empty( $this->model->options['post_types'] ) && is_array( $this->model->options['post_types'] ) ) {
+			if ( ! empty( $this->model->options['post_types'] ) ) {
```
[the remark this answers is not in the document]
Screenfeed: Indeed but this still prevents doing `array_merge()` + `array_combine()` for nothing.

include/translated-term.php
```diff
@@ -197,7 +197,7 @@ public function get_translated_object_types( $filter = true ) {
 		if ( false === $taxonomies ) {
 			$taxonomies = array( 'category' => 'category', 'post_tag' => 'post_tag' );
 
-			if ( ! empty( $this->model->options['taxonomies'] ) && is_array( $this->model->options['taxonomies'] ) ) {
+			if ( ! empty( $this->model->options['taxonomies'] ) ) {
```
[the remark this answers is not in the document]
Screenfeed: Same answer 😅

tests/phpunit/includes/options-trait.php
```diff
@@ -0,0 +1,49 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $options Initial options.
+	 * @return Options
+	 */
+	protected static function create_options( array $options = array() ): Options {
+		if ( ! empty( $options ) ) {
+			update_option( Options::OPTION_NAME, $options );
+		}
```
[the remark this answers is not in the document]
Screenfeed: Doing this will trigger the option validations: if `domains` is in the list, this will require the languages to be ready. This is the reason I did this.

install/install.php
```diff
@@ -112,16 +87,16 @@ public static function get_default_options() {
 	 * @return void
 	 */
 	protected function _activate() {
-		if ( $options = get_option( 'polylang' ) ) {
-			// Check if we will be able to upgrade
+		$options = $this->get_options();
```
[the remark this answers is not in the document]
Screenfeed: The `Options` class is not available during the activation hook.

install/install.php
```diff
@@ -112,16 +87,16 @@ public static function get_default_options() {
...
-				$upgrade->can_activate();
+				( new PLL_Upgrade( $options ) )->can_activate();
 			}
-		}
-		// Defines default values for options in case this is the first installation
-		else {
-			update_option( 'polylang', self::get_default_options() );
+		} else {
+			$options['version'] = POLYLANG_VERSION;
+			$this->save_options( $options );
```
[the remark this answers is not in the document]
Screenfeed: Because [I removed the hook on `shutdown`](https://github.com/polylang/polylang/pull/1481/files#diff-355427764f2ae86e58c72233eb32272eeb381e9ade0c017d4d0f4f8c28ce353fR152) to not interfere with another instance of `Options`.
Maybe there is better way to deal with all this.

The pain point is that during the activation hook there is no existing instance of `Options`, while there is one during the upgrade process.

settings/settings-module.php
Screenfeed: This method has two jobs:
1. Format the data the way the option classes need it.
2. Limit to options handled by the current settings class.

This means all settings classes need to extend this method. And in the spirit of the 2nd point, for me it makes sense to return an empty array instead of all the data.
Does it make sense to you?

settings/settings-url.php
Screenfeed: No. This class handles only these options, so:
1. We limit the options saved to this list.
2. The `array_merge()` could be removed though.

include/class-polylang.php
Screenfeed: It could but I'm not sure it should: registering options for PLL would be in `Options` then, but registering options for PLL Pro would be outside.
I have no strong feelings about this though, so we can do whatever you prefer.

integrations/domain-mapping/domain-mapping.php
```diff
@@ -37,10 +37,8 @@ public function __construct() {
 	 * @since 2.2
 	 */
 	public function dm_redirect_to_mapped_domain() {
-		$options = get_option( 'polylang' );
-
 		// The language is set from the subdomain or domain name
-		if ( $options['force_lang'] > 1 ) {
+		if ( PLL()->options['force_lang'] > 1 ) {
```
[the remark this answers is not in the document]
Screenfeed: Indeed. Fixed in [7eb7493](https://github.com/polylang/polylang/pull/1481/commits/7eb749301a53da6bbd30a52df367850fe479b477).
This brings a slight change in the behavior (there was no check in the `else` part).

tests/phpunit/tests/test-filters-links.php
```diff
@@ -35,6 +27,11 @@ public function set_up() {
 		register_taxonomy( 'trtax', 'trcpt' ); // translated custom tax
 		register_post_type( 'cpt', array( 'public' => true, 'has_archive' => true ) ); // *untranslated* custom post type with archives
 		register_taxonomy( 'tax', 'cpt' ); // *untranslated* custom tax
+
+		self::$model->options['hide_default'] = true;
+		self::$model->options['post_types']   = array( 'trcpt' );
+		self::$model->options['taxonomies']   = array( 'trtax' );
```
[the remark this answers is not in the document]
Screenfeed: Done in [022f826](https://github.com/polylang/polylang/pull/1481/commits/022f826a2110a615ee0547dfb96a020308e096ae).

include/base.php
```diff
@@ -163,7 +162,7 @@ public function switch_blog( $new_blog_id, $prev_blog_id ) {
 	 * @return bool
 	 */
 	protected function is_active_on_current_site(): bool {
-		return pll_is_plugin_active( POLYLANG_BASENAME ) && get_option( 'polylang' );
+		return pll_is_plugin_active( POLYLANG_BASENAME ) && ! empty( $this->options['version'] );
```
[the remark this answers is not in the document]
Screenfeed: Removed in [49d742f](https://github.com/polylang/polylang/pull/1481/commits/49d742faf6fe20d1ac2c47912e0e93078db7236c).
However, this could bring an edge case: doing `switch_to_blog()` after PLL being deactivated.

tests/phpunit/includes/options-trait.php
```diff
@@ -0,0 +1,49 @@
...
+trait PLL_Options_Trait {
+	/**
+	 * Creates a new instance of the options, resets the values, and returns the instance.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $options Initial options.
+	 * @return Options
+	 */
+	protected static function create_reset_options( array $options = array() ): Options {
```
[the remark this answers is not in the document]
Screenfeed: Renamed in [85a2a40](https://github.com/polylang/polylang/pull/1481/commits/85a2a402ca720ae4aecec6c554ae241d4fca5533).

tests/phpunit/includes/options-trait.php
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * Returns a new instance of the options.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $options Initial options.
+	 * @return Options
+	 */
+	protected static function create_options( array $options = array() ): Options {
```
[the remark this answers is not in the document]
Screenfeed: Renamed in [85a2a40](https://github.com/polylang/polylang/pull/1481/commits/85a2a402ca720ae4aecec6c554ae241d4fca5533).

settings/settings-cpt.php
Screenfeed: Removed in [a6caec1](https://github.com/polylang/polylang/pull/1481/commits/a6caec13ed756fa831c107749755a95aa32982ad).

include/options/registry.php
```diff
@@ -0,0 +1,62 @@
...
+	 * @since 3.7
+	 *
+	 * @param Options $options Instance of the options.
+	 * @return void
+	 */
+	public static function register_options( $options ): void {
+		if ( ! $options instanceof Options ) {
+			// Somebody is messing up our hooks.
+			return;
+		}
```
Chouby, line 56: You meant some dev is firing the action `pll_init_options_for_blog` from it's own plugin sending an incorrect type? I suppose he/she would test with our plugin so I'm not against to sanction this with stric typing.

Chouby (changes requested):

include/options/registry.php
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Polylang's options registry.
+ *
+ * @since 3.7
+ */
+class Registry {
+	const OPTIONS = array(
+		// URL modifications.
+		'force_lang'          => Business\Force_Lang::class,
```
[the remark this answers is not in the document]
Screenfeed: Yes, unless you want to write 16 `use` statements:
```php
use WP_Syntex\Polylang\Options\Business\Force_Lang;
use WP_Syntex\Polylang\Options\Business\Domains;
use WP_Syntex\Polylang\Options\Business\Hide_Default;
...
```

include/options/business/media-support.php
```diff
@@ -0,0 +1,43 @@
...
+
+	/**
+	 * Returns the description used in the JSON schema.
+	 *
+	 * @since 3.7
+	 *
+	 * @return string
+	 */
+	protected function get_description(): string {
+		return __( 'Translate media: true to translate, false to not translate.', 'polylang' );
```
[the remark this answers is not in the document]
Screenfeed: Done in [7092d5d](https://github.com/polylang/polylang/pull/1481/commits/7092d5d20a45c91061a24eaaf11a972085e667a6).

include/options/registry.php
Screenfeed: Done in [a8b1ee5](https://github.com/polylang/polylang/pull/1481/commits/a8b1ee5b866c98481674a5a401c5693ea511a588).

include/options/business/media-support.php
```diff
@@ -0,0 +1,43 @@
...
+	 * @since 3.7
+	 *
+	 * @param string $key   Option key.
+	 * @param mixed  $value Optional. Option value.
+	 *
+	 * @phpstan-param non-falsy-string $key
+	 */
+	public function __construct( string $key, $value = null ) {
+		parent::__construct( $key, $value, false );
+	}
```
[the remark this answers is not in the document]
Screenfeed: By doing this, child classes can't change the default value anymore, since this constructor has now 2 parameters instead of 3: in child classes we can't call `parent::__construct( $key, $value, true );` (there is no 3rd parameter).
As an alternative, I suggest to not pass the default value as an argument, but set the property directly in (child) classes.

include/base.php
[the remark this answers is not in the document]
Screenfeed: Reverted in [2eb7ccf](https://github.com/polylang/polylang/pull/1481/commits/2eb7ccf53062f21e4ebff9bd9009aa3323e74d31).

tests/phpunit/includes/options-trait.php
```diff
@@ -0,0 +1,49 @@
...
+	 * @param array $options Initial options.
+	 * @return Options
+	 */
+	protected static function create_reset_options( array $options = array() ): Options {
+		if ( ! empty( $options ) ) {
+			update_option( Options::OPTION_NAME, $options );
+		} else {
+			delete_option( Options::OPTION_NAME );
+		}
+		return self::create_options();
```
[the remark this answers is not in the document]
Screenfeed: `update_options()` removed in [9ae25d8](https://github.com/polylang/polylang/pull/1481/commits/9ae25d855428aad4603001ef90696b2eef29248b).
However I couldn't move `create_options()` to `PLL_UnitTestCase_Trait`: `create_options()` is needed in `PLL_Multisites_TestCase`, `PLL_Context_Base`, and `PLL_UnitTest_Factory`.

install/install.php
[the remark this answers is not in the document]
Screenfeed: Done in [ea0bcd3](https://github.com/polylang/polylang/pull/1481/commits/ea0bcd34fc5e6ef6af5e96d23dee4277a39fa10d).

install/install.php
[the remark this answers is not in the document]
Screenfeed: Changed in [ea0bcd3](https://github.com/polylang/polylang/pull/1481/commits/ea0bcd34fc5e6ef6af5e96d23dee4277a39fa10d).

include/options/business/previous-version.php
```diff
@@ -0,0 +1,42 @@
...
+	 *
+	 * @param string $key   Option key.
+	 * @param mixed  $value Optional. Option value.
+	 *
+	 * @phpstan-param non-falsy-string $key
+	 */
+	public function __construct( string $key, $value = null ) {
+		parent::__construct( $key, $value, '' );
+	}
+
```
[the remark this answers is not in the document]
Screenfeed: Done in [e6e1420](https://github.com/polylang/polylang/pull/1481/commits/e6e1420341cf9c6815a6f55899280b504bc119a3).

include/options/business/post-types.php
```diff
@@ -14,7 +14,21 @@
...
+	 * @since 3.7
+	 *
+	 * @param string $key   Option key.
+	 * @param mixed  $value Optional. Option value.
+	 *
+	 * @phpstan-param non-falsy-string $key
+	 */
+	public function __construct( string $key, $value = null ) {
+		parent::__construct( $key, $value, array(), 'string' );
+	}
```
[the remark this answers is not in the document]
Screenfeed: Done in [e6e1420](https://github.com/polylang/polylang/pull/1481/commits/e6e1420341cf9c6815a6f55899280b504bc119a3).

include/options/business/media-support.php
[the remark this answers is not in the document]
Screenfeed: Done in [e6e1420](https://github.com/polylang/polylang/pull/1481/commits/e6e1420341cf9c6815a6f55899280b504bc119a3).

include/options/business/previous-version.php
```diff
@@ -0,0 +1,42 @@
...
+	 * @since 3.7
+	 *
+	 * @param string $key   Option key.
+	 * @param mixed  $value Optional. Option value.
+	 *
+	 * @phpstan-param non-falsy-string $key
+	 */
+	public function __construct( string $key, $value = null ) {
+		parent::__construct( $key, $value, '' );
+	}
```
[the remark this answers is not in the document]
Screenfeed: Done in [e6e1420](https://github.com/polylang/polylang/pull/1481/commits/e6e1420341cf9c6815a6f55899280b504bc119a3).

include/options/abstract-option.php
```diff
@@ -227,9 +199,18 @@ protected function prepare( $value ) {
...
 	}
 
+	/**
+	 * Returns the default value.
+	 *
+	 * @since 3.7
+	 *
+	 * @return mixed
+	 */
+	abstract protected function get_default();
```
Hug0-Drelon, line 212:
```suggestion
	abstract protected function default();
```
To follow the same namgin as for `key()`, what do you think?

include/options/abstract-option.php
```diff
@@ -240,4 +221,13 @@ protected function sanitize( $value, Options $options ) { // phpcs:ignore Variab
 	 * @phpstan-return array{type: SchemaType}
 	 */
 	abstract protected function get_specific_schema(): array;
```
Hug0-Drelon, line 223:
```suggestion
	abstract protected function get_data_structure(): array;
```
So it's a bit more precise IMHO.

include/options/abstract-option.php
```diff
@@ -240,4 +221,13 @@ protected function sanitize( $value, Options $options ) { // phpcs:ignore Variab
...
+
+	/**
+	 * Returns the description used in the JSON schema.
+	 *
+	 * @since 3.7
+	 *
+	 * @return string
+	 */
+	abstract protected function get_description(): string;
```
Hug0-Drelon, line 232:
```suggestion
	abstract protected function description(): string;
```
Same as for `default()`.

include/options/options.php
```diff
@@ -116,13 +108,7 @@ public function register( string $class_name, string $key, $default, string $des
...
 
 			// Option raw value exists in database, use it.
-			$options[ $key ] = new $class_name(
-				$key,
-				$options[ $key ],
-				$default,
-				$description,
-				...$args
-			);
+			$options[ $key ] = new $class_name( $options[ $key ] );
```
Hug0-Drelon, line 111: I think it could be written only once with an improved condition around line 98, what do you think?

include/options/primitive/abstract-list.php
```diff
@@ -0,0 +1,81 @@
...
+
+namespace WP_Syntex\Polylang\Options\Primitive;
+
+use WP_Syntex\Polylang\Options\Abstract_Option;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Class defining single list option, default value type to mixed.
+ * For convenience, no empty or falsy values are allowed.
```
Hug0-Drelon, line 14:
```suggestion
```
Not sure that still applies.

include/options/registry.php
```diff
@@ -0,0 +1,53 @@
...
+
+	/**
+	 * Registers Polylang's options.
+	 *
+	 * @since 3.7
+	 *
+	 * @param Options $options Instance of the options.
+	 * @return void
+	 */
+	public static function register_options( Options $options ): void {
```
Hug0-Drelon, line 50:
```suggestion
	public static function register( Options $options ): void {
```
With namespaces no need to repeat ourselves IMHO.

tests/phpunit/includes/options-trait.php
```diff
@@ -0,0 +1,34 @@
...
+trait PLL_Options_Trait {
+	/**
+	 * Creates a new instance of the options, resets the values, and returns the instance.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $options Initial options.
+	 * @return Options
+	 */
+	protected static function create_options( array $options = array() ): Options {
```
Hug0-Drelon, line 19: You kept the option trait in the end? Not moved into the test case trait?

Hug0-Drelon (changes requested):

include/options/abstract-option.php
Screenfeed: We also have `get_schema()`, `get_errors()`, and `get_specific_schema()`/`get_data_structure()`.
So...?

include/options/options.php
Screenfeed: Like this?

```php
if ( ! array_key_exists( $key, $options ) || ! $options[ $key ] instanceof Abstract_Option ) {
	$options[ $key ] = new $class_name( $options[ $key ] ?? null );
}
```

I'm not sure to like it.

Screenfeed:
@Hug0-Drelon I'm not against the renaming. Let's see about all this on Monday with @Chouby.

include/options/abstract-option.php
Chouby: I like it.

include/options/abstract-option.php
Chouby: I prefer to keep as it is.

include/options/options.php
Hug0-Drelon: Yup! But it's a matter of taste, you get to choose 😉

include/options/abstract-option.php
Screenfeed: Done in [fa60b4f](https://github.com/polylang/polylang/pull/1481/commits/fa60b4f804c1c5aba7a54695a84fbede0fed7670).

tests/phpunit/includes/options-trait.php
Hug0-Drelon: I see, thanks!

include/options/registry.php
Screenfeed: Good idea. Done in [a22ed80](https://github.com/polylang/polylang/pull/1481/commits/a22ed80c00d4facfc7d389f3390e1a31e4d7327c).

Hug0-Drelon (approved):

PR #1485, opened 2024-05-31
Tests about option schemas

Screenfeed:
This adds some tests about option schemas, to make sure they work as intended.

### DIFF
--- a/tests/phpunit/includes/options-trait.php
+++ b/tests/phpunit/includes/options-trait.php
@@ -14,9 +14,14 @@ trait PLL_Options_Trait {
 	 * @since 3.7
 	 *
 	 * @param array $options Initial options.
-	 * @return Options
+	 * @return Options|array An instance of `Options` for PLL 3.7+, an array otherwise.
 	 */
-	protected static function create_options( array $options = array() ): Options {
+	protected static function create_options( array $options = array() ) {
+		if ( ! class_exists( Options::class ) ) {
+			// Backward compatibility with Polylang < 3.7. Required for PLLWC.
+			return array_merge( PLL_Install::get_default_options(), $options );
+		}
+
 		if ( ! empty( $options ) ) {
 			update_option( Options::OPTION_NAME, $options );
 		} else {
--- a/tests/phpunit/tests/test-option-schema.php
+++ b/tests/phpunit/tests/test-option-schema.php
@@ -469,7 +469,7 @@ class Option_Schema_Test extends WP_UnitTestCase {
 
 		$this->assertSame( $sanitized_value, $option->get() );
 
-		$valid = rest_validate_value_from_schema( $value, $option->get_schema(), $option->key() );
+		$valid = rest_validate_value_from_schema( $value, $option->get_schema(), $option::key() );
 
 		if ( is_wp_error( $expected_valid ) ) {
 			$this->assertInstanceOf( WP_Error::class, $valid );
--- a/tests/phpunit/tests/test-option-schema.php
+++ b/tests/phpunit/tests/test-option-schema.php
@@ -1,0 +1,481 @@
+<?php
+
+use WP_Syntex\Polylang\Options\Business;
+
+/**
+ * Test the schema of all classes extending {@see WP_Syntex\Polylang\Options\Abstract_Option}.
+ */
+class Option_Schema_Test extends WP_UnitTestCase {
+	/**
+	 * @dataProvider boolean_provider
+	 *
+	 * @param mixed         $value           The value to test.
+	 * @param bool          $sanitized_value Sanitized value.
+	 * @param true|WP_Error $expected_valid  Validation result.
+	 */
+	public function test_boolean( $value, bool $sanitized_value, $expected_valid ) {
+		$classes = array(
+			Business\Browser::class,
+			Business\Hide_Default::class,
+			Business\Media_Support::class,
+			Business\Redirect_Lang::class,
+			Business\Rewrite::class,
+		);
+		foreach ( $classes as $class ) {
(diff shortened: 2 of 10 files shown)


tests/phpunit/tests/test-option-schema.php
```diff
@@ -0,0 +1,481 @@
...
+	 * @param mixed         $value           The value to test.
+	 * @param mixed         $sanitized_value Sanitized value.
+	 * @param true|WP_Error $expected_valid  Validation result.
+	 */
+	private function test_option( string $class, $value, $sanitized_value, $expected_valid ) {
+		$option = new $class( $value );
+
+		$this->assertSame( $sanitized_value, $option->get() );
+
+		$valid = rest_validate_value_from_schema( $value, $option->get_schema(), $option->key() );
```
Hug0-Drelon, line 472: Why not using `set()` and `get_errors()` instead?
So the behavior of `PLL_Abstract_Option` is really tested. That way it seems to me we're testing WordPress.
What do you think?

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-option-schema.php
```diff
@@ -0,0 +1,481 @@
...
+
+		if ( is_wp_error( $expected_valid ) ) {
+			$this->assertInstanceOf( WP_Error::class, $valid );
+			$this->assertSame( sprintf( $expected_valid->get_error_message(), $option::key() ), $valid->get_error_message() );
```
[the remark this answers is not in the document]
Screenfeed: I chose to test error messages because they are more specific than the error codes.
For example in [`object_types_provider()`](https://github.com/polylang/polylang/pull/1485/files#diff-b1dc03f80f14f0030d72410b8382e69fa70ff837a6c26144a6e1c3c181a64365R270-R288) there are 2 `rest_invalid_type`, and the last one also mentions that the second value is the issue (`%s[1]`).
However, we may not need to be that much specific, what do you think?

tests/phpunit/tests/test-option-schema.php
Screenfeed: This would require to create an instance of `Options`, `PLL_Model`, `PLL_Frontend` (some validations need `PLL()`).
My idea was to focus only on the schemas.

tests/phpunit/tests/test-option-schema.php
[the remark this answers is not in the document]
Hug0-Drelon: I agree, there is no need to be that specific. Even with duplicate codes, I think it's sufficient because we can assert it behave fine.
Testing errors messages is more UI-like IMHO.

tests/phpunit/tests/test-option-schema.php
Hug0-Drelon: `PLL()` and consort could be mocked as well. But as you see fit.

tests/phpunit/includes/options-trait.php
```diff
@@ -14,9 +14,14 @@ trait PLL_Options_Trait {
 	 * @since 3.7
```
Hug0-Drelon, line 14: The tag is now meaningless IMHO.

tests/phpunit/tests/test-option-schema.php
```diff
@@ -0,0 +1,481 @@
+<?php
+
+use WP_Syntex\Polylang\Options\Business;
+
+/**
+ * Test the schema of all classes extending {@see WP_Syntex\Polylang\Options\Abstract_Option}.
+ */
+class Option_Schema_Test extends WP_UnitTestCase {
```
Hug0-Drelon, line 8: Since there is no factory used, no set up and all. Is `WP_UnitTestCase` useful?
Could we just use `PHPUnit_Adapter_TestCase`?
Sorry to think about it only now... It's more a generic question than a change request.

Hug0-Drelon (changes requested):

tests/phpunit/includes/options-trait.php
Screenfeed: Don't you dare touch my `@since` tags! 🔪

tests/phpunit/tests/test-option-schema.php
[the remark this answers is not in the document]
Screenfeed: Done in [afedd6d](https://github.com/polylang/polylang/pull/1485/commits/afedd6dc3e467b865fb776d74d98b96b541bdd87).

tests/phpunit/tests/test-option-schema.php
Screenfeed: Done in [17a1da4](https://github.com/polylang/polylang/pull/1485/commits/17a1da42b091dd0a1844d070d9c027a83f19315d).

Hug0-Drelon (approved):

tests/phpunit/includes/options-trait.php
```diff
@@ -14,9 +14,14 @@ trait PLL_Options_Trait {
...
 	 * @param array $options Initial options.
-	 * @return Options
+	 * @return Options|array An instance of `Options` for PLL 3.7+, an array otherwise.
 	 */
-	protected static function create_options( array $options = array() ): Options {
+	protected static function create_options( array $options = array() ) {
+		if ( ! class_exists( Options::class ) ) {
+			// Backward compatibility with Polylang < 3.7. Required for PLLWC.
+			return array_merge( PLL_Install::get_default_options(), $options );
+		}
```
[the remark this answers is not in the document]
Screenfeed: Don't we use a test suite separated from PLL version? https://github.com/polylang/polylang-wc/blob/53c644045d385cef87121d1158ce102f271fd997/tests/bin/install-plugins.sh#L59-L63

PR #1487, opened 2024-06-03
Avoids infinite loop when current_user_can() is called during get_terms filter

manooweb:
Reported in Help Scout tickets:
- https://secure.helpscout.net/conversation/2612537760/29308?folderId=6991790
- https://secure.helpscout.net/conversation/2609801223/29268/?folderId=6991790
- https://secure.helpscout.net/conversation/2609169616/29248?folderId=6991790

It's due to a conflict with Content control plugin since its 2.3.0 release.

## What happens?
The issue occurs on frontend when the use is logged out.
Conflict with Content Control aka CC plugin and WordPress and Polylang.
 
- CC is hooked on `template_redirect` and execute a `current_user_can()` WordPress function with `edit_post` capability.
- During `current_user_can()`  process WordPress read the `wp_page_for_privacy_policy` option under the hood in the `map_meta_cap()` function. See https://github.com/WordPress/WordPress/blob/6.5.3/wp-includes/capabilities.php#L273
Called from `WP_User::has_cap()` method itself called in `user_can()` function itself called in `current_user_can()` function
- Polylang is hooked on `wp_page_for_privacy_policy` option read to translate it. See https://github.com/polylang/polylang/blob/3.6.2/include/filters.php#L364. Then calls `PLL_Translatable_Object::get_language()` which calls `PLL_Translatable_Object::get_object_term` which calls `wp_get_object_terms()` WordPress function and then `get_terms()` WordPress function.
- CC is also hooked on `get_terms` filter and excute again the same `current_user_can()` WordPress function with `edit_post` capability
- Then we restart at the first step. We are in the infinite loop.

## To reproduce
Can be reproduced with the snippet below:

```php
<?php
add_filter( 'get_terms', 'pll_get_terms_29308', 10, 4 );

function pll_get_terms_29308( $terms, $taxonomy, $query_vars, $query ) {
	current_user_can( 'edit_post', 1 );

	return $terms;
}

```
With this snippet the issue also occurs with a logged in user (Back and front end).

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -361,7 +361,14 @@ class PLL_Filters {
 	 * @return int
 	 */
 	public function translate_page_for_privacy_policy( $id ) {
-		return empty( $this->curlang ) ? $id : $this->model->post->get( $id, $this->curlang );
+		static $once = false; // Avoid infinite loop.
+
+		if ( ! $once && ! empty( $this->curlang ) ) {
+			$once = true;
+			return $this->model->post->get( $id, $this->curlang );
+		}
+
+		return $id;
 	}
 
 	/**
@@ -376,7 +383,10 @@ class PLL_Filters {
 	 * @return array
 	 */
 	public function fix_privacy_policy_page_editing( $caps, $cap, $user_id, $args ) {
-		if ( in_array( $cap, array( 'edit_page', 'edit_post', 'delete_page', 'delete_post' ) ) ) {
+		static $once = true; // Avoid infinite loop.
+
+		if ( ! $once && in_array( $cap, array( 'edit_page', 'edit_post', 'delete_page', 'delete_post' ) ) ) {
+			$once = true;
 			$privacy_page = get_option( 'wp_page_for_privacy_policy' );
 			if ( $privacy_page && array_intersect( $args, $this->model->post->get_translations( $privacy_page ) ) ) {
 				$caps = array_merge( $caps, map_meta_cap( 'manage_privacy_options', $user_id ) );
--- a/tests/phpunit/tests/test-admin-filters.php
+++ b/tests/phpunit/tests/test-admin-filters.php
@@ -132,22 +132,21 @@ class Admin_Filters_Test extends PLL_UnitTestCase {
 	}
 
 	public function test_wp_page_for_privacy_policy_option_read_should_not_trigger_infinite_loop() {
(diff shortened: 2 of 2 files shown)


include/filters.php
```diff
@@ -361,7 +361,14 @@ public function translate_user_email( $email ) {
...
 	public function translate_page_for_privacy_policy( $id ) {
-		return empty( $this->curlang ) ? $id : $this->model->post->get( $id, $this->curlang );
+		static $once = false; // Avoid infinite loop.
+
+		if ( ! $once && ! empty( $this->curlang ) ) {
+			$once = true;
+			return $this->model->post->get( $id, $this->curlang );
+		}
+
+		return $id;
```
Chouby, line 371: Why is this necessary? I understand the modification of the second method as it deals with caps, but why this one?
Moreover, I fear that this introduces a bug as I expect the result to be wrong at the second call.

include/filters.php
```diff
@@ -376,7 +383,10 @@ public function translate_page_for_privacy_policy( $id ) {
 	 * @return array
 	 */
 	public function fix_privacy_policy_page_editing( $caps, $cap, $user_id, $args ) {
-		if ( in_array( $cap, array( 'edit_page', 'edit_post', 'delete_page', 'delete_post' ) ) ) {
+		static $once = true; // Avoid infinite loop.
```
Chouby, line 386: 1. precise the context for the infinite loop. I suggest:
```
// Avoid infinite loop if a capability check is made in a method hooked to `get_translations()`.
```
2. I fear you are introducing new bugs in case we check capabilities for a different user or a different page after the first call.

Chouby (changes requested):

include/filters.php
manooweb: With the snippet I proposed, the two modification are necessary otherwise there is always the infinite loop.
As you pointed out, I think we could introduce some bugs.
Indeed when working out on the test, I didn't succeed to develop a test as close as the real execution i.e. with a privacy policy page saved as an option precisely because `update_option()` is called in the same process of `get_option()` in the test. 
So I have no other idea to prevent this infinite loop.

manooweb:
The fix is two specific to the conflict and maybe too difficult to fix with other third party plugin conflict.
Also see @Chouby pull request proposed to code-atlantic/content control plugin https://github.com/code-atlantic/content-control/pull/108

PR #1489, opened 2024-06-06
Fix a regression in usage of the `pll_flag` filter

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/2111

Before #690 it was possible to use the filter `pll_flag` to modify only `$flag["url']` without modifying `$flag['src']`. `$flag['src']` was indeed always computed after the filter was applied. The cited PR introduced a regression by prefilling `$flag['src']` before the filter is applied, thus making it necessary when modifying the url to modify `$flag['src']` too.

This PR recomputes `$flag['src']` if the url has been modified and `$flag['src']` has not been modified.

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -373,7 +373,10 @@ class PLL_Language extends PLL_Language_Deprecated {
 	 * }
 	 */
 	public static function get_flag_information( $code ) {
-		$default_flag = array( 'url' => '' );
+		$default_flag = array(
+			'url' => '',
+			'src' => '',
+		);
 
 		// Polylang builtin flags.
 		if ( ! empty( $code ) && is_readable( POLYLANG_DIR . ( $file = '/flags/' . $code . '.png' ) ) ) {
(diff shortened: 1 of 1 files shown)


include/language.php
```diff
@@ -405,11 +405,11 @@ public static function get_flag_information( $code ) {
 		 * }
 		 * @param string $code Flag code.
 		 */
-		$flag = apply_filters( 'pll_flag', $flag, $code );
+		$flag = apply_filters( 'pll_flag', $default_flag, $code );
 
 		$flag['url'] = esc_url_raw( $flag['url'] );
 
-		if ( empty( $flag['src'] ) ) {
+		if ( empty( $flag['src'] ) || ( $flag['src'] === $default_flag['src'] && $flag['url'] !== $default_flag['url'] ) ) {
```
Screenfeed, line 412: `$default_flag['src']` may not exist if `PLL_ENCODED_FLAGS` is not defined. Maybe it should be set L376.

Screenfeed (changes requested):

include/language.php
Chouby: Good catch. I didn't pay attention that PHPStan warned about that due to other errors.

Screenfeed (approved):

PR #1490, opened 2024-06-07
Avoid term duplication when same name and empty slug.

Marianne380:
Fix https://github.com/polylang/polylang-wc/issues/694.

Returns the current empty slug if the term exists with the same name and an empty slug.

To reproduce the default behavior of WordPress when providing a term with a name that already exists and no slug.

Also added a `PLL_Term_Slug` class to share `set_pre_term_slug()` code from `PLL_Share_Term_Slug` and `PLL_Crud_Terms`.

Add a test to test the initial issue.


include/crud-terms.php
```diff
@@ -293,54 +293,19 @@ public function set_pre_term_name( $name ) {
...
-			$parent = apply_filters( 'pll_inserted_term_parent', 0, $taxonomy, $slug );
-		}
-
-		$term_id = (int) $this->model->term_exists_by_slug( $slug, $lang, $taxonomy, $parent );
-
 		// If no term exist in the given language with that slug, it can be created.
-		if ( ! $term_id ) {
-			$slug .= '-' . $lang->slug;
+		if ( ! $term_data['term_id'] ) {
+			$term_data['slug'] .= '-' . $term_data['lang']->slug;
```
Hug0-Drelon, line 306: Is it possible to pass the separator in parameter and return the whole slug from `PLL_Term_Slug` method?
So it avoid to return an array and allow to move these lines inside your new class?

Hug0-Drelon (changes requested):

include/crud-terms.php
```diff
@@ -293,54 +293,21 @@ public function set_pre_term_name( $name ) {
...
-		 * @param string            $slug     Term slug
-		 */
-		$lang = apply_filters( 'pll_inserted_term_language', null, $taxonomy, $slug );
-
-		if ( ! $lang instanceof PLL_Language ) {
-			return $slug;
+		$term_slug = new PLL_Term_Slug( $this->model );
+		$term_slug->set_slug( $slug );
+		$term_slug->set_taxonomy( $taxonomy );
+		$term_slug->set_name( $this->pre_term_name );
```
Marianne380, line 299: @Hug0-Drelon Would you prefer a global setter like `set_term_data()` that sets everything at once, or each item one by one like here?

Hug0-Drelon: What about passing all to thé constructor?

include/term-slug.php
```diff
@@ -0,0 +1,186 @@
...
+	 * Constructor
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Model $model    Instance of the current PLL_Model.
+	 * @param string    $slug     The term slug.
+	 * @param string    $taxonomy The term taxonomy.
+	 * @param string    $name     The term name.*
+	 */
+	public function __construct( PLL_Model $model, $slug, $taxonomy, $name ) {
```
Hug0-Drelon, line 51: Let's type hint all parameters, shall we?

include/term-slug.php
```diff
@@ -0,0 +1,186 @@
...
+		$this->model = $model;
+		$this->slug = $slug;
+		$this->taxonomy = $taxonomy;
+		$this->name = $name;
```
Hug0-Drelon, line 55: Could you align the equal signs? My eyes are bleeding 😛

include/crud-terms.php
```diff
@@ -293,54 +293,17 @@ public function set_pre_term_name( $name ) {
...
-
 		// If no term exist in the given language with that slug, it can be created.
-		if ( ! $term_id ) {
-			$slug .= '-' . $lang->slug;
+		if ( ! $term_slug->get_term_id() ) {
+			return $term_slug->get_suffixed_slug( '-' );
 		}
 
-		return $slug;
+		return $term_slug->get_slug();
```
Hug0-Drelon, line 307: Reading the code, `get_term_id()` and `get_suffixed_slug()` seem useless to me now.
We could internalize the logic even more by removing these method and keeping only `get_slug()` (with an extra parameter for the suffix).
And on Polylang Pro we could use `get_slug()` without any parameter depending on its specific condition (with all the globals).
To go a bit further, i'd rename `get_slug()` into  get()` since "slug" is already in the class name. I think this is something we tend to do nowadays.

Hug0-Drelon (changes requested):

include/term-slug.php
Marianne380: They come from `pre_term_slug` and `pre_term_name`, do you think we can ?

include/term-slug.php
Marianne380: Oops, yes I can 😁

include/crud-terms.php
Marianne380: Something like this https://github.com/polylang/polylang/pull/1490/commits/109de5cefd885a2b12ec37dbc120792f8f36d307 ?

Marianne380: I think I still need the `$term_id` on the PLL Pro side to be able to compare it with the global `ID`.
See https://github.com/polylang/polylang-pro/blob/3.6.2/modules/share-slug/share-term-slug.php#L224.

include/term-slug.php
Hug0-Drelon: If you ensure types are right before creating the object, there should be no problem. What do you think?

include/crud-terms.php
Hug0-Drelon: Indeed, let's keep the ID getter and use it only in Polylang Pro then. Is that ok for you?

Hug0-Drelon (changes requested):

include/term-slug.php
```diff
@@ -155,19 +160,34 @@ private function maybe_get_parent_suffix() {
 	 *
 	 * @since 3.7
 	 *
-	 * @param string $separator The separator for the slug suffix, or empty.
+	 * @param string $separator   The separator for the slug suffix, or empty.
+	 * @param bool   $suffix_slug Should force suffix slug, or not, default false.
 	 * @return string
 	 */
-	public function get( string $separator = '' ): string {
+	public function get( string $separator = '', bool $suffix_slug = false ): string {
```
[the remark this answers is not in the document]
Marianne380: Yes but actually it is to handle another case.

When we're modifying an existing term in `PLL_Share_Term_Slug` we have to find a way of forcing the suffixed slug, because if we already have a `$term_id ` [here](https://github.com/polylang/polylang/blob/17f1540d3976556bdf1857105c355b6364cee1a4/include/term-slug.php#L173) we'll return the slug without a suffix.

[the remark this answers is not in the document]
Hug0-Drelon: I thought that's what we'd want.
So in Polylang Pro:
- pass `can_add_suffix()`
- get the `term_id` and compare it to the globals
- if the test passes, call `get( self::TERM_SLUG_SEPARATOR )` because at this point we do want to add a suffix and the `term_id` is already guessed and do exists.

tests/phpunit/tests/test-admin-filters-term.php
```diff
@@ -755,4 +755,15 @@ public function test_update_term_slug() {
 
 		unset( $_REQUEST, $_POST );
 	}
+
+	/**
+	 * @ticket #694 {@see https://github.com/polylang/polylang-wc/issues/694}
+	 */
+	public function test_create_term_with_empty_slug_and_already_existing_name() {
+		$en = self::factory()->term->create( array( 'taxonomy' => 'category', 'name' => 'Cats', 'slug' => 'cat' ) );
```
Hug0-Drelon, line 763:
```suggestion
		self::factory()->term->create( array( 'taxonomy' => 'category', 'name' => 'Cats', 'slug' => 'cat' ) );
```
To fix last PHPCS error.

include/term-slug.php
```diff
@@ -0,0 +1,179 @@
...
+	public function get( string $separator = '' ): string {
+		if ( empty( $separator ) ) {
+			return $this->slug;
+		}
+
+		return $this->slug . $separator . $this->lang->slug;
+	}
+
+	/**
+	 * Returns the existed term ID.
```
Hug0-Drelon, line 170:
```suggestion
	 * Returns the existing term ID.
```

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

Chouby (changes requested):

include/term-slug.php
```diff
@@ -0,0 +1,179 @@
...
+		 * @param string            $taxonomy Term taxonomy.
+		 * @param string            $slug     Term slug
+		 */
+		$lang = apply_filters( 'pll_inserted_term_language', null, $this->taxonomy, $this->slug );
+		if ( ! $lang instanceof PLL_Language ) {
+			return false;
+		}
+		$this->lang = $lang;
+
+		$this->parent = 0;
```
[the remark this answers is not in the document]
Marianne380: We need the parent to get the existing term by slug, actually in `get_term_id()` if we keep this method.

include/term-slug.php
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	/**
+	 * Returns the existing term ID.
+	 *
+	 * @since 3.7
+	 *
+	 * @return int
+	 */
+	public function get_term_id(): int {
```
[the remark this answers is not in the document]
Marianne380: I think it's better to keep it to get the existing term by slug since we need the parent, see [here](https://github.com/polylang/polylang/pull/1490#discussion_r1647713347).

phpstan.neon.dist
```diff
@@ -16,6 +16,7 @@ parameters:
 		- **/view*.php
 		- include/widget-calendar.php
 		- install/plugin-updater.php
+	treatPhpDocTypesAsCertain: false
```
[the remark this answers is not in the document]
Marianne380: Looks like it is not needed anymore, it was because PHPStan assumed that `$taxonomy` and `$this->pre_term_name` were string and that the following check was useless, see https://github.com/polylang/polylang/actions/runs/9562954062/job/26360443848?pr=1490.

Chouby (changes requested):

include/term-slug.php
```diff
@@ -0,0 +1,244 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Translated_Term $translated_term Term translation object.
+	 * @param string              $slug            The term slug.
+	 * @param string              $taxonomy        The term taxonomy.
+	 * @param string              $name            The term name.
+	 */
+	public function __construct( PLL_Translated_Term $translated_term, string $slug, string $taxonomy, string $name ) {
+		$this->translated_term = $translated_term;
```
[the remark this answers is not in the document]
Marianne380: Yep, same thing, but we missed the use of this class for SQL queries.

However, is there any point in keeping this code in `PLL_Model`?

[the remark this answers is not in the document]
Hug0-Drelon: We discussed it earlier and agree on using `PLL_Translated_Term` for the following reasons:
- It goes in the right direction regarding single responsibility of `PLL_Model`, especially if we want to merge `PLL_Amin_Model` and `PLL_Model` and reduce its size.
- Type is stronger using `PLL_Translated_Term` instead of `PLL_Model`.

There is another solution but it'd imply to create static methods in `PLL_Translated_Object` (which seems feasible because a lot of properties are hardcoded by child classes). This way we'd have only `_construct( string $slug, string $name, string $taxonomy )`.

Marianne380:
@Chouby I had to reset `treatPhpDocTypesAsCertain` to `false` in `phpstan.neon.dist`, because we had the following error :

  `Else branch is unreachable because ternary operator condition is always true.
 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your phpstan.neon.dist.` 

in this [line](https://github.com/polylang/polylang/blob/58a2bdda6a9dcec3e233ffdb94ecc8ed421ffec9/include/crud-terms.php#L288).

It assumed that PHPStan doc was certain.

include/crud-terms.php
```diff
@@ -50,6 +50,11 @@ class PLL_CRUD_Terms {
 	 */
 	private $pre_term_name = '';
 
+	/**
+	 * @var int
+	 */
+	private $pre_term_id = 0;
+
```
Chouby, line 57: It's never used so you don't need it.

include/crud-terms.php
```diff
@@ -293,54 +300,12 @@ public function set_pre_term_name( $name ) {
...
-			$parent = apply_filters( 'pll_inserted_term_parent', 0, $taxonomy, $slug );
-		}
-
-		$term_id = (int) $this->model->term_exists_by_slug( $slug, $lang, $taxonomy, $parent );
-
-		// If no term exist in the given language with that slug, it can be created.
-		if ( ! $term_id ) {
-			$slug .= '-' . $lang->slug;
-		}
+		$term_slug = new PLL_Term_Slug( $this->model, $slug, $taxonomy, $this->pre_term_name, $this->pre_term_id );
```
Chouby, line 307:
```suggestion
		$term_slug = new PLL_Term_Slug( $this->model, $slug, $taxonomy, $this->pre_term_name );
```

include/term-slug.php
```diff
@@ -0,0 +1,177 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Model $model    Instance of PLL_Model.
+	 * @param string    $slug     The term slug.
+	 * @param string    $taxonomy The term taxonomy.
+	 * @param string    $name     The term name.
+	 * @param int       $term_id  The term ID, if exists.
+	 */
+	public function __construct( PLL_Model $model, string $slug, string $taxonomy, string $name, int $term_id ) {
```
Chouby, line 57:
```suggestion
	public function __construct( PLL_Model $model, string $slug, string $taxonomy, string $name, int $term_id = 0 ) {
```

include/term-slug.php
```diff
@@ -0,0 +1,177 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Model $model    Instance of PLL_Model.
+	 * @param string    $slug     The term slug.
+	 * @param string    $taxonomy The term taxonomy.
+	 * @param string    $name     The term name.
+	 * @param int       $term_id  The term ID, if exists.
```
Chouby, line 55: Maybe the comment should be adapted, as in Polylang you always pass 0.

include/term-slug.php
```diff
@@ -0,0 +1,177 @@
...
+	 * @return string The suffixed slug, or not if the lang isn't defined.
+	 */
+	public function get_suffixed_slug( string $separator ): string {
+		if ( ! $this->can_add_suffix() ) {
+			return $this->slug;
+		}
+
+		$term_id = (int) $this->model->term_exists_by_slug( $this->slug, $this->lang, $this->taxonomy, $this->parent );
+
+		if ( ! $term_id || $this->term_id === $term_id ) {
```
Chouby, line 171: This line requires a comment to explain exactly what's done.

Chouby (changes requested):

include/term-slug.php
```diff
@@ -0,0 +1,178 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * @since 3.7
```
Hug0-Drelon, line 7: Please add a description to the class.

Hug0-Drelon (approved):

PR #1493, opened 2024-06-10
Fix the latest phpstan errors

manooweb:
See https://github.com/polylang/polylang/actions/runs/9418621146/job/25946498156

### DIFF
--- a/include/translatable-objects.php
+++ b/include/translatable-objects.php
@@ -37,6 +37,14 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	 *
 	 * @param PLL_Translatable_Object $object The translatable object to register.
 	 * @return PLL_Translatable_Object
+	 *
+	 * @phpstan-return (
+	 *     $object is PLL_Translated_Post ? PLL_Translated_Post : (
+	 *         $object is PLL_Translated_Term ? PLL_Translated_Term : (
+	 *             PLL_Translatable_Object
+	 *         )
+	 *     )
+	 * )
 	 */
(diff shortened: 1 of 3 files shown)


include/translatable-objects.php
```diff
@@ -37,6 +37,14 @@ class PLL_Translatable_Objects implements IteratorAggregate {
...
 	 * @param PLL_Translatable_Object $object The translatable object to register.
 	 * @return PLL_Translatable_Object
+	 *
+	 * @phpstan-return (
+	 *     $object is PLL_Translated_Post ? PLL_Translated_Post : (
+	 *         $object is PLL_Translated_Term ? PLL_Translated_Term : (
+	 *             PLL_Translatable_Object
+	 *         )
+	 *     )
+	 * )
```
Hug0-Drelon, line 47: I think using [generics](https://phpstan.org/blog/generics-in-php-using-phpdocs#type-variables) is a better option here:
- Not aware of implementations
- Work with other types than `PLL_Translated_Post` and `PLL_Translated_Term`
- No need to update the doc in case we add a new `PLL_Translatable_Object` implementation?

Hug0-Drelon (changes requested):

include/translatable-objects.php
manooweb: Any idea @Screenfeed ?

Hug0-Drelon: Well, as discussed there's no way to use generics...

Hug0-Drelon (approved):

PR #1498, opened 2024-06-13
Fix tests on multisite

Screenfeed:
1. Fix: lock options of a deleted blog. When a blog is deleted, its options are removed from `Options` and locked, to prevent any operations on them. This prevents, for example, saving them on `shutdown` et triggering a fatal error because the DB table doesn't exist anymore.
2. Fix: type-cast `$this->options['domains']` as an array in `PLL_Links_Domain::get_hosts()` to make sure we never loop over `null`.
3. Improvement: rename `Language_Slug` into `Default_Lang`. Since we have one class per option, `Language_Slug` doesn't make sense anymore.
4. Fix tests on multisite:
  - During the setup, update options (via `update_option()`) on `switch_blog` before `Options`' callback is called: this allows options to be ready before `Options` read them.
  - Set the default language and the plugin's version into the options.
  - Don't delete options (stored during the setup) by creating a new instance of `Options` at the beginning of each test (a new instance is created by previous options are re-used).

> [!WARNING]
> Restore Github workflows before merging.

### DIFF
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -200,7 +200,7 @@ class Options implements \ArrayAccess {
 	 */
 	public function save_all(): void {
 		// Find blog with modified options.
-		$modified = array_filter( $this->modified );
+		$modified = $this->get_modified();
 
 		if ( empty( $modified ) ) {
 			// Not modified, or locked.
@@ -241,6 +241,11 @@ class Options implements \ArrayAccess {
 		}
 
 		unset( $this->modified[ $this->current_blog_id ] );
+
+		if ( is_multisite() && ! get_site( $this->current_blog_id ) ) { // Cached by `$this->get_modified()` if called from `$this->save_all()`.
+			// Deleted. Should not happen if called from `$this->save_all()`.
+			return false;
+		}
 
 		$options = get_option( self::OPTION_NAME, array() );
 
@@ -532,5 +537,52 @@ class Options implements \ArrayAccess {
 	 */
 	private function is_locked( int $blog_id ): bool {
 		return isset( $this->modified[ $blog_id ] ) && false === $this->modified[ $blog_id ];
+	}
+
+	/**
+	 * Returns the list of modified sites.
+	 * On multisite, sites are cached.
+	 * /!\ At this point, some sites may not appear as locked while they have been deleted: this can happen if they are
+	 * not deleted with `wp_delete_site()`. Those sites are locked here.
+	 *
+	 * @since 3.7
+	 *
+	 * @return bool[]
+	 * @phpstan-return array<int, true>
+	 */
+	private function get_modified(): array {
+		$modified = array_filter( $this->modified );
+
+		if ( empty( $modified ) ) {
+			// Not modified, or locked.
+			return $modified;
+		}
+
+		// Cleanup deleted sites and cache existing ones.
+		if ( ! is_multisite() || count( $modified ) === 1 ) {
+			// Multisite with only 1 site: no need to cache or verify existence.
+			return $modified;
+		}
+
+		// Fetch all the data instead of only the IDs, so it is cached.
+		$sites = get_sites(
+			array(
+				'site__in' => array_keys( $modified ),
+				'number'   => count( $modified ),
+			)
+		);
+
+		$modified = array();
+		// Keep only existing blogs.
+		foreach ( $sites as $site ) {
+			$modified[ $site->id ] = true;
+		}
+		// Make sure that `$this->modified` is up-to-date about the "locked" status.
+		foreach ( $this->modified as $blog_id => $_yup ) {
+			if ( ! isset( $modified[ $blog_id ] ) ) {
+				$this->modified[ $blog_id ] = false;
+			}
+		}
+		return $modified;
 	}
 }
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -531,6 +531,7 @@ class Options implements \ArrayAccess {
 		);
 
 		// Keep only existing blogs.
+		$this->modifed = array();
 		foreach ( $sites as $site ) {
 			$this->modified[ $site->id ] = true;
 		}
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -531,7 +531,7 @@ class Options implements \ArrayAccess {
 		);
 
 		// Keep only existing blogs.
-		$this->modifed = array();
+		$this->modified = array();
 		foreach ( $sites as $site ) {
 			$this->modified[ $site->id ] = true;
 		}
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -37,12 +37,9 @@ class Options implements \ArrayAccess {
 
 	/**
 	 * Tells if the options have been modified, by blog ID.
-	 * - Not set: not modified.
-	 * - True: modified.
-	 * - False: locked, must not be updated.
 	 *
 	 * @var bool[]
-	 * @phpstan-var array<int, bool>
+	 * @phpstan-var array<int, true>
 	 */
 	private $modified = array();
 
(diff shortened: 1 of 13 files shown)


include/Options/Options.php
Hug0-Drelon, line 1: Why not just deleting options for the blog instead of locking it?
You're afraid something tries to access options of the blog right after it's removed?

Screenfeed: To prevent a `switch_to_blog()` to register the options again for example.

Screenfeed: However there is a caveat with the "lock" thing: if a site is deleted with something else than `wp_delete_site()`, this won't work. Using `! get_site( $this->current_blog_id )` in `save()` and/or `save_all()` would work 100% of the time but requires a query to the DB 🤔

Hug0-Drelon: Well, using something other than `wp_delete_site()` to delete a blog seems to be a bad idea in the first place 😅 
Also, calling `switch_to_blog()`  on a deleted site will lead to database errors (I didn't dive deeply in WP code but I didn't see blog existence test...).
So after the last proposal might be the one to follow. Even though it adds an extra query because it's down on `shutdown` there should be no overhead for the end user.

What do you think?

Screenfeed: You mean using `get_site()`? We may need to ensure the sites are in cache because we're going to loop through them. All this may be overkill but I'm for it, it would prevent errors, and so, support.

Screenfeed: So, I added [1d6cf37](https://github.com/polylang/polylang/pull/1498/commits/1d6cf37a5ff64c95eb98262e81cd4f016fdb2e66).

include/Options/Options.php
```diff
@@ -533,4 +538,51 @@ public function offsetUnset( $offset ): void {
...
+	 * On multisite, sites are cached.
+	 * /!\ At this point, some sites may not appear as locked while they have been deleted: this can happen if they are
+	 * not deleted with `wp_delete_site()`. Those sites are locked here.
+	 *
+	 * @since 3.7
+	 *
+	 * @return bool[]
+	 * @phpstan-return array<int, true>
+	 */
+	private function get_modified(): array {
```
Hug0-Drelon, line 553: I'm not sure this method is really needed.
Instead we could use `get_site()` as we said.
It looks like `WP_Site::get_instance()` hits the cache and we can bet that is a blog is in our modified options it has been fetched already (so the cache is primed).
Also, even if the cache is empty and the query is made, since it's done on `shutdown` it's not a big deal IMHO.

Hug0-Drelon: But maybe you'd like an opinion from @Chouby?

Hug0-Drelon (changes requested):

Chouby (changes requested):

tests/phpunit/includes/testcase-multisites.php
```diff
@@ -307,4 +353,55 @@ protected function get_pll_frontend_env( $options = null, bool $init = true ): P
...
+		}
+	}
+
+	/**
+	 * Creates a new instance of the options, resets the values, and returns the instance.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array|null $options Optional. Initial options.
+	 *                            Use `null` to not delete previous options.
```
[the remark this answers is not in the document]
Screenfeed: Yes we need it in several places.
`get_pll_frontend_env( null )`/`get_pll_admin_env( null )` => `set_options( null )` => `create_options( null )`.

The main raison:
In `PLL_Multisites_TestCase::set_up()` we call `set_up_blog_with_pll()` several times, which stores options for the given blogs. These options, we don't want to delete them in the test by calling again `get_pll_frontend_env( null )`/`get_pll_admin_env( null )`.

include/Options/Options.php
Screenfeed: This is to lower the chances to call `get_sites()` because I think `wp_delete_site()` will be used in most cases.

Screenfeed: Simplified in [264175a](https://github.com/polylang/polylang/pull/1498/commits/264175a61160b67511fdbfb89b9cd7ccb0c0a386).

include/Options/Options.php
```diff
@@ -269,6 +296,11 @@ function ( $value ) {
 	public function merge( array $values ): WP_Error {
 		$errors = new WP_Error();
 
+		if ( $this->is_locked( $this->current_blog_id ) ) {
+			$errors->add( 'pll_blog_deleted', __( 'This site has been deleted.', 'polylang' ) );
+			return $errors;
```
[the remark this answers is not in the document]
Screenfeed: Removed in [264175a](https://github.com/polylang/polylang/pull/1498/commits/264175a61160b67511fdbfb89b9cd7ccb0c0a386).

include/Options/Options.php
```diff
@@ -36,6 +37,9 @@ class Options implements \ArrayAccess {
 
 	/**
 	 * Tells if the options have been modified, by blog ID.
+	 * - Not set: not modified.
+	 * - True: modified.
+	 * - False: locked, must not be updated.
```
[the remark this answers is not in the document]
Screenfeed: Not needed anymore after [264175a](https://github.com/polylang/polylang/pull/1498/commits/264175a61160b67511fdbfb89b9cd7ccb0c0a386).

include/Options/Options.php
```diff
@@ -490,4 +499,42 @@ public function offsetSet( $offset, $value ): void {
...
+			array(
+				'site__in' => array_keys( $this->modified ),
+				'number'   => count( $this->modified ),
+			)
+		);
+
+		// Keep only existing blogs.
+		foreach ( $sites as $site ) {
+			$this->modified[ $site->id ] = true;
+		}
```
Chouby, line 536: It looks like that the code doesn't remove deleted sites as described in method comment:
```suggestion
		// Keep only existing blogs.
		$this->modifed = array();
		foreach ( $sites as $site ) {
			$this->modified[ $site->id ] = true;
		}
```

include/Options/Options.php
```diff
@@ -490,4 +499,42 @@ public function offsetSet( $offset, $value ): void {
...
+	 */
+	private function get_modified(): array {
+		if ( empty( $this->modified ) ) {
+			// Not modified.
+			return $this->modified;
+		}
+
+		// Cleanup deleted sites and cache existing ones.
+		if ( ! is_multisite() || count( $this->modified ) === 1 ) {
+			// Not multisite, or multisite with only 1 site: no need to cache or verify existence.
```
Chouby, line 521: I don't understand the thing with `count( $this->modified ) === 1`.
What do you think of:
1. I am on multisite with N>2 sites.
2. `switch_to_blog( $id )`
3. Modify the options for this site
4. Restore current blog
5. `wp_delete_site( $id )`

There are more than one site (the comment is wrong) and I modified the options of a deleted site but the code checking that is bypassed.

Chouby (changes requested):

include/Options/Options.php
Screenfeed: Ah! Good catch. I think this line disappeared while I reworked this.

Screenfeed: Fixed in [79df102](https://github.com/polylang/polylang/pull/1498/commits/79df1025bdfe8599a5c7c177e36ae8aba660addc) and [7861f45](https://github.com/polylang/polylang/pull/1498/commits/7861f4534dfb0203d3d3abc1bf3e16c4fde6cc3e).

include/Options/Options.php
Screenfeed: Fixed in [cb39b22](https://github.com/polylang/polylang/pull/1498/commits/cb39b22a18bd30cd2122e33b836fe89b1865c649).

PR #1500, opened 2024-06-26
Adds wrapper for `wp_insert_term` and `wp_update_term`.

Marianne380:
Adds wrapper for `wp_insert_term` and `wp_update_term` with the language.

This allows term slug handling when adding or updating a term.

See https://github.com/polylang/polylang-pro/pull/2135.

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -561,7 +561,7 @@ function pll_count_posts( $lang, $args = array() ) {
 }
 
 /**
- * Wrapper of `wp_insert_term` with the language.
+ * Wraps `wp_insert_term` with language feature.
  *
  * @since 3.7
  *
--- a/include/api.php
+++ b/include/api.php
@@ -579,7 +579,7 @@ function pll_count_posts( $lang, $args = array() ) {
  * } $args
  */
 function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
-	return PLL()->model->term->insert_term( $term, $taxonomy, $language, $args );
+	return PLL()->model->term->insert( $term, $taxonomy, $language, $args );
 }
 
 /**
@@ -594,7 +594,7 @@ function pll_insert_term( string $term, string $taxonomy, PLL_Language $language
  * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
  */
 function pll_update_term( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
-	return PLL()->model->term->update_term( $term_id, $taxonomy, $language, $args );
+	return PLL()->model->term->update( $term_id, $taxonomy, $language, $args );
 }
 
 
--- a/include/api.php
+++ b/include/api.php
@@ -583,7 +583,7 @@ function pll_insert_term( string $term, string $taxonomy, PLL_Language $language
  * @param int          $term_id  The ID of the term.
  * @param string       $taxonomy The taxonomy of the term.
  * @param PLL_Language $language The term language.
- * @param array        $args     The taxonomy of the term.
+ * @param array        $args     Array of arguments for updating a term.
  * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
  */
 function pll_update_term( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
--- a/include/api.php
+++ b/include/api.php
@@ -609,11 +609,12 @@ function pll_insert_term( string $term, string $taxonomy, $language, array $args
  *     @type string              $description  The term description. Default empty string.
  *     @type int                 $parent       The id of the parent term. Default 0.
  *     @type string              $slug         The term slug to use. Default empty string.
+ *     @type string              $name         The term name.
  *     @type PLL_Language|string $lang         The term language object or slug.
  *     @type string[]            $translations The translation group to assign to the term with language slug as keys and `term_id` as values.
  * }
  * @return array|WP_Error {
- *     An array of the new term data, `WP_Error` otherwise.
+ *     An array containing the `term_id` and `term_taxonomy_id`, `WP_Error` otherwise.
  *
  *     @type int        $term_id          The new term ID.
  *     @type int|string $term_taxonomy_id The new term taxonomy ID. Can be a numeric string.
--- a/include/api.php
+++ b/include/api.php
@@ -576,7 +576,7 @@ function pll_insert_term( string $term, string $taxonomy, PLL_Language $language
 }
 
 /**
- * Wrapper of `wp_update_term` with the language.
+ * Wraps `wp_update_term` with language feature.
  *
  * @since 3.7
  *
--- a/include/api.php
+++ b/include/api.php
@@ -570,6 +570,13 @@ function pll_count_posts( $lang, $args = array() ) {
  * @param PLL_Language $language The term language.
  * @param array|string $args     Array or query string of arguments for inserting a term.
  * @return array|WP_Error An array of the new term, WP_Error otherwise.
+ *
+ * @phpstan-param array{
+ *   alias_of?: string,
+ *   description?: string,
+ *   parent?: int,
+ *   slug?: string,
+ * } $args
  */
 function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
 	return PLL()->model->term->insert_term( $term, $taxonomy, $language, $args );
--- a/include/api.php
+++ b/include/api.php
@@ -620,16 +620,6 @@ function pll_insert_term( string $term, string $taxonomy, $language, array $args
  * }
  */
 function pll_update_term( int $term_id, array $args = array() ) {
-	$language = isset( $args['lang'] ) ? PLL()->model->get_language( $args['lang'] ) : null;
-
-	if ( false === $language ) {
-		return new WP_Error( 'invalid_language', __( 'Please provide a valid language.', 'polylang' ) );
-	}
-
-	if ( $language instanceof PLL_Language ) {
-		$args['lang'] = $language;
-	}
-
 	return PLL()->model->term->update( $term_id, $args );
 }
 
--- a/include/api.php
+++ b/include/api.php
@@ -570,13 +570,6 @@ function pll_count_posts( $lang, $args = array() ) {
  * @param PLL_Language $language The term language.
  * @param array|string $args     Array or query string of arguments for inserting a term.
  * @return array|WP_Error An array of the new term, WP_Error otherwise.
- *
- * @phpstan-param array{
- *   alias_of?: string,
- *   description?: string,
- *   parent?: int,
- *   slug?: string,
- * } $args
  */
 function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
 	return PLL()->model->term->insert( $term, $taxonomy, $language, $args );
--- a/include/api.php
+++ b/include/api.php
@@ -604,12 +604,13 @@ function pll_insert_term( string $term, string $taxonomy, $language, array $args
  * @param array $args {
  *     Optional. Array of arguments for updating a term.
  *
- *     @type string       $alias_of    Slug of the term to make this term an alias of.
- *                                     Default empty string. Accepts a term slug.
- *     @type string       $description The term description. Default empty string.
- *     @type int          $parent      The id of the parent term. Default 0.
- *     @type string       $slug        The term slug to use. Default empty string.
- *     @type PLL_Language $lang        The term language object.
+ *     @type string              $alias_of     Slug of the term to make this term an alias of.
+ *                                             Default empty string. Accepts a term slug.
+ *     @type string              $description  The term description. Default empty string.
+ *     @type int                 $parent       The id of the parent term. Default 0.
+ *     @type string              $slug         The term slug to use. Default empty string.
+ *     @type PLL_Language|string $lang         The term language object or slug.
+ *     @type string[]            $translations The translation group to assign to the term with language slug as keys and `term_id` as values.
  * }
  * @return array|WP_Error {
  *     An array of the new term data, `WP_Error` otherwise.
@@ -619,6 +620,16 @@ function pll_insert_term( string $term, string $taxonomy, $language, array $args
  * }
  */
 function pll_update_term( int $term_id, array $args = array() ) {
+	$language = isset( $args['lang'] ) ? PLL()->model->get_language( $args['lang'] ) : null;
+
+	if ( false === $language ) {
+		return new WP_Error( 'invalid_language', __( 'Please provide a valid language.', 'polylang' ) );
+	}
+
+	if ( $language instanceof PLL_Language ) {
+		$args['lang'] = $language;
+	}
+
 	return PLL()->model->term->update( $term_id, $args );
 }
 
--- a/include/api.php
+++ b/include/api.php
@@ -565,13 +565,33 @@ function pll_count_posts( $lang, $args = array() ) {
  *
  * @since 3.7
  *
- * @param string       $term     The term name to add.
- * @param string       $taxonomy The taxonomy to which to add the term.
- * @param PLL_Language $language The term language.
- * @param array|string $args     Array or query string of arguments for inserting a term.
- * @return array|WP_Error An array of the new term, WP_Error otherwise.
+ * @param string              $term     The term name to add.
+ * @param string              $taxonomy The taxonomy to which to add the term.
+ * @param PLL_Language|string $language The term language object or slug.
+ * @param array               $args {
+ *            Optional. Array of arguments for inserting a term.
+ *
+ *     @type string   $alias_of     Slug of the term to make this term an alias of.
+ *                                  Default empty string. Accepts a term slug.
+ *     @type string   $description  The term description. Default empty string.
+ *     @type int      $parent       The id of the parent term. Default 0.
+ *     @type string   $slug         The term slug to use. Default empty string.
+ *     @type string[] $translations The translation group to assign to the term with language slug as keys and `term_id` as values.
+ * }
+ * @return array|WP_Error {
+ *     An array of the new term data, `WP_Error` otherwise.
+ *
+ *     @type int        $term_id          The new term ID.
+ *     @type int|string $term_taxonomy_id The new term taxonomy ID. Can be a numeric string.
+ * }
  */
-function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
+function pll_insert_term( string $term, string $taxonomy, $language, array $args = array() ) {
+	$language = PLL()->model->get_language( $language );
+
+	if ( ! $language instanceof PLL_Language ) {
+		return new WP_Error( 'invalid_language', __( 'Please provide a valid language.', 'polylang' ) );
+	}
+
 	return PLL()->model->term->insert( $term, $taxonomy, $language, $args );
 }
 
@@ -580,14 +600,26 @@ function pll_insert_term( string $term, string $taxonomy, PLL_Language $language
  *
  * @since 3.7
  *
- * @param int          $term_id  The ID of the term.
- * @param string       $taxonomy The taxonomy of the term.
- * @param PLL_Language $language The term language.
- * @param array        $args     Array of arguments for updating a term.
- * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
+ * @param int   $term_id The ID of the term.
+ * @param array $args {
+ *     Optional. Array of arguments for updating a term.
+ *
+ *     @type string       $alias_of    Slug of the term to make this term an alias of.
+ *                                     Default empty string. Accepts a term slug.
+ *     @type string       $description The term description. Default empty string.
+ *     @type int          $parent      The id of the parent term. Default 0.
+ *     @type string       $slug        The term slug to use. Default empty string.
+ *     @type PLL_Language $lang        The term language object.
+ * }
+ * @return array|WP_Error {
+ *     An array of the new term data, `WP_Error` otherwise.
+ *
+ *     @type int        $term_id          The new term ID.
+ *     @type int|string $term_taxonomy_id The new term taxonomy ID. Can be a numeric string.
+ * }
  */
-function pll_update_term( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
-	return PLL()->model->term->update( $term_id, $taxonomy, $language, $args );
+function pll_update_term( int $term_id, array $args = array() ) {
+	return PLL()->model->term->update( $term_id, $args );
 }
 
 
--- a/include/api.php
+++ b/include/api.php
@@ -561,6 +561,37 @@ function pll_count_posts( $lang, $args = array() ) {
 }
 
 /**
+ * Wrapper of `wp_insert_term` with the language.
+ *
+ * @since 3.7
+ *
+ * @param string       $term     The term name to add.
+ * @param string       $taxonomy The taxonomy to which to add the term.
+ * @param PLL_Language $language The term language.
+ * @param array|string $args     Array or query string of arguments for inserting a term.
+ * @return array|WP_Error An array of the new term, WP_Error otherwise.
+ */
+function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
+	return PLL()->model->term->insert_term( $term, $taxonomy, $language, $args );
+}
+
+/**
+ * Wrapper of `wp_update_term` with the language.
+ *
+ * @since 3.7
+ *
+ * @param int          $term_id  The ID of the term.
+ * @param string       $taxonomy The taxonomy of the term.
+ * @param PLL_Language $language The term language.
+ * @param array        $args     The taxonomy of the term.
+ * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
+ */
+function pll_update_term( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
+	return PLL()->model->term->update_term( $term_id, $taxonomy, $language, $args );
+}
+
+
+/**
  * Allows to access the Polylang instance.
  * However, it is always preferable to use API functions
  * as internal methods may be changed without prior notice.
--- a/include/translated-term.php
+++ b/include/translated-term.php
@@ -367,8 +367,7 @@ class PLL_Translated_Term extends PLL_Translated_Object implements PLL_Translata
 			return $term;
 		}
 
-		$term_id = (int) $term['term_id'];
-		$this->model->term->set_language( $term_id, $language );
+		$this->model->term->set_language( (int) $term['term_id'], $language );
 
 		return $term;
 	}
--- a/include/translated-term.php
+++ b/include/translated-term.php
@@ -358,9 +358,8 @@ class PLL_Translated_Term extends PLL_Translated_Object implements PLL_Translata
 			return $language;
 		};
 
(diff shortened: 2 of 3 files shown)


include/translated-term.php
```diff
@@ -338,4 +345,57 @@ protected function get_db_infos() {
...
+	}
+
+	/**
+	 * Filters the currently inserted term language to suffix the term slug or not.
+	 *
+	 * @since 3.7
+	 *
+	 * @return PLL_Language Overridden language object.
+	 */
+	public function set_language_for_term_slug() {
```
Hug0-Drelon, line 405: I'd rather not add a new property for a specific piece of data in this very generic class.
Maybe a closure as callback is more suitable?

include/translated-term.php
```diff
@@ -338,4 +345,57 @@ protected function get_db_infos() {
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param string       $term     The term name to add.
+	 * @param string       $taxonomy The taxonomy to which to add the term.
+	 * @param PLL_Language $language The term language.
+	 * @param array|string $args     Array or query string of arguments for inserting a term.
+	 * @return array|WP_Error An array of the new term, WP_Error otherwise.
+	 */
+	public function insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
```
Hug0-Drelon, line 367: I wonder if it should set the term language as well. What do you think?

Hug0-Drelon (changes requested):
This new feature should be thoroughly tested IMHO.
Since you'll go on vacations, I propose to write tests if you like.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+
+	/**
+	 * Wrapper of `wp_insert_term` with the language.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string       $term     The term name to add.
+	 * @param string       $taxonomy The taxonomy to which to add the term.
+	 * @param PLL_Language $language The term language.
+	 * @param array|string $args     The taxonomy of the term.
```
Hug0-Drelon, line 350:
```suggestion
	 * @param array|string $args     Array or query string of arguments for inserting a term.
```
According to WP doc.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+	 * @return array|WP_Error An array of the new term, WP_Error otherwise.
+	 *
+	 * @phpstan-param array{
+	 *   alias_of?: string,
+	 *   description?: string,
+	 *   parent?: int,
+	 *   slug?: string,
+	 * } $args
```
Hug0-Drelon, line 358: Not correct since it can also be a string.
I suggest to remove this PHPStan doc since it's WP business. What do you think?

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
 			'default_alias' => 't',
 		);
 	}
+
+	/**
+	 * Wrapper of `wp_insert_term` with the language.
```
Hug0-Drelon, line 343:
```suggestion
	 * Wraps `wp_insert_term` with language feature.
```
Just a suggestion.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+		$term = wp_insert_term( $term, $taxonomy, $args );
+		remove_filter( 'pll_inserted_term_language', $set_language_for_term_slug, 20 );
+
+		if ( is_wp_error( $term ) ) {
+			// Something went wrong!
+			return $term;
+		}
+
+		$term_id = (int) $term['term_id'];
+		$this->model->term->set_language( $term_id, $language );
```
Hug0-Drelon, line 375:
```suggestion
		$this->model->term->set_language( (int) $term['term_id'], $language );
```
Shorter.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+	 *   parent?: int,
+	 *   slug?: string,
+	 * } $args
+	 */
+	public function insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
+		$set_language_for_term_slug = function () use ( $language ) {
+			return $language;
+		};
+
+		add_filter( 'pll_inserted_term_language', $set_language_for_term_slug, 20 ); // After Polylang's filter.
```
Hug0-Drelon, line 365: We should also filter `pll_insterted_term_parent`. Indeed, if there is one it can be found in `$args['parent']`.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+		}
+
+		$term_id = (int) $term['term_id'];
+		$this->model->term->set_language( $term_id, $language );
+
+		return $term;
+	}
+
+	/**
+	 * Wrapper of `wp_update_term` with the language.
```
Hug0-Drelon, line 381:
```suggestion
	 * Wraps `wp_update_term` with language feature.
```

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+	 * @return array|WP_Error An array of the new term, WP_Error otherwise.
+	 *
+	 * @phpstan-param array{
+	 *   alias_of?: string,
+	 *   description?: string,
+	 *   parent?: int,
+	 *   slug?: string,
+	 * } $args
+	 */
+	public function insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
```
Hug0-Drelon, line 360:
```suggestion
	public function insert( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
```
The term "term" is already in the class name `PLL_Translated_Term` 😅

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+
+	/**
+	 * Wrapper of `wp_update_term` with the language.
+	 *
+	 * @since 3.7
+	 *
+	 * @param int          $term_id  The ID of the term.
+	 * @param string       $taxonomy The taxonomy of the term.
+	 * @param PLL_Language $language The term language.
+	 * @param array        $args     The taxonomy of the term.
```
Hug0-Drelon, line 388:
```suggestion
	 * @param array        $args     Array of arguments for updating a term.
```
According to WP doc.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+	 * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
+	 */
+	public function update_term( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
```
Hug0-Drelon, line 391:
```suggestion
	public function update( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
```
Same as above.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+
+		$term = get_term( $term_id );
+		if ( ! $term instanceof WP_Term ) {
+			return new WP_Error( 'invalid_term', __( 'Empty Term.', 'polylang' ) );
+		}
+
+		$term_parent              = $term->parent;
+		$get_inserted_term_parent = function () use ( $term_parent ) {
+			return $term_parent;
+		};
```
Hug0-Drelon, line 404:
```suggestion
		$get_inserted_term_parent = function () use ( $term_parent ) {
			return $term->parent;
		};
```
Shorter.

include/translated-term.php
```diff
@@ -338,4 +338,81 @@ protected function get_db_infos() {
...
+
+		// Set term parent and language for suffixed slugs.
+		add_filter( 'pll_inserted_term_language', $set_language_for_term_slug, 20 ); // After Polylang's filter.
```
Hug0-Drelon, line 407: Why do we have to come after Polylang filter? The other callback doesn't check there is already a language added?

Hug0-Drelon (changes requested):

include/translated-term.php
Marianne380: Actually I don't remember why we have this priority. 

It seems to me that the other callback indeed check if there is already a language and returns it if this is the case.

Hug0-Drelon: So we could use the default priority, right?

Marianne380: Yep, it's done.

include/translated-term.php
Hug0-Drelon: Well, it might be an error in WP stubs... Let's take care of this later 😉

include/api.php
```diff
@@ -560,6 +560,44 @@ function pll_count_posts( $lang, $args = array() ) {
 	return PLL()->model->count_posts( $lang, $args );
 }
 
+/**
+ * Wrapper of `wp_insert_term` with the language.
```
Hug0-Drelon, line 564:
```suggestion
 * Wraps `wp_insert_term` with language feature.
```
Sorry I didn't propose earlier.

include/api.php
```diff
@@ -560,6 +560,44 @@ function pll_count_posts( $lang, $args = array() ) {
...
+ * @param PLL_Language $language The term language.
+ * @param array|string $args     Array or query string of arguments for inserting a term.
+ * @return array|WP_Error An array of the new term, WP_Error otherwise.
+ *
+ * @phpstan-param array{
+ *   alias_of?: string,
+ *   description?: string,
+ *   parent?: int,
+ *   slug?: string,
+ * } $args
```
Hug0-Drelon, line 579:
```suggestion
```
Same as the method you created.

include/api.php
```diff
@@ -560,6 +560,44 @@ function pll_count_posts( $lang, $args = array() ) {
...
+ */
+function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
+	return PLL()->model->term->insert( $term, $taxonomy, $language, $args );
+}
+
+/**
+ * Wrapper of `wp_update_term` with the language.
```
Hug0-Drelon, line 586:
```suggestion
 * Wraps `wp_update_term` with language feature.
```

include/api.php
```diff
@@ -560,6 +560,44 @@ function pll_count_posts( $lang, $args = array() ) {
...
+ *
+ * @since 3.7
+ *
+ * @param int          $term_id  The ID of the term.
+ * @param string       $taxonomy The taxonomy of the term.
+ * @param PLL_Language $language The term language.
+ * @param array        $args     The taxonomy of the term.
```
Hug0-Drelon, line 593:
```suggestion
 * @param array        $args     Array of arguments for updating a term.
```

Hug0-Drelon (commented):
I have one last question.
Should `update()` take care of changing/checking the term language as well? Given that changing a term language might impact its translation group...
I'm not against your opinion on that one @Chouby.

Marianne380:
> I have one last question. Should `update()` take care of changing/checking the term language as well? Given that changing a term language might impact its translation group... I'm not against your opinion on that one @Chouby.

That's a good question, I'd tend to say yes to follow the logic of PLL_Translated_Term::insert().

Hug0-Drelon (approved):
LGTM
Although I opened an issue in [`wordpress-stubs`](https://github.com/php-stubs/wordpress-stubs/issues/185) for the remaining PHPStan error.

include/translated-term.php
```diff
@@ -338,4 +338,82 @@ protected function get_db_infos() {
...
+
+		$get_inserted_term_parent = function () use ( $term ) {
+			return $term->parent;
+		};
+
+		// Set term parent and language for suffixed slugs.
+		add_filter( 'pll_inserted_term_language', $set_language_for_term_slug );
+		add_filter( 'pll_inserted_term_parent', $get_inserted_term_parent );
+
+		$tr_term = wp_update_term( $term_id, $taxonomy, $args );
```
Chouby, line 411: It looks strange to name this var `$tr_term` which in the plugin generally refers to "translated term".

Chouby (changes requested):

include/api.php
```diff
@@ -560,6 +560,37 @@ function pll_count_posts( $lang, $args = array() ) {
...
+ *
+ * @since 3.7
+ *
+ * @param string       $term     The term name to add.
+ * @param string       $taxonomy The taxonomy to which to add the term.
+ * @param PLL_Language $language The term language.
+ * @param array|string $args     Array or query string of arguments for inserting a term.
+ * @return array|WP_Error An array of the new term, WP_Error otherwise.
+ */
+function pll_insert_term( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
```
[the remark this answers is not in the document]
Hug0-Drelon: Done in df4f4a35.

include/translated-term.php
```diff
@@ -338,4 +338,82 @@ protected function get_db_infos() {
...
+	/**
+	 * Wraps `wp_insert_term` with language feature.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string       $term     The term name to add.
+	 * @param string       $taxonomy The taxonomy to which to add the term.
+	 * @param PLL_Language $language The term language.
+	 * @param array|string $args     Array or query string of arguments for inserting a term.
+	 * @return array|WP_Error An array of the new term, WP_Error otherwise.
```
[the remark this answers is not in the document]
Hug0-Drelon: Done in df4f4a35.

include/translated-term.php
```diff
@@ -338,4 +338,82 @@ protected function get_db_infos() {
...
+	 */
+	public function insert( string $term, string $taxonomy, PLL_Language $language, $args = array() ) {
```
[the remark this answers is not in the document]
Hug0-Drelon: Done in df4f4a35.

include/translated-term.php
```diff
@@ -338,4 +338,82 @@ protected function get_db_infos() {
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param int          $term_id  The ID of the term.
+	 * @param string       $taxonomy The taxonomy of the term.
+	 * @param PLL_Language $language The term language.
+	 * @param array        $args     Array of arguments for updating a term.
+	 * @return array|WP_Error An array containing the term_id and term_taxonomy_id, WP_Error otherwise.
+	 */
+	public function update( int $term_id, string $taxonomy, PLL_Language $language, array $args = array() ) {
```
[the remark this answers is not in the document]
Hug0-Drelon: Done in df4f4a35.
Even though adding the language to the array of arguments doesn't simplify our code IMHO.
Indeed, we have to have two different `WP_Error` objects (one in the API, one in `PLL_Translated_Term`) for invalid language provided. Also, it's doesn't go the right way regarding strong typing.

include/translated-term.php
Hug0-Drelon: Oops, copy/paste drama.
Changed in df4f4a35.

Hug0-Drelon:
@Chouby Should I look into the parent slug thing thing this PR? If not, should I leaver the test as it is or make it fail?

tests/phpunit/tests/test-api.php
```diff
@@ -67,4 +79,92 @@ public function test_translated_objects_get_translation( $type ) {
...
+			$this->assertIsArray( $result, "The term in {$this->pll_env->model->get_language( $language )->name} could not be created." );
+			$this->assertArrayHasKey( 'term_id', $result );
+			$this->assertArrayHasKey( 'term_taxonomy_id', $result );
+
+			$translations[ $language ] = $result['term_id'];
+			$term                      = get_term( $result['term_id'], $taxonomy );
+
+			/*
+			 * @FIXME Polylang doesn't reuse parent slug and add a suffix with language slug instead...
+			 */
```
Chouby, line 133: You can remove the comment and open an issue instead.

include/translated-term.php
```diff
@@ -338,4 +338,120 @@ protected function get_db_infos() {
...
+
+		// Set term parent and language for suffixed slugs.
+		add_filter( 'pll_inserted_term_language', $set_language_for_term_slug );
+		add_filter( 'pll_inserted_term_parent', $get_inserted_term_parent );
+
+		$term = wp_update_term( $term->term_id, $term->taxonomy, $args );
+
+		// Clean up!
+		remove_filter( 'pll_inserted_term_parent', $get_inserted_term_parent );
+		remove_filter( 'pll_inserted_term_language', $set_language_for_term_slug );
```
Chouby, line 453: Most of the code is duplicate from the previous method. I wonder if we can find a way around this.

tests/phpunit/tests/test-api.php
```diff
@@ -67,4 +79,92 @@ public function test_translated_objects_get_translation( $type ) {
 		$this->assertSame( 0, PLL()->model->$type->get_translation( $objects['en'], 'de' ) );
 		$this->assertSame( 0, PLL()->model->$type->get_translation( $objects['en'], 'chti' ) );
 	}
+
+	/**
+	 * @covers ::pll_insert_term
```
Chouby, line 84: Should we really start to use this annotation?

Chouby (changes requested):

include/translated-term.php
Hug0-Drelon: I found one in e5692087.

include/translated-term.php
```diff
@@ -338,4 +338,120 @@ protected function get_db_infos() {
...
+	 * @return array|WP_Error An array containing the `term_id` and `term_taxonomy_id`,
+	 *                        WP_Error otherwise.                  WP_Error otherwise.
+	 */
+	public function update( int $term_id, array $args = array() ) {
+		$term = get_term( $term_id );
+		if ( ! $term instanceof WP_Term ) {
+			return new WP_Error( 'invalid_term', __( 'Empty Term.', 'polylang' ) );
+		}
+
+		$language = ! empty( $args['lang'] ) ? $this->model->get_language( $args['lang'] ) : null;
```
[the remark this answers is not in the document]
Hug0-Drelon: Done in [0568df6](https://github.com/polylang/polylang/pull/1500/commits/0568df60b5f3aeb1d1a7f5f4a15b4897cf55fe01).

include/translated-term.php
```diff
@@ -338,4 +338,120 @@ protected function get_db_infos() {
...
+	 *                                     Default empty string. Accepts a term slug.
+	 *     @type string       $description The term description. Default empty string.
+	 *     @type int          $parent      The id of the parent term. Default 0.
+	 *     @type string       $slug        The term slug to use. Default empty string.
+	 *     @type PLL_Language $lang        The term language object.
+	 * }
+	 * @return array|WP_Error An array containing the `term_id` and `term_taxonomy_id`,
+	 *                        WP_Error otherwise.                  WP_Error otherwise.
+	 */
+	public function update( int $term_id, array $args = array() ) {
```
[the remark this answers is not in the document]
Hug0-Drelon: Done in [0568df6](https://github.com/polylang/polylang/pull/1500/commits/0568df60b5f3aeb1d1a7f5f4a15b4897cf55fe01).

tests/phpunit/tests/test-api.php
Hug0-Drelon: 👍 
The [issue](https://github.com/polylang/polylang-pro/issues/2193).

tests/phpunit/tests/test-api.php
Hug0-Drelon: As you see fit.
It's already done in this very class.
Here is the [documentation](https://docs.phpunit.de/en/10.5/annotations.html#covers) for some insight.

I stumbled upon that when doing a PR on WordPress, which uses that kind of annotation as a standard, so I thought "Why not?".

include/translated-term.php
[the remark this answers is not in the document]
Hug0-Drelon: I think we thought about it as a team during your vacation, waiting your approval.
I must have forgot to tell you about that...

include/translated-term.php
```diff
@@ -338,4 +338,181 @@ protected function get_db_infos() {
...
+		if ( $enabled ) {
+			// Clean up!
+			remove_filter( 'pll_inserted_term_language', $language_callback );
+			remove_filter( 'pll_inserted_term_parent', $parent_callback );
+			$enabled = false;
+		} else {
+			// Set term parent and language for suffixed slugs.
+			add_filter( 'pll_inserted_term_language', $language_callback );
+			add_filter( 'pll_inserted_term_parent', $parent_callback );
+			$enabled = true;
```
Chouby, line 485: You can have a common instruction with `$enabled = ! $enabled`.

include/translated-term.php
```diff
@@ -338,4 +338,181 @@ protected function get_db_infos() {
...
+
+		$old_language = $this->get_language( $term_id );
+		if ( false === $old_language ) {
+			return new WP_Error( 'broken_language', __( 'Given term has a broken language.', 'polylang' ) );
+		}
+
+		if ( is_null( $new_language ) ) {
+			$new_language = $old_language;
+		}
+
```
Chouby, line 436: We don't need this as this is done in `set_language()`.
```suggestion
```

include/translated-term.php
```diff
@@ -338,4 +338,181 @@ protected function get_db_infos() {
...
+	public function update( int $term_id, array $args = array() ) {
+		$term = get_term( $term_id );
+		if ( ! $term instanceof WP_Term ) {
+			return new WP_Error( 'invalid_term', __( 'Empty Term.', 'polylang' ) );
+		}
+
+		$new_language = ! empty( $args['lang'] ) ? $this->model->get_language( $args['lang'] ) : null;
+		if ( false === $new_language ) {
+			return new WP_Error( 'invalid_language', __( 'Please provide a valid language.', 'polylang' ) );
+		}
```
Chouby, line 426: To simplify and improve the readability,I would manage the language in a unique block instead of 2 separate:
```php
	$language = $this->get_language( $term_id );
	if ( ! empty( $args['lang'] ) ) {
		$language = $this->model->get_language( $args['lang'] );
		if ( ! $language instance of PLL_Language ) {
			return new WP_Error( 'invalid_language', __( 'Please provide a valid language.', 'polylang' ) );
		}
		
		$this->set_language( $term_id, $language );
	}
```

tests/phpunit/tests/test-api.php
Chouby: Indeed I see that it's used in a few other places. I missed them when they were added (probably because I often don't review tests).

In my opinion it's misused. What it says is that your test covers only `pll_insert_term()`, which is wrong as it's more a test of `PLL_Translated_Term::insert()`.

I believe that this annotation is adapted for unit tests (or "pseudo" unit tests as a true unit test should not need it). It's certainly not adapted to wrappers like `pll_insert_term()`.

I never used it because most of the tests are integration tests and I like to know what lines are executed even indirectly.

include/api.php
```diff
@@ -560,6 +560,80 @@ function pll_count_posts( $lang, $args = array() ) {
...
+	$language = isset( $args['lang'] ) ? PLL()->model->get_language( $args['lang'] ) : null;
+
+	if ( false === $language ) {
+		return new WP_Error( 'invalid_language', __( 'Please provide a valid language.', 'polylang' ) );
+	}
+
+	if ( $language instanceof PLL_Language ) {
+		$args['lang'] = $language;
+	}
+
```
Chouby, line 632: Not need if done in the wrapped method.
```suggestion
```

Chouby (changes requested):

include/translated-term.php
Hug0-Drelon: Done in b54df4000a24ad2807375b7fbb6d623ea1dfe64c.

include/api.php
Hug0-Drelon: Removed in b54df4000a24ad2807375b7fbb6d623ea1dfe64c.

include/translated-term.php
Hug0-Drelon: Fixed in f588583dfb3748c15d32ad6acaeb6379dff49c62.

include/translated-term.php
```diff
@@ -338,4 +338,181 @@ protected function get_db_infos() {
...
+	 * Toggles Polylang term slug management.
+	 * Must be used before and after any term slug modification or insertion.
+	 *
+	 * @since 3.7
+	 *
+	 * @param callable $language_callback The language callback for term slug modification.
+	 * @param callable $parent_callback   The parent callback for term slug modification.
+	 * @return void
+	 */
+	private function toggle_slug_management( $language_callback, $parent_callback ): void {
```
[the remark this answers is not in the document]
Hug0-Drelon: Renamed in 4a74dbcff9c7017ad68f9bafb0edaebb651719c8.
Refactored in aa79ef99fa8e1cf6cb0dc59fda80e6b3f1fff7d2.

include/translated-term.php
Hug0-Drelon: I must say I'm not very fond of this.
With `$enabled = false` or `$enabled = true` I can directly tell what value is assigned.
Also, toggle function might already be quite confusion as it is, I suggest to keep it.
Do you agree?

include/translated-term.php
```diff
@@ -340,4 +340,190 @@ protected function get_db_infos() {
...
+			remove_filter( 'pll_inserted_term_language', $this->get_slug_management_language_callback( $language ) );
+			remove_filter( 'pll_inserted_term_parent', $this->get_slug_management_parent_callback( $parent ) );
+			$enabled = false;
+		} else {
+			// Set term parent and language for suffixed slugs.
+			add_filter( 'pll_inserted_term_language', $this->get_slug_management_language_callback( $language ) );
+			add_filter( 'pll_inserted_term_parent', $this->get_slug_management_parent_callback( $parent ) );
+			$enabled = true;
+		}
+	}
```
Chouby, line 478: Wouldn't this be much simpler?
```suggestion
	private function toggle_inserted_term_filters( $language, $parent ): void {
		static $callback = array();

		if ( ! empty( $callback ) ) {
			// Clean up!
			remove_filter( 'pll_inserted_term_language', $callback[ $language ] );
			remove_filter( 'pll_inserted_term_parent', $callback[ $parent ] );
			$callback = array();
			return;
		}
		
		$callback[ $language ] = function () use ( $language ) {
			return $language;
		};

		$callback[ $parent ] = function () use ( $parent ) {
			return $parent;
		};

		// Set term parent and language for suffixed slugs.
		add_filter( 'pll_inserted_term_language', $callback[ $language ] );
		add_filter( 'pll_inserted_term_parent', $callback[ $parent ] );
	}
```

Chouby (changes requested):

include/api.php
```diff
@@ -560,6 +560,70 @@ function pll_count_posts( $lang, $args = array() ) {
...
+ * @param array $args {
+ *     Optional. Array of arguments for updating a term.
+ *
+ *     @type string              $alias_of     Slug of the term to make this term an alias of.
+ *                                             Default empty string. Accepts a term slug.
+ *     @type string              $description  The term description. Default empty string.
+ *     @type int                 $parent       The id of the parent term. Default 0.
+ *     @type string              $slug         The term slug to use. Default empty string.
+ *     @type PLL_Language|string $lang         The term language object or slug.
+ *     @type string[]            $translations The translation group to assign to the term with language slug as keys and `term_id` as values.
```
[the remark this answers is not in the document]
Hug0-Drelon: Indeed, forgot to add it...
Added in dc355a95.

include/api.php
```diff
@@ -560,6 +560,70 @@ function pll_count_posts( $lang, $args = array() ) {
...
+ * }
+ * @return array|WP_Error {
+ *     An array of the new term data, `WP_Error` otherwise.
```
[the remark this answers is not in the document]
Hug0-Drelon: Fixed in dc355a95.

include/translated-term.php
Hug0-Drelon: Good idea, done in dc355a95.
Even though I had to adapt a bit to avoid side effects in case the method is called consecutively with different languages/parents when adding the filters. Also fixed an illegal offset type with language obects.

PR #1503, opened 2024-07-01
Make `PLL_Admin_Model`'s methods available in all contexts

Screenfeed:
This PR makes `PLL_Admin_Model`'s methods, like `update_language()`, available to frontend and REST.
This is a required first step to bring languages to REST.

To prevent to create a gigournous class, sub-models have been created:
- `Model\Languages` is dedicated to... languages.
- `Model\Post_Types` and `Model\Taxonomies` regroup methods related to translated post types, translated taxonomies, and filtered taxonomies.

The original classes `PLL_Admin_Model` and `PLL_Model` are kept to prevent breaking things, especially if things like `$this->model instanceof PLL_Admin_Model` are used.

The magic method `PLL_Model::__call()` is created for backwards compatibility. Old methods are not deprecated yet but may be in a future release.
The `@method` tags in `PLL_Model`'s header tells PHPStan about the methods handled by `__call()`. They are also used by IDE's.

Note:
`update_secondary_language_terms()` keeps its `_languages` part because the name describes "terms of secondary languages", not "secondary terms of languages".


include/Models/Language_Model.php
```diff
@@ -0,0 +1,1107 @@
...
+		$this->options['language_taxonomies'] = array_merge( $known_taxonomies, $new_taxonomies );
+	}
+
+	/**
+	 * Cleans language cache.
+	 *
+	 * @since 3.7
+	 * @return void
+	 */
+	public function clean_cache(): void {
```
Hug0-Drelon, line 777: This method should have the same signature as `PLL_Model::clean_languages_cahce()` IMHO.
So we'll be able to deprecate the other and use this one instead. What do you think?

include/model.php
```diff
@@ -47,35 +67,38 @@ class PLL_Model {
...
+	 * @since 3.7 Type of parameter `$options` changed from `array` to `Options`.
 	 *
-	 * @param array $options Polylang options.
+	 * @param Options $options Polylang options.
 	 */
-	public function __construct( &$options ) {
-		$this->options = &$options;
+	public function __construct( Options &$options ) {
+		$this->options                   = &$options;
+		$this->cache                     = new PLL_Cache();
```
Hug0-Drelon, line 95: Could we have a cache per object instead of a shared one?
i.e. one for `Language_Model`, one for `PLL_Translated_Post` and another `PLL_Translated_Term`.

include/model.php
```diff
@@ -367,7 +251,7 @@ public function is_translated_post_type( $post_type ) {
 	 * @param bool $filter True if we should return only valid registered taxonomies.
 	 * @return string[] Array of registered taxonomy names for which Polylang manages languages and translations.
 	 */
-	public function get_translated_taxonomies( $filter = true ) {
+	public function get_translated_taxonomies( bool $filter = true ): array {
```
Hug0-Drelon, line 254: Type hinting might be risky on such public method, isn't it?

include/model.php
Hug0-Drelon, line 1: Since you extracted methods for taxonomies, I suggest you do the same for posts. What do you think?

Hug0-Drelon (changes requested):
About naming, I think we should take the opportunity to get rid of the "Model" term.
Polylang is *not* an MVC.
Also, with namespaces, no need to repeat ourselves in classnames.
Regarding `Language_Model` class for instance, it's more a "Repository" IMHO, what do you think to call `Languages_Repository`?

include/Models/Language_Model.php
Screenfeed: Personally I prefer without parameters, so it doesn't depend on a WP hook: this way, the class doesn't mix hooks and processes.

include/model.php
Screenfeed: The same object was used everywhere so far, any reason to separate everything?

include/model.php
Screenfeed: The only time this parameter is used is in an integration test in PLL Pro.

include/model.php
Screenfeed: [This](https://github.com/polylang/polylang/pull/1503/files#diff-725ca425bf5aa9b32e14d0012cdb8459e1ad1d73df87e9e1820f6aafb8bc1231R222)?

Screenfeed:
> About naming, I think we should take the opportunity to get rid of the "Model" term. Polylang is _not_ an MVC. Also, with namespaces, no need to repeat ourselves in classnames. Regarding `Language_Model` class for instance, it's more a "Repository" IMHO, what do you think to call `Languages_Repository`?

I removed the `_Model` suffix from the class names in [c1d9549](https://github.com/polylang/polylang/pull/1503/commits/c1d9549461c083fa147de4fc183cf548b8464b4e). Let's wait for @Chouby for the `Repository` thing.

include/model.php
Hug0-Drelon: The purpose is to remove `PLL_Model::$cache` property which is useless to this very class and only useful others.

include/model.php
Hug0-Drelon: I meant for third party, but maybe I'm too zealous.

include/model.php
Hug0-Drelon: Yes, as well as `is_translated_post_type()` and `count_posts()`.

include/Models/Language_Model.php
Hug0-Drelon: I get your point 👍

include/model.php
Screenfeed: I don't understand what you want.

Hug0-Drelon: Create a class `Filtered_Post_Types` gathering methods such as `get_translated_post_types()`, `is_translated_post_type()` and `count_posts()`.

include/model.php
Screenfeed: `PLL_Model::$cache` is used in `PLL_Model::count_posts()` :/

include/model.php
Screenfeed: The class `Filtered_Taxonomies` is not related to `get_translated_taxonomies()` and `is_translated_taxonomies()`, which stayed in `PLL_Model`.

include/model.php
Hug0-Drelon: Damn...

include/Models/Language.php
Hug0-Drelon, line 1: Can we use plural here?
We already have a class named `PLL_Language` which represent a single language object, here we can manage several.
We also might want to change prop name `language_model` for `languages_model` as well.

Hug0-Drelon (changes requested):

include/Models/Language.php
Screenfeed: Done in [f03419a](https://github.com/polylang/polylang/pull/1503/commits/f03419a4f79033d300cdb3620a68d32a8d422a11).

include/model.php
Screenfeed: Since the doc was already mentioning a boolean, I think we can keep it.

Hug0-Drelon (approved):

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @param mixed $value `term_id`, `term_taxonomy_id`, `slug`, `locale`, or `w3c` of the queried language.
+	 *                     `term_id` and `term_taxonomy_id` can be fetched for any language taxonomy.
+	 *                     /!\ For the `term_taxonomy_id`, prefix the ID by `tt:` (ex: `"tt:{$tt_id}"`),
+	 *                     this is to prevent confusion between `term_id` and `term_taxonomy_id`.
+	 * @return PLL_Language|false Language object, false if no language found.
+	 */
+	public function get( $value ) {
+		if ( is_object( $value ) ) {
+			return $value instanceof PLL_Language ? $value : $this->get( $value->term_id ); // Will force cast to PLL_Language.
+		}
```
Chouby, line 95: Hopefully this should fix the PHPStan ignored error that you modified.
```suggestion
		if ( $value instanceof PLL_Language ) {
			return $value;
		}
		
		if ( $value instanceof WP_Term ) {
			return $this->get( $value->term_id ); // Will force cast to PLL_Language.
		}
```

include/model.php
```diff
@@ -90,134 +113,48 @@ public function __construct( &$options ) {
...
+			'update_translations'                => array( $this->languages_model, 'update_translations' ),
+			'maybe_create_language_terms'        => array( $this->languages_model, 'maybe_create_language_terms' ),
+			'get_filtered_taxonomies'            => array( $this->filtered_taxonomies_model, 'get_filtered_taxonomies' ),
+			'is_filtered_taxonomy'               => array( $this->filtered_taxonomies_model, 'is_filtered_taxonomy' ),
+			'get_filtered_taxonomies_query_vars' => array( $this->filtered_taxonomies_model, 'get_query_vars' ),
+		);
 
-			$this->is_creating_language_objects = false;
+		if ( isset( $methods[ $name ] ) ) {
+			return call_user_func_array( $methods[ $name ], $arguments );
```
Chouby, line 145: Shouldn't we deprecate these methods?

include/model.php
```diff
@@ -231,10 +168,9 @@ public function set_languages_ready() {
 	 * @param string $taxonomy taxonomy name
 	 * @return void
 	 */
-	public function clean_languages_cache( $term = 0, $taxonomy = null ) {
+	public function clean_languages_cache( $term = 0, $taxonomy = null ): void {
 		if ( empty( $taxonomy ) || 'language' === $taxonomy ) {
-			delete_transient( 'pll_languages_list' );
-			$this->cache->clean();
+			$this->languages_model->clean_cache();
```
Chouby, line 173: Any reason to keep this method here and not to move it *completely* to the Languages class?

include/model.php
```diff
@@ -335,7 +219,7 @@ public function terms_clauses( $clauses, $lang ) {
 	 * @param bool $filter True if we should return only valid registered post types.
 	 * @return string[] Post type names for which Polylang manages languages and translations.
 	 */
-	public function get_translated_post_types( $filter = true ) {
+	public function get_translated_post_types( bool $filter = true ): array {
```
Chouby, line 222: I would not type the param.

include/model.php
Chouby: I am thinking as Hugo.

include/model.php
```diff
@@ -467,10 +287,10 @@ public function get_filtered_taxonomies_query_vars() {
 	 *
 	 * @phpstan-return int<0, max>
 	 */
-	public function term_exists( $term_name, $taxonomy, $parent, $language ) {
+	public function term_exists( string $term_name, string $taxonomy, int $parent, $language ): int {
```
Chouby, line 290: Same

include/translatable-object.php
```diff
@@ -75,12 +93,19 @@ abstract class PLL_Translatable_Object {
...
+	 * @param Options         $options         Polylang's options.
+	 * @param PLL_Cache       $cache           Internal non persistent cache object.
 	 *
-	 * @param PLL_Model $model Instance of `PLL_Model`, passed by reference.
+	 * @phpstan-param PLL_Cache<mixed> $cache
 	 */
-	public function __construct( PLL_Model &$model ) {
-		$this->model          = $model;
-		$this->tax_to_cache[] = $this->tax_language;
+	public function __construct( Languages_Model $languages_model, Options $options, PLL_Cache $cache ) {
```
Chouby, line 104: I would not have changed the signature. What advantage to you see?

phpstan-baseline.neon
```diff
@@ -528,7 +523,7 @@ parameters:
 		-
 			message: "#^Access to an undefined property object\\:\\:\\$term_id\\.$#"
 			count: 1
-			path: include/model.php
+			path: include/Models/Languages.php
```
Chouby, line 526: I proposed a way to fix this instead.

tests/phpunit/includes/doing-it-wrong-trait.php
```diff
@@ -34,7 +36,7 @@ public static function filter_doing_it_wrong_trigger_error() {
 	 * @return void
 	 */
 	public function doing_it_wrong_run( $function, $message = '', $version = '' ) {
-		if ( 'PLL_Model::get_languages_list()' === $function ) {
+		if ( 'WP_Syntex\Polylang\Models\Languages::get_languages_list()' === $function ) {
```
Chouby, line 39: Isn't it possible to use `use` directives in a trait?

Chouby (changes requested):

admin/admin-model.php
```diff
@@ -8,528 +8,4 @@
...
-		do_action( 'pll_update_default_lang', $slug );
-
-		// Update options
-		$this->options['default_lang'] = $slug;
-
-		$this->clean_languages_cache();
-		flush_rewrite_rules();
-	}
-}
+class PLL_Admin_Model extends PLL_Model {}
```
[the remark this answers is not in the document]
Screenfeed: Backwards compatibility.
If we remove it, there are 16 files to modify in PLL (including tests), 3 in PLL Pro, 2 in PLLWC (only in tests).
I'm not against it but it implies a lots of files to modify.
Also, the filter `pll_model` has no use anymore then.

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	/**
+	 * Deletes a language.
+	 *
+	 * @since 1.2
+	 * @since 3.7 Moved from `PLL_Admin_Model::delete_language()` to `WP_Syntex\Polylang\Models\Languages::delete()`.
+	 *
+	 * @param int $lang_id Language term_id.
+	 * @return bool
+	 */
+	public function delete( int $lang_id ): bool {
```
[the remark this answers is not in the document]
Screenfeed: It still is:

```php
function foo( int $int ): int {
	return $int;
}
var_dump( foo( '6' ) ); // int 6
```

include/model.php
Screenfeed: Yes: the one from the languages class has no parameters, whose are used only for the hooks. It's a sort of decoupling the languages class from the WP hooks.

tests/phpunit/includes/doing-it-wrong-trait.php
Screenfeed: We can. You want to it `Languages::class . '::get_list()'`?

include/model.php
```diff
@@ -47,35 +67,38 @@ class PLL_Model {
...
 
 	/**
-	 * Flag set to true during the language objects creation.
+	 * Model for the languages.
 	 *
-	 * @var bool
+	 * @var Models\Languages
 	 */
-	private $is_creating_language_objects = false;
+	public $languages_model;
```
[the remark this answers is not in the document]
Screenfeed: Done in [25b010e](https://github.com/polylang/polylang/pull/1503/commits/25b010ee0c71eafd141734a70730e1b09cd9774c).

include/model.php
```diff
@@ -47,35 +67,38 @@ class PLL_Model {
...
 
 	/**
-	 * Tells if {@see PLL_Model::get_languages_list()} can be used.
+	 * Model for taxonomies filtered by Polylang.
 	 *
-	 * @var bool
+	 * @var Models\Filtered_Taxonomies
 	 */
-	private $languages_ready = false;
+	public $filtered_taxonomies_model;
```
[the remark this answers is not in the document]
Screenfeed: Done in [25b010e](https://github.com/polylang/polylang/pull/1503/commits/25b010ee0c71eafd141734a70730e1b09cd9774c).

include/Models/Filtered_Taxonomies.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+namespace WP_Syntex\Polylang\Models;
```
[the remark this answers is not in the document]
Screenfeed: Done in [2b6becf](https://github.com/polylang/polylang/pull/1503/commits/2b6becf59e8d6f7290bc113abf019a490a1847fd).

include/Models/Filtered_Taxonomies.php
```diff
@@ -0,0 +1,82 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Model for taxonomies filtered by Polylang.
+ *
+ * @since 3.7
+ */
+class Filtered_Taxonomies {
```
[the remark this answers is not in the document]
Screenfeed: Done in [9d9e84c](https://github.com/polylang/polylang/pull/1503/commits/9d9e84c1130cec321d1c6b3fd8c7de87f0eea83e).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @since 0.5
+	 * @since 3.7 Moved from `PLL_Admin_Model::update_translations()` to `WP_Syntex\Polylang\Models\Languages::update_translations()`.
+	 *
+	 * @param string $old_slug The old language slug.
+	 * @param string $new_slug Optional, the new language slug, if not set it means that the language has been deleted.
+	 * @return void
+	 *
+	 * @phpstan-param non-empty-string $old_slug
+	 */
+	public function update_translations( string $old_slug, string $new_slug = '' ): void {
```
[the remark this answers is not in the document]
Screenfeed: Done in [3dfc419](https://github.com/polylang/polylang/pull/1503/commits/3dfc41930755ab95da948b3aaf448c3dbd52946c).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 *     name: non-empty-string,
+	 *     slug: non-empty-string,
+	 *     locale: non-empty-string,
+	 *     rtl: bool|0|1,
+	 *     term_group: int|numeric-string,
+	 *     lang_id?: int|numeric-string,
+	 *     flag?: non-empty-string
+	 * } $args
+	 */
+	private function build_metas( array $args ): string {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e5e5821](https://github.com/polylang/polylang/pull/1503/commits/e5e582124e7a35ad411fbd6642b428f50de00fdd).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @return WP_Error
+	 *
+	 * @phpstan-param array{
+	 *     locale: string,
+	 *     slug: string,
+	 *     name: string,
+	 *     flag?: string
+	 * } $args
+	 */
+	private function validate_lang( array $args, ?PLL_Language $lang = null ): WP_Error {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e5e5821](https://github.com/polylang/polylang/pull/1503/commits/e5e582124e7a35ad411fbd6642b428f50de00fdd).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @param PLL_Language|null $language   Optional. A language object. Required to update the existing terms.
+	 * @param string[]          $taxonomies Optional. List of language taxonomies to deal with. An empty value means
+	 *                                      all of them. Defaults to all taxonomies.
+	 * @return void
+	 *
+	 * @phpstan-param non-empty-string $slug
+	 * @phpstan-param non-empty-string $name
+	 * @phpstan-param array<non-empty-string> $taxonomies
+	 */
+	private function update_secondary_language_terms( string $slug, string $name, ?PLL_Language $language = null, array $taxonomies = array() ): void {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e5e5821](https://github.com/polylang/polylang/pull/1503/commits/e5e582124e7a35ad411fbd6642b428f50de00fdd).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * Stores the list in a db transient and in a `PLL_Cache` object.
+	 *
+	 * @since 3.4
+	 * @since 3.7 Moved from `PLL_Model::get_languages_from_taxonomies()` to `WP_Syntex\Polylang\Models\Languages::get_languages_from_taxonomies()`.
+	 *
+	 * @return PLL_Language[] An array of `PLL_Language` objects, array keys are the type.
+	 *
+	 * @phpstan-return list<PLL_Language>
+	 */
+	private function get_languages_from_taxonomies(): array {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e5e5821](https://github.com/polylang/polylang/pull/1503/commits/e5e582124e7a35ad411fbd6642b428f50de00fdd).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * Returns the list of existing language terms.
+	 * - Returns all terms, that are or not assigned to posts.
+	 * - Terms are ordered by `term_group` and `term_id` (see `WP_Syntex\Polylang\Models\Languages::filter_language_terms_orderby()`).
+	 *
+	 * @since 3.2.3
+	 * @since 3.7 Moved from `PLL_Model::get_language_terms()` to `WP_Syntex\Polylang\Models\Languages::get_language_terms()`.
+	 *
+	 * @return WP_Term[]
+	 */
+	private function get_language_terms(): array {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e5e5821](https://github.com/polylang/polylang/pull/1503/commits/e5e582124e7a35ad411fbd6642b428f50de00fdd).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @since 3.2.3
+	 * @since 3.7 Moved from `PLL_Model::filter_language_terms_orderby()` to `WP_Syntex\Polylang\Models\Languages::filter_language_terms_orderby()`.
+	 *            Visibility changed from `public` to `private`.
+	 *
+	 * @param  string   $orderby    `ORDERBY` clause of the terms query.
+	 * @param  array    $args       An array of term query arguments.
+	 * @param  string[] $taxonomies An array of taxonomy names.
+	 * @return string
+	 */
+	private function filter_language_terms_orderby( $orderby, $args, $taxonomies ) {
```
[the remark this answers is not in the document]
Screenfeed: Done in [e5e5821](https://github.com/polylang/polylang/pull/1503/commits/e5e582124e7a35ad411fbd6642b428f50de00fdd).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @since 3.7 Moved from `PLL_Admin_Model::build_language_metas()` to `WP_Syntex\Polylang\Models\Languages::build_metas()`.
+	 *
+	 * @param array $args {
+	 *   Arguments used to build the language metas.
+	 *
+	 *   @type string   $name       Language name (used only for display).
+	 *   @type string   $slug       Language code (ideally 2-letters ISO 639-1 language code).
+	 *   @type string   $locale     WordPress locale. If something wrong is used for the locale, the .mo files will not
+	 *                              be loaded...
+	 *   @type bool|int $rtl        True if rtl language, false otherwise.
```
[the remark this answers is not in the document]
Screenfeed: Done in [c81bae3](https://github.com/polylang/polylang/pull/1503/commits/c81bae3c414f4e933579ee21aab8028f64296547).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+		 */
+		$languages = apply_filters_deprecated( 'pll_languages_list', array( $languages, $this ), '3.4', 'pll_additional_language_data' );
+
+		if ( ! $this->are_languages_ready() ) {
+			// Do not cache an incomplete list.
+			/** @var list<PLL_Language> $languages */
+			return $languages;
+		}
+
+		/**
```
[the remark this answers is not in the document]
Screenfeed: Fixed in [52452c1](https://github.com/polylang/polylang/pull/1503/commits/52452c1302be09cea1988fe85ca2e3730280618c).

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @since 3.7 Moved from `PLL_Model::get_languages_list()` to `WP_Syntex\Polylang\Models\Languages::get_languages_list()`.
+	 *
+	 * @param array $args {
+	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`).
+	 *   @type bool   $hide_default Hides default language from the list (default to `false`).
+	 *   @type string $fields       Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 * }
+	 * @return array List of PLL_Language objects or PLL_Language object properties.
+	 */
+	public function get_languages_list( array $args = array() ): array {
```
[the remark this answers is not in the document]
Screenfeed: Done in [c624862](https://github.com/polylang/polylang/pull/1503/commits/c624862329909d41f955ac109bec395ab9bb2975).

phpstan-baseline.neon
Screenfeed: Done in [9a346d1](https://github.com/polylang/polylang/pull/1503/commits/9a346d1fdb65dcda6297c0852d15e292a3d41860).

include/Models/Languages.php
Screenfeed: Done in [9a346d1](https://github.com/polylang/polylang/pull/1503/commits/9a346d1fdb65dcda6297c0852d15e292a3d41860).

include/translatable-object.php
Screenfeed: The idea was to prevent a recursion (not storing the model in a property). But now I see that it can be achieved without changing the signature.

include/Models/Languages.php
```diff
@@ -0,0 +1,1107 @@
...
+	 * @param PLL_Translatable_Objects $translatable_objects Translatable objects registry.
+	 * @param PLL_Cache                $cache                Internal non persistent cache object.
+	 *
+	 * @phpstan-param PLL_Cache<mixed> $cache
+	 */
+	public function __construct( Options $options, PLL_Translatable_Objects $translatable_objects, PLL_Cache $cache ) {
+		$this->options              = $options;
+		$this->translatable_objects = $translatable_objects;
+		$this->cache                = $cache;
+	}
```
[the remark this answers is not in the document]
Screenfeed: As discussed, we keep it like this, for the coming container.

include/translatable-object.php
Screenfeed: Reverted the signature in [fe71443](https://github.com/polylang/polylang/pull/1503/commits/fe71443a5a677d64e9262b7a96fb13d46cd2ca4d), for now.

include/model.php
Screenfeed: Reverted in [1b6f7ed](https://github.com/polylang/polylang/pull/1503/commits/1b6f7ed2f498c60d999730b027609e01bfcffd28).

include/Models/Filtered_Taxonomies.php
```diff
@@ -0,0 +1,82 @@
...
+	/**
+	 * Return taxonomies that need to be filtered (post_format like).
+	 *
+	 * @since 1.7
+	 * @since 3.7 Moved from `PLL_Model::get_filtered_taxonomies()` to `WP_Syntex\Polylang\Models\Filtered_Taxonomies::get_filtered_taxonomies()`.
+	 *
+	 * @param bool $filter True if we should return only valid registered taxonomies.
+	 * @return string[] Array of registered taxonomy names.
+	 */
+	public function get_filtered_taxonomies( bool $filter = true ): array {
```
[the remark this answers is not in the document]
Screenfeed: Reverted in [1b6f7ed](https://github.com/polylang/polylang/pull/1503/commits/1b6f7ed2f498c60d999730b027609e01bfcffd28).

include/Models/Languages.php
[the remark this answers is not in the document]
Screenfeed: Reverted in [1b6f7ed](https://github.com/polylang/polylang/pull/1503/commits/1b6f7ed2f498c60d999730b027609e01bfcffd28).

include/model.php
Screenfeed: Reverted in [1b6f7ed](https://github.com/polylang/polylang/pull/1503/commits/1b6f7ed2f498c60d999730b027609e01bfcffd28).

include/model.php
Screenfeed: Reverted in [1b6f7ed](https://github.com/polylang/polylang/pull/1503/commits/1b6f7ed2f498c60d999730b027609e01bfcffd28).

include/Model/Languages.php
```diff
@@ -0,0 +1,1114 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param Options                  $options              Polylang's options.
+	 * @param PLL_Translatable_Objects $translatable_objects Translatable objects registry.
+	 * @param PLL_Cache                $cache                Internal non persistent cache object.
+	 *
+	 * @phpstan-param PLL_Cache<mixed> $cache
+	 */
+	public function __construct( Options $options, PLL_Translatable_Objects $translatable_objects, PLL_Cache $cache ) {
```
Chouby, line 73: Would it make sense to have a cache internal to this class instead of passing the existing `PLL_Model`'s cache?

include/Model/Languages.php
```diff
@@ -0,0 +1,1114 @@
...
+	 * @return PLL_Language|false Language object, false if no language found.
+	 *
+	 * @phpstan-param PLL_Language|WP_Term|int|string $value
+	 */
+	public function get( $value ) {
+		if ( $value instanceof PLL_Language ) {
+			return $value;
+		}
+
+		if ( $value instanceof WP_Term ) {
```
Chouby, line 99: I propose to re-introduce the comment:
```suggestion
		// Cast WP_Term to PLL_Language.
		if ( $value instanceof WP_Term ) {
```

tests/phpunit/includes/doing-it-wrong-trait.php
Chouby: No. It was a bad idea.

include/Model/Languages.php
Screenfeed: The current cache object is currently shared among the main model, languages, post types, and taxonomies models.
I think the one for the languages model can be separated without issues, but `clean_languages_cache()` will probably still need to reset all caches.

Screenfeed: Done in [c33dc1e](https://github.com/polylang/polylang/pull/1503/commits/c33dc1e28de389a649db39c6beb24b54ebf09ddc). Take a look at the hooks I modified in `PLL_Model`'s constructor.

Chouby: Reverted in 562bb4d5fe74efb868c535e1c80ec9e04a6a3682

include/translatable-object.php
```diff
@@ -79,8 +97,10 @@ abstract class PLL_Translatable_Object {
 	 * @param PLL_Model $model Instance of `PLL_Model`, passed by reference.
 	 */
 	public function __construct( PLL_Model &$model ) {
-		$this->model          = $model;
-		$this->tax_to_cache[] = $this->tax_language;
+		$this->languages_model = $model->languages;
```
Chouby, line 100:
```suggestion
		$this->languages = $model->languages;
```

include/translated-object.php
```diff
@@ -26,7 +29,7 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 	 *
 	 * @since 1.8
 	 *
-	 * @param PLL_Model $model Instance of `PLL_Model`.
+	 * @param PLL_Model $model Instance of `PLL_Model`, passed by reference.
```
Chouby, line 32: I see that you modified the comment. Wouldn't this work if we simply remove the reference instead?

include/translated-post.php
```diff
@@ -57,7 +60,7 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 	 *
 	 * @since 1.8
 	 *
-	 * @param PLL_Model $model Instance of `PLL_Model`.
+	 * @param PLL_Model $model Instance of `PLL_Model`, passed by reference.
```
Chouby, line 63: Same as above.

Chouby (changes requested):

include/translatable-object.php
Screenfeed: Done in [8b93699](https://github.com/polylang/polylang/pull/1503/commits/8b93699e5e88c947b96e9805938be8b631939b5d).

include/translated-object.php
Screenfeed: Done in [1e8dabc](https://github.com/polylang/polylang/pull/1503/commits/1e8dabc6f3f835db2659d7f5047b063e8a832151).

PR #1504, opened 2024-07-05
Add filters on Yoast integration

sebastienserre:
Should fix https://github.com/polylang/polylang-pro/issues/2172


Marianne380:
Replaced by https://github.com/polylang/polylang/pull/1513.

PR #1505, opened 2024-07-09
Introduce a REST API for languages

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/1763.
Requires [new models](https://github.com/polylang/polylang/pull/1503).
Replaces [this PR in PLL Pro](https://github.com/polylang/polylang-pro/pull/2174).

`PLL_Admin_Default_Term` has been split into two classes, so it can be used in REST context.


modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+	}
+
+	/**
+	 * Registers the routes for languages.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	public function register_routes(): void {
```
Hug0-Drelon, line 60: I wonder if we should set `$override` parameter of `register_rest_route()` to `true`, so our code takes precedence. What do you think?

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+					'permission_callback' => array( $this, 'create_item_permissions_check' ),
+					'args'                => $this->get_endpoint_args_for_item_schema( WP_REST_Server::CREATABLE ),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
+			)
+		);
+
+		register_rest_route(
+			$this->namespace,
+			"/{$this->rest_base}/(?P<code>[a-z_-]+)",
```
Hug0-Drelon, line 82: That makes me think of something, does language slug accept figures? If so here is what I suggest:
```suggestion
			"/{$this->rest_base}/(?P<code>[a-zA-Z0-9_-]+)",
```

modules/REST/V2/Languages.php
Hug0-Drelon, line 1: We should state if our controller allows batch. See `WP_REST_Posts_Controller::$allow_batch` and `$args['allow_batch']` in `register_rest_route()`.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+					'callback'            => array( $this, 'delete_item' ),
+					'permission_callback' => array( $this, 'delete_item_permissions_check' ),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
+			)
+		);
+	}
+
+	/**
+	 * Retrieves a collection of languages.
```
Hug0-Drelon, line 115: As of today, we do not provide collection parameters, so the comment is not quite accurate I think. Here is what I propose instead:
```suggestion
	 * Retrieves all languages.
```
Also I wonder what happens for deactivated languages? And if we should show them or  not here.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+use WP_REST_Server;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Allows to access languages via the REST API.
+ *
+ * @since 3.7
+ *
+ * @phpstan-type Fields array{
```
Hug0-Drelon, line 24:
```suggestion
 * @phpstan-type EditableFields array{
```
Otherwise I might think it can be field to retrieve from a request. Like the `_filed=x` in WP routes.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+	 * @since 3.7
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 *
+	 * @phpstan-template T of Fields
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function create_item( $request ) {
+		if ( ! empty( $request['id'] ) ) {
```
Hug0-Drelon, line 147: Can this happen?
I would expect it to be WordPress business to filter out such requests.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+		 *     slug: string,
+		 *     rtl: 0|1,
+		 *     flag: string,
+		 *     term_group: 0|1
+		 * } $prepared
+		 */
+		$result = $this->model->languages_model->add( $prepared );
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
```
Hug0-Drelon, line 200: Should we take care of error codes here, so it's more suitable for REST.
Like you've done in other places with `rest_invalid_locale` for instance and a proper status.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 *
+	 * @phpstan-template T of Fields
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function get_item( $request ) {
+		$language = $this->get_language( $request['code'] );
+
+		if ( is_wp_error( $language ) ) {
+			return $language;
```
Hug0-Drelon, line 225: Same question as above regarding the error codes and stati.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+
+		if ( is_wp_error( $result ) ) {
+			return $result;
+		}
+
+		/** @var PLL_Language */
+		$language = $this->model->languages_model->get( $prepared['slug'] );
+		$response = $this->prepare_item_for_response( $language, $request );
+
+		return rest_ensure_response( $response );
```
Hug0-Drelon, line 207:
```suggestion
		return rest_ensure_response( $response );
```
As we said, this is not needed IMHO.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+	public function get_item( $request ) {
+		$language = $this->get_language( $request['code'] );
+
+		if ( is_wp_error( $language ) ) {
+			return $language;
+		}
+
+		$response = $this->prepare_item_for_response( $language, $request );
+
+		return rest_ensure_response( $response );
```
Hug0-Drelon, line 230:
```suggestion
		return $response;
```
As we said, this is not needed IMHO.

modules/REST/V2/Languages.php
Hug0-Drelon, line 1: Everywhere we return a `WP_Error`, I suggest to use `rest_convert_error_to_response()` so we return only one type of object (i.e. `WP_REST_Response`) through *all* methods of the controller.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+	 * @since 3.7
+	 *
+	 * @param PLL_Language    $item    Language object.
+	 * @param WP_REST_Request $request Request object.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 *
+	 * @phpstan-template T of Fields
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function prepare_item_for_response( $item, $request ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
```
Hug0-Drelon, line 406: Should we use `rest_is_field_included()` here? Like WP does.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+					'default'     => true,
+				),
+			),
+		);
+
+		$taxonomies = $this->model->translatable_objects->get_taxonomy_names( array( 'language' ) );
+
+		foreach ( $taxonomies as $taxonomy ) {
+			if ( 'language' === $taxonomy ) {
+				$description = __( 'Properties for posts, pages, etc in this language.', 'polylang' );
```
Hug0-Drelon, line 627:
```suggestion
				$description = __( 'Properties for post types translated by Polylang in this language.', 'polylang' );
```
What do you think?

tests/phpunit/tests/test-rest-languages.php
```diff
@@ -0,0 +1,243 @@
...
+
+		$values = array(
+			'locale'    => 'fr_BE',
+			'name'      => 'François',
+			'new_code'  => 'fra',
+			'direction' => 'rtl',
+			'flag'      => 'be',
+			'order'     => 22,
+		);
+		$request = new WP_REST_Request( 'PATCH', '/pll/v2/languages/fr' );
```
Hug0-Drelon, line 179: We should also test with `PUT` IMHO.
One of these accept only some fields.

tests/phpunit/tests/test-rest-languages.php
```diff
@@ -0,0 +1,243 @@
...
+		$this->assertIsArray( $data );
+		$this->assertArrayHasKey( 'deleted', $data );
+		$this->assertTrue( $data['deleted'] );
+		$this->assertArrayHasKey( 'previous', $data );
+		$this->assertIsArray( $data['previous'] );
+
+		$this->assertFalse( $this->pll_env->model->get_language( 'fr_FR' ) );
+	}
+
+	public function test_permissions() {
```
Hug0-Drelon, line 216: Maybe we should have a test for `403` as well, what do you think?

tests/phpunit/tests/test-rest-languages.php
Hug0-Drelon, line 1: Maybe some days we should add more test for error paths.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+					array( 'status' => 400 )
+				);
+			}
+
+			$language = $languages[ $request['locale'] ];
+			$defaults = array(
+				'name'       => $language['name'],
+				'slug'       => $language['code'],
+				'rtl'        => (int) ( 'rtl' === $language['dir'] ),
+				'flag'       => $language['flag'],
```
Hug0-Drelon, line 175: I wonder if `flag` should be a proper path to a custom flag here.
Maybe that's something to discuss with the other devs.

Hug0-Drelon (changes requested):

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+			/**
+			 * @phpstan-var array{
+			 *     lang_id: int,
+			 *     locale: string,
+			 *     name: string,
+			 *     slug: string,
+			 *     rtl: 0|1,
+			 *     term_group: int
+			 * } $prepared
+			 */
```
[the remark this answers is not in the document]
Screenfeed: Because:

```
Parameter #1 $args of method WP_Syntex\Polylang\Models\Languages::update() expects array{lang_id: int|numeric-string, name: string, ...}, non-empty-array given.
```

It seems difficult to properly tag `prepare_item_for_database()`.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,737 @@
...
+			$update = $this->model->languages_model->update( $prepared );
+
+			if ( is_wp_error( $update ) ) {
+				return $update;
+			}
+
+			/** @var PLL_Language */
```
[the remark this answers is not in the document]
Screenfeed: We know the language exists (`$this->get_language()`), and the update went well `$update = $this->model->languages_model->update()`. Do you think we can still get `false` here?

modules/REST/API.php
```diff
@@ -0,0 +1,56 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Setup the REST API endpoints and filters.
+ *
+ * @since 3.7
+ *
+ * @phpstan-import-type Fields from V2\Languages as RestFields
+ * @phpstan-template T of RestFields
```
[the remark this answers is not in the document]
Screenfeed: This is because of the stubs:
```php
    /**
     * @phpstan-template T of array
     * @phpstan-implements ArrayAccess<key-of<T>, value-of<T>>
     */
    class WP_REST_Request implements \ArrayAccess
    {
```

modules/REST/V2/Languages.php
Screenfeed: We never used it so far. I'm not against but I'm not sure in which cases it could be useful though 🤔

modules/REST/V2/Languages.php
Screenfeed: It's the pattern we use for the slug: https://github.com/polylang/polylang/blob/e647cbe80b72416ed00b172c7b0a9cd078cb59c0/admin/admin-model.php#L367.

modules/REST/V2/Languages.php
Screenfeed: There's no reason to, the said methods extend parent methods from `WP_REST_Controller`, which return `WP_REST_Response|WP_Error`.

modules/REST/V2/Languages.php
Screenfeed: Done in [d36a21f](https://github.com/polylang/polylang/pull/1505/commits/d36a21f9dbdce3e81d4bb91085d9e9d18b0db310).

modules/REST/V2/Languages.php
Screenfeed: Done in [946eb0e](https://github.com/polylang/polylang/pull/1505/commits/946eb0e2e47f6c42d8414315f58dd79ed43efcdc).

modules/REST/V2/Languages.php
Screenfeed: This can happen currently. What would you do to tell WP it's forbidden?

modules/REST/V2/Languages.php
Screenfeed: OK for the status code.
However for the error code I don't think it's a good idea, we would loose information. Or maybe we could prefix the code with `rest_`? I'm not sure this worth it.
`add()` can return these errors:
```
'pll_invalid_locale'  => 'Enter a valid WordPress locale'
'pll_invalid_slug'    => 'The language code contains invalid characters'
'pll_non_unique_slug' => 'The language code must be unique'
'pll_invalid_name'    => 'The language must have a name'
'pll_invalid_flag'    => 'The flag does not exist'
'pll_add_language'    => 'Impossible to add the language.' (wp_insert_term() failed)
```

What do you think, replace `pll_` by `rest_`?

modules/REST/V2/Languages.php
Screenfeed: I have no idea what it is used for 😂

modules/REST/API.php
```diff
@@ -0,0 +1,56 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Model $model Polylang's model.
+	 */
+	public function __construct( PLL_Model $model ) {
+		$this->model = $model;
+		add_action( 'rest_api_init', array( $this, 'init' ) );
```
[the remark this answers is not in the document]
Screenfeed: I have a slight preference for adding the hooks from within the class, but I have no strong position on this.
As an alternative we could do this, but it's probably too much.
```php
class API {
    public function __construct( PLL_Model $model ) {} // Just sets the model.
    public function init(): self {}                    // Adds hooks.
    public function rest_api_init(): void {}           // Instantiates `V2\Languages()`.
}
```

tests/phpunit/tests/test-rest-languages.php
```diff
@@ -0,0 +1,243 @@
...
+
+		$links_model = self::$model->get_links_model();
+		$links_model->init();
+
+		$this->server  = $wp_rest_server = new Spy_REST_Server();
+		$this->pll_env = new PLL_REST_Request( $links_model );
+
+		$this->pll_env->rest         = new API( $this->pll_env->model );
+		$this->pll_env->default_term = new PLL_Admin_Default_Term( $this->pll_env );
+		$this->pll_env->default_term->add_hooks();
```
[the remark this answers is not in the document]
Screenfeed: ??

tests/phpunit/tests/test-rest-languages.php
Screenfeed: Done in [01cc7b0](https://github.com/polylang/polylang/pull/1505/commits/01cc7b08079755bc0d0840c47fa1e005c34a863f).

tests/phpunit/tests/test-rest-languages.php
Screenfeed: Done in [418d7a0](https://github.com/polylang/polylang/pull/1505/commits/418d7a081fb0670b016e85b4ddf7ca89d4eee83e).

modules/REST/V2/Languages.php
Screenfeed: I have no idea about what it is used for 😅

Screenfeed: After our conversation, we'll see that later.

modules/REST/V2/Languages.php
Screenfeed: To be discussed with the team.

modules/REST/V2/Languages.php
Screenfeed: To be discussed with the team.

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Following our discussion, it seems unlikely to get `false`, so it seems simpler (and easier to maintain) to keeps things as they are.

modules/REST/API.php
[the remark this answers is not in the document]
Screenfeed: OK, it was necessary because `V2\Languages` was declaring the template at the class level:
```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------- 
  Line   modules/REST/API.php                                                                                                                    
 ------ ---------------------------------------------------------------------------------------------------------------------------------------- 
  :28    Property WP_Syntex\Polylang\REST\API::$languages with generic class WP_Syntex\Polylang\REST\V2\Languages does not specify its types: T  
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
```
But since it is declared at the method level now, it is useless.
Removed in [273acf3](https://github.com/polylang/polylang/pull/1505/commits/273acf30cb40711424df79dd965860c5713ff258).

modules/REST/V2/Languages.php
Screenfeed: Done in [8610884](https://github.com/polylang/polylang/pull/1505/commits/8610884aa391bcd85d906bc6c16f51663210afc5).

modules/REST/V2/Languages.php
Screenfeed: Done in [4d6e713](https://github.com/polylang/polylang/pull/1505/commits/4d6e7137232a89b55be4b13a7c9d4a4c300642d5).

modules/REST/API.php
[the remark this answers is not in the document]
Screenfeed: Moved in [8783377](https://github.com/polylang/polylang/pull/1505/commits/8783377866744e5ab196faf60944dcb0970444ea).

tests/phpunit/tests/test-rest-languages.php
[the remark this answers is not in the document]
Screenfeed: Yes. Done in [b14d1e6](https://github.com/polylang/polylang/pull/1505/commits/b14d1e6cb09920bdfc372e2c519a494af20c2ac6).

modules/REST/V2/Languages.php
Screenfeed: Not that it brings much, but done in b14d1e6.

frontend/frontend.php
```diff
@@ -116,6 +121,9 @@ public function init() {
 
 		$this->links = new PLL_Frontend_Links( $this );
 
+		$this->default_term = new PLL_Default_Term( $this );
```
Hug0-Drelon, line 124: Not sur it's needed on front.
Could we just use `PLL_Admin_Default_Term()` in REST?

phpstan.neon.dist
```diff
@@ -57,3 +57,9 @@ parameters:
 			message: "#^Parameter \\#1 \\$callback of function call_user_func_array expects callable\\(\\): mixed, array\\{.+\\} given\\.$#"
 			count: 1
 			path: include/model.php
+
+		# Ignored because PHPStan seems confused about WP_REST_Request's template.
```
Hug0-Drelon, line 61: I think it legit, indeed `$request['code']` *can* return `null`.
It's the same as doing `$request->get_param( 'code ' )`, which returns `null` if `code` isn't set in the request.

Hug0-Drelon (changes requested):

phpstan.neon.dist
Screenfeed: It can't be `null` if it's in the route. Right?

frontend/frontend.php
Screenfeed: That's something I discussed with @Chouby and we decided to split the class for the 2 contexts.

phpstan.neon.dist
Hug0-Drelon: Well, we have to be sure to have a 404 error if we do a `PUT` request on `pll/v2/languages/` without code.
Iff so, I agree with you.

Screenfeed: I added some failure tests in [58e7fa9](https://github.com/polylang/polylang/pull/1505/commits/58e7fa9aada7a5433e1e6965baee94457812b418).

include/default-term.php
```diff
@@ -0,0 +1,226 @@
...
+	 * @var PLL_Model
+	 */
+	protected $model;
+
+	/**
+	 * Preferred language to assign to new contents.
+	 *
+	 * @var PLL_Language|null
+	 */
+	protected $curlang;
```
Hug0-Drelon, line 25: What made you change `$ref_lang` for `$curlang`?
It's not the same thing in an admin context I think.

modules/REST/API.php
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	/**
+	 * Init filters and new endpoints.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	public function rest_api_init(): void {
```
Hug0-Drelon, line 48:
```suggestion
	public function init(): void {
```
No need to repeat ourselves IMHO.

modules/REST/API.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @since 3.7
+	 *
+	 * @param PLL_Model $model Polylang's model.
+	 */
+	public function __construct( PLL_Model $model ) {
+		$this->model = $model;
+	}
+
+	/**
+	 * Init filters and new endpoints.
```
Hug0-Drelon, line 42:
```suggestion
	 * Adds hooks and registers endpoints.
```

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,767 @@
...
+/**
+ * Languages REST controller.
+ *
+ * @since 3.7
+ */
+class Languages extends WP_REST_Controller {
+	/**
+	 * @var PLL_Model
+	 */
+	public $model;
```
Hug0-Drelon, line 28: I suggest to have `$languages` instead.
And get rid of `$model->translatable_objects` usage later on.
Instead we could store taxonomies directly in the constructor and use them we need (in only one place actually).
Then the code would be simpler IMHO, only `$this->languages` all along.

What do you think?

tests/phpunit/tests/test-rest-languages.php
```diff
@@ -0,0 +1,338 @@
...
+	/**
+	 * @testWith ["PUT"]
+	 *           ["DELETE"]
+	 *
+	 * @param string $method
+	 * @return void
+	 */
+	public function test_invalid_param( string $method ) {
+		wp_set_current_user( self::$administrator );
+		self::create_language( 'fr_FR' );
```
Hug0-Drelon, line 301: Please use the factory.

Hug0-Drelon (changes requested):

include/default-term.php
Screenfeed: But `$ref_lang` doesn't mean anything in non-admin context, so I wasn't sure which one to use. In doubt, I guess we should keep `$ref_lang` 🤔

Hug0-Drelon: Sorry, I meant `$pref_lang`... 😅

Hug0-Drelon: But I get what you say. Now I don't know what's better :/

modules/REST/API.php
Screenfeed: You're right. Done in [acce644](https://github.com/polylang/polylang/pull/1505/commits/acce644159f1c493703b79be153fce4676bdc850).

modules/REST/V2/Languages.php
Screenfeed: Good catch. Done in [5bdc9f0](https://github.com/polylang/polylang/pull/1505/commits/5bdc9f0c3157a1ac76c0f84ff6bb3a73392599a3).

tests/phpunit/tests/test-rest-languages.php
Screenfeed: Done in [898088f](https://github.com/polylang/polylang/pull/1505/commits/898088fc25c51c88bd330f763601683fcc3d8ee9).

include/default-term.php
Screenfeed: And since the current language is set to it, that's why I changed to `$curlang`.

Hug0-Drelon (approved):

Chouby (changes requested):

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+	 */
+	public function register_routes(): void {
+		register_rest_route(
+			$this->namespace,
+			"/{$this->rest_base}",
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_items' ),
+					'permission_callback' => '__return_true',
```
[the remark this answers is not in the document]
Hug0-Drelon: So any third party won't be able to use languages in REST?
Like a headless theme or a externalized CMS solution for instance?

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+				),
+				array(
+					'methods'             => WP_REST_Server::CREATABLE,
+					'callback'            => array( $this, 'create_item' ),
+					'permission_callback' => array( $this, 'create_item_permissions_check' ),
+					'args'                => $this->get_endpoint_args_for_item_schema( WP_REST_Server::CREATABLE ),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
+			),
+			true
```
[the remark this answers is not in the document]
Screenfeed: If they do it after we create it, they will still have the choice to merge or override, won't they? Doesn't this argument say that WE override a pre-existing route?

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+	 * @param int      $status_code Optional. A status code. Default is 400.
+	 * @return WP_Error
+	 */
+	private function restify_wp_error( WP_Error $error, int $status_code = 400 ): WP_Error {
+		return new WP_Error(
+			(string) preg_replace( '@^pll_@', 'rest_', (string) $error->get_error_code() ),
+			$error->get_error_message(),
+			array( 'status' => $status_code )
+		);
+	}
```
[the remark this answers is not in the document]
Screenfeed: We thought that standardizing the error codes would be good.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.7
+	 *
+	 * @param Languages_Model          $languages            Languages' model.
+	 * @param PLL_Translatable_Objects $translatable_objects Translatable objects registry.
+	 */
+	public function __construct( Languages_Model $languages, PLL_Translatable_Objects $translatable_objects ) {
+		$this->namespace            = 'pll/v2';
```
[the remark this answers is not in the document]
Screenfeed: Because the v1 exists in Pro and returns different data.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+				$prepared = array_merge(
+					array(
+						'locale'     => $language->locale,
+						'name'       => $language->name,
+						'rtl'        => (bool) $language->is_rtl,
+						'term_group' => $language->term_group,
+					),
+					$prepared
+				);
+			}
```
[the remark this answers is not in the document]
Screenfeed: But the part from `update_item()` and the one from `create_item()` are totally different, how would you achieve this?

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+			'locale'          => 'locale',
+			'w3c'             => 'w3c',
+			'facebook'        => 'facebook',
+			'is_rtl'          => 'direction',
+			'term_group'      => 'order',
+			'flag_code'       => 'flag',
+			'flag_url'        => 'flag_url',
+			'flag'            => 'flag_tag',
+			'custom_flag_url' => 'custom_flag_url',
+			'custom_flag'     => 'custom_flag_tag',
```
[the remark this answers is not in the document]
Screenfeed: Fait enough. Done in [548d56d](https://github.com/polylang/polylang/pull/1505/commits/548d56d4a93a147e52b4d243e4b9a9efa49b770b).

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+			'home_url'        => 'home_url',
+			'search_url'      => 'search_url',
+			'page_on_front'   => 'page_on_front',
+			'page_for_posts'  => 'page_for_posts',
+			'fallbacks'       => 'fallbacks',
+			'term_props'      => 'term_props',
+		);
+
+		foreach ( $props as $language_prop => $rest_item_prop ) {
+			if ( ! isset( $data[ $language_prop ] ) || ! rest_is_field_included( $rest_item_prop, $fields ) ) {
```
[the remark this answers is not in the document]
Screenfeed: Indeed. Done in [4955b03](https://github.com/polylang/polylang/pull/1505/commits/4955b03617bffb60d9bff1ccae972ebd8762081b).

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+					'enum'        => array( 'ltr', 'rtl' ),
+					'context'     => array( 'view', 'edit' ),
+				),
+				'order'           => array(
+					'description' => __( 'Position of the language in the language switcher.', 'polylang' ),
+					'type'        => 'integer',
+					'context'     => array( 'view', 'edit' ),
+				),
+				'flag'            => array(
+					'description' => __( 'Flag code corresponding to ISO 3166-1 (for example: en).', 'polylang' ),
```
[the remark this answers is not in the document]
Screenfeed: Done in [adf5589](https://github.com/polylang/polylang/pull/1505/commits/adf5589b1dc58f582099c67ffa3d2a357c99215f).

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+
+	/**
+	 * Registers the routes for languages.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	public function register_routes(): void {
+		register_rest_route(
```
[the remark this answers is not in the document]
Screenfeed: Done in [4dcfc73](https://github.com/polylang/polylang/pull/1505/commits/4dcfc738cda455ccdc366a56a420c5307c55f57f).

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+				/**
+				 * Filters the description to use for language term properties.
+				 *
+				 * @since 3.7
+				 *
+				 * @param string $description The description.
+				 */
+				$description = apply_filters( "pll_{$taxonomy}_properties_description", '' );
+			}
+
```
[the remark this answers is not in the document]
Screenfeed: Good idea. Done in [4ea43de](https://github.com/polylang/polylang/pull/1505/commits/4ea43def63cdce464ae2e4771c9716abe1b0eb37).
However, I didn't use the taxonomy description because it doesn't seem fit for this. Instead I created a dedicated class method.

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Currently we use `__return_true` in PLL Pro.

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+					'args'                => $this->get_endpoint_args_for_item_schema( WP_REST_Server::CREATABLE ),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
+			),
+			true
+		);
+
+		register_rest_route(
+			$this->namespace,
+			"/{$this->rest_base}/(?P<code>[a-z_-]+)",
```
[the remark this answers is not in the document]
Screenfeed: Isn't that similar to what [WP does](https://github.com/WordPress/WordPress/blob/cd610922d012929786bf04d8a58cfea6b51ce7a4/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L114-L119)?

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Done in [8a02f54](https://github.com/polylang/polylang/pull/1505/commits/8a02f54ccee133b58fa8a91553d975af9dbc07f5).

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Done in [7465015](https://github.com/polylang/polylang/pull/1505/commits/7465015de130371cce97cc9bd056c34a1a1774c6).

modules/REST/V2/Languages.php
```diff
@@ -0,0 +1,775 @@
...
+				'args'   => array(
+					'code' => array(
+						'description' => __( 'Unique code for the language.', 'polylang' ),
+						'type'        => 'string',
+					),
+				),
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_item' ),
+					'permission_callback' => '__return_true',
```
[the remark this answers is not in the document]
Screenfeed: Done in [7465015](https://github.com/polylang/polylang/pull/1505/commits/7465015de130371cce97cc9bd056c34a1a1774c6).

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Done in [fd1913a](https://github.com/polylang/polylang/pull/1505/commits/fd1913a1eeceb41973bf4c8357b2e5d9428927eb).

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Done in [b43909c](https://github.com/polylang/polylang/pull/1505/commits/b43909ce4f41d59b5128ea84ae4ec9b6ef07bf1c).

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Screenfeed: Done in [b43909c](https://github.com/polylang/polylang/pull/1505/commits/b43909ce4f41d59b5128ea84ae4ec9b6ef07bf1c).

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+					'callback'            => array( $this, 'update_item' ),
+					'permission_callback' => array( $this, 'update_item_permissions_check' ),
+					'args'                => $this->get_endpoint_args_for_item_schema( WP_REST_Server::EDITABLE ),
+				),
+				array(
+					'methods'             => WP_REST_Server::DELETABLE,
+					'callback'            => array( $this, 'delete_item' ),
+					'permission_callback' => array( $this, 'delete_item_permissions_check' ),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
```
Chouby, line 109: `allow_batch` could be added here.

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+				),
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_item' ),
+					'permission_callback' => array( $this, 'get_item_permissions_check' ),
+					'args'                => array(
+						'context' => $this->get_context_param( array( 'default' => 'view' ) ),
+					),
+				),
+				'schema' => array( $this, 'get_public_item_schema' ),
```
Chouby, line 131: `allow_batch` could be added here.

modules/REST/V2/Languages.php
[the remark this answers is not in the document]
Chouby: I suggest to proceed [as done in WordPress](https://github.com/WordPress/WordPress/blob/6.6.1/wp-includes/rest-api/endpoints/class-wp-rest-posts-controller.php#L1196) by checking `$request['id']`.

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+	 *     slug: non-empty-string
+	 * }
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function get_item( $request ) {
+		$language = $this->get_language( $request );
+
+		if ( is_wp_error( $language ) ) {
+			return $language;
+		}
```
Chouby, line 259: Shouldn't we call `add_status_to_error()`?

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+	 *     flag_code?: non-empty-string
+	 * }
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function update_item( $request ) {
+		$language = $this->get_language( $request );
+
+		if ( is_wp_error( $language ) ) {
+			return $language;
+		}
```
Chouby, line 288: Shouldn't we call `add_status_to_error()`?

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+	 *     term_id: int
+	 * }
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function delete_item( $request ) {
+		$language = $this->get_language( $request );
+
+		if ( is_wp_error( $language ) ) {
+			return $language;
+		}
```
Chouby, line 347: Shouldn't we call `add_status_to_error()`?

Chouby (changes requested):

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+					),
+				),
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_item' ),
+					'permission_callback' => array( $this, 'get_item_permissions_check' ),
+					'args'                => array(
+						'context' => $this->get_context_param( array( 'default' => 'view' ) ),
+					),
+				),
```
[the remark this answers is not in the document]
Screenfeed: We explored this possibility but in the end it just decreases readability (because we have to add some `array_merge()`), and the code is shorter by only 5 or 6 lines.

modules/REST/V1/Languages.php
Screenfeed: No, the error comes from `$this->get_language()`, which already includes the status code.

modules/REST/V1/Languages.php
Screenfeed: Same as above.

modules/REST/V1/Languages.php
Screenfeed: Same as above.

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,809 @@
...
+	public function register_routes(): void {
+		register_rest_route(
+			$this->namespace,
+			"/{$this->rest_base}",
+			array(
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_items' ),
+					'permission_callback' => array( $this, 'get_items_permissions_check' ),
+					'allow_batch'         => array( 'v1' => true ),
```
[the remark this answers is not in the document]
Screenfeed: I was sure I did it like this =_='
Fixed in [97f252f](https://github.com/polylang/polylang/pull/1505/commits/97f252fe332b2ce1b268b1286e05a17022e14184).

modules/REST/V1/Languages.php
Screenfeed: Done in [97f252f](https://github.com/polylang/polylang/pull/1505/commits/97f252fe332b2ce1b268b1286e05a17022e14184).

modules/REST/V1/Languages.php
Screenfeed: Done in [97f252f](https://github.com/polylang/polylang/pull/1505/commits/97f252fe332b2ce1b268b1286e05a17022e14184).

include/translated-post.php
```diff
@@ -373,6 +373,18 @@ public function get_untranslated( $type, PLL_Language $untranslated_in, PLL_Lang
...
+	/**
+	 * Returns the description to use for the "language properties" in the REST API.
+	 *
+	 * @since 3.7
+	 * @see WP_Syntex\Polylang\REST\V2\Languages::get_item_schema()
+	 *
+	 * @return string
+	 */
+	public function get_rest_description(): string {
+		return __( 'Properties for post types translated by Polylang in this language.', 'polylang' );
```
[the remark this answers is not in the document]
Screenfeed: Done in [60ae8e2](https://github.com/polylang/polylang/pull/1505/commits/60ae8e27a57cff220df3b9b6c7719a7454f6bf0f).

include/translated-term.php
```diff
@@ -315,6 +315,18 @@ public function set_language_in_mass( $ids, $lang ) {
...
+	/**
+	 * Returns the description to use for the "language properties" in the REST API.
+	 *
+	 * @since 3.7
+	 * @see WP_Syntex\Polylang\REST\V2\Languages::get_item_schema()
+	 *
+	 * @return string
+	 */
+	public function get_rest_description(): string {
+		return __( 'Properties for taxonomy terms in this language.', 'polylang' );
```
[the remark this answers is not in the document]
Screenfeed: Done in [60ae8e2](https://github.com/polylang/polylang/pull/1505/commits/60ae8e27a57cff220df3b9b6c7719a7454f6bf0f).

modules/REST/V1/Languages.php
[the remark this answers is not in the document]
Screenfeed: Done in [594b9d2](https://github.com/polylang/polylang/pull/1505/commits/594b9d2151e3652a1871829d96ab4f88c8b94788).

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,865 @@
...
+			'is_rtl'         => 'rtl',
+			'term_group'     => 'term_group',
+			'no_default_cat' => 'no_default_cat',
+		);
+
+		foreach ( $props as $rest_item_prop => $prepared_prop ) {
+			if ( isset( $request[ $rest_item_prop ] ) && rest_is_field_included( $rest_item_prop, $fields ) ) {
+				$prepared->$prepared_prop = $request[ $rest_item_prop ];
+			}
+		}
```
Chouby, line 790: This method prepares for the database, so I don't believe that we should deal with `get_fields_for response()`

Here is a proposal for the complete method (it merges what you did with `map_data_for_database()` which indeed does what I expect in this method). My understanding from the interface provided by the parent class is that this method must prepare the item for create or update. This method is thus expected to be directly called by `create()` and `update()`. 

Hopefully this proposal should help simplify all the PHPStan doc.

```suggestion
		if ( isset( $request['term_id'] ) ) {
			$language = $this->languages->get( (int) $request['term_id'] );
			
			if ( is_wp_error( $language ) ) {
				return $language;
			}
			
			$prepared->lang_id = $term_id;
			$prepared->slug    = $request['slug'] ?? $language->slug;
			$prepared->rtl     = $request['is_rtl'] ?? (bool) $language->is_rtl;	
			$prepared->flag    = $request['flag_code'] ?? $language->flag_code;	
		} else {
			// Allow to create a language from our default list with only the locale.
			if ( ! isset( $request['slug'], $request['name'], $request['is_rtl'] ) ) {
				$languages = include POLYLANG_DIR . '/settings/languages.php';

				if ( empty( $languages[ $request['locale'] ] ) ) {
					return new WP_Error(
						'pll_rest_invalid_locale',
						__( 'The locale is invalid.', 'polylang' ),
						array( 'status' => 400 )
					);
				}
					
				$language = (object) $languages[ $request['locale'] ];
			}		
			
			$prepared->slug = $request['slug'] ?? $language->code;
			$prepared->rtl  = $request['is_rtl'] ?? 'rtl' === $language->dir;
			$prepared->flag = $request['flag_code'] ?? $language->flag;
		}


		$prepared->name       = $request['name'] ?? $language->name;
		$prepared->locale     = $request['locale'] ?? $language->locale;
		$prepared->term_group = $request['term_group'] ?? $language->term_group ?? 0;

		if ( isset( request['no_default_cat'] ) ) {
			$prepared->no_default_cat = $request['no_default_cat'];
		}  
```

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,865 @@
...
+ *     lang_id?: int,
+ *     name?: non-empty-string,
+ *     slug?: non-empty-string,
+ *     locale?: non-empty-string,
+ *     flag?: non-empty-string,
+ *     rtl?: bool,
+ *     term_group?: int,
+ *     no_default_cat?: bool
+ * }
+ */
```
Chouby, line 81: I'm surprised by that we suddenly need **all** these different types. In my previous review, it was much simpler. Does this increased complexity come from he introduction of `map_data_for_database()`?

Chouby (changes requested):

modules/REST/V1/Languages.php
Screenfeed: It's worse. PHPStan is not smart enough.

Screenfeed: Done in [90cf12d](https://github.com/polylang/polylang/pull/1505/commits/90cf12d64ae9fc42a0f7a06374502b9b3041a49e).

IMHO it's not really better because PHPStan is completely failing:

1. For example, with your code as it is, there are new issues `$language may not be defined` (lines 800 to 808).
2. Doing `$args = (array) $prepared;`: `$args` looses `$prepared`'s "shape", `$args` is a classic array. Using `get_object_vars()` doesn't help.
3. The thing with `WP_REST_Request<T>` is making all this really difficult.

modules/REST/V1/Languages.php
Screenfeed: See my previous comment for some clarifications.

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,803 @@
...
+		 */
+		$args   = (array) $prepared;
+		$result = $this->languages->add( $args );
+
+		if ( is_wp_error( $result ) ) {
+			return $this->add_status_to_error( $result );
+		}
+
+		/** @var PLL_Language */
+		$language = $this->languages->get( $prepared->locale );
```
Chouby, line 208: This should allow to further simplify the PHPStan doc by removing https://github.com/polylang/polylang/pull/1505/files#diff-8c31521867cfd1b3643b3ef5d9d0a74ae82e7ec401b2da6ffd5b904d84f3a804R181-R189 which becomes useless
```suggestion
		$language = $this->languages->get( $prepared->locale );
```

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,803 @@
...
+		 * } $prepared
+		 * @phpstan-var array{
+		 *    locale: non-empty-string,
+		 *    slug: non-empty-string,
+		 *    name: non-empty-string,
+		 *    rtl: bool,
+		 *    term_group: int,
+		 *    flag: non-empty-string,
+		 *    no_default_cat: bool
+		 * } $args
```
Chouby, line 198: Is it possible to create a type of https://github.com/polylang/polylang/blob/ce8452c0dd20c64551d2be333f693c3fb6b4df17/include/Model/Languages.php#L144-L152, import it and use it here?

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,803 @@
...
+		 * } $prepared
+		 * @phpstan-var array{
+		 *     lang_id: int,
+		 *     locale: non-empty-string,
+		 *     slug: non-empty-string,
+		 *     name: non-empty-string,
+		 *     rtl: bool,
+		 *     term_group: int,
+		 *     flag?: non-empty-string
+		 * } $args
```
Chouby, line 270: Same as abvove

modules/REST/V1/Languages.php
```diff
@@ -0,0 +1,803 @@
...
+		 */
+		$args   = (array) $prepared;
+		$update = $this->languages->update( $args );
+
+		if ( is_wp_error( $update ) ) {
+			return $this->add_status_to_error( $update );
+		}
+
+		/** @var PLL_Language */
+		$language = $this->languages->get( $prepared->lang_id );
```
Chouby, line 280: With the same objective as above:
```suggestion
		$language = $this->languages->get( $args['lang_id'] );
```

Chouby (changes requested):

modules/REST/V1/Languages.php
Screenfeed: Possible but that's forcing the type `add()` wants, instead of providing the type `prepare_item_for_database()` returns (they're not exactly the same).

modules/REST/V1/Languages.php
Screenfeed: ?

modules/REST/V1/Languages.php
Screenfeed: Done in [c860600](https://github.com/polylang/polylang/pull/1505/commits/c86060005637a400d23e9a516817dfa8bb3e84f4).

PR #1507, opened 2024-07-10
Add translation update number in the SIte Health

sebastienserre:
The idea is to know if translations updates are missing... this can help to understand why some strings are not translated.
I'm leaving as a draft... because I've code commented... that I need to speak with @Chouby about

The commented code is listing the plugins/theme names and locale which are needing to be updated... the problem is on my test website with lots of plugins & themes... it becomes unreadable.
https://www.loom.com/share/e4a6145974994c91ae5f1fccbd95fbee

In the meantime, our customers don't have 750 plugins and 150 theme installed....

### DIFF
--- a/modules/site-health/admin-site-health.php
+++ b/modules/site-health/admin-site-health.php
@@ -422,6 +422,28 @@ class PLL_Admin_Site_Health {
 			$fields['term-no-lang']['value'] = $this->format_array( $terms_no_lang );
 		}
 
+		// translations update available
+		$translation_updates_nb = 0;
+		if ( is_countable( wp_get_translation_updates() ) ) {
+		//	$translation_updates = $this->get_translations_update_list( wp_get_translation_updates() );
+			$translation_updates_nb = sizeof( wp_get_translation_updates() );
+		}
+
+		if ( $translation_updates_nb > 0 ) {
+			$fields['translations_update'] = array(
+				'label' => __( 'Translation updates' ),
+				/* translators: the placeholder is the number of available translation update. */
+				'value' => sprintf( _n( '%d translation update is available', '%d translations update are available', $translation_updates_nb ), $translation_updates_nb ),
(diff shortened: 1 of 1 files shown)


Screenfeed (approved):
I don't like  how it's displayed but OK.

Chouby:
I'm all in for this new feature but not as is:
1. The PR adds a list of available translations updates. We should add missing translations too.
2. The PR relies on Polylang main locales. We should account for locale fallbacks too.
3. The list should be displayed vertically as, for example, the current active plugins.
4. I would add one or more separate sections dedicated to this information (I propose to debate this in a future meeting).

sebastienserre:
I've added section (WordPress / Plugins / Themes) to separate the type of updates but both columns needs "something" to be correctly exported.

In backoffice, it doesn't matter:

but once copied to clipboard we have an "undefined":
```
### pll_translation ###

translation_plugins: undefined
translation_acf-extended: A translation is missing for fr_FR .
translation_connect-your-nuki-smartlock: A translation is updatable for fr_FR .
translation_404page: A translation is missing for fr_FR .
```

In the PR I have added an equal sign but it should be better to add a  text...

```
### pll_translation ###

translation_plugins: =
translation_acf-extended: A translation is missing for fr_FR .
translation_connect-your-nuki-smartlock: A translation is updatable for fr_FR .
translation_404page: A translation is missing for fr_FR .
```
Any idea is welcome (ping @Elisa3232 )

sebastienserre:
After a week of testing, I came to the conclusion we should only list translation from wp.org.
On Premium plugin and themes, there are too many edge cases.
- Some have a Domain Path in their header... but load the translation on a complete different path
- Some have a no Domain Path
- Getting all path on `load_text_domain()` results by having a list of path with lots without any file....

I prefer having secure information in the Site Health report than something we can't count in.

PR #1514, opened 2024-07-24
Fix tests

Screenfeed:
Fixes [these tests](https://github.com/polylang/polylang/actions/runs/10073677265/job/27848225144?pr=1503).

The issue caused by languages being created in `Default_Term_Test::set_up()`, but deleted only in `PLL_UnitTestCase_Trait::wpTearDownAfterClass()`.

Also, while investigating external factors, like globals, I noticed that `PLL_UnitTestCase_Trait::tear_down()` were'nt cleaning all globals, so some globals are added to `PLL_UnitTestCase_Trait::tear_down()` and some tests from `Sync_Test` have been fixed.

[EDIT from @Hug0-Drelon ] Took the opportunity to add `$this->reset__SERVER();` in `tear_down()` when applicable (i.e. when `go_to()` is called in tests.

### DIFF
--- a/tests/phpunit/includes/testcase-trait.php
+++ b/tests/phpunit/includes/testcase-trait.php
@@ -145,6 +145,7 @@ trait PLL_UnitTestCase_Trait {
 		}
 
 		$_REQUEST = array(); // WP Cleans up only $_POST and $_GET.
+		$this->reset__SERVER();
 
 		parent::tear_down();
 	}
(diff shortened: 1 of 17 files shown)


Hug0-Drelon (approved):
LGTM, even though I cannot found why it break only in nightlies...

Chouby (changes requested):

PR #1515, opened 2024-07-30
Improvements proposal for the default term in backend

Chouby:
Context: `PLL_Admin_Default_Term` is split in #1505.

- Introduces a new filter `pll_first_language_term_column` to filter specifically the output of the first language column in the terms list table. This allows to remove the method `get_first_language_column()` in `PLL_Admin_Default_term` which duplicates the same method in `PLL_Admin_Filters_Columns`.
- Simplifies the method `get_first_language_column()`.
- Removes the constructor of `PLL_Admin_Default_Term`. The only difference with the parent constructor was the current language assigned to`$tpref_lang` instead of `$curlang` but in fact both are the same in the terms list table which is the only screen where the class acts.
- Adapts 2 tests. 

Considering new the size of `PLL_Admin_Default_Term`, we could envisage to merge it back to `PLL_Default_Term`:
- Pro: Further reduces the code size
- Con: Doesn't separate the code specific to the backend.

### DIFF
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -88,20 +88,18 @@ class PLL_Admin_Filters_Columns {
 	}
 
 	/**
-	 * Returns the first language column in the posts, pages and media library tables
+	 * Returns the first language column in posts, pages, media, categories and tags tables.
 	 *
 	 * @since 0.9
 	 *
-	 * @return string first language column name
+	 * @return string first language column name.
 	 */
 	protected function get_first_language_column() {
-		$columns = array();
-
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[] = 'language_' . $language->slug;
+			return 'language_' . $language->slug;
 		}
 
-		return empty( $columns ) ? '' : reset( $columns );
+		return '';
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


admin/admin-filters-columns.php
```diff
@@ -341,6 +335,21 @@ public function term_column( $out, $column, $term_id ) {
 			$out .= $this->links->new_term_translation_link( $term_id, $taxonomy, $post_type, $language );
 		}
 
+		if ( $column == $this->get_first_language_column() ) {
```
Screenfeed, line 338: Since we're editing this part...

```suggestion
		if ( $this->get_first_language_column() === $column ) {
```

admin/admin-filters-columns.php
```diff
@@ -341,6 +335,21 @@ public function term_column( $out, $column, $term_id ) {
...
+			$out .= sprintf( '<div class="hidden" id="lang_%d">%s</div>', intval( $term_id ), esc_html( $lang->slug ) );
+
+			/**
+			 * Filters the output of the first language column in the terms list table.
+			 *
+			 * @since 3.7
+			 *
+			 * @param string $output  First language column output.
+			 * @param int    $term_id Term ID
```
Screenfeed, line 347:
```suggestion
			 * @param int    $term_id Term ID.
```

Screenfeed (changes requested):

PR #1518, opened 2024-08-07
Improve .typos.toml

szepeviktor:
- use TAB-s to [indent lines](https://github.com/polylang/polylang/blob/62e0163522925b89572c9e8f32380b2268160b1c/.editorconfig#L11)
- allow French word only in context

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,13 +1,13 @@
 [files]
 extend-exclude = [
-    ".git/",
-    ".phpunit.result.cache",
-    "flags/??.png",
-    "node_modules/",
-    "settings/flags.php",
-    "settings/languages.php",
-    "tmp/",
-    "vendor/",
+	".git/",
+	".phpunit.result.cache",
+	"flags/??.png",
+	"node_modules/",
+	"settings/flags.php",
+	"settings/languages.php",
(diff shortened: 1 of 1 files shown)


szepeviktor:
@Chouby Friendly ping 🏓

Screenfeed:
@szepeviktor What is the advantage of moving some words from `default.extend-words` to `default.extend-ignore-re`?

szepeviktor:
In `words` "exemple" will match any occurrence of exemple.
As a regexp "Etiquette exemple" will match only that French phrase.

Screenfeed:
I see. So if we use "exemple" in another French phrase, we will have to add this phrase as a regex, but this prevents typos where we could write "exemple" instead of "example". And since `words` cannot contain space characters, we cannot keep "Etiquette exemple" in it.
Thank you for the explanation.

Screenfeed:
If you don't mind, I'll push a commit to group and sort things.

Screenfeed (approved):

szepeviktor:
OK.

Screenfeed:
And sorry for the delay Viktor.

Chouby:
Thank you @szepeviktor :)

PR #1521, opened 2024-08-13
Fix REST Languages collection

Chouby:
In #1505, we introduced a REST endpoint for languages. However the response of `get_items()` is malformed. It includes an array of `WP_REST_Response` when we expect an array of items. The call to [`prepare_response_for_collection()`](https://developer.wordpress.org/reference/classes/wp_rest_controller/prepare_response_for_collection/) was forgotten.
The corresponding test is fixed as the expectation was wrong too.

### DIFF
--- a/modules/REST/V1/Languages.php
+++ b/modules/REST/V1/Languages.php
@@ -141,13 +141,14 @@ class Languages extends WP_REST_Controller {
 	 * @phpstan-param WP_REST_Request<T> $request
 	 */
 	public function get_items( $request ) {
-		$languages = array();
+		$response = array();
 
 		foreach ( $this->languages->get_list() as $language ) {
-			$languages[] = $this->prepare_item_for_response( $language, $request );
+			$language   = $this->prepare_item_for_response( $language, $request );
+			$response[] = $this->prepare_response_for_collection( $language );
 		}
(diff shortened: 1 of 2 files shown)


modules/REST/V1/Languages.php
```diff
@@ -141,13 +141,14 @@ public function register_routes(): void {
 	 * @phpstan-param WP_REST_Request<T> $request
 	 */
 	public function get_items( $request ) {
-		$languages = array();
+		$response = array();
 
 		foreach ( $this->languages->get_list() as $language ) {
-			$languages[] = $this->prepare_item_for_response( $language, $request );
+			$language   = $this->prepare_item_for_response( $language, $request );
+			$response[] = $this->prepare_response_for_collection( $language );
```
Hug0-Drelon, line 148: Do we want to ensure we get an array here? Or is it overkill?

Hug0-Drelon (changes requested):

modules/REST/V1/Languages.php
Hug0-Drelon: As discusses in Slack, overkill.

Hug0-Drelon (approved):

PR #1529, opened 2024-09-04
Adds an event when the lang has been changed on `post_lang_choice`.

Marianne380:
See https://github.com/polylang/polylang-pro/pull/2243.

Adds a custom event `onPostLangChoice` with `lang` and `dir` as data (and use them when we can in the listener).

Moreover, fixed a small issue where the `w3c` locale was not passed as it should be in the language switcher (and so add it to the language elements).

### DIFF
--- a/include/language.php
+++ b/include/language.php
@@ -571,7 +571,12 @@ class PLL_Language extends PLL_Language_Deprecated {
 	 * @return stdClass Converted `PLL_Language` object.
 	 */
 	public function to_std_class() {
-		return (object) $this->to_array();
+		$object = (object) $this->to_array();
+
+		// Sets the w3c locale as the main locale.
+		$object->locale = $this->get_locale( 'display' );
+
+		return $object;
 	}
 
 	/**
--- a/include/language.php
+++ b/include/language.php
@@ -571,12 +571,7 @@ class PLL_Language extends PLL_Language_Deprecated {
 	 * @return stdClass Converted `PLL_Language` object.
 	 */
 	public function to_std_class() {
-		$object = (object) $this->to_array();
-
-		// Sets the w3c locale as the main locale.
-		$object->locale = $this->get_locale( 'display' );
-
-		return $object;
+		return (object) $this->to_array();
 	}
 
 	/**
--- a/include/switcher.php
+++ b/include/switcher.php
@@ -122,12 +122,13 @@ class PLL_Switcher {
 		$out   = array();
 
 		foreach ( $this->links->model->get_languages_list( array( 'hide_empty' => $args['hide_if_empty'] ) ) as $language ) {
-			$id = (int) $language->term_id;
-			$order = (int) $language->term_group;
-			$slug = $language->slug;
-			$locale = $language->get_locale( 'display' );
+			$id           = (int) $language->term_id;
+			$order        = (int) $language->term_group;
+			$slug         = $language->slug;
+			$locale       = $language->get_locale( 'display' );
+			$w3c          = $language->get_locale( 'display' );
 			$item_classes = array( 'lang-item', 'lang-item-' . $id, 'lang-item-' . esc_attr( $slug ) );
-			$classes = isset( $args['classes'] ) && is_array( $args['classes'] ) ?
+			$classes      = isset( $args['classes'] ) && is_array( $args['classes'] ) ?
 				array_merge(
 					$item_classes,
 					$args['classes']
@@ -183,7 +184,7 @@ class PLL_Switcher {
 				$first = false;
 			}
 
-			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
+			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'w3c', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
 		}
 
 		return $out;
--- a/include/switcher.php
+++ b/include/switcher.php
@@ -127,6 +127,7 @@ class PLL_Switcher {
 			$slug         = $language->slug;
 			$locale       = $language->get_locale( 'display' );
 			$w3c          = $language->get_locale( 'display' );
+			$is_rtl       = $language->is_rtl;
 			$item_classes = array( 'lang-item', 'lang-item-' . $id, 'lang-item-' . esc_attr( $slug ) );
 			$classes      = isset( $args['classes'] ) && is_array( $args['classes'] ) ?
 				array_merge(
@@ -184,7 +185,7 @@ class PLL_Switcher {
 				$first = false;
 			}
 
-			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'w3c', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
+			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'is_rtl', 'w3c', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
 		}
 
 		return $out;
--- a/include/switcher.php
+++ b/include/switcher.php
@@ -126,7 +126,6 @@ class PLL_Switcher {
 			$order        = (int) $language->term_group;
 			$slug         = $language->slug;
 			$locale       = $language->get_locale( 'display' );
-			$w3c          = $language->get_locale( 'display' );
 			$is_rtl       = $language->is_rtl;
 			$item_classes = array( 'lang-item', 'lang-item-' . $id, 'lang-item-' . esc_attr( $slug ) );
 			$classes      = isset( $args['classes'] ) && is_array( $args['classes'] ) ?
@@ -185,7 +184,7 @@ class PLL_Switcher {
 				$first = false;
 			}
 
-			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'is_rtl', 'w3c', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
+			$out[ $slug ] = compact( 'id', 'order', 'slug', 'locale', 'is_rtl', 'name', 'url', 'flag', 'current_lang', 'no_translation', 'classes', 'link_classes' );
 		}
 
 		return $out;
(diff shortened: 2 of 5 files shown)


manooweb (changes requested):

js/src/classic-editor.js
```diff
@@ -200,6 +200,11 @@ jQuery(
 
 								pll.media.resetAllAttachmentsCollections();
 							}
+						).done( function () {
+								// Creates an event once the language has been successfully changed.
+								const on_lang_change_event = new Event( 'onLangChange' );
+								document.dispatchEvent( on_lang_change_event );
```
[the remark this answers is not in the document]
manooweb: What do you think about naming the event `onPostLangChoice`?

manooweb (approved):

js/src/classic-editor.js
```diff
@@ -199,6 +199,10 @@ jQuery(
 								$( '#content_ifr' ).contents().find( 'body' ).attr( 'dir', dir );
 
 								pll.media.resetAllAttachmentsCollections();
+
+								// Creates an event once the language has been successfully changed.
+								const onPostLangChoice = new Event( 'onPostLangChoice' );
+								document.dispatchEvent( onPostLangChoice );
```
Chouby, line 205: OK to add this new event, but we should use it everywhere, including here.

Chouby (changes requested):

js/src/classic-editor.js
```diff
@@ -216,6 +203,29 @@ jQuery(
...
+						tagBox.get( id );
+					}
+				);
+
+				// Modifies the text direction
+				$( 'body' ).removeClass( 'pll-dir-rtl' ).removeClass( 'pll-dir-ltr' ).addClass( 'pll-dir-' + dir );
+				$( '#content_ifr' ).contents().find( 'html' ).attr( 'lang', lang ).attr( 'dir', dir );
+				$( '#content_ifr' ).contents().find( 'body' ).attr( 'dir', dir );
+
+				pll.media.resetAllAttachmentsCollections();
```
Hug0-Drelon, line 225: Could all this data be passed in the custom event?
Using the CustomEvent API ([see](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent)).

Hug0-Drelon (changes requested):

include/walker-dropdown.php
```diff
@@ -36,7 +36,7 @@ public function start_el( &$output, $element, $depth = 0, $args = array(), $curr
 		$output .= sprintf(
 			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
 			'url' === $value_type ? esc_url( $element->$value_type ) : esc_attr( $element->$value_type ),
-			! empty( $element->locale ) ? sprintf( ' lang="%s"', esc_attr( $element->locale ) ) : '',
+			! empty( $element->w3c ) ? sprintf( ' lang="%s"', esc_attr( $element->w3c ) ) : '',
```
Hug0-Drelon, line 39: Why did you removed the `locale` and use `w3c` instead? Sorry I might have miss something...

Marianne380: Because it's the `w3c` that should be passed normally, it's a bug introduced in an old [PR](https://github.com/polylang/polylang/pull/1039).

Moreover, it was causing a bug in our JS code, as we had WordPress and W3C locales in our HTML that we were comparing, and which were not the same.

Hug0-Drelon: I should have read the PR description correctly, sorry 🙈

include/walker-dropdown.php
```diff
@@ -34,11 +34,12 @@ class PLL_Walker_Dropdown extends PLL_Walker {
...
 		$value_type = $args['value'];
 		$output .= sprintf(
-			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
+			"\t" . '<option value="%1$s"%2$s%3$s data-pll-lang="%5$s">%4$s</option>' . "\n",
 			'url' === $value_type ? esc_url( $element->$value_type ) : esc_attr( $element->$value_type ),
 			! empty( $element->w3c ) ? sprintf( ' lang="%s"', esc_attr( $element->w3c ) ) : '',
 			selected( isset( $args['selected'] ) && $args['selected'] === $element->$value_type, true, false ),
-			esc_html( $element->name )
+			esc_html( $element->name ),
+			esc_html( wp_json_encode( $element ) )
```
Hug0-Drelon: If `$element` is an instance of  `PLL_Language`, we should use the method `to_array()` and re-cast into object IMHO.

Marianne380: As discussed, already done [here](https://github.com/polylang/polylang/blob/8b4d211152b3e5bd0037d1639d6537bc35303f5e/include/walker.php#L40).

Hug0-Drelon (approved):

include/switcher.php
```diff
@@ -122,12 +122,13 @@ protected function get_elements( $args ) {
...
 		foreach ( $this->links->model->get_languages_list( array( 'hide_empty' => $args['hide_if_empty'] ) ) as $language ) {
-			$id = (int) $language->term_id;
-			$order = (int) $language->term_group;
-			$slug = $language->slug;
-			$locale = $language->get_locale( 'display' );
+			$id           = (int) $language->term_id;
+			$order        = (int) $language->term_group;
+			$slug         = $language->slug;
+			$locale       = $language->get_locale( 'display' );
+			$w3c          = $language->get_locale( 'display' );
```
Chouby, line 129: It's a bit strange to add a 2nd property with the same value as an existing one. Why?
If I well understood the issue, it's not here that it should be fixed.

include/walker-dropdown.php
```diff
@@ -34,11 +34,12 @@ class PLL_Walker_Dropdown extends PLL_Walker {
...
 		$output .= sprintf(
-			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
+			"\t" . '<option value="%1$s"%2$s%3$s data-pll-lang="%5$s">%4$s</option>' . "\n",
 			'url' === $value_type ? esc_url( $element->$value_type ) : esc_attr( $element->$value_type ),
-			! empty( $element->locale ) ? sprintf( ' lang="%s"', esc_attr( $element->locale ) ) : '',
+			! empty( $element->w3c ) ? sprintf( ' lang="%s"', esc_attr( $element->w3c ) ) : '',
 			selected( isset( $args['selected'] ) && $args['selected'] === $element->$value_type, true, false ),
-			esc_html( $element->name )
+			esc_html( $element->name ),
+			esc_html( (string) wp_json_encode( $element ) )
```
Chouby, line 42: Outputting the complete language is a bit heavy. Moreover depending on what's been passed to the walker, we won't have the properties. Shouldn't we pass only useful properties?

include/walker-dropdown.php
```diff
@@ -34,11 +34,12 @@ class PLL_Walker_Dropdown extends PLL_Walker {
 	public function start_el( &$output, $element, $depth = 0, $args = array(), $current_object_id = 0 ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UnusedVariable
 		$value_type = $args['value'];
 		$output .= sprintf(
-			"\t" . '<option value="%1$s"%2$s%3$s>%4$s</option>' . "\n",
+			"\t" . '<option value="%1$s"%2$s%3$s data-pll-lang="%5$s">%4$s</option>' . "\n",
```
Chouby, line 37: I don't believer we need a prefix here:
```suggestion
			"\t" . '<option value="%1$s"%2$s%3$s data-lang="%5$s">%4$s</option>' . "\n",
```

js/src/classic-editor.js
```diff
@@ -216,6 +207,31 @@ jQuery(
 			}
 		);
 
+		// Listen to `onPostLangChoice` to perform actions after the language has been changed.
+		document.addEventListener(
+			'onPostLangChoice',
+			( e ) => {
+				// Update the old language with the new one to be able to compare it in the next changing.
+				initializeLanguageOldValue();
+				// modifies the language in the tag cloud
```
Chouby, line 216:
```suggestion

				// Modifies the language in the tag cloud.
```

js/src/classic-editor.js
```diff
@@ -216,6 +207,31 @@ jQuery(
...
+				$( '.tagcloud-link' ).each(
+					function () {
+						var id = $( this ).attr( 'id' );
+						tagBox.get( id );
+					}
+				);
+
+				let dir = e.detail.lang.is_rtl ? 'rtl' : 'ltr'
+
+				// Modifies the text direction
```
Chouby, line 226:
```suggestion
				// Modifies the text direction.
				let dir = e.detail.lang.is_rtl ? 'rtl' : 'ltr'

```

js/src/classic-editor.js
```diff
@@ -216,6 +207,31 @@ jQuery(
...
+				$( 'body' ).removeClass( 'pll-dir-rtl' ).removeClass( 'pll-dir-ltr' ).addClass( 'pll-dir-' + dir );
+				$( '#content_ifr' ).contents().find( 'html' ).attr( 'lang', e.detail.lang.locale ).attr( 'dir', dir );
+				$( '#content_ifr' ).contents().find( 'body' ).attr( 'dir', dir );
+
+				pll.media.resetAllAttachmentsCollections();
```
Chouby, line 231: This is another action which would benefit a new comment.

Chouby (changes requested):

include/switcher.php
Marianne380: Since we're getting the `w3c` [here](https://github.com/polylang/polylang/blob/c4fcd02aa52cb40eedfec8d2bc512c68e275140b/include/walker-dropdown.php#L39), we have to set it.

js/src/classic-editor.js
Marianne380: I'm not sure to understand, do you want me to add a comment to say what this line does for the media?

include/language.php
```diff
@@ -571,7 +571,12 @@ public function to_array( $context = 'display' ) {
 	 * @return stdClass Converted `PLL_Language` object.
 	 */
 	public function to_std_class() {
-		return (object) $this->to_array();
+		$object = (object) $this->to_array();
+
+		// Sets the w3c locale as the main locale.
+		$object->locale = $this->get_locale( 'display' );
```
Chouby, line 577: Please not here. There is a risk that we need to convert the object outside the case you are handling today.
This should be done in `PLL_Walker` as discussed.

Chouby (changes requested):

include/language.php
Marianne380: See https://github.com/polylang/polylang/pull/1529/commits/1654f17ed8a33f0af294b86bbb81e820377ec796

PR #1532, opened 2024-09-09
Harmonize context checks in `PLL_Admin_Sync`

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2087.

This PR harmonizes the way the context (`$_GET['from_post']`, etc) is validated in `PLL_Admin_Sync`'s methods, by introducing the method `check_context()`.
The new method is public and static, so it can be reused in PLL Pro and PLLWC.

Some tests have been fixed to set `$_GET['post_type']`.

### DIFF
--- a/modules/sync/admin-sync.php
+++ b/modules/sync/admin-sync.php
@@ -254,7 +254,7 @@ class PLL_Admin_Sync extends PLL_Sync {
 	 * @param array $post_data A post array.
 	 * @return array
 	 *
-	 * @phpstan-return array{}|array{from_post_id: int<0,max>, new_lang: string}
+	 * @phpstan-return array{}|array{from_post_id: int<0,max>, new_lang: string}|never
 	 */
 	public static function get_data_from_request( array $post_data ): array {
 		if ( ! isset( $GLOBALS['pagenow'], $_GET['_wpnonce'], $_GET['from_post'], $_GET['new_lang'], $_GET['post_type'], $post_data['post_type'] ) ) {
@@ -270,10 +270,7 @@ class PLL_Admin_Sync extends PLL_Sync {
 		}
 
 		// Capability check already done in post-new.php.
-		if ( ! wp_verify_nonce( wp_unslash( $_GET['_wpnonce'] ), 'new-post-translation' ) ) {
-			return array();
-		}
-
+		check_admin_referer( 'new-post-translation' );
 		return array(
 			'from_post_id' => $_GET['from_post'] >= 1 ? abs( (int) $_GET['from_post'] ) : 0,
 			'new_lang'     => sanitize_key( $_GET['new_lang'] ),
--- a/modules/sync/admin-sync.php
+++ b/modules/sync/admin-sync.php
@@ -26,28 +26,40 @@ class PLL_Admin_Sync extends PLL_Sync {
 	}
 
 	/**
-	 * Translate post parent if exists when using "Add new" ( translation )
+	 * Translates the post parent if it exists when using "Add new" (translation).
 	 *
 	 * @since 0.6
 	 *
-	 * @param int   $post_parent Post parent ID
-	 * @param int   $post_id     Post ID, unused
-	 * @param array $postarr     Array of parsed post data
+	 * @param int   $post_parent Post parent ID.
+	 * @param int   $post_id     Post ID, unused.
+	 * @param array $postarr     Array of parsed post data.
 	 * @return int
 	 */
 	public function wp_insert_post_parent( $post_parent, $post_id, $postarr ) {
-		if ( isset( $_GET['from_post'], $_GET['new_lang'], $_GET['post_type'] ) ) {
-			check_admin_referer( 'new-post-translation' );
-			// Make sure not to impact media translations created at the same time
-			if ( $_GET['post_type'] === $postarr['post_type'] && ( $id = wp_get_post_parent_id( (int) $_GET['from_post'] ) ) && $parent = $this->model->post->get_translation( $id, sanitize_key( $_GET['new_lang'] ) ) ) {
-				$post_parent = $parent;
-			}
+		$context_data = $this->check_context( $postarr );
+
+		if ( empty( $context_data ) ) {
+			return $post_parent;
 		}
-		return $post_parent;
+
+		// Make sure not to impact media translations created at the same time.
+		$parent_id = wp_get_post_parent_id( $context_data['from_post_id'] );
+
+		if ( empty( $parent_id ) ) {
+			return $post_parent;
+		}
+
+		$tr_parent = $this->model->post->get_translation( $parent_id, $context_data['new_lang'] );
+
+		if ( empty( $tr_parent ) ) {
+			return $post_parent;
+		}
+
+		return $tr_parent;
 	}
 
 	/**
-	 * Copy menu order, comment, ping status and optionally the date when creating a new translation
+	 * Copies menu order, comment, ping status and optionally the date when creating a new translation.
 	 *
 	 * @since 2.5
 	 *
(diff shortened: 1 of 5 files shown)


modules/sync/admin-sync.php
```diff
@@ -230,4 +250,35 @@ public function __call( $func, $args ) {
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $post_data       A post array.
+	 * @param bool  $check_post_type Optional. Whether to check the post type in `$_GET`. Default is `true`.
+	 * @return array
+	 *
+	 * @phpstan-return array{}|array{from_post_id: int, new_lang: string}|never
+	 */
+	private function check_context( array $post_data, bool $check_post_type = true ): array {
```
Hug0-Drelon, line 265: What about `check_globals()`?

Hug0-Drelon:
About removing `check_admin_referer()`, looks fine to check for `$_GET` instead as you said.
But we have to ensure we do not get a form (POST request), and have the nonce in `$_POST` instead of `$_GET`. Do you think it could happen?

modules/sync/admin-sync.php
```diff
@@ -127,14 +142,30 @@ protected function get_fields_to_sync( $post ) {
...
+			// 2nd case for quick edit.
+			$post_type = sanitize_key( $_REQUEST['post_type'] ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		}
+
+		// Make sure not to impact media translations when creating them at the same time as post
+		if ( in_array( 'post_parent', $this->options['sync'], true ) && ( ! isset( $post_type ) || $post_type !== $post->post_type ) ) {
+			unset( $postarr['post_parent'] );
+		}
+
+		$context_data = $this->check_context( (array) $post, false );
```
Hug0-Drelon, line 157: In what cases `$_REQUEST['post_type']` isn't set?

modules/sync/admin-sync.php
```diff
@@ -230,4 +250,35 @@ public function __call( $func, $args ) {
...
+	 * @return array
+	 *
+	 * @phpstan-return array{}|array{from_post_id: int, new_lang: string}|never
+	 */
+	private function check_context( array $post_data, bool $check_post_type = true ): array {
+		if ( ! isset( $GLOBALS['pagenow'], $_GET['from_post'], $_GET['new_lang'], $post_data['post_type'] ) ) {
+			return array();
+		}
+
+		if ( $check_post_type && ( empty( $_GET['post_type'] ) || $_GET['post_type'] !== $post_data['post_type'] ) ) {
```
Hug0-Drelon, line 270: Instead of having `$check_post_type`, what do you think to check `$post_data['post_type']` only when `$_GET['post_type']` is not empty?

Hug0-Drelon (changes requested):

modules/sync/admin-sync.php
Screenfeed: Seeing [this](https://github.com/polylang/polylang/blob/dcdd8e55295d95be185b2305524cae0cfc05586c/admin/admin-links.php#L100-L112), in none. This is the only place where the nonce is created, so the pos type should always be set. In that case there is no reason to not check the post type.

modules/sync/admin-sync.php
```diff
@@ -230,4 +245,41 @@ public function __call( $func, $args ) {
...
+	 * Checks the context, and returns the value of `from_post_id` and `new_lang`.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $post_data A post array.
+	 * @return array
+	 *
+	 * @phpstan-return array{}|array{from_post_id: int<0,max>, new_lang: string}
+	 */
+	private function check_context( array $post_data ): array {
```
Hug0-Drelon, line 259: If we keep this method name, then it should return a `WP_Error|array` IMHO.
Otherwise we should rename it `get_data_from_request()` or something. What do you think?

Hug0-Drelon (changes requested):

modules/sync/admin-sync.php
Screenfeed: Changed in [b46f995](https://github.com/polylang/polylang/pull/1532/commits/b46f9958e3ecf95f93599dbc87d873c8c5471465).

modules/sync/admin-sync.php
Screenfeed: OK why not, I renamed in [8f152ac](https://github.com/polylang/polylang/pull/1532/commits/8f152acd9e6556e79d3473247637a6717e892465).

Hug0-Drelon (approved):

modules/sync/admin-sync.php
```diff
@@ -269,7 +269,7 @@ private function get_data_from_request( array $post_data ): array {
 			return array();
 		}
 
-		if ( $post_data['post_type'] !== $_GET['post_type'] || ! $this->model->is_translated_post_type( $post_data['post_type'] ) ) {
+		if ( $post_data['post_type'] !== $_GET['post_type'] || ! pll_is_translated_post_type( $post_data['post_type'] ) ) {
```
Hug0-Drelon, line 272: Not sure @Chouby will be very fond of it.
I wonder if we should have a dedicated function for this.
Let's see!

Screenfeed: Are you with me or against me?!
Woman Screaming At Cat

modules/sync/admin-sync.php
```diff
@@ -230,4 +245,42 @@ public function __call( $func, $args ) {
...
+			return array();
+		}
+
+		if ( 'post-new.php' !== $GLOBALS['pagenow'] ) {
+			return array();
+		}
+
+		if ( ! is_string( $_GET['_wpnonce'] ) || ! is_numeric( $_GET['from_post'] ) || ! is_string( $_GET['new_lang'] ) ) {
+			return array();
+		}
```
Chouby, line 270: I don't believe these tests are useful.

modules/sync/admin-sync.php
```diff
@@ -230,4 +245,42 @@ public function __call( $func, $args ) {
...
+
+		if ( $post_data['post_type'] !== $_GET['post_type'] || ! pll_is_translated_post_type( $post_data['post_type'] ) ) {
```
Chouby, line 272: I prefer not to use API functions inside the core classes
```suggestion
		if ( $post_data['post_type'] !== $_GET['post_type'] || ! $this->model->is_translated_post_type( $post_data['post_type'] ) ) {
```

modules/sync/admin-sync.php
```diff
@@ -230,4 +245,42 @@ public function __call( $func, $args ) {
...
+			return array();
+		}
+
+		// Capability check already done in post-new.php.
+		if ( ! wp_verify_nonce( wp_unslash( $_GET['_wpnonce'] ), 'new-post-translation' ) ) {
+			return array();
+		}
```
Chouby, line 279: We must use `check_admin_referer()` instead. This is our nonce and thus we are expect to check it ourselves and stop the execution if someone is cheating so I believe that `wp_verify_nonce()` is not the best in this case.
```suggestion
		check_admin_referer( 'new-post-translation' );
```

modules/sync/admin-sync.php
```diff
@@ -230,4 +245,42 @@ public function __call( $func, $args ) {
...
+
+		return array(
+			'from_post_id' => $_GET['from_post'] >= 1 ? abs( (int) $_GET['from_post'] ) : 0,
+			'new_lang'     => sanitize_key( $_GET['new_lang'] ),
+		);
```
Chouby, line 284: I propose to return a `WP_Post` and a `PLL_Language`. If one of them is not valid, then return an empty array.

modules/sync/admin-sync.php
```diff
@@ -230,4 +245,42 @@ public function __call( $func, $args ) {
...
+	 * Returns some data (`from_post_id` and `new_lang`) from the current request.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $post_data A post array.
+	 * @return array
+	 *
+	 * @phpstan-return array{}|array{from_post_id: int<0,max>, new_lang: string}
+	 */
+	public static function get_data_from_request( array $post_data ): array {
```
Chouby, line 259: We use only the post type, so passing only the post type should be sufficient.

Chouby (changes requested):

modules/sync/admin-sync.php
Screenfeed: Can't use `$this` in a static method, and making the method non-static would make it difficult to reuse.

modules/sync/admin-sync.php
Screenfeed: Done in [8c89b5b](https://github.com/polylang/polylang/pull/1532/commits/8c89b5bf8ab7967b2a0e6841f5e98f600eee55a6).

modules/sync/admin-sync.php
Screenfeed: Done in [d8b274f](https://github.com/polylang/polylang/pull/1532/commits/d8b274fa1c7d2c802af63077c3fbe680f09f0fcf).

modules/sync/admin-sync.php
Screenfeed: Done in [72716d8](https://github.com/polylang/polylang/pull/1532/commits/72716d857d47221b40bbd3e18b063d5c6fcd7e6d).

modules/sync/admin-sync.php
Screenfeed: It was to prevent php warnings/errors. Done in [56b7965](https://github.com/polylang/polylang/pull/1532/commits/56b7965ca6298a1d5bed48b59863b20d47f9a06a).

modules/sync/admin-sync.php
Chouby: Ok. Let's improve that.
Since we want this method to be called from outside, I would move it outside the class. Indeed, this class is dedicated to hooks and now I prefer not to have classes mixing hooks and "true" public methods.  My proposal is `PLL_Admin_Links` as this is the class where we generate the info read in this method. I would rename it to something like `get_info_from_new_post_translation()` or `get_data_from_new_post_translation_request()` (close to your `pll_...` proposal, or something mixing these...
I would keep it non-static and use `$this->model->is_translated_post_type()` as the other methods in the class. I don't mind if the class is available only from admin side. It's already the case when we generate links.

Screenfeed: Done in [60d8ed4](https://github.com/polylang/polylang/pull/1532/commits/60d8ed4a7318c9c067bc352837f64dfd8b79ac4e).

PR #1533, opened 2024-09-09
Filter `/wp-block-editor/v1/navigation-fallback` by language when applicable.

Hug0-Drelon:
## What?
Fixes partially https://github.com/polylang/polylang-pro/issues/2211

## How?
- Allow third party to modified filterable REST routes with a new filter `pll_filtered_rest_routes`.
- Improve the code of `PLL_Filter_REST_Routes` a bit.

### DIFF
--- a/include/filter-rest-routes.php
+++ b/include/filter-rest-routes.php
@@ -42,6 +42,10 @@ class PLL_Filter_REST_Routes {
 
 		// Adds search REST endpoint.
 		$this->filtered_routes['search'] = 'wp/v2/search';
+
+		if ( $this->model->post_types->is_translated( 'wp_navigation' ) ) {
+			$this->filtered_routes['navigation-fallback'] = 'wp-block-editor/v1/navigation-fallback';
+		}
 	}
 
 	/**
--- a/include/filter-rest-routes.php
+++ b/include/filter-rest-routes.php
@@ -10,15 +10,7 @@
  */
 class PLL_Filter_REST_Routes {
 	/**
-	 * REST routes filterable by language ordered by entity type.
-	 *
-	 * @var string[]
-	 * @phpstan-var array<string, string>
-	 */
-	private $filtered_entities = array();
-
-	/**
-	 * Other REST routes filterable by language.
+	 * Cached REST routes filterable by language ordered by types.
 	 *
 	 * @var string[]
 	 * @phpstan-var array<string, string>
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-preload-paths.php
```diff
@@ -117,4 +117,27 @@ public function test_search_rest_route_is_filterable() {
...
+			'pll_get_post_types',
+			function ( $post_types ) {
+				return array_merge(
+					$post_types,
+					array( 'wp_navigation' => 'wp_navigation' )
+				);
+			}
+		);
+		$this->pll_admin->model->cache->clean( 'post_types' );
+		$this->pll_admin->block_editor->filter_rest_routes = new PLL_Filter_REST_Routes( $this->pll_admin->model );
```
Marianne380, line 135: Why do you have to re init `PLL_Filter_REST_Routes` ?

Hug0-Drelon: Because the routes are initialized in its constructor, and the test momentary adds `wp_navigation` to translated post types.

Marianne380 (approved):

include/filter-rest-routes.php
```diff
@@ -42,6 +42,10 @@ public function __construct( PLL_Model $model ) {
 
 		// Adds search REST endpoint.
 		$this->filtered_routes['search'] = 'wp/v2/search';
+
+		if ( $this->model->post_types->is_translated( 'wp_navigation' ) ) {
+			$this->filtered_routes['navigation-fallback'] = 'wp-block-editor/v1/navigation-fallback';
+		}
```
Chouby, line 48: `wp_navigation` is always translatable`.

tests/phpunit/tests/test-preload-paths.php
Chouby: We should not need to manually add `wp_navigation` as it's always translated. Would the test work as expected by instantiating `PLL_Sync_Navigation`?

Chouby (changes requested):

include/filter-rest-routes.php
Hug0-Drelon: Not in Polylang Free I'm afraid.

tests/phpunit/tests/test-preload-paths.php
Hug0-Drelon: This is the case in Polylang Pro, not Free.

Hug0-Drelon:
> Also I suggest some enhancements for this class: Move `'wp/v2/search'` assignment to `get()`. Add the filter in this method. Make `extract_filtered_rest_entities()` return the evaluated array. Keep only one property used in `get()` for the cache.

Done in [30e73ba](https://github.com/polylang/polylang/pull/1533/commits/30e73ba7f1ca5893e0dc02243325e30c2947bdf2).

PR #1546, opened 2024-09-23
Allow numbers in language code

Chouby:
`es-419` is a valid locale (Spanish for Latin America and Caribbean islands). It would make sense to use `es-419` as language code for this locale. Thus this PR allow numbers in language code. Interesting side effect: we will have the same characters allowed as in `sanitize_key()` that we use to sanitize language codes.

Furthermore, to avoid repeating ourselves, the PR introduces constants for the patterns validating the local and language slug.

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -22,7 +22,7 @@ defined( 'ABSPATH' ) || exit;
  */
 class Languages {
 	public const LOCALE_PATTERN = '^[a-z]{2,3}(?:_[A-Z]{2})?(?:_[a-z0-9]+)?$';
-	public const SLUG_PATTERN   = '^[a-z0-9_-]+$';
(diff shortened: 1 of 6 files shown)


Screenfeed (changes requested):
The same pattern is used in several places to match language code/slug, they should all be modified the same way.

By searching `[a-z_-]+`:
- In PLL: 7 matches.
- In PLL Pro: 2 matches (1 in JS).
- In PLL WC: 0 matches.

Screenfeed (approved):

Chrystll:
OK, a format like es-419 works. 
@Chouby from what I understand, 419 alone cannot be used because a language code must always start with a language identifier, and "419" alone is not recognized as a language. Is it right?

Chouby:
Yes the first character is mandatorily a lowercase letter `[a-z]` For the other characters, the same letters are accepted + numbers + `-` + `_`.

PR #1548, opened 2024-09-30
Use 2 constants for locale and slug patterns

Chouby:
Follow up of #1546.

Use 2 different constants instead of trimming the first and last characters when the pattern is included in a larger one.
See also https://github.com/polylang/polylang-pro/pull/2279#pullrequestreview-2328186474


Screenfeed (approved):
String concat, great idea 👍 
🤭

PR #1550, opened 2024-10-01
Don't use textdomains until `after_setup_theme`

Screenfeed:
Since WP 6.7-beta, any call to `__()`, `load_*_textdomains()` and `_load_textdomain_just_in_time()` before the hook `after_setup_theme` ends with a `_doing_it_wrong()` notice.

This is caused by [this commit](https://github.com/WordPress/WordPress/commit/48f12f72df8bb2c11570158d32e7fa2e6f5dcea6) about [this ticket](https://core.trac.wordpress.org/ticket/44937).

In this PR:
1. `load_plugin_textdomain()` is removed from `PLL_Admin_Base::init_user()`.

2. ~~`Options\Business\Sync::get_data_structure()` used to call `PLL_Settings_Sync::list_metas_to_sync()`, which contains several `__()`. A method `PLL_Settings_Sync::get_synchronizable_elements()` has been created, without any `__()`.~~ Moved to another PR.

3. When registering language taxonomies, don't provide labels.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -104,7 +104,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 
 		// Filter admin language for users
 		// We must not call user info before WordPress defines user roles in wp-settings.php
-		add_action( 'after_setup_theme', array( $this, 'init_user' ) );
+		add_action( 'setup_theme', array( $this, 'init_user' ) );
 		add_filter( 'request', array( $this, 'request' ) );
 
 		// Adds the languages in admin bar
@@ -468,9 +468,6 @@ abstract class PLL_Admin_Base extends PLL_Base {
 		$this->pref_lang = apply_filters( 'pll_admin_preferred_language', $this->pref_lang );
 
 		$this->set_current_language();
-
-		// Plugin i18n, only needed for backend.
-		load_plugin_textdomain( 'polylang' );
 	}
 
 	/**
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -104,7 +104,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 
 		// Filter admin language for users
 		// We must not call user info before WordPress defines user roles in wp-settings.php
-		add_action( 'setup_theme', array( $this, 'init_user' ) );
+		add_action( 'after_setup_theme', array( $this, 'init_user' ) );
 		add_filter( 'request', array( $this, 'request' ) );
 
 		// Adds the languages in admin bar
--- a/include/Options/Business/Sync.php
+++ b/include/Options/Business/Sync.php
@@ -37,14 +37,16 @@ class Sync extends Abstract_List {
 	 *
 	 * @return array Partial schema.
 	 *
-	 * @phpstan-return array{type: 'array', items: array{type: SchemaType, enum: non-empty-list<non-falsy-string>}}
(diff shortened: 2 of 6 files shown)


manooweb:
> 3. When registering language taxonomies, don't translate the labels.

If I correctly understood we won't translate these labels anymore?
Has it become useless? I don't see the impact.

include/translated-post.php
```diff
@@ -82,9 +82,9 @@ protected function register_language_taxonomy(): void {
 			(array) $this->get_translated_object_types(),
 			array(
 				'labels' => array(
-					'name'          => __( 'Languages', 'polylang' ),
-					'singular_name' => __( 'Language', 'polylang' ),
-					'all_items'     => __( 'All languages', 'polylang' ),
+					'name'          => 'Languages',
+					'singular_name' => 'Language',
+					'all_items'     => 'All languages',
```
manooweb, line 87: See my global [comment](https://github.com/polylang/polylang/pull/1550#issuecomment-2387814900)
I don't see what it really imply to remove translated labels here.

manooweb (changes requested):

include/translated-post.php
Screenfeed: As far as I can tell, these labels are not used anywhere. I think we could even remove them entirely. @Chouby doesn't recall any use of them neither.

manooweb (approved):

admin/admin-base.php
```diff
@@ -104,7 +104,7 @@ public function init() {
 
 		// Filter admin language for users
 		// We must not call user info before WordPress defines user roles in wp-settings.php
-		add_action( 'setup_theme', array( $this, 'init_user' ) );
+		add_action( 'after_setup_theme', array( $this, 'init_user' ) );
```
Chouby, line 107: Instead of doing that we may just remove our call to `load_plugin_textdomain()` in `init_user()` as this useless for a while and rely only on `_load_textdomain_just_in_time()`.

include/Options/Business/Sync.php
```diff
@@ -37,16 +37,14 @@ public static function key(): string {
...
 	 */
 	protected function get_data_structure(): array {
-		/** @phpstan-var non-empty-array<non-falsy-string> */
-		$enum = array_keys( \PLL_Settings_Sync::list_metas_to_sync() );
 		return array(
 			'type'  => 'array',
 			'items' => array(
 				'type' => $this->get_type(),
-				'enum' => $enum,
+				'enum' => \PLL_Settings_Sync::get_synchronizable_elements(),
```
Chouby, line 47: Could you move all the related changes to a separate PR? I'd like to cherry-pick this PR to the 3.6.x branch to avoid the notices in the next minor release.

Moreover, instead of duplicating code from `list_metas_to_sync()` to `get_synchronizable_elements()`, I suggest to prevent loading the translation as we don't need them anyway.
I did not tests but I expect something like this to work:
```php
add_filter( 'lang_dir_for_domain', '__return_false' ); // Prevents loading the translations too early.
remove_filter( 'lang_dir_for_domain', '__return_false' );
```

Chouby (changes requested):

include/Options/Business/Sync.php
Screenfeed: Reverted in [795fae4](https://github.com/polylang/polylang/pull/1550/commits/795fae4ca38df08eef6eb6c72059949762d7c758).

PR #1552, opened 2024-10-01
New hook `pll_upgrade`.

Hug0-Drelon:
## What?
Create a hook applied during Polylang upgrade process, motivated by https://github.com/polylang/polylang-pro/pull/2283.

### DIFF
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -114,14 +114,11 @@ class PLL_Upgrade {
 		}
 
 		/**
-		 * Fires once Polylang has been upgraded.
+		 * Fires after Polylang has been upgraded and before the new version is saved in options.
 		 *
 		 * @since 3.7
-		 *
-		 * @param $from_version Base version.
-		 * @param $to_version   New version.
 		 */
-		do_action( 'pll_upgrade', $this->options['version'], POLYLANG_VERSION );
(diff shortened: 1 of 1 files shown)


install/upgrade.php
```diff
@@ -113,6 +113,16 @@ public function _upgrade() {
...
 
+		/**
+		 * Fires once Polylang has been upgraded.
+		 *
+		 * @since 3.7
+		 *
+		 * @param $from_version Base version.
+		 * @param $to_version   New version.
+		 */
+		do_action( 'pll_upgrade', $this->options['version'], POLYLANG_VERSION );
```
Chouby, line 124: I believe that we don't need the additional parameters. They can be easily accessed by other ways if needed (as you did in Polylang Pro. What do you think?
```suggestion
		/**
		 * Fires after Polylang has been upgraded and before the new version is saved in options.
		 *
		 * @since 3.7
		 */
		do_action( 'pll_upgrade' );
```

Chouby (changes requested):

install/upgrade.php
Hug0-Drelon: Sure!
I've proposed this parameters with third party developers in mind. But the lesser the better!

PR #1555, opened 2024-10-03
Remove backward compatibility with WP < 6.2

Chouby:
- `Requests_IDNAEncoder` replaced by `WpOrg\Requests\IdnaEncoder`
- Object term cache storing ids instead of terms
- Use `WP_Block_Editor_Context::$name`

### DIFF
--- a/admin/admin-block-editor.php
+++ b/admin/admin-block-editor.php
@@ -48,11 +48,7 @@ class PLL_Admin_Block_Editor {
 			return $preload_paths;
 		}
 
-		// Backward compatibility with WP < 6.0 where `WP_Block_Editor_Context::$name` doesn't exist yet.
-		if (
-			( property_exists( $context, 'name' ) && 'core/edit-post' !== $context->name )
-			|| ! $context->post instanceof WP_Post
-		) {
+		if ( 'core/edit-post' !== $context->name ) {
 			// Do nothing if not post editor.
 			return $preload_paths;
 		}
--- a/include/links-domain.php
+++ b/include/links-domain.php
@@ -107,12 +107,8 @@ class PLL_Links_Domain extends PLL_Links_Abstract_Domain {
 			// The function idn_to_ascii() is much faster than the WordPress method.
 			if ( function_exists( 'idn_to_ascii' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 				$hosts[ $lang ] = idn_to_ascii( $host, 0, INTL_IDNA_VARIANT_UTS46 );
-			} elseif ( class_exists( 'WpOrg\Requests\IdnaEncoder' ) ) {
-				// Since WP 6.2.
-				$hosts[ $lang ] = \WpOrg\Requests\IdnaEncoder::encode( $host );
 			} else {
-				// Backward compatibility with WP < 6.2.
-				$hosts[ $lang ] = Requests_IDNAEncoder::encode( $host );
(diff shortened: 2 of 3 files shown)


admin/admin-block-editor.php
```diff
@@ -48,11 +48,7 @@ public function filter_preload_paths( $preload_paths, $context ) {
 			return $preload_paths;
 		}
 
-		// Backward compatibility with WP < 6.0 where `WP_Block_Editor_Context::$name` doesn't exist yet.
-		if (
-			( property_exists( $context, 'name' ) && 'core/edit-post' !== $context->name )
-			|| ! $context->post instanceof WP_Post
-		) {
+		if ( 'core/edit-post' !== $context->name ) {
```
Hug0-Drelon, line 51:
```suggestion
		if ( 'core/edit-post' !== $context->name || ! $context->post instanceof WP_Post ) {
```
So it behaves like before.

include/links-domain.php
```diff
@@ -107,12 +107,8 @@ public function get_hosts() {
 			// The function idn_to_ascii() is much faster than the WordPress method.
 			if ( function_exists( 'idn_to_ascii' ) && defined( 'INTL_IDNA_VARIANT_UTS46' ) ) {
 				$hosts[ $lang ] = idn_to_ascii( $host, 0, INTL_IDNA_VARIANT_UTS46 );
-			} elseif ( class_exists( 'WpOrg\Requests\IdnaEncoder' ) ) {
-				// Since WP 6.2.
-				$hosts[ $lang ] = \WpOrg\Requests\IdnaEncoder::encode( $host );
 			} else {
-				// Backward compatibility with WP < 6.2.
-				$hosts[ $lang ] = Requests_IDNAEncoder::encode( $host );
+				$hosts[ $lang ] = \WpOrg\Requests\IdnaEncoder::encode( $host ); // Since WP 6.2.
```
Hug0-Drelon, line 111:
```suggestion
				$hosts[ $lang ] = (string) \WpOrg\Requests\IdnaEncoder::encode( $host ); // Since WP 6.2.
```
According to the doc.

Hug0-Drelon (changes requested):

include/links-domain.php
Chouby: I'm not against this, but is it useful? According to the doc, we already have a string https://github.com/WordPress/wordpress-develop/blob/6.6.2/src/wp-includes/Requests/src/IdnaEncoder.php#L58

admin/admin-block-editor.php
Hug0-Drelon: Only `property_exists()` was here for backward compatibility.
Ensuring `$context->post` is a `WP_Post` is type safety.
On the other hand, `$context->name` should be `'core/edit-post'` regarding the screen to filter.
Because it could be badly set or change on the fly since it's a public prop.
See the doc [here](https://github.com/WordPress/wordpress-develop/blob/31ce381cc8a0f63af53c5f107f626763c3d29cb1/src/wp-includes/class-wp-block-editor-context.php#L39).

Hug0-Drelon: Bonus, it will remove PHPStan errors 😅

include/links-domain.php
Hug0-Drelon: I've suggested that because of the following error:
```
Method PLL_Links_Domain::get_hosts() should return array<string> but returns array<string|false>.
```
That's weird, could `wordpress-phpstan` have a different return type for it?

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

PR #1556, opened 2024-10-03
Remove phpcs:ignore WordPress.DB.PreparedSQL in Model\Languages

Chouby:
This PR reformats SQL queries to avoid `phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared` when possible.

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -884,7 +884,7 @@ class Languages {
 
 				if ( ! empty( $tr[ $old_slug ] ) ) {
 					if ( $new_slug ) {
-						$tr[ $new_slug ] = $tr[ $old_slug ]; // Suppress this for delete
+						$tr[ $new_slug ] = $tr[ $old_slug ]; // Suppress this for delete.
 					} else {
 						$dr['id'][] = (int) $tr[ $old_slug ];
 						$dr['tt'][] = (int) $term->term_taxonomy_id;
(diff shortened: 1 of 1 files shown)


include/Model/Languages.php
```diff
@@ -884,45 +884,71 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
...
-				"UPDATE $wpdb->term_taxonomy
-				SET description = ( CASE term_id " . implode( ' ', $ut['case'] ) . ' END ) ' . // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
-				'WHERE term_id IN ( ' . implode( ',', $ut['in'] ) . ' )' // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+				$wpdb->prepare(
+					sprintf(
+						"UPDATE {$wpdb->term_taxonomy} SET description = ( CASE term_id %s END ) WHERE term_id IN (%s)",
+						implode( ' ', array_fill( 0, count( $ut['case'] ), 'WHEN %d THEN %s' ) ),
+						implode( ',', array_fill( 0, count( $ut['in'] ), '%d' ) )
+					),
+					array_merge( array_merge( ...$ut['case'] ), $ut['in'] )
```
Hug0-Drelon, line 950: Is this the same result you'd expect? I'ma asking, before we lose readability here and I couldn't wrap my head around what we want to achieve here 😅 
```suggestion
					array_unique( ...$ut['case'] )+ $ut['in']
```

Hug0-Drelon (changes requested):

include/Model/Languages.php
Chouby: We are passing parameters to `sprintf()` which must be ordered correctly to match the placeholders.
Let's say we have 2 terms to update, then the arrays could look like this:
```
$ut['case'] = array
	array( 32, 'serialized value' ),
	array( 45, 'other serialized value' ),
);

$ut['in'] = array( 32, 45 );
```
We want to obtain a flat array of parameters like this:
```
array( 32, 'serialized value', 45, 'other serialized value', 32, 45 )
```
So as you see your proposal is not correct as it will remove some items.

Hug0-Drelon (approved):

PR #1558, opened 2024-10-04
Reformat SQL queries using `PLL_Db_Tools`

Chouby:
This is is an opinionated PR reformatting some SQL queries.
1. SQL queries are rewritten using directly `prepare()` instead of being split and using `PLL_Db_Tools`
2. `PLL_Db_Tools` is removed as it's not used anymore.
3. More controversial, I proposed in https://github.com/polylang/polylang/commit/f1a58005e137f0e740086b460f6043b6414585e2 that `get_objects_with_no_lang_sql()` returns the result of the query instead of the SQL query string. The purpose is to have the query done where it is prepared allow WPCS to correctly analyze the code. This however imply to change the signature of some methods.


include/translatable-object-with-types-trait.php
```diff
@@ -14,36 +14,49 @@
...
 	 * @param int   $limit        Max number of objects to return. `-1` to return all of them.
 	 * @param array $args         An array of translated object types.
-	 * @return string
+	 * @return string[]
 	 *
 	 * @phpstan-param array<positive-int> $language_ids
 	 * @phpstan-param -1|positive-int $limit
 	 * @phpstan-param array<string> $args
 	 */
 	protected function get_objects_with_no_lang_sql( array $language_ids, $limit, array $args = array() ) {
```
Hug0-Drelon, line 31:
```suggestion
	protected function get_raw_objects_with_no_lang( array $language_ids, $limit, array $args = array() ) {
```
No it's method business whether or not a custom SQL query is made.

include/translatable-object.php
```diff
@@ -391,24 +391,30 @@ public function get_objects_with_no_lang( $limit, array $args = array() ) {
...
 	 * @return string[] An array of numeric object IDs.
+	 *
+	 * @phpstan-param array<positive-int> $language_ids
+	 * @phpstan-param -1|positive-int $limit
+	 * @phpstan-param array<empty> $args
 	 */
-	protected function query_objects_with_no_lang( $sql ) {
-		$key          = md5( $sql );
+	protected function query_objects_with_no_lang( array $language_ids, $limit, array $args = array() ) {
+		$key          = md5( maybe_serialize( $language_ids ) . maybe_serialize( $args ) . $limit );
```
Hug0-Drelon, line 417:
```suggestion
		$key          = md5( (string) json_encode( $language_ids + $args ) . $limit );
```
To improve perf a bit.

include/translatable-object.php
```diff
@@ -460,30 +466,41 @@ public function sanitize_int_ids_list( $ids ) {
...
+					WHERE %%i NOT IN (
+						SELECT object_id FROM {$wpdb->term_relationships} WHERE term_taxonomy_id IN (%s)
+					)
+					LIMIT %%d",
+					implode( ',', array_fill( 0, count( $language_ids ), '%d' ) )
+				),
+				array_merge(
+					array( $db['id_column'], $db['table'], $db['id_column'] ),
+					$language_ids,
+					array( $limit >= 1 ? $limit : 4294967295 )
```
Hug0-Drelon, line 501: `$limit` could erase a value in `$language_ids` here, right?
Would it be better to concatenate (`+`) arrays instead of merging them?

Hug0-Drelon (changes requested):

include/translatable-object.php
Chouby: I prefer not to merge the 2 different arrays

include/translatable-object.php
Chouby: We need to ensure that we have keep the expected number of elements. `+` doesn't ensure that unlike `array_merge()` (because I used numeric keys).

Hug0-Drelon (approved):

PR #1559, opened 2024-10-04
Fix (most of) PHPStan errors in WPML compatibility code

Chouby:
All is in the title.

### DIFF
--- a/modules/wpml/wpml-compat.php
+++ b/modules/wpml/wpml-compat.php
@@ -40,11 +40,9 @@ class PLL_WPML_Compat {
 		require_once __DIR__ . '/wpml-legacy-api.php';
 		$this->api = new PLL_WPML_API();
 
-		self::$strings = get_option( 'polylang_wpml_strings', array() );
+		$strings = get_option( 'polylang_wpml_strings' );
 
-		if ( ! is_array( self::$strings ) ) {
-			self::$strings = array(); // In case the serialized option is corrupted.
-		}
+		self::$strings = is_array( $strings ) ? $strings : array();
 
 		add_action( 'pll_language_defined', array( $this, 'define_constants' ) );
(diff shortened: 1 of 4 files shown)


modules/wpml/wpml-compat.php
```diff
@@ -40,11 +40,9 @@ protected function __construct() {
 		require_once __DIR__ . '/wpml-legacy-api.php';
 		$this->api = new PLL_WPML_API();
 
-		self::$strings = get_option( 'polylang_wpml_strings', array() );
+		$strings = get_option( 'polylang_wpml_strings' );
 
-		if ( ! is_array( self::$strings ) ) {
-			self::$strings = array(); // In case the serialized option is corrupted.
-		}
+		self::$strings = is_array( $strings ) ? $strings : array();
```
Hug0-Drelon, line 45: Should it bail out early if the option type is not as expected? (i.e. not adding actions).

Hug0-Drelon (changes requested):

modules/wpml/wpml-compat.php
Chouby: I did not change how setting up action were grouped.
Two of them are not related and must be added anyway.

Do you prefer something like:
```php
		$strings = get_option( 'polylang_wpml_strings' );

		if ( is_array( $strings ) ) {
			self::$strings = $strings;
			add_filter( 'pll_get_strings', array( $this, 'get_strings' ) );
		}

		add_action( 'pll_language_defined', array( $this, 'define_constants' ) );
		add_action( 'pll_no_language_defined', array( $this, 'define_constants' ) );
```

Hug0-Drelon: What you suggest LGTM.

Hug0-Drelon (approved):

PR #1561, opened 2024-10-09
Introduce get_data_from_new_media_translation_request()

Chouby:
Follows up #1532 
This PR introduces the new method `PLL_Admin_Links::get_data_from_new_media_translation_request()` similar to `PLL_Admin_Links::get_data_from_new_post_translation_request()` but for media.

### DIFF
--- a/admin/admin-links.php
+++ b/admin/admin-links.php
@@ -256,25 +256,7 @@ class PLL_Admin_Links extends PLL_Links {
 
 		// Capability check already done in post-new.php.
 		check_admin_referer( 'new-post-translation' );
-
-		$post_id   = (int) $_GET['from_post'];
-		$lang_slug = sanitize_key( $_GET['new_lang'] );
-
-		if ( $post_id <= 0 || empty( $lang_slug ) ) {
-			return array();
-		}
-
-		$post = get_post( $post_id );
-		$lang = $this->model->get_language( $lang_slug );
-
-		if ( empty( $post ) || empty( $lang ) ) {
-			return array();
-		}
-
-		return array(
-			'from_post' => $post,
-			'new_lang'  => $lang,
-		);
+		return $this->get_objects_from_new_post_translation_request( (int) $_GET['from_post'], sanitize_key( $_GET['new_lang'] ) );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


admin/admin-links.php
```diff
@@ -272,4 +276,47 @@ public function get_data_from_new_post_translation_request( string $post_type ):
...
+		$lang = $this->model->get_language( $lang_slug );
+
+		if ( empty( $post ) || empty( $lang ) ) {
+			return array();
+		}
+
+		return array(
+			'from_post' => $post,
+			'new_lang'  => $lang,
+		);
```
Screenfeed, line 320: Maybe we could share this part with `get_data_from_new_post_translation_request()`.

Ex:
```php
	private function get_translation_request_data_objects( int $post_id, string $lang_slug ): array {
		if ( $post_id <= 0 || empty( $lang_slug ) ) {
			return array();
		}

		$post = get_post( $post_id );
		$lang = $this->model->get_language( $lang_slug );

		if ( empty( $post ) || empty( $lang ) ) {
			return array();
		}

		return array(
			'from_post' => $post,
			'new_lang'  => $lang,
		);
	}
```

Hug0-Drelon: The returned key should be `from_media` instead of `from_post`, am I right?

Hug0-Drelon (changes requested):

admin/admin-links.php
Screenfeed: STOP RUINING EVERYTHING! 😭

Screenfeed: On second thought, we could keep `from_post` for both, it would require only few adjustments for the media 🤔

Hug0-Drelon: After working on https://github.com/polylang/polylang-pro/pull/2274 which depends on this very PR, I think we should keep it that way.
After all a media is a post, and it allows to reuse completely what's done in `PLL_ACF_Post`.

Chouby: I intentionally kept `from_post` for media too. I'm often annoyed with thinks like post `ID` but term `term_id` or other similar things in Polylang related to early decisions.

Hug0-Drelon (approved):

Hug0-Drelon (approved):

PR #1562, opened 2024-10-09
Use parallel processing and cache for PHPCS

Chouby:
All is in the title. On my laptop, the parallel processing decreases the time to check from 38 s to 6s.
On GitHub the time goes from 32s in the previous PR to 15s in this PR. Not that much.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -5,6 +5,8 @@
 	<arg name="extensions" value="php,js" />
 	<arg name="colors"/>
 	<arg value="ps"/>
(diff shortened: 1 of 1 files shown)


Hug0-Drelon (approved):
As per this [discussion](https://wpsyntex.slack.com/archives/C017SSTK4KY/p1728466192640739), it looks promising!

manooweb:
Also tested on Polylang-Pro. Nice find 💪👍
Time divide by 4 after the first pass on my slow machine.

PR #1563, opened 2024-10-14
Update plugin updater to version 1.9.4

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/2272


Screenfeed:
image
image

PR #1565, opened 2024-10-14
Exclude `PLL_Plugin_Updater` from Typos and Rector tools

Screenfeed:
This PR prevents issues reported in [this comment](https://github.com/polylang/polylang/pull/1563#issuecomment-2410660688).

image
image

Since the class `PLL_Plugin_Updater` is not from us, but a copy/paste from EDD, the solution suggested here is to exclude it from Typos and Rector tools.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -3,6 +3,7 @@ extend-exclude = [
     ".git/",
     ".phpunit.result.cache",
     "flags/??.png",
+	"install/plugin-updater.php",
     "node_modules/",
     "settings/flags.php",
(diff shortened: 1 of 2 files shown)


szepeviktor:
> Modified version with 'polylang' text domain and missing comments for translators.

That file is already modified. Why not fix a single typo in it?

Screenfeed:
Indeed I missed this part and thought only the class name changed.

szepeviktor:
All right.
Please see #1518

PR #1569, opened 2024-11-04
Polylang's options in the REST API

Screenfeed:
This introduced Polylang's options into the REST API.

## What it does

One route: `/wp-json/pll/v1/settings`. It it available only in the `edit` context and requires the user capability `manage_options`.
This route allows to:
- Retrieve all options with `GET`. No arguments (except for the context), no collections.
- Update one or several options at once with `POST`/`PUT`/`PATCH`. Arguments are "option name"/"option value" pairs.

## About the value returned by an update

- If everything goes right, the request returns the updated options (even if the value didn't change).
- In case of an error, it returns an object representing a `WP_Error`.
  - If it's a format error, The-CMS-That-Must-Not-Be-Named blocks everything and nothing is saved. An error is returned.
  - If other cases: options without errors are still updated. Example: we want to update 4 options but 1 returns an error: 3 are saved and an error is returned.

## Schema improvements

- Add `readonly` to `first_activation`, `previous_version`, `version`. This disallows these properties to be modified in REST.
- Add `default` values.
- Remove the `context` from the main schema: since it isn't used directly and all the sub-schema include the context, it is useless and can be removed.

### DIFF
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -156,7 +156,7 @@ abstract class Abstract_Option {
 				'$schema'     => 'http://json-schema.org/draft-04/schema#',
 				'title'       => static::key(),
 				'description' => $this->get_description(),
-				'context'     => array( 'edit' ),
+				'context'     => array( 'view', 'edit', 'embed' ),
 			),
 			$this->get_data_structure()
 		);
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -156,7 +156,7 @@ abstract class Abstract_Option {
 				'$schema'     => 'http://json-schema.org/draft-04/schema#',
 				'title'       => static::key(),
 				'description' => $this->get_description(),
-				'context'     => array( 'view', 'edit', 'embed' ),
+				'context'     => array( 'edit' ),
 				'default'     => $this->get_default(),
 			),
 			$this->get_data_structure()
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -17,8 +17,6 @@ defined( 'ABSPATH' ) || exit;
  *
  * @phpstan-type SchemaType 'string'|'null'|'number'|'integer'|'boolean'|'array'|'object'
  * @phpstan-type Schema array{
- *     '$schema': non-falsy-string,
- *     title: non-falsy-string,
  *     type: SchemaType,
  *     context: array<non-falsy-string>
  * }
@@ -153,8 +151,6 @@ abstract class Abstract_Option {
 
 		$this->schema = array_merge(
 			array(
-				'$schema'     => 'http://json-schema.org/draft-04/schema#',
-				'title'       => static::key(),
 				'description' => $this->get_description(),
 				'context'     => array( 'edit' ),
 				'default'     => $this->get_default(),
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -157,6 +157,7 @@ abstract class Abstract_Option {
 				'title'       => static::key(),
 				'description' => $this->get_description(),
 				'context'     => array( 'view', 'edit', 'embed' ),
+				'default'     => $this->get_default(),
 			),
 			$this->get_data_structure()
 		);
--- a/include/Options/Business/Previous_Version.php
+++ b/include/Options/Business/Previous_Version.php
@@ -5,8 +5,6 @@
 
 namespace WP_Syntex\Polylang\Options\Business;
 
-use WP_Syntex\Polylang\Options\Primitive\Abstract_String;
-
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -14,7 +12,7 @@ defined( 'ABSPATH' ) || exit;
  *
  * @since 3.7
  */
-class Previous_Version extends Abstract_String {
+class Previous_Version extends Version {
 	/**
 	 * Returns option key.
 	 *
@@ -37,21 +35,5 @@ class Previous_Version extends Abstract_String {
 	 */
 	protected function get_description(): string {
 		return __( "Polylang's previous version.", 'polylang' );
-	}
-
-	/**
-	 * Returns the JSON schema part specific to this option.
-	 *
-	 * @since 3.7
-	 *
-	 * @return array Partial schema.
-	 *
-	 * @phpstan-return array{type: 'string', readonly: true, readonly: true}
-	 */
-	protected function get_data_structure(): array {
-		$string_schema             = parent::get_data_structure();
-		$string_schema['readonly'] = true;
-
-		return $string_schema;
 	}
 }
--- a/include/Options/Business/Previous_Version.php
+++ b/include/Options/Business/Previous_Version.php
@@ -38,4 +38,20 @@ class Previous_Version extends Abstract_String {
 	protected function get_description(): string {
 		return __( "Polylang's previous version.", 'polylang' );
 	}
+
+	/**
+	 * Returns the JSON schema part specific to this option.
+	 *
+	 * @since 3.7
+	 *
+	 * @return array Partial schema.
+	 *
+	 * @phpstan-return array{type: 'string', readonly: true}
+	 */
+	protected function get_data_structure(): array {
+		$string_schema             = parent::get_data_structure();
+		$string_schema['readonly'] = true;
+
+		return $string_schema;
+	}
 }
--- a/include/Options/Business/Previous_Version.php
+++ b/include/Options/Business/Previous_Version.php
@@ -46,7 +46,7 @@ class Previous_Version extends Abstract_String {
 	 *
 	 * @return array Partial schema.
 	 *
-	 * @phpstan-return array{type: 'string', readonly: true}
+	 * @phpstan-return array{type: 'string', readonly: true, readonly: true}
 	 */
 	protected function get_data_structure(): array {
 		$string_schema             = parent::get_data_structure();
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -335,7 +335,6 @@ class Options implements ArrayAccess, IteratorAggregate {
 		}
 
 		$properties = array();
-		$default    = array();
 
 		if ( ! empty( $this->options[ $this->current_blog_id ] ) ) {
 			foreach ( $this->options[ $this->current_blog_id ] as $option ) {
@@ -349,10 +348,6 @@ class Options implements ArrayAccess, IteratorAggregate {
 				unset( $sub_schema['title'], $sub_schema['$schema'] );
 
 				$properties[ $option->key() ] = $sub_schema;
-
-				if ( isset( $sub_schema['default'] ) ) {
-					$default[ $option->key() ] = $sub_schema['default'];
-				}
 			}
 		}
 
@@ -361,7 +356,6 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'title'                => static::OPTION_NAME,
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
-			'default'              => $default,
 			'properties'           => $properties,
 			'additionalProperties' => true,
 		);
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -361,7 +361,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'title'                => static::OPTION_NAME,
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
-			'context'              => array( 'view', 'edit', 'embed' ),
+			'context'              => array( 'edit' ),
 			'default'              => $default,
 			'required'             => array_keys( $properties ),
 			'properties'           => $properties,
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -356,7 +356,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'title'                => static::OPTION_NAME,
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
-			'context'              => array( 'edit' ),
+			'context'              => array( 'view', 'edit', 'embed' ),
 			'properties'           => $properties,
 			'additionalProperties' => false,
 		);
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -358,6 +358,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'type'                 => 'object',
 			'context'              => array( 'view', 'edit', 'embed' ),
 			'default'              => array(),
+			'required'             => array_keys( $properties ),
 			'properties'           => $properties,
 			'additionalProperties' => false,
 		);
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -357,6 +357,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
 			'context'              => array( 'view', 'edit', 'embed' ),
+			'default'              => array(),
 			'properties'           => $properties,
 			'additionalProperties' => false,
 		);
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -361,7 +361,6 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'title'                => static::OPTION_NAME,
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
-			'context'              => array( 'edit' ),
 			'default'              => $default,
 			'properties'           => $properties,
 			'additionalProperties' => true,
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -342,12 +342,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 					continue;
 				}
 
-				$sub_schema = $option->get_schema();
-
-				// Cleanup.
-				unset( $sub_schema['title'], $sub_schema['$schema'] );
-
-				$properties[ $option->key() ] = $sub_schema;
+				$properties[ $option->key() ] = $option->get_schema();
 			}
 		}
 
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -360,7 +360,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'default'              => array(),
 			'required'             => array_keys( $properties ),
 			'properties'           => $properties,
-			'additionalProperties' => false,
+			'additionalProperties' => true,
 		);
 
 		return $this->schema[ $this->current_blog_id ];
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -357,7 +357,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
 			'properties'           => $properties,
-			'additionalProperties' => true,
+			'additionalProperties' => false,
 		);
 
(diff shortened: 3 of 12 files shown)


include/Options/Options.php
```diff
@@ -356,9 +361,11 @@ public function get_schema(): array {
 			'title'                => static::OPTION_NAME,
 			'description'          => __( 'Polylang options', 'polylang' ),
 			'type'                 => 'object',
-			'context'              => array( 'edit' ),
+			'context'              => array( 'view', 'edit', 'embed' ),
+			'default'              => $default,
```
Hug0-Drelon, line 365: Why did you add the `default` values all at once at the top level instead of having a `default` key on each sub-instances?
Regarding the spec it seems doable, [see](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-00#rfc.section.9.2).

Hug0-Drelon (changes requested):
- I disagree with having `view`, `edit` and `embed` as allowed contexts in *all* options. Indeed, I don't see any reason to display an option value in an embed request (no HTML and no content of any sort). Also, I we don't implement `_link` we must not provide embed response according to the [documentation](https://developer.wordpress.org/rest-api/using-the-rest-api/linking-and-embedding/).
About `view`, I means to me "request made on frontend" and I don't see any reason to display plugin options on frontend as it is right now.

- About making *all* properties required, we force us to pass *all* options when making a `PUT` request to update even a single option. Is it what you want?

include/Options/Options.php
Screenfeed: Not "instead", it is also done [in each sub-option](https://github.com/polylang/polylang/pull/1569/files#diff-271c92425f5047de724b1a26c3011a77ba0a637d63e23c0ba7cb61816093c90cR160). This one is in case of the whole option is considered not valid.

include/Options/Abstract_Option.php
Hug0-Drelon, line 1: Reading both classes side to side, I wonder if having `context` set *only* on `Options` is sufficient.
Do you think it is?

Screenfeed: Good point.

Reading `rest_filter_response_by_context()`, I understand:
1. The context in `Options` is used if our options are only a part of the REST response. This would happen if we want to add our options to `/wp-json/wp/v2/settings`.
2. The context in `Abstract_Option` is used if our options are the whole REST response. This would happen if we create our route that returns all of our options.

Hug0-Drelon: I see, thanks for the insight. Let's go the way you suggest!

Screenfeed: Done in [754a5d4](https://github.com/polylang/polylang/pull/1569/commits/754a5d4cd8d26fefbf4b7a48984c4b47faf6fedc).

Hug0-Drelon (approved):

Screenfeed:
I'm holding this PR for the final revue: this work should not be done outside the REST API context, it may present results different from what we expected.

modules/REST/API.php
```diff
@@ -48,5 +55,8 @@ public function __construct( PLL_Model $model ) {
 	public function init(): void {
 		$this->languages = new V1\Languages( $this->model->languages, $this->model->translatable_objects );
 		$this->languages->register_routes();
+
+		$this->options = new V1\Options( $this->model->options );
+		$this->options->register_routes();
```
Hug0-Drelon, line 60: Can `register_route()` return `\Options` object? And this line can be removed.
Same for languages.

modules/REST/Abstract_Controller.php
```diff
@@ -0,0 +1,45 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param int $version Optional. Version of the API. Default is `1`.
+	 *
+	 * @phpstan-param positive-int $version
+	 */
+	public function __construct( int $version = 1 ) {
+		$this->namespace = "pll/v{$version}";
+	}
```
Hug0-Drelon, line 30: Since this is *all* V1, I suggest to remove the constructor for now.

modules/REST/Abstract_Controller.php
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * Adds a status code to the given error and returns the error.
+	 *
+	 * @since 3.7
+	 *
+	 * @param WP_Error $error       A `WP_Error` object.
+	 * @param int      $status_code Optional. A status code. Default is 400.
+	 * @return WP_Error
+	 */
+	protected function add_status_to_error( WP_Error $error, int $status_code = 400 ): WP_Error {
```
Hug0-Drelon, line 41: If this all what languages and options share, a trait is sufficient? What do you think?

modules/REST/V1/Options.php
```diff
@@ -0,0 +1,218 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.7
+	 *
+	 * @param Options_Registry $options Options registry.
+	 */
+	public function __construct( Options_Registry $options ) {
+		parent::__construct();
+		$this->rest_base = 'options';
```
Hug0-Drelon, line 37: I thnk about it only now, but should we use the same term as WordPress does? i.e. `settings`?

modules/REST/V1/Options.php
```diff
@@ -0,0 +1,218 @@
...
+	 *
+	 * @return void
+	 */
+	public function register_routes(): void {
+		register_rest_route(
+			$this->namespace,
+			"/{$this->rest_base}",
+			array(
+				'args'        => array(
+					'context' => $this->get_context_param( array( 'default' => 'view' ) ),
```
Hug0-Drelon, line 54: I'd use `array( 'default' => 'edit' )` instead, what do you think?

modules/REST/V1/Options.php
```diff
@@ -0,0 +1,218 @@
...
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function get_items( $request ) {
+		return $this->prepare_item_for_response( $this->options->get_all(), $request );
+	}
+
+	/**
+	 * Updates option(s).
+	 * This allows to unpdate one or several options.
+	 * Note: to update an array value, use `option=val1,val2`.
```
Hug0-Drelon, line 91: A JSON body can handle array shape easily, right?

modules/REST/V1/Options.php
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	/**
+	 * Tells if options can be edited.
+	 *
+	 * @since 3.7
+	 *
+	 * @return bool
+	 */
+	protected function check_update_permission(): bool {
```
Hug0-Drelon, line 215: This method can be removed if I read correctly.
It's used only in `get_items_permissions_check()` and thus the code can be put in it directly.

tests/phpunit/tests/test-rest-options.php
```diff
@@ -0,0 +1,194 @@
...
+				add_action( 'rest_api_init', array( $polylang->rest, 'init' ) );
+			}
+		);
+
+		$this->pll_env = ( new PLL_Context_Rest() )->get();
+		$this->server  = $GLOBALS['wp_rest_server'];
+	}
+
+	public function tear_down() {
+		parent::tear_down();
```
Hug0-Drelon, line 40: I think it is a good practice to call `parent::tear_down();` at the end.

tests/phpunit/tests/test-rest-options.php
Hug0-Drelon, line 1: Should wa have a test with the `DELETE` method? Ensuring a correct error code is returner?

Hug0-Drelon (changes requested):

modules/REST/API.php
Screenfeed: `WP_REST_Controller::register_routes()` doesn't return anything. On the other hand, the doc doesn't provide any `@return` tag, so I guess it can return anything?

modules/REST/Abstract_Controller.php
Screenfeed: How do you suggest to set the property then?

modules/REST/Abstract_Controller.php
Screenfeed: I'm not sure to prefer a trait over an abstract class here, since we'll always extend `WP_REST_Controller` anyway.

modules/REST/V1/Options.php
Screenfeed: It's a good idea I think, so it's consistent.

modules/REST/V1/Options.php
Screenfeed: So we don't need to pass it by default? Why not, good idea.

modules/REST/API.php
Screenfeed: However, now we have this:

```
 ------ ---------------------------------------------------------------------------------- 
  Line   modules/REST/API.php                                                              
 ------ ---------------------------------------------------------------------------------- 
  :28    Property WP_Syntex\Polylang\REST\API::$languages is never read, only written.     
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties   
  :35    Property WP_Syntex\Polylang\REST\API::$options is never read, only written.       
         💡 See: https://phpstan.org/developing-extensions/always-read-written-properties   
 ------ ----------------------------------------------------------------------------------
```

But I notices something: in the class `API` the properties are `public $model;`, `private $languages`, and `private $options`. I think they should be the opposite:
- `private $model;` because we don't need to access it from the outside,
- `public $languages;` and `public $options;`, so their hooks can be manipulated from the outside.

And by doing so, PHPStan doesn't trigger the previous errors.
What do you think?

modules/REST/Abstract_Controller.php
Hug0-Drelon: Directly in the class.

modules/REST/API.php
Hug0-Drelon: I forgot that `register_routes()` is defined in `WP_REST_Controller`... 
But your suggestion sounds good!

tests/phpunit/tests/test-rest-options.php
Screenfeed: I thought about this but we'd test WP in that case, not PLL.

modules/REST/V1/Options.php
Screenfeed: Done in [fb165da](https://github.com/polylang/polylang/pull/1569/commits/fb165da750a8c4703a47bf3cd47eeb6b3df15e49).

modules/REST/V1/Options.php
Screenfeed: Done in [2e04b6e](https://github.com/polylang/polylang/pull/1569/commits/2e04b6e45506c889bbc7b39c45072e3814c4d395).

modules/REST/API.php
Screenfeed: Done in [509c086](https://github.com/polylang/polylang/pull/1569/commits/509c0862442619a8f7d89b472d4cb95ed2295dff) and [1ee0c2b](https://github.com/polylang/polylang/pull/1569/commits/1ee0c2bda35b1f054b1338a8c936f370d36285bd).

tests/phpunit/tests/test-rest-options.php
Screenfeed: Indeed, copied from another class. Done in [e6cffa4](https://github.com/polylang/polylang/pull/1569/commits/e6cffa4c1150156ddb552a230dee799f10b1fc7f).

modules/REST/V1/Options.php
Screenfeed: Indeed. This is when passed in the URL. It is misleading so I removed it in [c74b5ba](https://github.com/polylang/polylang/pull/1569/commits/c74b5badb0b1df7aec6a3cc74431ae4b3a4dd6df).

modules/REST/V1/Options.php
Screenfeed: Right, I thought it was in `WP_REST_Controller` and should be implemented. Done in [73c7d58](https://github.com/polylang/polylang/pull/1569/commits/73c7d58bd3b984272a06520d2d9d4bf02dc9ad68).

modules/REST/Abstract_Controller.php
Screenfeed: Done in [12f1146](https://github.com/polylang/polylang/pull/1569/commits/12f11463a675f3797307b8a81b5d231e17de35a7).

modules/REST/V1/Options.php
Hug0-Drelon: We shouldn't be aware of that, this is WP business IMHO.
We should only do `$request->get_param( 'foo' )`.

Screenfeed: ???

modules/REST/V1/Options.php
```diff
@@ -0,0 +1,208 @@
...
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function update_items( $request ) {
+		$response = array();
+		$errors   = new WP_Error();
+		$schema   = $this->options->get_schema();
+		$options  = array_intersect_key( $this->options->get_all(), $request->get_params() ); // Remove `context`.
+
+		foreach ( $options as $option_name => $current_value ) {
+			if ( ! empty( $schema['properties'][ $option_name ]['readonly'] ) ) {
```
Hug0-Drelon, line 109: Isn't it checked automatically by WordPress?
I cannot found where WP does such a thing in a concrete controller

modules/REST/V1/Options.php
```diff
@@ -0,0 +1,208 @@
...
+use WP_Syntex\Polylang\REST\Abstract_Controller;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Options REST controller.
+ *
+ * @since 3.7
+ */
+class Options extends Abstract_Controller {
```
Hug0-Drelon, line 22:
```suggestion
class Settings extends Abstract_Controller {
```
So it matches the namespace of the route.

Hug0-Drelon (changes requested):

modules/REST/V1/Options.php
Screenfeed: WP does strictly nothing at this point.
The only moment WP uses it is in `rest_get_endpoint_args_for_schema()`.

modules/REST/V1/Options.php
Screenfeed: Done in [0fa8170](https://github.com/polylang/polylang/pull/1569/commits/0fa8170a4ab309e3e0c5614ce765762d84587602).

modules/REST/V1/Options.php
Hug0-Drelon: I mean for WordPress core routes, such as for posts.

Screenfeed: Ha! Haven't tried. They probably use `rest_get_endpoint_args_for_schema()` I suppose.

Screenfeed: They simply ignore the fields that are `readonly`.

Hug0-Drelon: For simplicity, should we do the same here?

Screenfeed: Done in [cf488c7](https://github.com/polylang/polylang/pull/1569/commits/cf488c7434d3c32899152ad4371b784f49e4d2d4).

Hug0-Drelon (approved):

Chouby (changes requested):

modules/REST/V1/Settings.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @phpstan-template T of array
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function update_items( $request ) {
+		$response = array();
+		$errors   = new WP_Error();
+		$schema   = $this->options->get_schema();
+		$options  = array_intersect_key(
+			$request->get_params(),
+			$this->options->get_all(), // Remove `context`.
```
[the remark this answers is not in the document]
Screenfeed: `$request->get_params()` returns all arguments passed to the API, which include the `context` key and the new option values (and possibly other things). So by using `array_intersect_key()` I keep only the keys that correspond to options.
But after taking another look at this, I realized that `rest_get_endpoint_args_for_schema( $schema, WP_REST_Server::EDITABLE )` returns only the options, without the context key, so this line is useless.

include/Options/Business/Previous_Version.php
```diff
@@ -38,4 +38,20 @@ public static function key(): string {
...
+	 *
+	 * @return array Partial schema.
+	 *
+	 * @phpstan-return array{type: 'string', readonly: true, readonly: true}
+	 */
+	protected function get_data_structure(): array {
+		$string_schema             = parent::get_data_structure();
+		$string_schema['readonly'] = true;
+
+		return $string_schema;
```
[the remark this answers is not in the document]
Screenfeed: Done in [f29b274](https://github.com/polylang/polylang/pull/1569/commits/f29b2747c95919a432852b6e341fc68ba66be592).

modules/REST/V1/Settings.php
```diff
@@ -0,0 +1,203 @@
...
+				),
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_items' ),
+					'permission_callback' => array( $this, 'get_items_permissions_check' ),
+				),
+				array(
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => array( $this, 'update_items' ),
+					'permission_callback' => array( $this, 'update_items_permissions_check' ),
```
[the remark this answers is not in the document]
Screenfeed: Done in [72eb124](https://github.com/polylang/polylang/pull/1569/commits/72eb12497bb9302035c0e2362558ee04dcff631d).

modules/REST/V1/Settings.php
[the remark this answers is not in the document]
Screenfeed: Removed in [d20305c](https://github.com/polylang/polylang/pull/1569/commits/d20305c90c28710834285bf348f19e07044deb57).

modules/REST/V1/Settings.php
```diff
@@ -0,0 +1,203 @@
...
+ * @package Polylang
+ */
+
+namespace WP_Syntex\Polylang\REST\V1;
+
+use WP_Error;
+use WP_REST_Request;
+use WP_REST_Response;
+use WP_REST_Server;
+use WP_Syntex\Polylang\Options\Options as Options_Registry;
```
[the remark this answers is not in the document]
Screenfeed: Originally this class was also named `Options`.
Removed in [2c3d29c](https://github.com/polylang/polylang/pull/1569/commits/2c3d29cfba5cfa0c21c2fe045cbc17deaaf45897).

tests/phpunit/tests/test-rest-settings.php
```diff
@@ -0,0 +1,184 @@
...
+	private function get_request( string $method = 'GET', array $params = array() ): WP_REST_Request {
+		$request = new WP_REST_Request( $method, '/pll/v1/settings' );
+
+		foreach ( $params as $name => $value ) {
+			if ( isset( $value ) ) {
+				$request->set_param( $name, $value );
+			}
+		}
+
+		return $request;
```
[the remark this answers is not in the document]
Screenfeed: Done in [63a1b55](https://github.com/polylang/polylang/pull/1569/commits/63a1b55d8c23d1dad3fd1b1f640f0b30473badd1).

modules/REST/V1/Languages.php
```diff
@@ -53,9 +53,9 @@ public function __construct( Languages_Model $languages, PLL_Translatable_Object
 	 *
 	 * @since 3.7
 	 *
-	 * @return void
+	 * @return self
 	 */
-	public function register_routes(): void {
+	public function register_routes(): self {
```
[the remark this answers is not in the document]
Screenfeed: Reverted in [5c5319e](https://github.com/polylang/polylang/pull/1569/commits/5c5319ef41a966c0bdf6faf8a20ce8ceedf0af75).

PR #1571, opened 2024-11-12
Restrict option sanitization to the relevant string

Chouby:
Currently, an option sanitization callback is a applied to all string translations. If the callback doesn't make the necessary checks, this could lead to undesired changed to other strings translations.

First, a simple test is added to demonstrate the bug.

The proposed solution:
1. Add a new parameter to the filter `pll_sanitize_string_translation`: the string passed to `pll_register_string()`
2. For each option, store all registered strings in an array (in fact I propose to store md5 hashes). 
3. Use this array to check, in the method hooked to `pll_sanitize_string_translation`, that the string passed is included in the option.

### DIFF
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -21,9 +21,9 @@ class PLL_Translate_Option {
 	/**
 	 * Sanitization callback.
 	 *
-	 * @var callable
+	 * @var string
 	 */
-	private $sanitize_callback = 'sanitize_option';
+	private $sanitize_callback;
 
 	/**
 	 * Hashes for registered strings for this option.
@@ -98,7 +98,7 @@ class PLL_Translate_Option {
 		add_action( 'update_option_' . $name, array( $this, 'update_option' ) );
 
 		// Sanitizes translated strings.
-		if ( ! empty( $args['sanitize_callback'] ) && is_callable( $args['sanitize_callback'] ) ) {
+		if ( ! empty( $args['sanitize_callback'] ) ) {
 			$this->sanitize_callback = $args['sanitize_callback'];
 		}
 		add_filter( 'pll_sanitize_string_translation', array( $this, 'sanitize_option' ), 10, 4 );
@@ -426,6 +426,11 @@ class PLL_Translate_Option {
 			return $value;
 		}
 
-		return call_user_func( $this->sanitize_callback, $value, $name, $context, $original );
+		if ( is_callable( $this->sanitize_callback ) ) {
+			return call_user_func( $this->sanitize_callback, $value, $name, $context, $original );
+		}
+
+		/** @var string */
+		return sanitize_option( $name, $value );
 	}
 }
--- a/include/translate-option.php
+++ b/include/translate-option.php
@@ -14,9 +14,23 @@ class PLL_Translate_Option {
 	/**
 	 * Array of option keys to translate.
 	 *
-	 * @var array
+	 * @var string[]
 	 */
 	private $keys;
+
+	/**
+	 * Sanitization callback.
+	 *
+	 * @var string
+	 */
+	private $sanitize_callback;
+
+	/**
+	 * Hashes for registered strings for this option.
+	 *
+	 * @var string[]
+	 */
+	private $hashes = array();
 
 	/**
 	 * Used to prevent filtering when retrieving the raw value of the option.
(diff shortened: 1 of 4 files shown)


include/translate-option.php
```diff
@@ -14,10 +14,24 @@ class PLL_Translate_Option {
...
 	 *
-	 * @var array
+	 * @var string[]
 	 */
 	private $keys;
 
+	/**
+	 * Sanitization callback.
+	 *
+	 * @var string
```
Hug0-Drelon, line 24:
```suggestion
	 * @var callable
```

include/translate-option.php
```diff
@@ -84,11 +98,10 @@ public function __construct( $name, $keys = array(), $args = array() ) {
 		add_action( 'update_option_' . $name, array( $this, 'update_option' ) );
 
 		// Sanitizes translated strings.
-		if ( empty( $args['sanitize_callback'] ) ) {
-			add_filter( 'pll_sanitize_string_translation', array( $this, 'sanitize_option' ), 10, 2 );
-		} else {
-			add_filter( 'pll_sanitize_string_translation', $args['sanitize_callback'], 10, 3 );
+		if ( ! empty( $args['sanitize_callback'] ) ) {
```
Hug0-Drelon, line 101: Maybe it's out of scope, but it makes me think of a neat enhancement for this class.
We could pass the callback as a value for the keys, so each keys could have a different callback. But it might be hard to implement for array shaped callbacks...

Hug0-Drelon, line 101:
```suggestion
		if ( ! empty( $args['sanitize_callback'] ) && is_callable( $args['sanitize_callback'] ) ) {
```
For type safety.

include/translate-option.php
```diff
@@ -216,8 +229,10 @@ protected function register_string_recursive( $context, $option, $values, $key )
 					$this->register_string_recursive( $context, $n, $value, $key );
 				}
 			}
-		} else {
-			PLL_Admin_Strings::register_string( $option, $values, $context, true );
+		} elseif ( is_scalar( $values ) ) {
+			$string         = (string) $values;
+			$this->hashes[] = md5( "$string|$option|$context" );
```
Hug0-Drelon, line 234: I would have a dedicated method to get the hash. I see two benefits:
- Enhanced readability
- Easier maintenance (even though it's already encapsulated in *one* class)

Just a suggestion.

include/translate-option.php
```diff
@@ -14,10 +14,24 @@ class PLL_Translate_Option {
...
+	 * @var string[]
 	 */
 	private $keys;
 
+	/**
+	 * Sanitization callback.
+	 *
+	 * @var string
+	 */
+	private $sanitize_callback;
```
Hug0-Drelon, line 26:
```suggestion
	private $sanitize_callback = 'sanitize_option';
```
See later the improvement suggested.

include/translate-option.php
```diff
@@ -395,17 +410,27 @@ protected function check_value_recursive( $old_values, $values, $key, $mo ) {
...
+		if ( ! in_array( md5( "$original|$name|$context" ), $this->hashes, true ) ) {
+			return $value;
+		}
+
+		if ( is_callable( $this->sanitize_callback ) ) {
+			return call_user_func( $this->sanitize_callback, $value, $name, $context, $original );
+		}
+
+		/** @var string */
+		return sanitize_option( $name, $value );
```
Hug0-Drelon, line 434:
```suggestion
		return call_user_func( $this->sanitize_callback, $value, $name, $context, $original );
```
Follows previous suggestions.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

Chouby:
Rebase + force push to resolve a conflict.

Chouby:
@Hug0-Drelon I had to revert your suggestions. because of the different order of parameters between `sanitize_option()` and the custom callback. See the corresponding failing test: https://github.com/polylang/polylang/actions/runs/12668890902/job/35305219984

Hug0-Drelon:
> @Hug0-Drelon I had to revert your suggestions. because of the different order of parameters between `sanitize_option()` and the custom callback. See the corresponding failing test: https://github.com/polylang/polylang/actions/runs/12668890902/job/35305219984

Oops I didn't see that coming... 👍

Hug0-Drelon (approved):

PR #1574, opened 2024-11-18
Use empty string instead of duplicate for non-translated strings.

Hug0-Drelon:
## What?
Fixes partially https://github.com/polylang/polylang-pro/issues/2308.
Removes non-translated strings from the dashboard. Use empty string instead.

## How
- Ensure to use empty string on display
- Force to empty string when adding entries
- Migration from Polylang < 3.7 to ensure steady behavior.

### DIFF
--- a/include/mo.php
+++ b/include/mo.php
@@ -74,4 +74,29 @@ class PLL_MO extends MO {
 	public function delete_entry( $string ) {
 		unset( $this->entries[ $string ] );
 	}
+
+	/**
+	 * Translates a string or returns false if the translation is not found.
+	 * Contrary to `self::translate()`, this method doesn't fallback to the source string.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $source The source string to translate.
+	 * @return string|false The translated string or false if not found.
+	 */
+	public function translate_or_fail( string $source ) {
+		$entry = $this->translate_entry(
+			new Translation_Entry(
+				array(
+					'singular' => $source,
+				)
+			)
+		);
+
+		if ( ! $entry instanceof Translation_Entry || empty( $entry->translations ) ) {
+			return false;
+		}
+
+		return $entry->translations[0];
+	}
 }
--- a/include/mo.php
+++ b/include/mo.php
@@ -77,19 +77,19 @@ class PLL_MO extends MO {
 
 	/**
 	 * Translates a string or returns false if the translation is not found.
-	 * Contrary to `self::translate()`, this method doesn't fallback to the source string.
+	 * Contrary to `self::translate()`, this method doesn't fallback to the source string but returns empty string instead.
 	 *
 	 * @since 3.7
 	 *
 	 * @param string $source The source string to translate.
-	 * @return string|false The translated string or false if not found.
+	 * @return string The translated string or empty string if not found.
 	 */
-	public function translate_or_fail( string $source ) {
+	public function translate_if_any( string $source ) {
 		$entry = new Translation_Entry( array( 'singular' => $source ) );
 		$entry = $this->translate_entry( $entry );
 
 		if ( ! $entry instanceof Translation_Entry || empty( $entry->translations ) ) {
-			return false;
+			return '';
 		}
 
 		return $entry->translations[0];
--- a/include/mo.php
+++ b/include/mo.php
@@ -85,13 +85,8 @@ class PLL_MO extends MO {
 	 * @return string|false The translated string or false if not found.
 	 */
 	public function translate_or_fail( string $source ) {
-		$entry = $this->translate_entry(
-			new Translation_Entry(
-				array(
-					'singular' => $source,
-				)
-			)
-		);
+		$entry = new Translation_Entry( array( 'singular' => $source ) );
+		$entry = $this->translate_entry( $entry );
 
 		if ( ! $entry instanceof Translation_Entry || empty( $entry->translations ) ) {
 			return false;
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+use WP_Syntex\Polylang\Options\Options;
+
 /**
  * Manages Polylang upgrades
  *
@@ -212,6 +214,7 @@ class PLL_Upgrade {
 	/**
 	 * Upgrades if the previous version is < 3.7.
 	 * Hides the "The language is set from content" option if it isn't the one selected.
+	 * Cleans up strings translations so we don't store translations duplicated from the source.
 	 *
 	 * @since 3.7
 	 *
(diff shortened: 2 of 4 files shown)


Screenfeed (approved):

install/upgrade.php
```diff
@@ -222,6 +225,24 @@ protected function upgrade_3_7() {
...
+			foreach ( $mo->entries as $entry ) {
+				if ( $entry->singular === $entry->translations[0] ) {
+					$entry->translations = array( '' );
+				}
+
+				$mo->add_entry_or_merge( $entry );
+			}
+
+			$mo->export_to_db( $language );
+		}
```
Chouby, line 245: 1. To clarify and avoid mixing unrelated things, I would separate `upgrade_3_7()` in 2 methods as we did for 3.4.
2. I would manage the upgrade at a lower level (dealing directly with the meta). This is what I always did in the past. The reasoning is that, once merged, we will *never* touch this code anymore. While other parts of Polylang will continue to evolve and this *may* break this. I want to be sure that the upgrade process will continue to work in far future.

settings/table-string.php
```diff
@@ -141,7 +141,9 @@ public function column_translations( $item ) {
 				esc_attr( $key ),
 				esc_attr( $item['row'] ),
 				esc_html( $languages[ $key ] ),
-				format_to_edit( $translation ) // Don't interpret special chars.
+				format_to_edit(
+					$translation === $item['string'] ? '' : $translation
```
Chouby, line 145: That means that if a user manually copy the original string in the translation, we will remove it.
Do we really want do that?

Hug0-Drelon: Ah! You got me!
In fact I've done that to tests the behavior on the Pro.
But you're right to point it out, I don't think we should prevent anyone to duplicate a string on purpose. Don't you agree?

install/upgrade.php
Hug0-Drelon: 👍 
See 9c1f1d8e

settings/table-string.php
Chouby: This is what I had in mind.

Hug0-Drelon: Done in 1cca4c43.
Actually we to "translate" entry when preparing items for the table, but not using `Translations::translate()` otherwise we always get the source if set to empty string in DB...

So to summarize I removed the code forcing to empty string on display and on save.
Then when preparing items for the strings table, ensure to display what's from DB.

settings/table-string.php
```diff
@@ -300,8 +300,16 @@ public function prepare_items() {
...
+				$entry = new Translation_Entry(
+					array(
+						'singular' => $row['string'],
+					)
+				);
+
+				$entry                                                  = $mo[ $language->slug ]->translate_entry( $entry );
+				$translation                                            = $entry && ! empty( $entry->translations ) ? $entry->translations[0] : '';
+				$this->items[ $key ]['translations'][ $language->slug ] = $translation;
+				$this->items[ $key ]['row']                             = $key; // Store the row number for convenience
```
Screenfeed, line 312: Do you think it's better?
```suggestion
				$entry       = $mo[ $language->slug ]->translate_entry( $entry );
				$translation = $entry && ! empty( $entry->translations ) ? $entry->translations[0] : '';

				$this->items[ $key ]['translations'][ $language->slug ] = $translation;
				$this->items[ $key ]['row']                             = $key; // Store the row number for convenience
```

Hug0-Drelon: I have no preference.

Screenfeed (approved):

settings/table-string.php
```diff
@@ -300,8 +300,17 @@ public function prepare_items() {
...
-				$this->items[ $key ]['translations'][ $language->slug ] = $mo[ $language->slug ]->translate( $row['string'] );
-				$this->items[ $key ]['row'] = $key; // Store the row number for convenience
+				$entry = new Translation_Entry(
+					array(
+						'singular' => $row['string'],
+					)
+				);
+
+				$entry       = $mo[ $language->slug ]->translate_entry( $entry );
+				$translation = $entry && ! empty( $entry->translations ) ? $entry->translations[0] : '';
```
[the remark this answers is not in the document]
Hug0-Drelon: I'm all in!
New method created in 2bd77198.
Array of `PLL_MO` removed in 3ca3c530.

include/mo.php
```diff
@@ -74,4 +74,24 @@ public function import_from_db( $lang ) {
...
+	 *
+	 * @param string $source The source string to translate.
+	 * @return string|false The translated string or false if not found.
+	 */
+	public function translate_or_fail( string $source ) {
+		$entry = new Translation_Entry( array( 'singular' => $source ) );
+		$entry = $this->translate_entry( $entry );
+
+		if ( ! $entry instanceof Translation_Entry || empty( $entry->translations ) ) {
+			return false;
```
Chouby, line 92: I would rather always return a string.
```suggestion
			return '';
```

settings/table-string.php
```diff
@@ -196,14 +196,16 @@ protected function get_default_primary_column_name() {
...
 	 * @since 2.6
 	 *
-	 * @param PLL_MO[] $mos An array of PLL_MO objects.
-	 * @param string   $s   Searched string.
+	 * @param PLL_Language[] $languages An array of language objects.
+	 * @param string         $s         Searched string.
 	 * @return string[] Found strings.
 	 */
-	protected function search_in_translations( $mos, $s ) {
+	protected function search_in_translations( $languages, $s ) {
```
Chouby, line 203: I missed that the array of `PLL_MO` was used here. I thought that this array was something stupid that I made but finally it was not. It's ok as is now thanks to the cache introduced previously.

include/mo.php
```diff
@@ -74,4 +74,24 @@ public function import_from_db( $lang ) {
...
+	/**
+	 * Translates a string or returns false if the translation is not found.
+	 * Contrary to `self::translate()`, this method doesn't fallback to the source string.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $source The source string to translate.
+	 * @return string|false The translated string or false if not found.
+	 */
+	public function translate_or_fail( string $source ) {
```
Chouby, line 87: I'm not fond of the proposed name. `maybe_translate()`? `translate_if_exists()`? `translate_if_any()`?

Chouby (changes requested):

include/mo.php
Hug0-Drelon: Done in 813cda54.

include/mo.php
Hug0-Drelon: Done in 813cda54.

settings/table-string.php
Hug0-Drelon: Having the new cache in mind, I though it was worth the change because it simplified the code in `prepare_items()` as well IMHO.

PR #1576, opened 2024-11-20
Reformat SQL queries in `PLL_Translated_Object`

Chouby:
Follows up #1556, #1558.

This PR reformats SQL queries prepare them in only one place when possible thus avoiding `phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared`.

In some cases this ignore comment is replaced by `phpcs:ignore WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber` as least as long as WPCS doesn't handle correctly the SQL syntax `INSERT INTO ... VALUES ...`.

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -519,38 +519,60 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 		$count       = array();
 
 		foreach ( $translations as $t ) {
-			$term = uniqid( 'pll_' ); // the term name
-			$terms[] = $wpdb->prepare( '( %s, %s )', $term, $term );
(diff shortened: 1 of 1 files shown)


Chouby:
I haven't invented anything. This is the technique used to pass WPCS checks. See for example: https://github.com/WordPress/WordPress-Coding-Standards/blob/3.1.0/WordPress/Tests/DB/PreparedSQLUnitTest.1.inc#L26-L30. There is also https://core.trac.wordpress.org/ticket/54042 about the same topic. Currently focus is on `IN( ... )` only as this is the most frequent case. I opened an issue in WPCS to fully support it for `INSERT INTO ... VALUES ...`.

PR #1577, opened 2024-11-20
Reformat SQL queries for the  WP Importer

Chouby:
Same as #1576 for the WordPress importer.

This PR reformats SQL queries prepare them in only one place when possible thus avoiding `phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared`.

In some cases this ignore comment is replaced by `phpcs:ignore WordPress.DB.PreparedSQLPlaceholders.ReplacementsWrongNumber` as least as long as WPCS doesn't handle correctly the SQL syntax `INSERT INTO ... VALUES ...`.

### DIFF
--- a/integrations/wp-importer/wp-import.php
+++ b/integrations/wp-importer/wp-import.php
@@ -128,20 +128,18 @@ class PLL_WP_Import extends WP_Import {
 			$translations = maybe_unserialize( $term['term_description'] );
 			foreach ( $translations as $slug => $old_id ) {
 				if ( $old_id && ! empty( $this->processed_terms[ $old_id ] ) && $lang = PLL()->model->get_language( $slug ) ) {
-					// Language relationship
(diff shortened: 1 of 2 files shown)


integrations/wp-importer/wp-import.php
```diff
@@ -128,35 +128,40 @@ protected function remap_terms_relations( &$terms ) {
...
 			}
 		}
 
-		// Insert term_relationships
+		// Insert term_relationships.
 		if ( ! empty( $trs ) ) {
-			$trs = array_unique( $trs );
-
-			// Make sure we don't attempt to insert already existing term relationships
+			// Make sure we don't attempt to insert already existing term relationships.
```
Hug0-Drelon, line 142: `$trs = array_unique( $trs );` has been lost. Is it on purpose?

Chouby: I don't remember. I guess I found  it's useless and thus removed it.

Hug0-Drelon (approved):

PR #1581, opened 2024-11-25
Improve and fix `install-plugins.sh`

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2350.

Depends on https://github.com/polylang/wp-phpunit/pull/14.

Themes from the repository use a redirection.
Ex: `https://downloads.wordpress.org/theme/twentyfourteen.zip` currently redirects (302) to `https://downloads.wordpress.org/theme/twentyfourteen.4.1.zip`.

This breaks our current `download` bash function because it doesn't allow redirects when using curl ([ex](https://github.com/polylang/polylang/actions/runs/11935773152/job/33267859373)).

This PR's main aim is to fix this issue by allowing redirections with curl (done in https://github.com/polylang/wp-phpunit/pull/14).

Also in this PR:

- Don't try to unzip a file that doesn't exist: display an error message directly if a download fails (done in https://github.com/polylang/wp-phpunit/pull/14).
- Don't download Yoast SEO: we never use it.
- Add scripts to composer, like in PLL Pro. This allows to get rid of `install-wp-tests.sh`.
- Define some more constants in tests' bootstrap.
- Display an error when a theme or plugin is not installed, instead of simply skipping the test. This allows to be immediately warned if something goes wrong.
- The plugins and themes are not moved to WP's `wp-content` folder, but in `tmp/plugins` and `tmp/themes` instead, this matches what is done in PLL Pro.

> [!NOTE]
> This PR doesn't use the full capacities of PLL PHPUnit (like bootstraping tests) on purpose, this can be done subsequently.


tests/phpunit/tests/plugins/test-duplicate-post.php
```diff
@@ -6,16 +6,13 @@ class Duplicate_Post_Test extends PLL_UnitTestCase {
...
 	 */
 	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
-
-		if ( ! file_exists( DIR_TESTROOT . '/../duplicate-post/' ) ) {
-			self::markTestSkipped( 'This test requires the plugin Duplicate Post' );
-		}
-
-		require_once DIR_TESTROOT . '/../duplicate-post/duplicate-post.php';
-		require_once DIR_TESTROOT . '/../duplicate-post/admin-functions.php';
+		self::assertFileExists( PLL_TEST_EXT_PLUGINS_DIR . 'duplicate-post/duplicate-post.php', 'This test requires the plugin Duplicate Post.' );
```
Hug0-Drelon, line 9: Sorry I missed your question on Slack about that yesterday.
I must say I'm not very fond of.
Reason being I don't always install plugins/themes for integration tests locally. I don't think it's useful to run compatibility tests each time I modify code on my machine and it takes quite some times to run theses tests...
And with that code, the suite will fail every time.
Could we skip the tests gracefully instead?
It'll still be clear in CI whether or not compatibility tests run or not.

tests/phpunit/includes/bootstrap.php
Hug0-Drelon, line 1: Why not using `wp-content` even for external plugins or themes?

Hug0-Drelon (changes requested):

tests/phpunit/tests/plugins/test-duplicate-post.php
Screenfeed: I will be clear only if we go see the results of each action, because everything will be green.
However, we could add tags and create composer scripts to run tests without the plugins. But I think it should be in another PR.

tests/phpunit/includes/bootstrap.php
Screenfeed: Because that's how PLL PHPUnit works: we made that so plugins and themes are not deleted when we update the suite.

tests/phpunit/tests/plugins/test-duplicate-post.php
Hug0-Drelon: Then we should implement our own `markTestSkipped` that fails only in CI.

Hug0-Drelon: Something like `markTestSkippedOrDieInCI`, so it rhymes.

tests/phpunit/includes/bootstrap.php
Hug0-Drelon: I see, thanks.

tests/phpunit/tests/plugins/test-duplicate-post.php
Screenfeed: Done in [26e0b06](https://github.com/polylang/polylang/pull/1581/commits/26e0b0635aaa25bec11bcedd1bf37cd8ffa83e73).
An env var `PLL_SKIP_PLUGINS_TESTS` must be set in order to skip instead of asserting.

tests/phpunit/includes/testcase.php
```diff
@@ -44,17 +44,22 @@ protected function assert_has_language( $object, $language, $message = '' ) {
...
+	 *
+	 * @see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
 	 *
 	 * @param string $path    Path to the file.
 	 * @param string $message Error message.
 	 * @return void
 	 */
 	protected static function markTestSkippedIfFileNotExists( string $path, string $message = '' ): void {
-		if ( ! getenv( 'PLL_SKIP_PLUGINS_TESTS' ) ) {
+		if ( 'true' === getenv( 'GITHUB_ACTIONS' ) ) {
```
Hug0-Drelon, line 58: According to the documentation, it is *always* set to `'true'`. Do we need to check the value then?

Hug0-Drelon (changes requested):

tests/phpunit/includes/testcase.php
Screenfeed: Not needed per say, but I prefer using booleans in `if` statements. `if ( 'true' ) {` is "only" true-ish.
Do you fear they change the returned value?

Hug0-Drelon: It could, but in the other hand I agree with you on using bool value in if statements.
What about `false !== getenv( 'GITHUB_ACTIONS' )`?
`getenv()` returning `false` if `GITHUB_ACTIONS` doesn't exist at all.

Hug0-Drelon (approved):

tests/phpunit/includes/bootstrap.php
```diff
@@ -14,24 +15,51 @@ function () use ( $_root_dir, $_tests_dir ) {
...
+	define( 'PLL_TEST_WP_DIR', dirname( DIR_TESTROOT ) . '/wordpress/' );
+}
+
+if ( ! defined( 'PLL_TEST_THEMES_DIR' ) ) {
+	define( 'PLL_TEST_THEMES_DIR', PLL_TEST_WP_DIR . 'wp-content/themes/' );
+}
+
+if ( ! defined( 'PLL_TEST_PLUGINS_DIR' ) ) {
+	define( 'PLL_TEST_PLUGINS_DIR', PLL_TEST_WP_DIR . 'wp-content/plugins/' );
+}
```
Chouby, line 53:
```suggestion
if ( ! defined( 'PLL_TEST_WP_DIR' ) ) {
	define( 'PLL_TEST_WP_DIR', dirname( DIR_TESTROOT ) . '/wordpress/' );
}

if ( ! defined( 'PLL_TEST_THEMES_DIR' ) ) {
	define( 'PLL_TEST_THEMES_DIR', "{$_root_dir}/tmp/themes/" );
}

if ( ! defined( 'PLL_TEST_PLUGINS_DIR' ) ) {
	define( 'PLL_TEST_PLUGINS_DIR', "{$_root_dir}/tmp/plugins/" );
}
```

Having 2 constants for plugins and themes directories is confusing. I propose never to use the default plugins and themes directories and always use ours.
This may also help us reminder not to use default themes. See what I made with Twenty Twenty Three in 9a8512075124bc81e250d119b47f687a5392a519.

Screenfeed: Didn't know about [those themes](https://github.com/WordPress/wordpress-develop/tree/trunk/tests/phpunit/data/themedir1). Done in [2ec97c9](https://github.com/polylang/polylang/pull/1581/commits/2ec97c910da990da759e9cd914bd95b523b09d46).

tests/phpunit/includes/testcase.php
```diff
@@ -41,4 +41,27 @@ protected function assert_has_language( $object, $language, $message = '' ) {
...
+	 * - If defined (value in GitHub Actions is `'true'`): does an assertion.
+	 * - If not defined: skips if the file doesn't exist.
+	 *
+	 * @see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables
+	 *
+	 * @param string $path    Path to the file.
+	 * @param string $message Error message.
+	 * @return void
+	 */
+	protected static function markTestSkippedIfFileNotExists( string $path, string $message = '' ): void {
```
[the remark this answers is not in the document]
Screenfeed: Done in [ebe6590](https://github.com/polylang/polylang/pull/1581/commits/ebe65903da128eef1a5182c090d1ddb4c82c1f46), so it can be used in AJAX testcases and others.

PR #1582, opened 2024-11-26
Adds a class to allow to switch language.

Marianne380:
Following https://github.com/polylang/polylang-pro/pull/2239#discussion_r1854070658.

Specifically, add a `switch_language()` method that will change the current language and load strings translations.

### DIFF
--- a/include/switch-language.php
+++ b/include/switch-language.php
@@ -1,0 +1,101 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+/**
+ * Class to handle site language switch.
+ */
+class PLL_Switch_Language {
+
+	/**
+	 * Singleton instance.
+	 *
+	 * @var PLL_Switch_Language|null
+	 */
+	protected static $instance;
+
+	/**
+	 * The original language.
+	 *
+	 * @var PLL_Language|null
+	 */
+	private $original_language;
+
+	/**
+	 * Access to the single instance of the class.
+	 *
+	 * @since 3.7
+	 *
+	 * @return PLL_Switch_Language
+	 */
+	public static function instance() {
+		if ( empty( self::$instance ) ) {
+			self::$instance = new self();
+		}
+
+		return self::$instance;
+	}
+
+	/**
+	 * Switch the site to the given language.
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Language|string $language The language we want to switch to.
+	 * @return void
+	 */
+	public function switch_language( $language ) {
+		$language = PLL()->model->get_language( $language );
+		if ( ! $language instanceof PLL_Language ) {
+			return;
+		}
+
+		if ( PLL()->curlang === $language ) {
+			return;
+		}
+
+		if ( ! in_array( $language, PLL()->model->languages->get_list() ) ) {
+			return;
+		}
+
+		// Stores the original language.
+		$current_language        = empty( PLL()->curlang ) ? null : PLL()->curlang;
+		$this->original_language = null === $this->original_language ? $current_language : $this->original_language;
+
+		PLL()->curlang = $language;
+
+		PLL()->load_strings_translations( $language->slug );
+
+		/**
+		 * Fires when the language is switched.
+		 *
+		 * @since 3.7
+		 *
+		 * @param PLL_Language $language The new language.
+		 */
+		do_action( 'pll_change_language', $language );
+	}
+
+	/**
+	 * Gets the original language.
+	 *
+	 * @since 3.7
+	 *
+	 * @return PLL_Language|null The language if there is one, false otherwise.
+	 */
+	public function get_original_language() {
+		return $this->original_language;
+	}
+
+	/**
+	 * Restores the original language.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	public function restore_current_language() {
+		PLL()->curlang = $this->original_language;
+	}
+}


include/switch-language.php
```diff
@@ -0,0 +1,173 @@
...
+	 * Sets the current language.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $slug Current language slug.
+	 * @return void
+	 */
+	public static function set_current_language( $slug ) {
+		$language = self::$model->languages->get( $slug );
+		self::$current_language = ! empty( $language ) ? $language : null;
```
Screenfeed, line 61: What is the use of `$current_language`? Shouldn't we assign the new language to `self::$original_language` instead?
Actually I think `$original_language` is the only `$xxx_language` property we need in this class.

include/switch-language.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Model $model Instance of `PLL_Model`.
+	 * @return void
+	 */
+	public static function init( PLL_Model $model ) {
+		self::$model = $model;
+
+		add_action( 'pll_language_defined', array( static::class, 'set_current_language' ) );
```
Screenfeed, line 48: I think we should do this early.

```suggestion
		add_action( 'pll_language_defined', array( static::class, 'set_current_language' ), -1000 );
```

phpstan-baseline.neon
```diff
@@ -730,3 +730,12 @@ parameters:
...
 
+		-
+			message: "#^Offset 'file' does not exist on array\\{function\\: string, line\\?\\: int, file\\?\\: string, class\\?\\: class\\-string, type\\?\\: '\\-\\>'\\|'\\:\\:', args\\?\\: array, object\\?\\: object\\}\\.$#"
+			count: 1
+			path: include/base.php
+
+		-
+			message: "#^Offset 'line' does not exist on array\\{function\\: string, line\\?\\: int, file\\?\\: string, class\\?\\: class\\-string, type\\?\\: '\\-\\>'\\|'\\:\\:', args\\?\\: array, object\\?\\: object\\}\\.$#"
+			count: 1
+			path: include/base.php
```
Screenfeed, line 741: To prevent this, we can use:
```php
esc_html( $debug[0]['file'] ?? '' ),
absint( $debug[0]['line'] ?? 0 )
```

This is what we've done in `PLL_Model::__call()`.

include/switch-language.php
```diff
@@ -0,0 +1,173 @@
...
+	 * Loads user defined strings translations
+	 *
+	 * @since 1.2
+	 * @since 2.1.3 $locale parameter added.
+	 * @since 3.7   Moved from `PLL_Base`.
+	 *
+	 * @param string $locale Language locale or slug. Defaults to current locale.
+	 * @return void
+	 */
+	public static function load_strings_translations( $locale = '' ) {
```
Screenfeed, line 158: This parameter can be type-hinted:
```suggestion
	public static function load_strings_translations( string $locale = '' ): void {
```

Screenfeed (changes requested):
We can add several `: void` return type-hints in `PLL_Switch_Language`.

include/switch-language.php
Screenfeed, line 48: Maybe we should also hook `pll_no_language_defined` 🤔 
If we do, we must not forget that `PLL_REST_Request::$curlang` also accepts `false`.

include/switch-language.php
```diff
@@ -0,0 +1,173 @@
...
+	 * @return void
+	 */
+	public static function set_current_language( $slug ) {
+		$language = self::$model->languages->get( $slug );
+		self::$current_language = ! empty( $language ) ? $language : null;
+	}
+
+	/**
+	 * Switches to the given language.
+	 * Hooked on `pll_post_synchronized` at first.
```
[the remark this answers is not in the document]
Marianne380: [Here](https://github.com/polylang/polylang-pro/pull/2239/files#diff-2c7d1cb4435f5895a8afc187ec0bed19bc144797e4a61cf638fd8c567921e1c3R79). 

It would be weird to add a Polylang Pro action to PLL.

include/switch-language.php
```diff
@@ -0,0 +1,173 @@
...
+	/**
+	 * Switches the language back.
+	 * Hooked on `pll_post_synchronized` at last.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	public static function after_post_synchronized() {
+		self::$previous_language = self::$current_language;
```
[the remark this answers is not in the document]
Marianne380: I need it [here](https://github.com/polylang/polylang-pro/blob/300c1d24972ab3e1cb86a8c591f699a6e3ca033e/integrations/acf/acf-object.php#L239) to see if the current target language is different from the previous one, to see if the strings translations need to be reloaded.

include/switch-language.php
```diff
@@ -0,0 +1,173 @@
...
+	 *
+	 * @param PLL_Language|string|null $language The language we want to switch to.
+	 * @return void
+	 */
+	public static function switch_language( $language = null ) {
+		if ( null === $language ) {
+			self::$current_language = null;
+			return;
+		}
+
```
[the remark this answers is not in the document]
Marianne380: If the original language was `null`, it can be reset to `null` later (with a `restore_original_language()` with a `null` language, for example).

include/switch-language.php
Marianne380: How do you restore the original one if you only have one ?

include/switch-language.php
Marianne380: In this case we should make sure to get a `string` in `PLL_Base::__call` for the `return call_user_func_array( array( PLL_Switch_Language::class, $name ), $args );` no? (to avoid an error for those who misused it before)

include/switch-language.php
Marianne380: Why would we need that?

include/switch-language.php
Screenfeed: Yes.

include/switch-language.php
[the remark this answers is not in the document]
Screenfeed: I'll need more information about why we need it outside. If we still need it, I think `$previous_language` should be `private` and have a getter.

include/switch-language.php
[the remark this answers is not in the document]
Screenfeed: You're right 👍
Maybe we'll need to fire a hook here (I don't know if this would be useful but we may have a thought about it).

include/translated-post.php
```diff
@@ -371,6 +371,9 @@ public function create_media_translation( $post_id, $lang ) {
 		$translations[ $lang->slug ] = $tr_id;
 		$this->save_translations( $tr_id, $translations );
 
+		$switch_language = new PLL_Switch_Language();
+		$switch_language->switch_language( $lang );
```
Screenfeed, line 375:
```suggestion
		PLL_Switch_Language::switch_language( $lang );
```

PR #1586, opened 2024-12-03
Don't add translation entry when the translated string is empty

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/2368

## What?
This issue was revealed by new ACF compatibility and reported in https://github.com/polylang/polylang-pro/issues/2368.
However it is a more general issue.

## Why?
The issue appears since #1574 and as explained in [#2368 comment](https://github.com/polylang/polylang-pro/issues/2368#issuecomment-2514194399), it shows that WordPress translation process replaces any string of Polylang `pll_string` textdomain by its empty translated string instead of keeping the original string.

Because the entry key exists
https://github.com/WordPress/WordPress/blob/6.7.1/wp-includes/pomo/translations.php#L135
the value is returned as the translated string
https://github.com/WordPress/WordPress/blob/6.7.1/wp-includes/pomo/translations.php#L155
(`$translated->translations[0]` instead of `$singular`)

## How?
Add a condition not to add a translation entry in the `pll_string` domain when the source string or its translated string are empty.

### DIFF
--- a/include/mo.php
+++ b/include/mo.php
@@ -85,7 +85,7 @@ class PLL_MO extends MO {
 		foreach ( $strings as $msg ) {
 			$entry = $this->make_entry( $msg[0], $msg[1] );
 
-			if ( '' !== $entry->singular ) {
+			if ( '' !== $entry->singular && '' !== $msg[1] ) {
 				$this->add_entry( $entry );
 			}
 		}
--- a/include/mo.php
+++ b/include/mo.php
@@ -83,9 +83,13 @@ class PLL_MO extends MO {
 		}
 
 		foreach ( $strings as $msg ) {
+			if ( '' === $msg[0] || '' === $msg[1] ) {
+				continue;
+			}
+
 			$entry = $this->make_entry( $msg[0], $msg[1] );
 
-			if ( '' !== $entry->singular && '' !== $msg[1] ) {
+			if ( '' !== $entry->singular ) {
 				$this->add_entry( $entry );
 			}
 		}
--- a/tests/phpunit/tests/test-strings.php
+++ b/tests/phpunit/tests/test-strings.php
@@ -289,4 +289,18 @@ class Strings_Test extends PLL_UnitTestCase {
 		$mo->import_from_db( $lang );
 		$this->assertSame( array( 'test', 'test 2' ), array_keys( $mo->entries ) );
 	}
+
+	public function test_translate_untranslated_string_should_return_original_string() {
+		$pll_admin = new PLL_Admin( $this->links_model );
+		$pll_admin->init();
+
+		$lang = self::$model->get_language( 'en' );
+		$mo   = new PLL_MO();
+		$mo->add_entry( $mo->make_entry( 'test', '' ) );
+		$mo->export_to_db( $lang );
+
+		do_action( 'pll_language_defined' );
+
+		$this->assertSame( 'test', __( 'test', 'pll_string' ) ); // PHPCS:ignore WordPress.WP.I18n
+	}
 }
--- a/tests/phpunit/tests/test-strings.php
+++ b/tests/phpunit/tests/test-strings.php
@@ -291,15 +291,14 @@ class Strings_Test extends PLL_UnitTestCase {
 	}
 
 	public function test_translate_untranslated_string_should_return_original_string() {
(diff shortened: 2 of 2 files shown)


include/mo.php
```diff
@@ -85,7 +85,7 @@ public function import_from_db( $lang ) {
 		foreach ( $strings as $msg ) {
 			$entry = $this->make_entry( $msg[0], $msg[1] );
 
-			if ( '' !== $entry->singular ) {
+			if ( '' !== $entry->singular && '' !== $msg[1] ) {
```
Hug0-Drelon, line 88: Let's improve the code and check `$msg[0]` and `$msg[1]` emptiness *before* we create the entry.

Hug0-Drelon (changes requested):

include/mo.php
manooweb: 🤔Not necessarily. `MO:make_entry()` empties entry singular property if the source string is empty
https://github.com/WordPress/WordPress/blob/6.7.1/wp-includes/pomo/mo.php#L329

manooweb: Done in 9394c3bbdc2db954690b59aa15c977caf5c4dca3

Hug0-Drelon (approved):

tests/phpunit/tests/test-strings.php
```diff
@@ -289,4 +289,18 @@ public function test_import_empty_string() {
...
+		$pll_admin->init();
+
+		$lang = self::$model->get_language( 'en' );
+		$mo   = new PLL_MO();
+		$mo->add_entry( $mo->make_entry( 'test', '' ) );
+		$mo->export_to_db( $lang );
+
+		do_action( 'pll_language_defined' );
+
+		$this->assertSame( 'test', __( 'test', 'pll_string' ) ); // PHPCS:ignore WordPress.WP.I18n
```
Chouby, line 304: Wouldn't this be better? 
```suggestion
		$this->assertSame( 'test', pll__( 'test' ) );
```

tests/phpunit/tests/test-strings.php
```diff
@@ -289,4 +289,18 @@ public function test_import_empty_string() {
...
+	public function test_translate_untranslated_string_should_return_original_string() {
+		$pll_admin = new PLL_Admin( $this->links_model );
+		$pll_admin->init();
+
+		$lang = self::$model->get_language( 'en' );
+		$mo   = new PLL_MO();
+		$mo->add_entry( $mo->make_entry( 'test', '' ) );
+		$mo->export_to_db( $lang );
+
+		do_action( 'pll_language_defined' );
```
Chouby, line 302: Why admin context? It's not useful to update strings in DB.
The action `pll_language_defined` without defining a language ?
If I look at previous tests, this is the kind of things usually done: https://github.com/polylang/polylang/blob/3d9bc5cd01f9dd5461e5783e62dd3ea86e0f9493/tests/phpunit/tests/test-strings.php#L198-L200

Chouby (changes requested):

tests/phpunit/tests/test-strings.php
manooweb: In fact, the issue #2368 was revealed in a admin context with a ACF compatibility issue where `__()` is callled instead of `pll__()`. It's why I chose this solution.
See 
https://github.com/polylang/polylang-pro/blob/new-acf-dev/integrations/acf/acf-translate-labels.php#L76-L78
and
https://github.com/AdvancedCustomFields/acf/blob/6.3.11/includes/api/api-helpers.php#L3103

I said to myself that this case could also happen in this way

tests/phpunit/tests/test-strings.php
manooweb: Because the issue #2368 was revealed in a admin context (not frontend) with a ACF compatibility issue and also without a specific language defined (default WordPress locale one).

To be closer to the reality, I propose to trigger `setup_theme` action as Polylang `init_user()` is hooked on.
Thus `pll_no_language_defined` is done instead of `pll_language_defined`.
Then we also trigger `PLL_Base::load_strings_translations()` through `PLL_OLT_Manager::load_textdomains()` hooked on `pll_no_language_defined`.

We can also use `pll_no_language_defined` if you prefer.

PR #1588, opened 2024-12-03
When updating options in REST, return all values

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/2373.

When updating options in REST, return all values instead of only the updated ones.

### DIFF
--- a/modules/REST/V1/Settings.php
+++ b/modules/REST/V1/Settings.php
@@ -98,10 +98,9 @@ class Settings extends Abstract_Controller {
 	 * @phpstan-param WP_REST_Request<T> $request
 	 */
 	public function update_item( $request ) {
-		$response = array();
-		$errors   = new WP_Error();
-		$schema   = $this->options->get_schema();
-		$options  = array_intersect_key(
+		$errors  = new WP_Error();
+		$schema  = $this->options->get_schema();
+		$options = array_intersect_key(
 			$request->get_params(),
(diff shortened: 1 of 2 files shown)


Hug0-Drelon (changes requested):
LGTM, but we may want to cover with a test.
What do you think @Screenfeed?

Screenfeed:
@Hug0-Drelon [There is](https://github.com/polylang/polylang/blob/5df10b0301ed92234a9af7596dd843df195d62a7/tests/phpunit/tests/test-rest-settings.php#L95), but it asserts the new values only.

Hug0-Drelon:
> @Hug0-Drelon [There is](https://github.com/polylang/polylang/blob/5df10b0301ed92234a9af7596dd843df195d62a7/tests/phpunit/tests/test-rest-settings.php#L95), but it asserts the new values only.

So the behavior isn't covered, right?

Screenfeed:
It doesn't test that all values are returned indeed, do you think it's worth testing?

Hug0-Drelon:
> It doesn't test that all values are returned indeed, do you think it's worth testing?

I think it worth doing that but in a tiny test, just asserting two arrays with only one value updated.

Screenfeed:
Done in [9fe1d3c](https://github.com/polylang/polylang/pull/1588/commits/9fe1d3c728b0e32a469ba83f280243c624a713ec).

Hug0-Drelon (approved):

PR #1589, opened 2024-12-04
Don't allow setting `default_lang` value with non-existing language.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2375

## Why?
Prevent setting bad default language value through REST.
I chose to manage the sanitization on option object (`Default_Lang`)  level, so any external code using this feature will benefit from it (e.g. Polylang CLI?).

## How?
- Implement `Default_Lang::sanitize()` and ensure term language with the given slug exists before updating the option.
- Update `default_lang` option in `Languages::update()` only once the langue term *is* really updated.
- Cover the expected behavior on option object level.
- Cover the expected behavior on REST response level.
- Fix `REST_Request_Test::test_should_define_default_language_when_language_is_invalid` test assigning bad default language using proper `create_option()` method.

### DIFF
--- a/include/Options/Business/Default_Lang.php
+++ b/include/Options/Business/Default_Lang.php
@@ -76,6 +76,7 @@ class Default_Lang extends Abstract_String {
 			return $value;
 		}
 
+		/** @var string $value */
 		if ( ! get_term_by( 'slug', $value, 'language' ) ) {
 			return new WP_Error( 'pll_invalid_language', sprintf( 'The language slug \'%s\' is not a valid language.', $value ) );
 		}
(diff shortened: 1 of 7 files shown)


include/Options/Business/Default_Lang.php
```diff
@@ -55,4 +57,30 @@ protected function get_data_structure(): array {
...
 	}
+
+	/**
+	 * Sanitizes option's value.
+	 * Can populate the `$errors` property with blocking and non-blocking errors: in case of non-blocking errors,
+	 * the value is sanitized and can be stored.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array   $value   Value to sanitize.
```
Screenfeed, line 68: Copy/Paste drama.

```suggestion
	 * @param string  $value   Value to sanitize.
```

Screenfeed (changes requested):

Screenfeed (approved):

PR #1590, opened 2024-12-04
Remove `no_default_cat` property from schema in editable language requests.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2382
The issue can be seen with either route `OPTIONS https://mysite.site/wp-json/pll/v1/languages` or  `GET https://mysite.site/wp-json/pll/v1/`.

## How?
- Override `get_endpoint_args_for_item_schema()` and remove `no_default_cat` property from the schema only for `EDITABLE` requests.
- Cover the behavior in `REST_langues_Test::test_no_default_cat_arg()`.
- Took the opportunity to improve `REST_langues_Test::test_create_language()` by testing with different `no_default_cat` values.

### DIFF
--- a/modules/REST/V1/Languages.php
+++ b/modules/REST/V1/Languages.php
@@ -637,7 +637,7 @@ class Languages extends Abstract_Controller {
 	 * @return array Endpoint arguments.
 	 */
 	public function get_endpoint_args_for_item_schema( $method = WP_REST_Server::CREATABLE ) {
-		$schema = $this->get_item_schema()
+		$schema = $this->get_item_schema();
 		if ( WP_REST_Server::CREATABLE !== $method ) {
 			unset( $schema['properties']['no_default_cat'] );
 		}
--- a/modules/REST/V1/Languages.php
+++ b/modules/REST/V1/Languages.php
@@ -629,13 +629,11 @@ class Languages extends Abstract_Controller {
 
 	/**
 	 * Retrieves an array of endpoint arguments from the item schema for the controller.
-	 * Ensures that the `no_default_cat` property is not returned for `EDITABLE` requests.
+	 * Ensures that the `no_default_cat` property is returned only for `CREATABLE` requests.
 	 *
 	 * @since 3.7
 	 *
-	 * @param string $method Optional. HTTP method of the request. The arguments for `CREATABLE` requests are
-	 *                       checked for required values and may fall-back to a given default, this is not done
-	 *                       on `EDITABLE` requests. Default WP_REST_Server::CREATABLE.
+	 * @param string $method Optional. HTTP method of the request. Default WP_REST_Server::CREATABLE.
 	 * @return array Endpoint arguments.
 	 */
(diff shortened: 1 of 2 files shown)


Screenfeed (approved):

modules/REST/V1/Languages.php
```diff
@@ -627,6 +627,28 @@ public function get_item_schema(): array {
...
+	 */
+	public function get_endpoint_args_for_item_schema( $method = WP_REST_Server::CREATABLE ) {
+		if ( WP_REST_Server::EDITABLE === $method ) {
+			$edit_schema = $this->get_item_schema();
+			unset( $edit_schema['properties']['no_default_cat'] );
+
+			return rest_get_endpoint_args_for_schema( $edit_schema, $method );
+		}
+
+		return parent::get_endpoint_args_for_item_schema( $method );
```
Chouby, line 649: What do you think of this slightly less verbose version?
Also I would check CREATABLE instead
```suggestion
		$schema = $this->get_item_schema()
		if ( WP_REST_Server::CREATABLE !== $method ) {
			unset( $schema['properties']['no_default_cat'] );
		}
		return rest_get_endpoint_args_for_schema( $schema, $method );
```

modules/REST/V1/Languages.php
```diff
@@ -627,6 +627,28 @@ public function get_item_schema(): array {
 		return $this->add_additional_fields_schema( $this->schema );
 	}
 
+	/**
+	 * Retrieves an array of endpoint arguments from the item schema for the controller.
+	 * Ensures that the `no_default_cat` property is not returned for `EDITABLE` requests.
```
Chouby, line 632:
```suggestion
	 * Ensures that the `no_default_cat` property is returned only for `CREATABLE` requests.
```

modules/REST/V1/Languages.php
```diff
@@ -627,6 +627,28 @@ public function get_item_schema(): array {
...
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $method Optional. HTTP method of the request. The arguments for `CREATABLE` requests are
+	 *                       checked for required values and may fall-back to a given default, this is not done
+	 *                       on `EDITABLE` requests. Default WP_REST_Server::CREATABLE.
```
Chouby, line 638: Extra comments are relevant to the WordPress function:
```suggestion
	 * @param string $method Optional. HTTP method of the request. Default WP_REST_Server::CREATABLE.
```

Chouby (changes requested):

modules/REST/V1/Languages.php
Hug0-Drelon: I'm ok for checking `CREATABLE` instead, that way we ensure covering *all* other methods.
But with your proposal, we may have a problem if `parent::get_endpoint_args_for_item_schema()` changes. We won't benefit from its code anymore, right?

Chouby: I thought to this too but we would have a problem anyway with the direct call to `rest_get_endpoint_args_for_schema()`inside the test.
Hopefully WP will change only the function and never the method.

PR #1591, opened 2024-12-04
Make REST settings context agnostic.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2376

## How?
- Remove `context` parameter from options schema.
- Remove checks for context in `Settings`.
- Adapt tests accordingly.

### DIFF
--- a/tests/phpunit/tests/test-rest-settings.php
+++ b/tests/phpunit/tests/test-rest-settings.php
@@ -144,16 +144,17 @@ class REST_Settings_Test extends PLL_UnitTestCase {
 	 * @param string $method
 	 * @return void
 	 */
-	public function test_wrong_context( string $method ) {
+	public function test_context_param_should_be_ignored( string $method ) {
 		wp_set_current_user( self::$administrator );
 
-		// Context 'view'.
 		$response = $this->dispatch_request( $method, array( 'context' => 'view' ) );
-		$this->assertSame( 400, $response->get_status() );
+		$this->assertSame( 200, $response->get_status() );
 
-		// Context 'embed'.
 		$response = $this->dispatch_request( $method, array( 'context' => 'embed' ) );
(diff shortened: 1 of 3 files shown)


Screenfeed:
*”Cough”* https://github.com/polylang/polylang/pull/1569#discussion_r1843903317

Hug0-Drelon:
> _”Cough”_ [#1569 (comment)](https://github.com/polylang/polylang/pull/1569#discussion_r1843903317)

It was the misunderstanding of the behavior of the method. I thought it was adding `edit` to the schema for *our* properties that didn't have that set. No properties passed along a request.
Moreover, the behavior was not tested at all and passed under the radar.

Also, we may need to rollback about this context, see https://github.com/polylang/polylang-pro/issues/2376#issuecomment-2518637200

tests/phpunit/tests/test-rest-settings.php
```diff
@@ -163,6 +163,15 @@ public function test_unknown_method() {
 		$this->assertSame( 404, $response->get_status() );
 	}
 
+	public function test_get_settings_with_no_context() {
```
Screenfeed, line 166: This could go in `test_wrong_context()`.

Screenfeed (changes requested):

Screenfeed:
>Moreover, the behavior was not tested at all and passed under the radar.

It was already tested (that's why we have 4 tests failing) and didn't passed under the radar (since it was expected).

tests/phpunit/tests/test-rest-settings.php
Hug0-Drelon: Not applicable anymore.

Hug0-Drelon:
The PR changes its purpose, now we don't care about `context` anymore.

PR #1594, opened 2024-12-05
Update default language upon setting change in REST

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2374.

This PR updates the default language when the `default_lang` setting is changed via the REST API.

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -624,10 +624,10 @@ class Languages {
 	 * @param string $slug New language slug.
 	 * @return WP_Error A `WP_Error` object containing possible errors during slug validation/sanitization.
 	 */
-	public function update_default( $default_lang ): WP_Error {
+	public function update_default( $slug ): WP_Error {
 		$prev_default_lang = $this->options->get( 'default_lang' );
 
-		if ( $prev_default_lang === $default_lang ) {
+		if ( $prev_default_lang === $slug ) {
 			return new WP_Error();
 		}
 
@@ -643,7 +643,7 @@ class Languages {
 			$menus = array();
 
 			foreach ( $this->options['nav_menus'][ $theme ] as $key => $loc ) {
-				$menus[ $key ] = empty( $loc[ $default_lang ] ) ? 0 : $loc[ $default_lang ];
+				$menus[ $key ] = empty( $loc[ $slug ] ) ? 0 : $loc[ $slug ];
 			}
 			set_theme_mod( 'nav_menu_locations', $menus );
 		}
@@ -655,10 +655,10 @@ class Languages {
 		 * @since 3.7 The previous default language's slug is passed as 2nd param.
 		 *            The default language is updated before this hook is fired.
 		 *
-		 * @param string $default_lang      New default language's slug.
+		 * @param string $slug              New default language's slug.
 		 * @param string $prev_default_lang Previous default language's slug.
 		 */
-		do_action( 'pll_update_default_lang', $default_lang, $prev_default_lang );
+		do_action( 'pll_update_default_lang', $slug, $prev_default_lang );
 
 		// Update options.
 
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -624,18 +624,17 @@ class Languages {
 	 * @param string $slug New language slug.
 	 * @return WP_Error A `WP_Error` object containing possible errors during slug validation/sanitization.
 	 */
-	public function update_default( $slug ): WP_Error {
+	public function update_default( $default_lang ): WP_Error {
 		$prev_default_lang = $this->options->get( 'default_lang' );
-		$errors            = $this->options->set( 'default_lang', $slug );
-
-		if ( $errors->has_errors() ) {
-			return $errors;
-		}
-
-		$default_lang = $this->options->get( 'default_lang' );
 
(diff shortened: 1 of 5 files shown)


modules/REST/V1/Settings.php
```diff
@@ -106,10 +114,19 @@ public function update_item( $request ) {
...
 		foreach ( $options as $option_name => $new_value ) {
-			$result = $this->options->set( $option_name, $new_value );
+			$prev_value = $this->options->get( $option_name );
+			$result     = $this->options->set( $option_name, $new_value );
 
 			if ( $result->has_errors() ) {
 				$errors->merge_from( $result );
+				continue;
+			}
+			if ( $prev_value === $new_value ) {
```
Hug0-Drelon, line 124: I wonder if `Options::set()` should return a specific error in such case.
It would prevent you to make this test and get the previous value above.

modules/REST/V1/Settings.php
```diff
@@ -106,10 +114,19 @@ public function update_item( $request ) {
...
+				continue;
+			}
+			// Default language.
```
Hug0-Drelon, line 127:
```suggestion
```
I think the code speaks for itself.

Hug0-Drelon (changes requested):
The is a flaw with `default_option`, you've fixed it for REST Settings. But I fear we'll duplicate the logic you implemented in other context (e.g. CLI).

An other solution would be to have some sort of middleware that calls `Languages::update_default()` when the corresponding options has changes (either directly in `Options`, or in `Default_Lang` or even through a new hook.

What do you think?

modules/REST/V1/Settings.php
Screenfeed: Am I not even allowed to manspain my own code?!
Done in [6af6129](https://github.com/polylang/polylang/pull/1594/commits/6af61290e152f4bb776b6b0087b4267436e24bd9) 😬

modules/REST/V1/Settings.php
Screenfeed: OMG this is actually the only place (outside `Options`) we use `set()` o.O

Screenfeed: I think we shouldn't, because this is not an error.

And concretely, [here](https://github.com/polylang/polylang/blob/6af61290e152f4bb776b6b0087b4267436e24bd9/settings/settings-module.php#L275) is the only place (so far) where we use `Options::merge()`, which internally uses `Options::set()`. If we do add an error for equal values, saving the settings in PLL's settings page will display a lot of these "errors", so we'll have to add more code to prevent this.

include/Model/Languages.php
```diff
@@ -602,22 +602,36 @@ public function get_default() {
 
 	/**
 	 * Updates the default language.
-	 * taking care to update the default category & the nav menu locations.
+	 * Taking care to update the default category & the nav menu locations.
```
Hug0-Drelon, line 605:
```suggestion
	 * Takes care to update default category, nav menu locations and to flush  cache and rewrite rules.
```

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

include/Model/Languages.php
```diff
@@ -602,22 +602,36 @@ public function get_default() {
...
+
+		if ( $errors->has_errors() ) {
+			return $errors;
+		}
+
+		$default_lang = $this->options->get( 'default_lang' );
+
+		if ( $prev_default_lang === $default_lang ) {
+			return new WP_Error();
+		}
```
Chouby, line 626: 1. Good idea to check that the new default language is different but I would do this first.
2. Why reading the new option once it has been set. If there are no errors it can be only what has been set. Do I miss something? 
```suggestion
	public function update_default( $default_lang ): WP_Error {
		$prev_default_lang = $this->options->get( 'default_lang' );

		if ( $prev_default_lang === $default_lang ) {
			return new WP_Error();
		}

		$errors = $this->options->set( 'default_lang', $slug );

		if ( $errors->has_errors() ) {
			return $errors;
		}
```

modules/REST/API.php
```diff
@@ -56,7 +56,7 @@ public function init(): void {
 		$this->languages = new V1\Languages( $this->model->languages, $this->model->translatable_objects );
 		$this->languages->register_routes();
 
-		$this->settings = new V1\Settings( $this->model->options );
+		$this->settings = new V1\Settings( $this->model->options, $this->model->languages );
```
Chouby, line 59: I would pass only one parameter: `$this->model`.

Chouby (changes requested):

include/Model/Languages.php
Screenfeed: On both points: because the sanitized value could differ from the raw value.
Except that it is not possible in this specific case: the validation will use the regex pattern, and the sanitization will only cast the value to a string. So you're right, since the sanitization won't modify the value, it seems safe to compare the old value at the beginning.

modules/REST/API.php
Screenfeed: Done in [96edeca](https://github.com/polylang/polylang/pull/1594/commits/96edeca6efffd98bf453b9bb9f1e1de236bdcb31).

Hug0-Drelon (approved):

PR #1595, opened 2024-12-12
Fix sub-options not being updated

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2379.

This PR provides an alternative to the following notation:

```php
$polylang->options['nav_menus']['twentytwentyone']['primary']['fr'] = $menu_id;
```

## What

Polylang's `Options` class implements `ArrayAccess`, which allows us to do:

```php
$polylang->options['foo'] = array( 'k' => 'v' );
```

Unfortunatly this doesn't allow to do:

```php
$polylang->options['foo']['k'] = 'v';
```

This works if `offsetGet()` returns a reference, aka `&offsetGet()`. However, even by doing so, `offsetSet()` is not called, which prevents sanitization of the data and setting the `$modified` property, which then doesn't trigger the DB saving process in `save_all()`.
Worst case scenario would be:

```php
$polylang->options['bar'] = 'baz'; // Sanitized, triggers `save_all()`.
$polylang->options['foo']['k'] = 'v'; // NOT sanitized, and the previous line will save this unsafe data into the DB.
```

### First solution

One possible solution is to replace all arrays by recursive implementations of `ArrayAccess`, where each `ArrayAccess::offsetSet()` calls `offsetSet()` of its parent. This works but comes with several issues, the main one being that array functions cannot be used anymore, and triggers a fatal error if do so (array functions only accept... arrays, not objects that implement `ArrayAccess`, so it's a type issue), so it's a big backward compatibility issue.

This has been discarded because it introduces too many issues.

### Second solution

The advantage of this solution is that it changes only assignation, it doesn't impact reading values, which allows a better backward compatibility.
This solution introduces 2 new methods: `Options::set_sub_value()` and `Options::unset_sub_value()`.
They can be used this way:

```php
$polylang->options->set_sub_value( 'nav_menus', array( 'twentytwentyone', 'primary', 'fr' ), $menu_id );
$polylang->options->unset_sub_value( 'nav_menus', array( 'twentytwentyone', 'primary', 'fr' ) );
```

However, this solution has been also discarded because it makes the code less readable.

### Third solution, the one in this PR

In the end we chose not to introduce new things, and simply change the way we set or unset sub-values:

```php
$nav_menus = $polylang->options->get( 'nav_menus' );
foreach ( $nav_menus as $theme => $menus ) {
    foreach ( $menus as $location => $ids ) {
        unset( $nav_menus[ $theme ][ $location ]['fr'] );
    }
}
$polylang->options->set( 'nav_menus', $nav_menus );
```

That way we keep the object notation on `$polylang->options` and use the array notation on the values themselves.

### About `expectException()` in tests

PHPUnit triggers a warning in these cases (because this disappears in PHPUnit 10):
```php
use PHPUnit\Framework\Error\Notice;
...
$this->expectNotice();
$this->expectException( Notice::class );
$this->expectException( \PHPUnit\Framework\Error\Notice::class );
```

No warning from PHPUnit, but Rector is not happy:
```php
$this->expectException( '\PHPUnit\Framework\Error\Notice' );
```

So my solution is to silence Rector in the config file.


admin/admin-nav-menu.php
```diff
@@ -262,14 +273,22 @@ public function theme_mod_nav_menu_locations( $menus ) {
 	 * @return void
 	 */
 	public function delete_nav_menu( $term_id ) {
-		foreach ( $this->options['nav_menus'] as $theme => $locations ) {
+		if ( empty( $this->options->get( 'nav_menus' ) ) ) {
```
Hug0-Drelon, line 276: This is to prevent calling `set()` with empty value?

include/Model/Languages.php
```diff
@@ -303,21 +303,27 @@ public function update( $args ) {
 			}
 
 			// Update menus locations.
-			if ( ! empty( $this->options['nav_menus'] ) ) {
-				foreach ( $this->options['nav_menus'] as $theme => $locations ) {
+			if ( ! empty( $this->options->get( 'nav_menus' ) ) ) {
+				$nav_menus = $this->options->get( 'nav_menus' );
```
Hug0-Drelon, line 307:
```suggestion
                         $nav_menus = $this->options->get( 'nav_menus' );
			if ( ! empty( $nav_menus ) ) {
```
I'm surprised PHPStan doesn't throw an error. How does he know the method is pure?

include/Model/Languages.php
```diff
@@ -303,21 +303,27 @@ public function update( $args ) {
...
+
+				$this->options->set( 'nav_menus', $nav_menus );
 			}
 
 			// Update domains.
-			if ( ! empty( $this->options['domains'][ $old_slug ] ) ) {
-				$this->options['domains'][ $slug ] = $this->options['domains'][ $old_slug ];
-				unset( $this->options['domains'][ $old_slug ] );
+			if ( ! empty( $this->options->get( 'domains' )[ $old_slug ] ) ) {
+				$domains = $this->options->get( 'domains' );
```
Hug0-Drelon, line 323:
```suggestion
                         $domains = $this->options->get( 'domains' );
			if ( ! empty( $domains ) ) {
```

include/Model/Languages.php
```diff
@@ -414,20 +420,26 @@ public function delete( $lang_id ): bool {
 		}
 
 		// Delete menus locations.
-		if ( ! empty( $this->options['nav_menus'] ) ) {
-			foreach ( $this->options['nav_menus'] as $theme => $locations ) {
+		if ( ! empty( $this->options->get( 'nav_menus' ) ) ) {
+			$nav_menus = $this->options->get( 'nav_menus' );
```
Hug0-Drelon, line 424:
```suggestion
                 $nav_menus = $this->options->get( 'nav_menus' );
		if ( ! empty( $nav_menus ) ) {
```

tests/phpunit/tests/Options/Options/test-OffsetSet.php
```diff
@@ -0,0 +1,34 @@
...
+	public function test_arrayaccess_should_not_set_sub_value() {
+		$options = self::create_options();
+
+		$this->assertSame( array(), $options->get( 'nav_menus' ), 'The initial value should be an empty array.' );
+
+		/*
+		 * This will break with PHPUnit 10.
+		 * @see https://github.com/sebastianbergmann/phpunit/issues/5062
+		 */
+		$this->expectException( '\PHPUnit\Framework\Error\Notice' );
```
Hug0-Drelon, line 27: Can't we use `try...catch` to avoid these expectations?

tests/phpunit/tests/Options/Options/test-OffsetUnset.php
```diff
@@ -0,0 +1,47 @@
...
+			)
+		);
+
+		$this->assertSame( $nav_menus, $options->get( 'nav_menus' ), 'The initial value should have the value we want to unset.' );
+
+		/*
+		 * This will break with PHPUnit 10.
+		 * @see https://github.com/sebastianbergmann/phpunit/issues/5062
+		 */
+		$this->expectException( '\PHPUnit\Framework\Error\Notice' );
```
Hug0-Drelon, line 40: Same question as above.

tests/phpunit/tests/Options/Options/test-OffsetUnset.php
```diff
@@ -0,0 +1,47 @@
...
+use PLL_UnitTestCase;
+
+/**
+ * Tests for `Options\Options->offsetUnset()`.
+ *
+ * @group options
+ */
+class OffsetUnset_Test extends PLL_UnitTestCase {
+	/**
+	 * Makes sure that `offsetUnset()` doesn't modify the option's value (because `offsetGet()` doesn't return a reference).
```
Hug0-Drelon, line 14: Could we make `offsetUnset()` behave as usual even though `offsetGet()` doesn't return a reference?

Hug0-Drelon (changes requested):

admin/admin-nav-menu.php
Screenfeed: Yes, to prevent calling `set()` while there is nothing to update.

include/Model/Languages.php
Screenfeed: Why would PHPStan throw an error?
I didn't do that because `get()` is cheap, but I don't mind changing.

include/Model/Languages.php
Screenfeed: Done in [d78b566](https://github.com/polylang/polylang/pull/1595/commits/d78b566c2ac53d015116da9415cfd409770b70ff).

include/Model/Languages.php
Screenfeed: Done in [d78b566](https://github.com/polylang/polylang/pull/1595/commits/d78b566c2ac53d015116da9415cfd409770b70ff).

tests/phpunit/tests/Options/Options/test-OffsetUnset.php
Screenfeed: No, the issue is the same: child's `offsetUnset()` won't trigger parent's `offsetUnset()`.

include/Model/Languages.php
Hug0-Drelon: PHPStan should throw an error because to consecutive calls to a user defined function (here `PLL_ABstract_Options::get()` could return different values.
Of course this is not the case here since `PLL_ABstract_Options::get()` is a pure method, but PHPStan doesn't know that.
So I'm surprised it doesn't throw any error.
You could kept the code as it is, it was more a general question.

Hug0-Drelon (approved):

admin/admin-nav-menu.php
```diff
@@ -262,14 +273,22 @@ public function theme_mod_nav_menu_locations( $menus ) {
...
+			return;
+		}
+
+		$nav_menus = $this->options->get( 'nav_menus' );
```
Chouby, line 280: To avoid one call:
```suggestion
		$nav_menus = $this->options->get( 'nav_menus' );

		if ( empty( $nav_menus ) ) {
			return;
		}
```

Chouby (changes requested):

PR #1596, opened 2024-12-13
Flush rewrite rules when specific options are updated in REST.

Hug0-Drelon:
## What?
There is an issue when updating some links related options through REST. For instance, updating `rewrite` options doesn't flush rewrite rules then breaks the frontend.

## How?
- Flush rewrite rules when options such as `rewrite`, `force_lang` or `hide_default` is updated in REST.
- Add a tests for `rewrite` case, tell me if more is needed.

### DIFF
--- a/modules/REST/V1/Settings.php
+++ b/modules/REST/V1/Settings.php
@@ -103,6 +103,10 @@ class Settings extends Abstract_Controller {
 		);
 
 		foreach ( $options as $option_name => $new_value ) {
+			if ( $new_value === $this->options->get( $option_name ) ) {
+				continue;
+			}
+
 			$result = $this->options->set( $option_name, $new_value );
 
 			if ( $result->has_errors() ) {
--- a/modules/REST/V1/Settings.php
+++ b/modules/REST/V1/Settings.php
@@ -103,14 +103,15 @@ class Settings extends Abstract_Controller {
 		);
 
 		foreach ( $options as $option_name => $new_value ) {
-			if ( $new_value === $this->options->get( $option_name ) ) {
-				continue;
-			}
-
-			$result = $this->options->set( $option_name, $new_value );
+			$previous_value = $this->options->get( $option_name );
+			$result         = $this->options->set( $option_name, $new_value );
 
 			if ( $result->has_errors() ) {
 				$errors->merge_from( $result );
+				continue;
+			}
+
+			if ( $previous_value === $this->options->get( $option_name ) ) {
 				continue;
 			}
 
--- a/modules/REST/V1/Settings.php
+++ b/modules/REST/V1/Settings.php
@@ -111,7 +111,7 @@ class Settings extends Abstract_Controller {
 				continue;
 			}
 
-			if ( $previous_value === $this->options->get( $option_name ) ) {
+			if ( $this->options->get( $option_name ) === $previous_value ) {
 				continue;
 			}
 
--- a/modules/REST/V1/Settings.php
+++ b/modules/REST/V1/Settings.php
@@ -107,6 +107,12 @@ class Settings extends Abstract_Controller {
 
 			if ( $result->has_errors() ) {
 				$errors->merge_from( $result );
+				continue;
+			}
+
+			$options_with_flush = array( 'rewrite', 'force_lang', 'hide_default' );
+			if ( in_array( $option_name, $options_with_flush, true ) ) {
+				flush_rewrite_rules();
 			}
 		}
 
--- a/modules/REST/V1/Settings.php
+++ b/modules/REST/V1/Settings.php
@@ -115,9 +115,11 @@ class Settings extends Abstract_Controller {
 				continue;
 			}
 
-			$options_with_flush = array( 'rewrite', 'force_lang', 'hide_default' );
-			if ( in_array( $option_name, $options_with_flush, true ) ) {
-				flush_rewrite_rules();
+			switch ( $option_name ) {
+				case 'rewrite':
+				case 'force_lang':
+				case 'hide_default':
+					flush_rewrite_rules();
 			}
 		}
 
--- a/tests/phpunit/tests/test-rest-settings.php
+++ b/tests/phpunit/tests/test-rest-settings.php
@@ -173,6 +173,45 @@ class REST_Settings_Test extends PLL_UnitTestCase {
 		$this->assertSame( 'pll_invalid_language', $response->get_data()['code'] );
 	}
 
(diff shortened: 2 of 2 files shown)


modules/REST/V1/Settings.php
```diff
@@ -107,6 +107,12 @@ public function update_item( $request ) {
 
 			if ( $result->has_errors() ) {
 				$errors->merge_from( $result );
+				continue;
+			}
+
+			$options_with_flush = array( 'rewrite', 'force_lang', 'hide_default' );
+			if ( in_array( $option_name, $options_with_flush, true ) ) {
+				flush_rewrite_rules();
 			}
```
Screenfeed, line 116: Maybe we should do this only if the value change?

```php
$prev_value = $this->options->get( $option_name );
$result     = $this->options->set( $option_name, $new_value );

if ( $result->has_errors() ) {
	$errors->merge_from( $result );
	continue;
}

if ( $this->options->get( $option_name ) === $prev_value ) {
	continue;
}

$options_with_flush = array( 'rewrite', 'force_lang', 'hide_default' );
if ( in_array( $option_name, $options_with_flush, true ) ) {
	flush_rewrite_rules();
}
```

What do you think?

tests/phpunit/tests/test-rest-settings.php
```diff
@@ -173,6 +173,45 @@ public function test_should_not_update_default_language_with_bad_value() {
...
+
+		// Redo frontend to get new rules.
+		$frontend->links_model->remove_filters();
+		$links_model = $this->pll_env->model->get_links_model();
+		$links_model->init();
+		$frontend = new PLL_Frontend( $links_model );
+		$frontend->init();
+		$rules_after_update = $wp_rewrite->rewrite_rules();
+
+		$this->assertNotSame( $rules_prior_to_update, $rules_after_update );
```
Screenfeed, line 211: I'm not sure this is useful, since we have the next line.
But if you prefer to keep it, maybe use `assertNotSameSetsWithIndex()` instead?

Screenfeed (changes requested):

modules/REST/V1/Settings.php
Hug0-Drelon: Why not, but I we got that way I propose to do nothing at all in such case.
See 84c2a709.

tests/phpunit/tests/test-rest-settings.php
Hug0-Drelon: You're right, it's pointless. I primarily used that to iterate in my test and forgot to remove it.
Removed in 499cde94.

modules/REST/V1/Settings.php
```diff
@@ -103,10 +103,20 @@ public function update_item( $request ) {
 		);
 
 		foreach ( $options as $option_name => $new_value ) {
+			if ( $new_value === $this->options->get( $option_name ) ) {
+				continue;
+			}
```
Screenfeed, line 108: This should be done after sanitization, so boolans can be casted from `1` to `true` by the sanitization for example.

Hug0-Drelon: Done in c16372da.

modules/REST/V1/Settings.php
```diff
@@ -103,17 +103,18 @@ public function update_item( $request ) {
...
-			$result = $this->options->set( $option_name, $new_value );
+			$previous_value = $this->options->get( $option_name );
+			$result         = $this->options->set( $option_name, $new_value );
 
 			if ( $result->has_errors() ) {
 				$errors->merge_from( $result );
 				continue;
 			}
 
+			if ( $previous_value === $this->options->get( $option_name ) ) {
```
Screenfeed, line 114: 😅

```suggestion
			if ( $this->options->get( $option_name ) === $previous_value ) {
```

Screenfeed (changes requested):

Screenfeed (approved):

Screenfeed (approved):

modules/REST/V1/Settings.php
```diff
@@ -103,10 +103,21 @@ public function update_item( $request ) {
...
+			$result         = $this->options->set( $option_name, $new_value );
 
 			if ( $result->has_errors() ) {
 				$errors->merge_from( $result );
+				continue;
+			}
+
+			if ( $this->options->get( $option_name ) === $previous_value ) {
+				continue;
+			}
```
Chouby, line 116: I suggest to avoid one call to `get()` and prevent useless `set()` if the new value is not different.
```php
			$previous_value = $this->options->get( $option_name );
			if ( $new_value === $previous_value ) {
				continue;
			}

			$result = $this->options->set( $option_name, $new_value );

			if ( $result->has_errors() ) {
				$errors->merge_from( $result );
				continue;
			}
```

modules/REST/V1/Settings.php
```diff
@@ -103,10 +103,21 @@ public function update_item( $request ) {
...
+
+			$options_with_flush = array( 'rewrite', 'force_lang', 'hide_default' );
+			if ( in_array( $option_name, $options_with_flush, true ) ) {
+				flush_rewrite_rules();
 			}
```
Chouby, line 121: What do you think of?
```suggestion
			switch ( $option_name ) {
				case 'rewrite':
				case 'force_lang':
				case 'hide_default':
					flush_rewrite_rules();
			}
```

modules/REST/V1/Settings.php
Screenfeed: In this case the sanitized value can be different from the raw value. Example: all the boolean values: `1` => `true`. So the comparison must be done after `set()`.

Screenfeed:
> Shouldn't we handle this directly at options level?

We considered doing this but we think we shouldn't: the `Options` class shouldn't do everything, we should keep a way to do the basic processes (sanitizing/validating/updating) alone. We may reconsider when adding support WP-CLI for example, but it seems too soon to do this now.

modules/REST/V1/Settings.php
Hug0-Drelon: I primarily does nothing regarding the previous value and perform the update anyway.
Your commentaries confirm what I think, it complicates things.
Also, WP does the insertion anyway in `WP_REST_Settings_Controller::update_item()` so we may want to do the same here.

Regarding the client, if it send update request with the a value already set I think that performance issues is on its side.

If we really want to go that way, then we need proper error management to tell the client no update has been performed but not blocking error is raised either.
Otherwise it could do the request over and over without any changes.

modules/REST/V1/Settings.php
Hug0-Drelon: I like it!

PR #1598, opened 2024-12-16
Display a notice about emptied strings on 3.7 upgrade.

Hug0-Drelon:
## What?
Polylang 3.7 will empty untranslated strings from database from now on. Let's add a notice as stated in https://github.com/polylang/polylang-pro/issues/2308, so user won't be lost.
Follows https://github.com/polylang/polylang/pull/1574

## How?
Use `PLL_Admin_Notices::add_notice()` on upgrade.


manooweb (approved):

admin/admin-notices.php
```diff
@@ -185,6 +185,14 @@ public function display_notices() {
 				$this->review_notice();
 			}
 
+			if (
+				isset( $GLOBALS['pagenow'] ) && 'admin.php' === $GLOBALS['pagenow'] &&
+				isset( $_GET['page'] ) && 'mlang_strings' === $_GET['page'] && // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				! static::is_dismissed( 'empty-strings-translations' )
+			) {
```
Chouby, line 192: I'm not fan. What do you think of:
```suggestion
			$allowed_screen = sanitize_title( __( 'Languages', 'polylang' ) ) . '_page_mlang_strings';
			if ( $this->can_display_notice( 'empty-strings-translations', (array) $allowed_screen ) && ! static::is_dismissed( 'empty-strings-translations' ) ) {
```
with the following change to `can_display_notice()`:
```php
	protected function can_display_notice( $notice, $allowed_screens = array() ) {
		$screen = get_current_screen();
		if ( empty( $screen ) ) {
			return false;
		}
		
		if ( empty( $allowed_screens ) ) {
			$screen_id       = sanitize_title( __( 'Languages', 'polylang' ) );
			$allowed_screens = array(
				'dashboard',
				'plugins',
				'toplevel_page_mlang',
				$screen_id . '_page_mlang_strings',
				$screen_id . '_page_mlang_settings',
			);
		}

		/**
		 * Filters admin notices which can be displayed.
		 *
		 * @since 2.7.0
		 *
		 * @param bool   $display Whether the notice should be displayed or not.
		 * @param string $notice  The notice name.
		 */
		return apply_filters( 'pll_can_display_notice', in_array( $screen->id, $allowed_screens, true ), $notice );
	}
```

Chouby (changes requested):

admin/admin-notices.php
Hug0-Drelon: Nice, done in 4ddfd36c.

Hug0-Drelon (commented):
> [!NOTE]
> Do not display the notice for install older than 3.7.

PR #1600, opened 2024-12-17
`nav_menus` and `domains` options: also update `0` IDs

Screenfeed:
Consider the `nav_menus` option containing this:

```php
array(
    'twentytwentyone' => array(
      'primary' => array(
        'fr' => 2738,
        'en' => 4317,
      ),
      'footer' => array(
        'fr' => 0,
        'en' => 0,
      ),
    ),
)
```

After modifying the `en` lang slug to `eng`:

```php
array(
    'twentytwentyone' => array(
      'primary' => array(
        'fr' => 2738,
        'eng' => 4317,
      ),
      'footer' => array(
        'fr' => 0,
        'en' => 0,
      ),
    ),
)
```

`['twentytwentyone']['footer']['en']` is not updated because it is `empty()`.

While not a big issue, this PR fixes this problem.

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -308,7 +308,7 @@ class Languages {
 			if ( ! empty( $nav_menus ) ) {
 				foreach ( $nav_menus as $theme => $locations ) {
 					foreach ( array_keys( $locations ) as $location ) {
-						if ( ! empty( $nav_menus[ $theme ][ $location ][ $old_slug ] ) ) {
+						if ( isset( $nav_menus[ $theme ][ $location ][ $old_slug ] ) ) {
 							$nav_menus[ $theme ][ $location ][ $slug ] = $nav_menus[ $theme ][ $location ][ $old_slug ];
 							unset( $nav_menus[ $theme ][ $location ][ $old_slug ] );
 						}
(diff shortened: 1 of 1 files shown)


include/Model/Languages.php
```diff
@@ -321,7 +321,7 @@ public function update( $args ) {
 			// Update domains.
 			$domains = $this->options->get( 'domains' );
 
-			if ( ! empty( $domains[ $old_slug ] ) ) {
+			if ( isset( $domains[ $old_slug ] ) ) {
```
Hug0-Drelon, line 324: Should we use `is_numeric()` or `is_int()`?

Hug0-Drelon (changes requested):

include/Model/Languages.php
Screenfeed: No need IMHO.
1. If the options have not been saved yet since the creation of this language, `$domains[ $old_slug ]` is not set, which means we must keep `isset()`.
2. In the other cases, the value of `$domains[ $old_slug ]` comes from options, which are already validated and sanitized, so we know it is an integer: no need of an extra `is_int()` then.

Hug0-Drelon: I see, sorry I messed up with menu.
I have to say I don't understand the link between what you're fixing and menu issue you're showing in the PR description.

Screenfeed: Probably because I'm the one who messed up, not you.

Hug0-Drelon: 🤣

Hug0-Drelon (approved):

PR #1601, opened 2024-12-20
Use determine_locale() instead of get_locale() to get HTML language attributes correctly on the log in page

manooweb:
fixes https://github.com/polylang/polylang-pro/issues/2420

## What?

It's seems that `determine_locale()`  WordPress function already manage `wp_lang` dropdown & cookie for the log in page.
See 
https://github.com/WordPress/WordPress/blob/6.8.2/wp-login.php#L393
and
https://github.com/WordPress/WordPress/blob/6.8.2/wp-login.php#L538

In `determine_locale()` WordPress handles specifically the login page See https://github.com/WordPress/WordPress/blob/6.8.2/wp-includes/l10n.php#L140-L148 and, in this case, prevents from passing to `get_locale()` function on which Polylang is hooked https://github.com/WordPress/WordPress/blob/6.8.2/wp-includes/l10n.php#L165-L167

> [!NOTE]
> On login page `WP::main()` method isn't called. So Polylang language cookie is never set when the page is called directly without having visit a frontend page before because we are hooked on `wp` action.
> See  https://github.com/polylang/polylang/blob/3.7.3/frontend/choose-lang.php#L69

## How?
- ~Adds a condition not to filter language attributes in this case.~
- Uses `determine_locale()` instead of `get_locale()` to correctly set the HTML language attribute to the W3C format if the language exists in Polylang. If the language doesn't exist in Polylang then it keeps the WordPress value which comes from translations See https://github.com/WordPress/WordPress/blob/6.8.2/wp-includes/general-template.php#L890-L894

> [!NOTE]
> Note that here if the value isn't translated, it is also `determine_locale()` which is used

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -74,7 +74,7 @@ class PLL_Filters {
 		add_filter( 'get_previous_post_where', array( $this, 'posts_where' ), 10, 5 );
 		add_filter( 'get_next_post_where', array( $this, 'posts_where' ), 10, 5 );
 
-		// Converts the locale to a valid W3C locale except on log in page.
+		// Converts the locale to a valid W3C locale.
 		add_filter( 'language_attributes', array( $this, 'language_attributes' ) );
 
 		// Translate the site title in emails sent to users
@@ -332,13 +332,7 @@ class PLL_Filters {
 	 * @return string
 	 */
 	public function language_attributes( $output ) {
-		if (
-			isset( $GLOBALS['pagenow'] ) && 'wp-login.php' === $GLOBALS['pagenow']
-			&& ( ! empty( $_GET['wp_lang'] ) || ! empty( $_COOKIE['wp_lang'] ) )  // phpcs:ignore WordPress.Security.NonceVerification.Recommended
-		) {
-			return $output;
-		}
-		if ( $language = $this->model->get_language( is_admin() ? get_user_locale() : get_locale() ) ) {
+		if ( $language = $this->model->get_language( is_admin() ? get_user_locale() : determine_locale() ) ) {
 			$output = str_replace( '"' . get_bloginfo( 'language' ) . '"', '"' . $language->get_locale( 'display' ) . '"', $output );
 		}
 		return $output;
--- a/include/filters.php
+++ b/include/filters.php
@@ -74,7 +74,7 @@ class PLL_Filters {
 		add_filter( 'get_previous_post_where', array( $this, 'posts_where' ), 10, 5 );
 		add_filter( 'get_next_post_where', array( $this, 'posts_where' ), 10, 5 );
 
-		// Converts the locale to a valid W3C locale
+		// Converts the locale to a valid W3C locale except on log in page.
 		add_filter( 'language_attributes', array( $this, 'language_attributes' ) );
 
 		// Translate the site title in emails sent to users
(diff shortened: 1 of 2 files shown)


include/filters.php
```diff
@@ -330,6 +330,12 @@ public function posts_where( $sql, $in_same_term, $excluded_terms, $taxonomy, $p
 	 * @return string
 	 */
 	public function language_attributes( $output ) {
+		if (
+			isset( $GLOBALS['pagenow'] ) && 'wp-login.php' === $GLOBALS['pagenow']
+			&& ( ! empty( $_GET['wp_lang'] ) || ! empty( $_COOKIE['wp_lang'] ) )  // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		) {
+			return $output;
```
Marianne380, line 337: Shouldn't we convert the WordPress locale to W3C locale in this case too?

Marianne380 (changes requested):

include/filters.php
manooweb: By using `determine_locale()` instead of `get_locale()`, the condition is useless and thus we ensure W3C conversion if the language exists in Polylang. Otherwise the WordPress value is kept.
See the latest commit: 42c25efb3bbdf3dba36cfc40be23c5d5060d01ef

> [!NOTE]
> A language can be displayed in the list even if it doesn't exist in Polylang. The list is based on the translation files existence.

include/filters.php
```diff
@@ -332,7 +332,7 @@ public function posts_where( $sql, $in_same_term, $excluded_terms, $taxonomy, $p
 	 * @return string
 	 */
 	public function language_attributes( $output ) {
-		if ( $language = $this->model->get_language( is_admin() ? get_user_locale() : get_locale() ) ) {
+		if ( $language = $this->model->get_language( is_admin() ? get_user_locale() : determine_locale() ) ) {
```
Marianne380: Can we take the opportunity to remove the assignment in the `if` statement?

Marianne380 (changes requested):

include/filters.php
manooweb: Sure! Done in 8d59f0ec543ad7511f822d3bee17566f8e09fbc1

Marianne380 (approved):

include/filters.php
```diff
@@ -332,10 +332,13 @@ public function posts_where( $sql, $in_same_term, $excluded_terms, $taxonomy, $p
 	 * @return string
 	 */
 	public function language_attributes( $output ) {
-		if ( $language = $this->model->get_language( is_admin() ? get_user_locale() : get_locale() ) ) {
-			$output = str_replace( '"' . get_bloginfo( 'language' ) . '"', '"' . $language->get_locale( 'display' ) . '"', $output );
+		$language = $this->model->get_language( is_admin() ? get_user_locale() : determine_locale() );
```
Chouby: IMHO `determine_locale()` should be sufficient. See https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/l10n.php#L153
```suggestion
		$language = $this->model->get_language( determine_locale() );
```

Could you add a test?

manooweb: Done in 9f54b15b084a9fc9df32b0eaa51170c63b2683ec & bc466d9d810afcff87c102f817f4c6f000b363a1

PR #1608, opened 2025-01-07
Loads strings translations with the media's target language on media creation.

Marianne380:
Linked to https://github.com/polylang/polylang-pro/pull/2239.

Strings translations must also be loaded in the target language for media, as done for posts in https://github.com/polylang/polylang-pro/pull/2239.

### DIFF
--- a/include/translated-post.php
+++ b/include/translated-post.php
@@ -371,6 +371,9 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 		$translations[ $lang->slug ] = $tr_id;
 		$this->save_translations( $tr_id, $translations );
 
+		// Loads the strings translations with the attachment's target language.
+		PLL()->load_strings_translations( $lang );
+
 		/**
 		 * Fires after a media translation is created
 		 *
--- a/include/translated-post.php
+++ b/include/translated-post.php
@@ -346,6 +346,10 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 			$post['post_parent'] = (int) $this->get_translation( $post['post_parent'], $lang->slug );
 		}
 		$post['tax_input'] = array( 'language' => array( $lang->slug ) ); // Assigns the language.
+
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -20,6 +20,8 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 		self::create_language( 'es_ES' );
 
 		self::$editor = $factory->user->create( array( 'role' => 'editor' ) );
+
+		require_once POLYLANG_DIR . '/include/api.php';
```
manooweb, line 24: Strange on my side only `$GLOBALS['polylang']` assignment is needed

manooweb (changes requested):

manooweb (approved):

include/translated-post.php
```diff
@@ -371,6 +371,9 @@ public function create_media_translation( $post_id, $lang ) {
 		$translations[ $lang->slug ] = $tr_id;
 		$this->save_translations( $tr_id, $translations );
 
+		// Loads the strings translations with the attachment's target language.
+		PLL()->load_strings_translations( $lang->slug );
+
```
Chouby, line 376: I would move this sooner before we start inserting anything in DB.
And reset to the previous translations before returning.

Chouby (changes requested):

include/translated-post.php
Marianne380: See https://github.com/polylang/polylang/pull/1608/commits/cabf48cf435191af0ae2b876c076ffae3830bee2.

tests/phpunit/tests/test-admin-filters-post.php
Marianne380: Indeed because the API is already loaded here https://github.com/polylang/polylang/blob/3.6.5/tests/phpunit/includes/testcase-assets.php#L2.

PR #1610, opened 2025-01-09
Make sure language slugs in the `nav_menus` option correspond to existing languages

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2413.

This PR validates the language slugs in the `nav_menus` option by making sure they correspond to existing languages.

Some code has been factorized in `Abstract_Option`, to be used in `Domains` and `Nav_Menus`.

Also, a bug has been fixed in `Domains::sanitize()`: the property `Abstract_Option::$value` is `private` and cannot be used in child-classes.

`Nav_Menus::sanitize()` keeps the previous sub-values if they are not provided by the new ones. This means the whole value must be sent to it to prevent unintantional removals.

### DIFF
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -6,6 +6,7 @@
 namespace WP_Syntex\Polylang\Options;
 
 use WP_Error;
+use WP_Term;
 use WP_Syntex\Polylang\Options\Options;
 
 defined( 'ABSPATH' ) || exit;
@@ -225,4 +226,58 @@ abstract class Abstract_Option {
 	 * @return string
 	 */
 	abstract protected function get_description(): string;
+
+	/**
+	 * Returns a list of language terms.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $fields Optional. Term fields to query for (see `WP_Term_Query` for possible values). Default is `'all'`.
+	 * @return array
+	 *
+	 * @phpstan-param 'all'|'all_with_object_id'|'names'|'slugs'|'id=>name'|'id=>slug'|'ids'|'tt_ids' $fields
+	 * @phpstan-return (
+	 *     $fields is 'all'|'all_with_object_id' ? list<WP_Term> : (
+	 *         $fields is 'names'|'slugs'|'id=>name'|'id=>slug' ? array<int, string> : array<int, int>
+	 *     )
+	 * )
+	 */
+	protected function get_language_terms( string $fields = 'all' ): array {
+		$language_terms = get_terms(
+			array(
+				'taxonomy'   => 'language',
+				'hide_empty' => false,
+				'fields'     => $fields,
+			)
+		);
+		return is_array( $language_terms ) ? $language_terms : array();
+	}
+
+	/**
+	 * Adds a non-blocking error warning about unknown language slugs.
+	 *
+	 * @since 3.7
+	 *
+	 * @param array $language_slugs List of language slugs.
+	 * @return void
+	 */
+	protected function add_unknown_languages_warning( array $language_slugs ): void {
+		$this->errors->add(
+			sprintf( 'pll_unknown_%s_languages', static::key() ),
+			sprintf(
+				/* translators: %s is a list of language slugs. */
+				_n( 'The language %s is unknown and has been discarded.', 'The languages %s are unknown and have been discarded.', count( $language_slugs ), 'polylang' ),
+				wp_sprintf_l(
+					'%l',
+					array_map(
+						function ( $slug ) {
+							return "<code>{$slug}</code>";
+						},
+						$language_slugs
+					)
+				)
+			),
+			'warning'
+		);
+	}
 }
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -228,6 +228,31 @@ abstract class Abstract_Option {
 	abstract protected function get_description(): string;
 
 	/**
+	 * Tells if the languages list is ready.
+	 * If not ready, a "doing it wrong" notice is fired.
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $method The name of the method using this. Used in the error message.
+	 * @return bool
+	 */
+	protected function are_languages_ready( string $method ): bool {
+		global $polylang;
+
+		if ( empty( $polylang ) || ! $polylang->model->languages->are_ready() ) {
+			// Access to global `$polylang` is required.
+			_doing_it_wrong(
+				esc_html( $method ),
+				esc_html( sprintf( 'The option \'%s\' cannot be set before the hook \'pll_init\'.', static::key() ) ),
+				'3.7'
+			);
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	 * Returns a list of language terms.
 	 *
 	 * @since 3.7
--- a/include/Options/Abstract_Option.php
+++ b/include/Options/Abstract_Option.php
@@ -228,31 +228,6 @@ abstract class Abstract_Option {
 	abstract protected function get_description(): string;
 
 	/**
-	 * Tells if the languages list is ready.
-	 * If not ready, a "doing it wrong" notice is fired.
-	 *
-	 * @since 3.7
-	 *
-	 * @param string $method The name of the method using this. Used in the error message.
-	 * @return bool
-	 */
-	protected function are_languages_ready( string $method ): bool {
-		global $polylang;
-
-		if ( empty( $polylang ) || ! $polylang->model->languages->are_ready() ) {
-			// Access to global `$polylang` is required.
-			_doing_it_wrong(
-				esc_html( $method ),
-				esc_html( sprintf( 'The option \'%s\' cannot be set before the hook \'pll_init\'.', static::key() ) ),
-				'3.7'
-			);
-			return false;
-		}
-
-		return true;
-	}
-
-	/**
 	 * Returns a list of language terms.
 	 *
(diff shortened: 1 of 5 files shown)


include/Options/Abstract_Option.php
```diff
@@ -225,4 +226,58 @@ abstract protected function get_data_structure(): array;
...
+	 * @return array
+	 *
+	 * @phpstan-param 'all'|'all_with_object_id'|'names'|'slugs'|'id=>name'|'id=>slug'|'ids'|'tt_ids' $fields
+	 * @phpstan-return (
+	 *     $fields is 'all'|'all_with_object_id' ? list<WP_Term> : (
+	 *         $fields is 'names'|'slugs'|'id=>name'|'id=>slug' ? array<int, string> : array<int, int>
+	 *     )
+	 * )
+	 */
+	protected function get_language_terms( string $fields = 'all' ): array {
```
Hug0-Drelon, line 245: The `$fields` parameter is overkill IMHO. It's only used once and can be filtered by the calling code easily.
In other words, it does things we don't need, adds complexity and maintenance ( `$fields` doc must be aligned with `get_terms()`).

include/Options/Business/Nav_Menus.php
```diff
@@ -15,6 +17,14 @@
  * Class defining navigation menus array option.
  *
  * @since 3.7
+ *
+ * @phpstan-type Value array<
```
Hug0-Drelon, line 21:
```suggestion
 * @phpstan-type NavMenusValue array<
```
I found it to be too generic.
Also, should you do that for Domains as well? (I saw you had to add inline PHPStan doc as well).

tests/phpunit/tests/Options/Options/test-Set.php
Hug0-Drelon, line 1: I must say I'm not sure to understand the "why" in `test_should_not_set_bad_nav_menus_language()` 😅 
Comments tell what we expect, not why some are discarded, some are kept...

Hug0-Drelon (changes requested):

include/Options/Abstract_Option.php
Screenfeed: I disagree. If it is used (at least once), that means there is at least another case where it isn't used (in `Domains`), so there are 2 different results.
If we remove it we will add complexity instead of removing it. In `Nav_Menus` we do:

```php
$language_terms = $this->get_language_terms( 'slugs' );
// ...
$unknown_langs = array_diff_key( $unknown_langs, array_flip( $language_terms ) );
```

Without this parameter we will have to create an intermediate array containing the language slugs, which is exactly what this parameter does.

include/Options/Business/Nav_Menus.php
Screenfeed: At the class level I don't think it's too generic, it's the type of the option the class handles. Other classes using it can alias it with `@phpstan-type Value from Nav_Menus as NavMenusValue`. But OK, if someday we need it outside this class, using a more specific name will prevent us to use the alias notation.

Screenfeed: Done in [d89ae43](https://github.com/polylang/polylang/pull/1610/commits/d89ae43bee89c9828ab8cd0f3084778b8285029f).

tests/phpunit/tests/Options/Options/test-Set.php
Screenfeed: It's basically a recursive `array_filter( array_merge() )`.
I improved few comments in [35f4bbc](https://github.com/polylang/polylang/pull/1610/commits/35f4bbc0e0174c9727cd45fe1f4ba8d553a10207).

include/Options/Abstract_Option.php
Hug0-Drelon: As you said, it's use only once so the complexity is added into a new method (used once).
Also, you could simplify using `wp_list_pluck()` for instance.

tests/phpunit/tests/Options/Options/test-Set.php
Hug0-Drelon: I wasn't talking about the `set()` behavior, I was talking about why some values are kept, other are replaced and so on.
But nevermind, it'll do the job like that.

include/Options/Business/Nav_Menus.php
Screenfeed: Done in [0f3c69b](https://github.com/polylang/polylang/pull/1610/commits/0f3c69b36f857e56af03748af9e52e288afcf0b5) for `Domains`. I didn't do it at first because it isn't a complex type, and it isn't used outside the class.

include/Options/Abstract_Option.php
Screenfeed: Done in [d995085](https://github.com/polylang/polylang/pull/1610/commits/d995085afda0b8a5df6f363f4412429b7d83df73).

include/Options/Business/Nav_Menus.php
```diff
@@ -73,6 +83,79 @@ protected function get_data_structure(): array {
...
+
+		/** @phpstan-var NavMenusValue $value */
+		if ( empty( $value ) ) {
+			// Nothing to validate.
+			return $value;
+		}
+
+		$unknown_langs  = array();
+		$language_terms = $this->get_language_terms();
+		$language_terms = \wp_list_pluck( $language_terms, 'slug' );
```
Hug0-Drelon, line 129:
```suggestion
		$language_terms = \wp_list_pluck( $this->get_language_terms(), 'slug' );
```

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

include/Options/Business/Nav_Menus.php
```diff
@@ -73,6 +83,78 @@ protected function get_data_structure(): array {
...
+		if ( empty( $polylang ) || ! $polylang->model->are_languages_ready() ) {
+			// Access to global `$polylang` is required.
+			_doing_it_wrong(
+				__METHOD__,
+				esc_html( sprintf( 'The option \'%s\' cannot be set before the hook \'pll_init\'.', static::key() ) ),
+				'3.7'
+			);
+			/** @phpstan-var NavMenusValue */
+			return $this->get();
+		}
```
Chouby, line 111: I don't see where `$polylang` is used outside this block. To avoid using a global, we could change `Languages::set_ready()` to fire an action `pll_languages_ready` instead of setting a boolean flag. The check could then be made with `did_action()`. Also this check could be moved to `Abstract_Option` to avoid duplicate code.

include/Options/Business/Nav_Menus.php
```diff
@@ -73,6 +83,78 @@ protected function get_data_structure(): array {
...
+				$value[ $theme_slug ][ $location ] = array();
+
+				foreach ( $language_terms as $lang_slug ) {
+					if ( ! empty( $menu_ids[ $lang_slug ] ) ) {
+						$value[ $theme_slug ][ $location ][ $lang_slug ] = $menu_ids[ $lang_slug ];
+					}
+				}
+
+				// Detect unknown languages.
+				$unknown_langs = array_merge( $unknown_langs, $menu_ids );
```
Chouby, line 142: A this point, the name of the variable `$unknown_langs` is confusing (yes I struggled to understand). We really know they are unknown only the line after.

Chouby (changes requested):

include/Options/Business/Nav_Menus.php
Screenfeed: You're right. Improved in [6114d37](https://github.com/polylang/polylang/pull/1610/commits/6114d37fc5993ecc0d8eab1de18502eb7ab27b57).

include/Options/Business/Nav_Menus.php
Screenfeed: Check factorized in [fe2f891](https://github.com/polylang/polylang/pull/1610/commits/fe2f891cf27fda8d354237f4d221493e3259a9e6).

Screenfeed: I removed these checks about languages being ready in 85ecc97.
They are not needed anymore because we retrieve the languages list directly from the language terms now.

Note: I tried your suggestion and it breaks some tests, probably because we use more than one model (so the 1st model "readiness" breaks the 2nd model's).

include/Options/Business/Nav_Menus.php
```diff
@@ -73,6 +83,65 @@ protected function get_data_structure(): array {
...
+		}
+
+		/** @phpstan-var NavMenusValue $value */
+		$unknown_langs = array_diff_key( $all_langs, array_flip( $language_terms ) );
+
+		// Detect invalid language slugs.
+		if ( ! empty( $unknown_langs ) ) {
+			// Non-blocking error.
+			$this->add_unknown_languages_warning( array_keys( $unknown_langs ) );
+		}
```
Chouby, line 140: Instead of using directly `add_unknown_languages_warning()`, I suggest to create a new method in `Abstract_Option` which would remove unknown languages from an array and add the warning when some is encountered.
`function remove_unknown_languages( array $languages ): array`
I expect this to factorize more code with `Domains` as the query to get all languages terms would be done in this new method.

What do you think?

Chouby (changes requested):

include/Options/Business/Nav_Menus.php
Screenfeed: As discussed, the `foreach` loops in `Domains` and `Nav_Menus` are too different:
- In `Domains` we fill new values, fallback to previous values, warn for missing languages.
- In `Nav_Menus` we only fill non-empty values.

PR #1612, opened 2025-01-10
Handle breaking changes in jQuery 4's AJAX prefilter

### LINKED ISSUE #1527: trim() is deprecated in the forthcoming jQuery 4
I tried your plugin with the [Test jQuery Updates](https://wordpress.org/plugins/wp-jquery-update-test/) plugin which ships with jQuery 4.0 beta, and Polylang throws a `trim() is not a function` error here:

https://github.com/polylang/polylang/blob/bcc11ff89075b8d483aa17f44458714f8999f890/admin/admin-base.php#L335C74-L335C78

### LINKED ISSUE #1603: Javascript exception in `addPolylangParametersAsString` for options.data=null
Hi everybody,

I got a problem using Polylang for admin-ajax requests on a website : 

> Uncaught TypeError: Cannot read properties of null (reading 'trim') at addPolylangParametersAsString

The reason is clear. In my case, options.data = null, and your code (admin-base.php:325) is :

```
function addPolylangParametersAsString() {
     if ( 'undefined' === typeof options.data || '' === options.data.trim() ) { /* -> 💥*/
          ...
     }
}
```
Maybe you could fix that using : 
```
function addPolylangParametersAsString() {
     if ( !options.data || '' === options.data.trim() ) { /* -> 👍*/
          ...
     }
}
```

Thanks by advance !

Casimir

Screenfeed:
Fixes #1527. Probably also fixes #1603.

jQuery 4 introduces [a breaking change](https://github.com/jquery/jquery/pull/5197) in the AJAX prefilter.

@manooweb explained it well [in this comment](https://github.com/polylang/polylang/issues/1527#issuecomment-2331756557). The TL;DR being: in jQ 3 the data is stringified before the prefilter, while in jQ 4 it is done after. Then we must expect several types of data in our prefilter.

The solution suggested here is to deal with the given data type without changing it (or changing it to a similar type):
- `undefined`, `null`, `String` => `string`.
- A plain `Object` => `Object`.

## Code moved

The current file `admin.js` is enqueued only in our settings pages and the wizard, so it has been renamed into `settings.js`.
The new JS code has been moved to a dedicated JS file `ajax-filter.js`, then imported in a new file `admin.js`, which is enqueued in all admin screens. In the future we expect to re-use `ajax-filter.js` in PLL Pro by creating a new JS package containing all the common code (for now we will duplicate the file).

## Code linted

PHPCS suddenly started linting the new file `settings.js` while it wasn't linting [the old file `admin.js`](https://github.com/polylang/polylang/blob/a4b64f5b8773d2f1a8d72f2f6ad476688ee73e36/js/src/admin.js#L1-L425).
1. `phpcbf` has been used to add some missing space characters.
2. Each insertion of HTML code has been verified, to make sure the inserted code is safe. Then those errors have been ignored and comments have been added to explain why.

### php callback using data in `$_POST`

```php
print_r( $_POST );
die();
```

| `$.ajax()` config | Data received by the php callback | Supported | Broken |
| ----------------- | --------------------------------- | :-------: | :----: |
|`{ data: [7, "a"] }`|`Array( [undefined] => )`|❌|No|
|`{}`|`Array( [pll] => 1 )`|✅|No|
|`{ data: null }`|`Array( [pll] => 1 )`|✅|No|
|`{ data: "" }`|`Array( [pll] => 1 )`|✅|No|
|`{ data: "A string" }`|`Array( [A_string] => null, [pll] => 1 )`|✅|No|
|`{ data: "foo=2" }`|`Array( [foo] => 2, [pll] => 1 )`|✅|No|
|`{ data: {foo: "bar"} }`|`Array( [foo] => bar, [pll] => 1 )`|✅|No|

> [!NOTE]
> The first case doesn't work because the first argument passed to `Object.assign()` is an `Array`.

### php callback using data in request's body

```php
print_r( file_get_contents( 'php://input' ) );
die();
```

| `$.ajax()` config | Data received by the php callback | Supported | Broken |
| ----------------- | --------------------------------- | :-------: | :----: |
|`{ data: "A string", contentType: "text/plain; charset=UTF-8" }`|`A string&pll=1`|❌|⚠️ Yes|
|`{ data: document.implementation.createDocument( null, "books" ), contentType: "application/xml; charset=UTF-8", processData: false } )`|`<books/>`|❌|No|

> [!CAUTION]  
> The first case is problematic (however I don't think this is a regression).

> [!NOTE]
> The second case doesn't work because the data is not a plain object.

### php callback using JSON data in request's body

```php
print_r( json_decode( file_get_contents( 'php://input' ), true ) );
die();
```

| `$.ajax()` config | Data received by the php callback | Supported | Broken |
| ----------------- | --------------------------------- | :-------: | :----: |
|`{ data: JSON.stringify( [7, "a"] ), contentType: "application/json; charset=UTF-8" } )`|`Array( [0] => 7, [1] => a )`|❌|No|
|`{ data: JSON.stringify( {foo: 2} ), contentType: "application/json; charset=UTF-8" }`|`Array( [foo] => 2, [pll] => 1 )`|✅|No|

> [!NOTE]
> The first case doesn't work because the first argument passed to `Object.assign()` is an `Array`.

> [!WARNING]
> A similar work must be done in [Polylang Pro](https://github.com/polylang/polylang-pro/blob/d7bed34992c13bb49540258e14b9b9851fb777b1/js/src/editors/site/ajax-filter.js#L26).

## Testing
We can use [Test jQuery Updates](https://wordpress.org/plugins/wp-jquery-update-test/) plugin to replace current jQuery version (3.7.1) with the current jQuery beta version (4.0.0-beta.2). Deactivate the plugin to return back to 3.7.1 version.

To test with third party plugin refer to older PRs:
- [Hubspot](https://wordpress.org/plugins/leadin/) referenced in #803 & #824. It seems that Huspot no longer use ajax but REST API endpoint instead now to handle account creation and connection.
- WooCommerce Tree Table Rate Shipping and WooCommerce License Manager referenced in https://github.com/polylang/polylang/pull/824#issuecomment-825450659
- [Ninja Forms](https://wordpress.org/plugins/ninja-forms/) referenced in https://github.com/polylang/polylang-pro/issues/821

> [!NOTE]
> There is also ajax requests in ACF. For example by using our `related post` block from our PLL ACF Blocks plugin. The ajax request is sent when selecting the block to fill ACF fields.

> [!CAUTION]  
> The new code should still working with jQuery 3.7.1


manooweb (approved):

manooweb (approved):

admin/admin-base.php
```diff
@@ -172,13 +171,16 @@ public function languages_page() {}
 	 * @return void
 	 */
 	public function admin_enqueue_scripts() {
+		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+
+		wp_enqueue_script( 'pll_admin', plugins_url( "/js/build/admin{$suffix}.js", POLYLANG_ROOT_FILE ), array( 'jquery' ), POLYLANG_VERSION, true );
+		wp_localize_script( 'pll_admin', 'pll_admin', array( 'ajax_filter' => $this->get_ajax_filter_data() ) );
```
Hug0-Drelon, line 177: `wp_add_inline_script()` is the way to go regarding such data transfer.
See what we've done in Polylang Pro [here](https://github.com/polylang/polylang-pro/blob/81bbe2aecf54cc1e531f9e3bb5f958a94647f571/modules/Editors/Screens/Abstract_Screen.php#L105-L121).

js/src/admin.js
```diff
@@ -2,425 +2,6 @@
...
-		$( '.if-js-closed' ).removeClass( 'if-js-closed' ).addClass( 'closed' );
-		// postboxes setup
-		if ( 'undefined' !== typeof postboxes ) {
-			postboxes.add_postbox_toggles( pagenow );
-		}
-	}
-);
+import ajaxFilter from './lib/ajax-filter.js';
 
+ajaxFilter( pll_admin.ajax_filter );
```
Hug0-Drelon, line 7:
```suggestion
ajaxFilter( pll_admin?.ajax_filter );
```
If someone mess around our code in PHP.

js/src/lib/ajax-filter.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Adds data to all ajax requests made with jQuery.
+ *
+ * @since 3.7
+ *
+ * @param {Object} data The data to add.
+ * @returns {void}
+ */
+const ajaxFilter = ( data ) => {
+	if ( 'undefined' === typeof jQuery ) {
```
Hug0-Drelon, line 14:
```suggestion
	if ( 'undefined' === typeof jQuery || ! data ) {
```
Follows my previous comment.

js/src/admin.js
```diff
@@ -2,425 +2,6 @@
...
-		// Manage closing the metabox.
-		// close postboxes that should be closed
-		$( '.if-js-closed' ).removeClass( 'if-js-closed' ).addClass( 'closed' );
-		// postboxes setup
-		if ( 'undefined' !== typeof postboxes ) {
-			postboxes.add_postbox_toggles( pagenow );
-		}
-	}
-);
+import ajaxFilter from './lib/ajax-filter.js';
```
Hug0-Drelon, line 5:
```suggestion
import ajaxFilter from './lib/ajax-filter';
```
Along that, please move `./lib/ajax-filter.js` into `./lib/ajax-filter/index.js`.
This is some sort of convention in JS, it will allow to refactor or do anything you want to `ajaxFilter` without impacting the imports.

js/src/lib/ajax-filter.js
```diff
@@ -0,0 +1,51 @@
...
+				} else if ( $.isPlainObject( options.data ) ) {
+					// An object.
+					options.data = Object.assign( options.data, data );
+				}
+			} );
+		}
+	);
+}
+
+export default ajaxFilter;
```
Hug0-Drelon, line 51: I tend to avoid default export nowadays, that we can always import using `{}` and don't need to think wether or not it is a default one.
Also, if some you want to add a new function, it doable more easily.
But I let you see what you prefer.

modules/wizard/wizard.php
```diff
@@ -377,17 +377,17 @@ public function add_step_licenses( $steps ) {
...
 		add_action( 'wp_ajax_pll_deactivate_license', array( $this, 'deactivate_license' ) );
 
-		// Be careful pll_admin script is enqueued here without dependency except jquery because only code useful for deactivate license button is needed.
-		// To be really loaded the script need to be passed to the $steps['licenses']['scripts'] array below with the same handle than in wp_enqueue_script().
-		wp_enqueue_script( 'pll_admin', plugins_url( '/js/build/admin' . $this->get_suffix() . '.js', POLYLANG_ROOT_FILE ), array( 'jquery' ), POLYLANG_VERSION, true );
-		wp_localize_script( 'pll_admin', 'pll_admin', array( 'dismiss_notice' => esc_html__( 'Dismiss this notice.', 'polylang' ) ) );
+		// Be careful `settings` script is enqueued here without dependency except jquery because only code useful for deactivate license button is needed.
+		// To be really loaded the script needs to be passed to the `$steps['licenses']['scripts']` array below with the same handle than in `wp_enqueue_script()`.
+		wp_enqueue_script( 'pll_settings', plugins_url( '/js/build/settings' . $this->get_suffix() . '.js', POLYLANG_ROOT_FILE ), array( 'jquery' ), POLYLANG_VERSION, true );
```
Hug0-Drelon, line 382: Did we check our other plugins for this script handle?

Hug0-Drelon (changes requested):

modules/wizard/wizard.php
Screenfeed: Yass.

manooweb:
> **Note:** On `/wp-admin/admin.php?page=mlang_wizard&step=languages`, I don’t see in the console that I'm using WordPress jQuery: 4.0.0-beta.2.

It isn't the case on my side:

image

image

admin/admin-base.php
Screenfeed: Done in [bdedea8](https://github.com/polylang/polylang/pull/1612/commits/bdedea8143bcad10147378afb852967a320ebe00).
I didn't add a global filter because there is already a filter in `get_ajax_filter_data()`. I don't think a filter for the whole object would be useful.

js/src/admin.js
Screenfeed: Done in [587484f](https://github.com/polylang/polylang/pull/1612/commits/587484f4d8d9a8638d92cb00b86246ce36075b6a).

js/src/admin.js
Screenfeed: Done in [d40f818](https://github.com/polylang/polylang/pull/1612/commits/d40f81858134431a4598d4f5db10ec5e69812487).

js/src/lib/ajax-filter.js
Screenfeed: Makes sense.
Done in [105b1a8](https://github.com/polylang/polylang/pull/1612/commits/105b1a80bbeb122b7588a4e2031f8b022ef9dea2).

admin/admin-base.php
Hug0-Drelon: Agree, no need to filter the value here.

js/src/lib/ajax-filter.js
Hug0-Drelon: Well, I thought it was a good idea to check for `length` prototype property as well.
But it seems to exists only for strings or arrays, so it'd fail for object.
Do you confirm we don't want that? Object must pass?

Screenfeed: Oh, right, I brainfarted. I thought it was available for objects and not for arrays for a moment >_<
Fixed in [cc6b1ee](https://github.com/polylang/polylang/pull/1612/commits/cc6b1eeedc8140db3a7a5443d79c7aa59787f1b1).

Hug0-Drelon: 🤣 
And also, if somedays prototype property are not found, we're in deep shit regarding Javascript in general 😅

Hug0-Drelon (approved):

js/src/lib/ajax-filter.js
Screenfeed: You know that I don't understand shit about what you just said, right? 😅

PR #1616, opened 2025-01-14
Disable the warning for static page translation when page translations are deactivated

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2296.

If the `page` post type is removed from the list of translatable post types, we shouldn't display the "translate your static front page" warning notice.

It can be done by adding this:
```php
add_filter(
	'pll_get_post_types',
	function ( $post_types ) {
		return array_diff( $post_types, array( 'page' ) );
	}
);
```

This PR prevents this notice (and other related hooks) if the `page` post type is not translated.

### DIFF
--- a/admin/admin-static-pages.php
+++ b/admin/admin-static-pages.php
@@ -26,15 +26,35 @@ class PLL_Admin_Static_Pages extends PLL_Static_Pages {
 
 		$this->links = &$polylang->links;
 
-		// Add post state for translations of the front page and posts page
+		add_action( 'pll_init', array( $this, 'init_page_hooks' ), 1000 ); // Hook late to allow other sources to filter `'pll_get_post_types'` easily.
+	}
+
+	/**
+	 * Launches hooks related to the `page` post type.
+	 *
+	 * @since 3.7
+	 *
+	 * @param PLL_Base $polylang The Polylang object.
+	 * @return void
+	 */
+	public function init_page_hooks( PLL_Base $polylang ): void {
+		$post_types = $polylang->model->post->get_translated_object_types();
+
+		if ( ! in_array( 'page', $post_types, true ) ) {
+			// No need of the following if the pages are not translated.
+			return;
+		}
+
+		// Add post state for translations of the front page and posts page.
 		add_filter( 'display_post_states', array( $this, 'display_post_states' ), 10, 2 );
 
-		// Refreshes the language cache when a static front page or page for for posts has been translated.
+		// Refresh the language cache when a static front page or page for for posts has been translated.
 		add_action( 'pll_save_post', array( $this, 'pll_save_post' ), 10, 3 );
 
-		// Prevents WP resetting the option
+		// Prevent WP resetting the option.
 		add_filter( 'pre_update_option_show_on_front', array( $this, 'update_show_on_front' ), 10, 2 );
 
+		// Add a notice to translate the static front page if it is not translated in all languages.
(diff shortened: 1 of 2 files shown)


admin/admin-static-pages.php
```diff
@@ -25,16 +30,34 @@ public function __construct( &$polylang ) {
...
+
+	/**
+	 * Launches hooks related to the `page` post type.
+	 *
+	 * @since 3.7
+	 *
+	 * @return void
+	 */
+	public function init_page_hooks(): void {
+		if ( ! in_array( 'page', $this->post->get_translated_object_types(), true ) ) {
```
Hug0-Drelon, line 46: Can't we use `Post_Types::is_translated_post_type()` instead? Seems more appropriate to me.

tests/phpunit/tests/test-admin-static-pages.php
```diff
@@ -0,0 +1,116 @@
...
+
+	/**
+	 * @param PLL_UnitTest_Factory $factory
+	 * @return void
+	 */
+	public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
+		parent::pllSetUpBeforeClass( $factory );
+
+		$factory->language->create_many( 2 );
+		self::$model->options['default_lang'] = 'en'; // Otherwise static model isn't aware of the created languages...
```
Hug0-Drelon, line 14: Since this is a whole new test class, I suggest you don't use any static model at all.
Either use `$factory->pll_model` or create your own instance for the tests.

Hug0-Drelon (changes requested):

admin/admin-static-pages.php
Screenfeed: 🤦 
Done in [dbf052c](https://github.com/polylang/polylang/pull/1616/commits/dbf052c2f48737f7320ba080ae68ed8b01d6ea2d).
Note: I could have used `PLL_Translated_Post::is_translated()` but it felt better suited to use `Post_Types` as you suggested.

tests/phpunit/tests/test-admin-static-pages.php
Screenfeed: Gomennasai Sensei 🙇 
Done in [5b32e84](https://github.com/polylang/polylang/pull/1616/commits/5b32e84b0bb397c839553c62d1540e78f5b6b232).

tests/phpunit/tests/test-admin-static-pages.php
```diff
@@ -0,0 +1,116 @@
...
+		if ( ! $translate_page_post_type ) {
+			// No warning notice if we don't translate the pages.
+			$this->assertStringNotContainsString( '/post-new.php?post_type=page', $notices );
+			return;
+		}
+
+		// Make sure there is something (our notice) that displays a link to create a translation.
+		$this->assertStringContainsString( '/post-new.php?post_type=page', $notices );
+
+		$this->assertNotFalse( preg_match( '@href="([^"]+/post-new\.php[^"]+)"@', $notices, $matches ) );
```
Hug0-Drelon, line 89: Can't we test the expected URL directly instead of parsing it?

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-admin-static-pages.php
Screenfeed: You can but this way is safer because it won't break if the query changes (new params or change in params order).

Hug0-Drelon (approved):

PR #1617, opened 2025-01-16
Do not relate an object to a term if the term does not exist.

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/2249.

## What ?

All categories are assigned to a post when the set default category no longer exists.

See the mentioned ticket in the issue.

## Why ?

When a term no longer exists but is still set as a `default_category`, a `wp_set_object_terms` is still made to this non-existent term, so we end up with a term that doesn't exist in `PLL_Crud_Posts::set_object_terms()`. 

So in our `get_terms()` we'll get all existing terms, since `$tt_ids` is empty.

## Fix

Don't proceed further if `$tt_iids` is empty, which means the term(s) don't exist.

### DIFF
--- a/tests/phpunit/tests/test-crud-posts.php
+++ b/tests/phpunit/tests/test-crud-posts.php
@@ -262,7 +262,7 @@ class CRUD_Posts_Test extends PLL_UnitTestCase {
 		// Create a post with this category so that `get_terms()` returns terms (since `hide_empty` is `true` by default).
 		self::factory()->post->create( array( 'post_category' => array( $en_cat ) ) );
 
-		update_option( 'default_category', '999' );
+		update_option( 'default_category', $en_cat + 999 ); // Use this number to make sure the category doesn't exist.
 
 		// Simulate the request that create the post and assigns language.
 		$post = get_default_post_to_edit( 'post', true );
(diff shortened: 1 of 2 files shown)


Hug0-Drelon (approved):

tests/phpunit/tests/test-crud-posts.php
```diff
@@ -245,4 +245,34 @@ public function test_existing_post_updated_with_tag_mixing_slug_and_name() {
...
+		$this->pll_admin->posts = new PLL_CRUD_Posts( $this->pll_admin );
+
+		// Create an english category.
+		$en_cat = self::factory()->term->create( array( 'taxonomy' => 'category', 'name' => 'English category' ) );
+		$this->pll_admin->model->term->set_language( $en_cat, 'en' );
+
+		// Create a post with this category so that `get_terms()` returns terms (since `hide_empty` is `true` by default).
+		self::factory()->post->create( array( 'post_category' => array( $en_cat ) ) );
+
+		update_option( 'default_category', '999' );
```
Chouby, line 265: Some day, someone unlucky will have `$en_cat === 999`. FYI I just got `1010` in my first test (so very close). So you must make sure that this will never happen. Either use a really huge number or something like `$en_cat + 999` that cannot exist yet. Also add a comment to explain why this number.

Chouby (changes requested):

PR #1622, opened 2025-02-04
Use `wp_add_inline_script()` to add our cookie script.

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/2469.

Finally, we decided to use `wp_add_inline_script()` to add our cookie script.

Using this method, the type attribute is managed by WordPress with `wp_get_inline_script_tag()`, see [here](https://github.com/WordPress/wordpress-develop/blob/6.7.2/src/wp-includes/script-loader.php#L2849).

We can also see that we have a filter `wp_inline_script_attributes` that allows attributes to be filtered in `current_theme_supports()`, which meets the customer's request.

### DIFF
--- a/integrations/cache/cache-compat.php
+++ b/integrations/cache/cache-compat.php
@@ -72,16 +72,14 @@ class PLL_Cache_Compat {
 
 		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
 
-		$script = "<script{$type_attr}>\n{$js}\n</script>\n";
-
 		/**
-		 * Filters the content of the script tag that creates the cookie.
+		 * Filters the script tag that creates the cookie.
 		 *
 		 * @since 3.7
 		 *
 		 * @param string $script The Script tag and content.
 		 */
-		echo apply_filters( 'pll_cookie_script', $script ); // phpcs:ignore WordPress.Security.EscapeOutput
+		echo apply_filters( 'pll_cookie_script', "<script{$type_attr}>\n{$js}\n</script>\n" ); // phpcs:ignore WordPress.Security.EscapeOutput
 	}
 
 	/**
--- a/integrations/cache/cache-compat.php
+++ b/integrations/cache/cache-compat.php
@@ -71,8 +71,8 @@ class PLL_Cache_Compat {
 		);
 
 		// Need to register prior to enqueue empty script and add extra code to it.
-		wp_register_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION );
-		wp_enqueue_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION, true );
+		wp_register_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION, true );
+		wp_enqueue_script( 'pll_cookie_script' );
 		wp_add_inline_script( 'pll_cookie_script', $js );
 	}
 
--- a/integrations/cache/cache-compat.php
+++ b/integrations/cache/cache-compat.php
@@ -17,7 +17,7 @@ class PLL_Cache_Compat {
 	 */
 	public function init() {
 		if ( PLL_COOKIE ) {
-			add_action( 'wp_print_footer_scripts', array( $this, 'add_cookie_script' ) );
+			add_action( 'wp_enqueue_scripts', array( $this, 'add_cookie_script' ) );
 		}
 
 		// Since version 3.0.5, WP Rocket does not serve the cached page if our cookie is not set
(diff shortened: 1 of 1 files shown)


integrations/cache/cache-compat.php
```diff
@@ -70,7 +72,16 @@ public function add_cookie_script() {
 
 		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
 
-		echo "<script{$type_attr}>\n{$js}\n</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
+		$script = "<script{$type_attr}>\n{$js}\n</script>\n";
+
+		/**
+		 * Filters the content of the script tag that creates the cookie.
```
Hug0-Drelon, line 78:
```suggestion
		 * Filters the script tag that creates the cookie.
```
Not only the content, any attributes attached to the tag as well.

integrations/cache/cache-compat.php
```diff
@@ -70,7 +72,16 @@ public function add_cookie_script() {
...
+		 *
+		 * @since 3.7
+		 *
+		 * @param string $script The Script tag and content.
+		 */
+		echo apply_filters( 'pll_cookie_script', $script ); // phpcs:ignore WordPress.Security.EscapeOutput
```
Hug0-Drelon, line 84:
```suggestion
		echo apply_filters( 'pll_cookie_script', "<script{$type_attr}>\n{$js}\n</script>\n" ); // phpcs:ignore WordPress.Security.EscapeOutput
```
If you chose to apply my suggestion, please delete previous `$script` assignation.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

Chouby (changes requested):
This class is quite old. Couldn't we use something like `wp_add_inline_script()` to add our script? That way if WordPress provides some way to filter the script then ok. If not, then it probably means that the need was too specific.

integrations/cache/cache-compat.php
```diff
@@ -68,9 +70,9 @@ public function add_cookie_script() {
 			esc_js( $expiration )
 		);
 
-		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
-
-		echo "<script{$type_attr}>\n{$js}\n</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
+		wp_register_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION );
```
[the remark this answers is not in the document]
Marianne380: It even seems mandatory to enqueue a script which has no source later on.

integrations/cache/cache-compat.php
```diff
@@ -70,6 +70,7 @@ public function add_cookie_script() {
 			esc_js( $expiration )
 		);
 
+		// Need to register the script to enqueue and add extra code to it.
```
Hug0-Drelon, line 73:
```suggestion
		// Need to register prior to enqueue empty script and add extra code to it.
```

Hug0-Drelon (approved):

integrations/cache/cache-compat.php
```diff
@@ -68,9 +70,10 @@ public function add_cookie_script() {
 			esc_js( $expiration )
 		);
 
-		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
-
-		echo "<script{$type_attr}>\n{$js}\n</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
+		// Need to register prior to enqueue empty script and add extra code to it.
+		wp_register_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION );
+		wp_enqueue_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION, true );
```
Chouby, line 75: You don't need to repeat what you did in `wp_register_script()`.
```suggestion
		wp_enqueue_script( 'pll_cookie_script' );
```

integrations/cache/cache-compat.php
```diff
@@ -68,9 +70,10 @@ public function add_cookie_script() {
 			esc_js( $expiration )
 		);
 
-		$type_attr = current_theme_supports( 'html5', 'script' ) ? '' : ' type="text/javascript"';
-
-		echo "<script{$type_attr}>\n{$js}\n</script>\n"; // phpcs:ignore WordPress.Security.EscapeOutput
+		// Need to register prior to enqueue empty script and add extra code to it.
+		wp_register_script( 'pll_cookie_script', '', array(), POLYLANG_VERSION );
```
Chouby, line 74: Keep the script in footer.

Chouby (changes requested):

Chrystll:
The change to wp_add_inline_script() did not break the logic.

The cookie is still injected and works on the browser side, when I check in Console > document.cookie
Capture d’écran du 2025-04-04 10-54-25

Test done in private browsing (not logged into WordPress).

PR #1625, opened 2025-02-10
Don't use translated labels in the sync option

manooweb:
Fixes https://github.com/polylang/polylang-pro/issues/2470

## How?
- Cherry pick the initial [commit](https://github.com/polylang/polylang/pull/1550/commits/b1ca2acf6ae20bf1d7031359ad20415c9875edb8) as proposed in https://github.com/polylang/polylang/pull/1550

### DIFF
--- a/include/Options/Business/Sync.php
+++ b/include/Options/Business/Sync.php
@@ -40,14 +40,11 @@ class Sync extends Abstract_List {
 	 * @phpstan-return array{type: 'array', items: array{type: SchemaType, enum: non-empty-list<non-falsy-string>}}
 	 */
 	protected function get_data_structure(): array {
-		add_filter( 'lang_dir_for_domain', '__return_false' ); // Prevents loading the translations too early.
-		$enum = array_keys( \PLL_Settings_Sync::list_metas_to_sync() );
-		remove_filter( 'lang_dir_for_domain', '__return_false' );
 		return array(
 			'type'  => 'array',
 			'items' => array(
 				'type' => $this->get_type(),
-				'enum' => $enum,
+				'enum' => \PLL_Settings_Sync::get_synchronizable_elements(),
 			),
 		);
 	}
--- a/modules/sync/settings-sync.php
+++ b/modules/sync/settings-sync.php
@@ -113,4 +113,29 @@ class PLL_Settings_Sync extends PLL_Settings_Module {
 			'_thumbnail_id'     => __( 'Featured image', 'polylang' ),
 		);
 	}
+
+	/**
+	 * Returns the list of elements that can be synced at the post level.
+	 *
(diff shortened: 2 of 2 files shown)


Screenfeed (approved):

include/Options/Business/Sync.php
```diff
@@ -40,14 +40,11 @@ public static function key(): string {
...
 	 */
 	protected function get_data_structure(): array {
-		add_filter( 'lang_dir_for_domain', '__return_false' ); // Prevents loading the translations too early.
-		$enum = array_keys( \PLL_Settings_Sync::list_metas_to_sync() );
-		remove_filter( 'lang_dir_for_domain', '__return_false' );
 		return array(
 			'type'  => 'array',
 			'items' => array(
 				'type' => $this->get_type(),
-				'enum' => $enum,
```
Chouby, line 50: As always, I'd like to avoid to duplicate code. Using `unload_textdomain()` added to the previous code should do the trick. Am I wrong?

Chouby (changes requested):

include/Options/Business/Sync.php
manooweb: I will check.

manooweb: > I will check.

It's ok.

include/Options/Business/Sync.php
```diff
@@ -43,6 +43,7 @@ protected function get_data_structure(): array {
 		add_filter( 'lang_dir_for_domain', '__return_false' ); // Prevents loading the translations too early.
 		$enum = array_keys( \PLL_Settings_Sync::list_metas_to_sync() );
 		remove_filter( 'lang_dir_for_domain', '__return_false' );
+		unload_textdomain( 'polylang' ); // To ensure string translation later in the process will be done.
```
Chouby, line 46: 1. I'm surprised that it works like this without allowing texdomain reload.  
2. "string translation" is confusing (as we use this expression for one of our feature), so I suggest a new comment:
3. Is it possible to write a test?
```suggestion
		unload_textdomain( 'polylang', true ); // Required to allow `_load_textdomain_just_in_time()` to load the translations.
```

Chouby (changes requested):

include/Options/Business/Sync.php
manooweb: I confirm that the 2nd parameter isn't needed to make `_load_textdomain_just_in_time()` reload the textdomain.
In this case
See https://github.com/WordPress/WordPress/blob/6.7.1/wp-includes/l10n.php#L921
Only one locale is concerned at this step.

I'm looking to write a test.

Screenfeed (approved):

PR #1626, opened 2025-02-17
Adds an event when the lang has been changed on term_lang_choice. #1529

Marianne380:
Adds a custom event `onTermLangChoice` the same way it was done for `onPostLangChoice`, see https://github.com/polylang/polylang/pull/1529.

### DIFF
--- a/js/src/term.js
+++ b/js/src/term.js
@@ -179,6 +179,8 @@ jQuery(
 				var value = $( this ).val();
 				var lang  = $( this ).children( 'option[value="' + value + '"]' ).attr( 'lang' );
 				var dir   = $( '.pll-translation-column > span[lang="' + lang + '"]' ).attr( 'dir' );
+				// The selected option in the dropdown list.
+				const selectedOption = event.target;
 
 				var data = {
 					action:     'term_lang_choice',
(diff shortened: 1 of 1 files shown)


js/src/term.js
```diff
@@ -221,11 +223,29 @@ jQuery(
...
+						);
+						document.dispatchEvent( onTermLangChoice );
 					}
 				);
 			}
 		);
+
+		// Listen to `onTermLangChoice` to perform actions after the language has been changed.
+		document.addEventListener(
+			'onPostLangChoice',
```
Hug0-Drelon, line 243:
```suggestion
			'onTermLangChoice',
```
Am I missing something? 😅

Hug0-Drelon (changes requested):

js/src/term.js
Marianne380: Nope, copy/paste.

js/src/term.js
```diff
@@ -179,6 +179,8 @@ jQuery(
 				var value = $( this ).val();
 				var lang  = $( this ).children( 'option[value="' + value + '"]' ).attr( 'lang' );
 				var dir   = $( '.pll-translation-column > span[lang="' + lang + '"]' ).attr( 'dir' );
```
Hug0-Drelon, line 181: This var becomes useless, doesn't it?

Hug0-Drelon (changes requested):

js/src/term.js
Marianne380: Indeed.

Hug0-Drelon (approved):

PR #1629, opened 2025-02-26
Fix properties not set in updater

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2513.

This PR fixes php warnings due to missing object properties.
These warnings are caused by the upgrade of `EDD_SL_Plugin_Updater` from 1.9.1 to 1.9.4: backward compatibility is incomplete. This happens if the data in cache is still in 1.9.1 format, without the new properties.
See https://github.com/polylang/polylang/commit/1fc74e2967d969c18b31ea30ba4532e3c7f0f53a.

### DIFF
--- a/install/plugin-updater.php
+++ b/install/plugin-updater.php
@@ -7,10 +7,12 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 /**
  * Allows plugins to use their own update API.
- * Modified version with 'polylang' text domain and missing comments for translators.
+ * Modified version with 'polylang' text domain, missing comments for translators, and a bug fix.
  *
  * @author Easy Digital Downloads
  * @version 1.9.4
+ *
+ * @see https://github.com/polylang/polylang-pro/pull/2516
  */
 class PLL_Plugin_Updater {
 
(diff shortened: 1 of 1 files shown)


Screenfeed:
@Chouby If my analysis is correct, I suppose it would.
However, only if the request returns all the required properties. Personally I would prefer to ensure everything is right. If you really don't want to modify this file, we can apply the same solution from https://github.com/polylang/updater/pull/14.

install/plugin-updater.php
```diff
@@ -151,6 +154,20 @@ public function get_repo_api_data() {
...
+			if ( ! isset( $version_info->$prop ) ) {
+				$version_info->$prop = '';
+			}
+		}
+
+		foreach ( array( 'icons', 'banners' ) as $prop ) {
+			if ( ! isset( $version_info->$prop ) ) {
+				$version_info->$prop = new stdClass();
+			}
+		}
```
Chouby, line 168: What do you think of
```suggestion
		$default = array(
			'url'          => '',
			'package       => '',
			'new_version'  => '',
			'tested'       => '',
			'requires'     => '',
			'requires_php' => '',
			'icons'        => new stdClass(),
			'banners'      => new stdClass(),
		);
		
		$version_info = (object) array_merge( $defaults, (array) $version_info );
```

Chouby (changes requested):

install/plugin-updater.php
Screenfeed: That's indeed a solution I thought about.

PR #1634, opened 2025-03-05
Fix test failing due to `get_calendar()`

Screenfeed:
This fixes [this test](https://github.com/polylang/polylang/actions/runs/13654297541/job/38169679134).

This is due to [this change](https://github.com/WordPress/WordPress/commit/f4c1c5778ece35d0ca6192a6343bac1dea9c158b) from [this ticket](https://core.trac.wordpress.org/ticket/34093).
Also [this change](https://core.trac.wordpress.org/changeset/59939) after [this comment](https://core.trac.wordpress.org/ticket/34093#comment:28).

The new `get_calendar()` is backward compatible with the previous parameter signature, and allows to filter the post type used in the calendar. This introduces a filter `get_calendar_args`, and ads a 2nd parameter to the filter `get_calendar`.


Hug0-Drelon (approved):
The code LGTM.

I've though about some improvement regarding the survey. Should we run WP core PHPUnit tests on `get_calendar()` on our own impelmentation?

PR #1637, opened 2025-03-10
Update import of strings translations with WP Importer.

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/2490.

Use term metas for the strings translations import and remove the old way with the post and its metas.

Also adds a test, particularly in the case where languages already exist and we import a file with different strings translations than the existing ones.


Hug0-Drelon (changes requested):
I think we should add some tests, this business domain doesn't seem well covered.

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -89,11 +89,31 @@ protected function _import_wp( $filename, $users = array(), $fetch_files = true
...
 	public function test_simple_import() {
-		$this->_import_wp( PLL_TEST_DATA_DIR . 'test-import.xml' );
-
+		$import_filepath      = PLL_TEST_DATA_DIR . 'test-import.xml';
+		$import_filepath_test = PLL_TEST_DATA_DIR . 'test-modified-import.xml';
+		$import_content  = strtr(
+			file_get_contents( $import_filepath ),
+			array(
+				'{lang_id_1}' => '<![CDATA[172]]>',
+				'{lang_id_2}' => '<![CDATA[174]]>',
```
Hug0-Drelon, line 98: Since you use hard coded IDs, why the need to have formatted data with `lag_id_*`?

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -116,4 +136,39 @@ public function test_simple_import() {
...
+
+		// Strings translations.
+		$en_mo = new PLL_MO();
+		$en_mo->import_from_db( self::$model->languages->get( 'en' ) );
+		$this->assertSame( 'WordPress EN', $en_mo->translate( 'WordPress' ) );
+		$fr_mo = new PLL_MO();
+		$fr_mo->import_from_db( self::$model->languages->get( 'fr' ) );
+		$this->assertSame( 'WordPress FR', $fr_mo->translate( 'WordPress' ) );
+
+		unlink( $import_filepath_test );
```
Hug0-Drelon, line 172: I should be done in `teard_down()` so even it the test fails, it is still called.

Hug0-Drelon (changes requested):

tests/phpunit/tests/plugins/test-wp-importer.php
Marianne380: Actually this is needed for the other test `test_simple_import_with_existing_languages()` because the languages are created before launching the import, so we need to inject the IDs into the XML file.

In this test, I've just restored the original IDs of the import file.

Hug0-Drelon (approved):

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -116,4 +140,37 @@ public function test_simple_import() {
 		$this->assertEquals( 'fr', self::$model->term->get_language( $fr->term_id )->slug );
 		$this->assertEqualSetsWithIndex( array( 'en' => $en->term_id, 'fr' => $fr->term_id ), self::$model->term->get_translations( $en->term_id ) );
 	}
+
+	/**
+	 * This test checks the import of strings translations with already existing languages.
+	 * The strings translations are not the same between those of the existing languages and those of the import file.
+	 */
+	public function test_simple_import_with_existing_languages() {
```
Chouby, line 148: I believe the test is incomplete. I would expect that it checks that imported strings translations are merged with existing ones and do not overwrite them.

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -89,11 +93,31 @@ protected function _import_wp( $filename, $users = array(), $fetch_files = true
 	}
 
 	public function test_simple_import() {
-		$this->_import_wp( PLL_TEST_DATA_DIR . 'test-import.xml' );
-
+		$import_filepath      = PLL_TEST_DATA_DIR . 'test-import.xml';
+		$import_filepath_test = PLL_TEST_DATA_DIR . 'test-modified-import.xml';
+		$import_content  = strtr(
```
Chouby, line 98:
```suggestion
		
		// Prepare the imported file by replacing language id placeholders by a meaningful int.
		$import_content  = strtr(
```

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -89,11 +93,31 @@ protected function _import_wp( $filename, $users = array(), $fetch_files = true
...
+			file_get_contents( $import_filepath ),
+			array(
+				'{lang_id_1}' => '<![CDATA[172]]>',
+				'{lang_id_2}' => '<![CDATA[174]]>',
```
Chouby, line 102: The `CDATA` being always present, I would keep it in the imported file and just have the numbers here.

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -116,4 +140,37 @@ public function test_simple_import() {
...
+	 * This test checks the import of strings translations with already existing languages.
+	 * The strings translations are not the same between those of the existing languages and those of the import file.
+	 */
+	public function test_simple_import_with_existing_languages() {
+		self::create_language( 'en_US' );
+		self::create_language( 'fr_FR' );
+
+		$import_filepath      = PLL_TEST_DATA_DIR . 'test-import.xml';
+		$import_filepath_test = PLL_TEST_DATA_DIR . 'test-modified-import.xml';
+		$import_content  = strtr(
```
Chouby, line 154:
```suggestion

		// Prepare the imported file by replacing language id placeholders by the id of the languages just created.
		$import_content  = strtr(
```

Chouby (changes requested):

tests/phpunit/tests/plugins/test-wp-importer.php
Marianne380: Okay, see https://github.com/polylang/polylang/pull/1637/commits/a755bb50bbaa119f9dc0cec65b8930686aa82961.

PR #1423, opened 2024-02-14
Translate the content of the widget custom html

Chouby:
WP 4.8.1 introduced the widget custom html which works more or less like the widget text.
The content honors the filter 'widget_text'.
However it is stored in DB with the key 'content' instead of 'text'. So we fail to register the string in the strings translation panel.

In this PR, we register the string corresponding to the widget content when the widget is displayed in all languages.
It also includes minor formatting enhancements.

PR #1641, opened 2025-03-17
Fix widget block wrongly added to string translation

Chouby:
See https://github.com/polylang/polylang-pro/issues/2552

In #1423, we introduced the possibility to translate the widget custom html which can be translated the same way as the widget text but stores the text with a different key.

This however introduces an undesired side effect that widget blocks are now registered in the strings translations because the content is stored exactly the same ways as the widget custom html text. These blocks are however not translated as they don't honor the filter `widget_text`. These are also much more complex to handle as we need to parse block to export them properly.

This PR checks the widget instance to avoid to mix the widget custom html and the widget block.


Hug0-Drelon (approved):

Chrystll:
I confirm that widget blocks no longer appear in Strings Translations, while legacy widgets can still be translated as expected.

PR #1644, opened 2025-03-24
Avoid fetching  `active_sitewide_plugins` on single site installation.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2557

## How?
- Split `Options::init_options_for_blog()` and create a new method `Options::on_blog_switch()` to manage multisite logic.
- Ensure `Options::__construct()` populates `Options::$current_blog_id` property correctly.
- Add a test to cover the regression.

### DIFF
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -78,7 +78,7 @@ class Options implements ArrayAccess, IteratorAggregate {
 		$this->current_blog_id = $this->blog_id;
 
 		// Handle options.
-		$this->init_options_for_blog( $this->blog_id );
+		$this->init_options_for_current_blog();
 
 		add_filter( 'pre_update_option_polylang', array( $this, 'protect_wp_option_storage' ), 1 );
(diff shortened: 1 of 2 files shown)


include/Options/Options.php
```diff
@@ -146,16 +146,6 @@ public function protect_wp_option_storage( $value ) {
 	 * @return void
 	 */
 	public function init_options_for_blog( $blog_id ): void {
```
Screenfeed, line 148: The `$blog_id` parameter is not needed anymore, we can use `$this->current_blog_id`.

Screenfeed (changes requested):

include/Options/Options.php
Hug0-Drelon: Good idea, I've been a bit further by making the method name more explicit and make it private.
See 1a723f14.

Screenfeed (approved):

include/Options/Options.php
Screenfeed: Good idea.

PR #1648, opened 2025-04-01
Improve performances of textdomain unloading

Screenfeed:
See https://github.com/polylang/polylang-pro/issues/2563.

In `Options\Primitive\Abstract_List\sync::get_data_structure()` we had a way to prevent WP to load the plugin's textdomain.
While it works, it doesn't prevent WP from fetching the `.mo` files, resulting in an unnecessary loss of performances.

## Numbers

By using the code provided in the issue to measure timings, here are the results, with 3 measurements for each case. The important value is the 1st timer.

Without the solution of this PR, without the "trick" added to `Options\Primitive\Abstract_List\sync::get_data_structure()`:

| | 1st timer | 2nd timer |
|---|---|---|
| Measure 1 | 8.8849067687988 | 0.072002410888672 |
| Measure 2 | 7.2968006134033 | 0.075101852416992 |
| Measure 3 | 6.9570541381836 | 0.077962875366211 |

Without the solution of this PR, but with the "trick" added to `Options\Primitive\Abstract_List\sync::get_data_structure()`:

| | 1st timer | 2nd timer |
|---|---|---|
| Measure 1 | 4.2450428009033 | 0.041961669921875 |
| Measure 2 | 2.2268295288086 | 0.040054321289062 |
| Measure 3 | 3.0200481414795 | 0.036954879760742 |

With the solution of this PR:

| | 1st timer | 2nd timer |
|---|---|---|
| Measure 1 | 0.010013580322266 | 0.037908554077148 |
| Measure 2 | 0.011920928955078 | 0.040054321289062 |
| Measure 3 | 0.0078678131103516 | 0.037908554077148 |


Hug0-Drelon (changes requested):

include/class-polylang.php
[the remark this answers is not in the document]
Chouby: What do you think of:
```php
		$GLOBALS['l10n_unloaded']['polylang'] = true;
		unset( $GLOBALS['l10n_unloaded']['polylang'] );
		return array(
			'type'  => 'array',
			'items' => array(
				'type' => $this->get_type(),
				'enum' => $enum,
			),
		);
```

The method proposed by @Hug0-Drelon  is smart. However it requires to modify `PLL_Settings_Sync::list_metas_to_sync()` to match a specific usage of the method.
In my opinion, if possible, it's better to avoid to modify a function because we want to use its returned value in a specific way.

The method proposed by @Screenfeed  prevents loading the translations too early but will later load them in all cases, even on frontend when we don't need to translate any string.

I'm often annoyed by `_load_textdomain_just_in_time()` but I must admit that it's useful for our use case (on frontend where our need to translate strings is required for only a few websites).

I take back the idea of using directly the global as `unload_textdomain()` cannot match our use case ( indeed calling `unload_texdomain( 'polylang', true )` after  `unload_texdomain( 'polylang' )` would be ineffective ).

include/Options/Business/Sync.php
```diff
@@ -40,9 +40,9 @@ public static function key(): string {
 	 * @phpstan-return array{type: 'array', items: array{type: SchemaType, enum: non-empty-list<non-falsy-string>}}
 	 */
 	protected function get_data_structure(): array {
-		add_filter( 'lang_dir_for_domain', '__return_false' ); // Prevents loading the translations too early.
+		$GLOBALS['l10n_unloaded']['polylang'] = true;
 		$enum = array_keys( \PLL_Settings_Sync::list_metas_to_sync() );
-		remove_filter( 'lang_dir_for_domain', '__return_false' );
+		unset( $GLOBALS['l10n_unloaded']['polylang'] );
 		unload_textdomain( 'polylang', true ); // Required to allow `_load_textdomain_just_in_time()` to load the translations.
```
Chouby, line 46:
```suggestion
```
I guess we don't need this line anymore. Do we ?

.github/workflows/integration-and-unit-tests.yml
```diff
@@ -10,7 +10,9 @@ on:
       - '**.yml'
       - '**.sh'
   pull_request:
-    branches: master
+    branches:
+      - master
+      - dev-3.8
```
Chouby, line 15: Given the small change, I propose to include this PR in v3.7

Chouby (changes requested):

include/Options/Business/Sync.php
Screenfeed: We do.

Screenfeed: Our texts are not translated otherwise (the test also fails).

Screenfeed:
Something is wrong with the rebase (who is surprised?).

include/Options/Business/Sync.php
Chouby: I was surprised but understand now. This is not really the global `$l10n_unloaded` which prevents the translation but this [piece of code](https://github.com/WordPress/wordpress-develop/blob/6.7.2/src/wp-includes/l10n.php#L1412-L1417).

With this in mind, I would now propose:
```php
		$GLOBALS['l10n']['polylang'] = new \NOOP_Translations();
		unset( $GLOBALS['l10n']['polylang'] );
		return array(
			'type'  => 'array',
			'items' => array(
				'type' => $this->get_type(),
				'enum' => $enum,
			),
		);
	}
```
What do you think?

Screenfeed:
Closed in favor of #1652.

PR #1650, opened 2025-04-07
Prevent DB queries to fetch the language terms

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/2562.

When the language slug is present in the URL, and then passed to the query, WP changes the "tax query by `slug`" by a "tax query by `term_taxonomy_id`. This creates 2 extra queries to fetch this ID.

These 2 extra queries are useless because the `term_taxonomy_id` of the language can be fetch from our `PLL_Language` object.

### DIFF
--- a/frontend/frontend.php
+++ b/frontend/frontend.php
@@ -88,10 +88,8 @@ class PLL_Frontend extends PLL_Base {
 		// Avoids the language being the queried object when querying multiple taxonomies
 		add_action( 'parse_tax_query', array( $this, 'parse_tax_query' ), 1 );
 
-		if ( $this->model->has_languages() ) {
-			// Prevent unnecessary queries to the DB.
-			add_action( 'parse_tax_query', array( $this, 'transform_query' ) );
-		}
+		// Prevent unnecessary queries to the DB.
+		add_action( 'parse_tax_query', array( $this, 'transform_query' ) );
 
 		// Filters posts by language
 		add_action( 'parse_query', array( $this, 'parse_query' ), 6 );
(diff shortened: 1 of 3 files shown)


manooweb (approved):

frontend/frontend.php
```diff
@@ -88,6 +88,11 @@ public function __construct( &$links_model ) {
 		// Avoids the language being the queried object when querying multiple taxonomies
 		add_action( 'parse_tax_query', array( $this, 'parse_tax_query' ), 1 );
 
+		if ( $this->model->has_languages() ) {
```
Chouby, line 91: I believe this test is useless because we already did it before (we can't reach this point without languages): https://github.com/polylang/polylang/blob/3.7-beta3/include/class-polylang.php#L219
Also we could merge this action to the previous one.

frontend/frontend.php
```diff
@@ -179,6 +184,60 @@ public function parse_tax_query( $query ) {
...
+	 *     'operator'         => 'IN',
+	 *     'include_children' => 1,
+	 * )
+	 *
+	 * @since 3.8
+	 *
+	 * @param WP_Query $query WP_Query object.
+	 * @return void
+	 */
+	public function transform_query( $query ): void {
```
Chouby, line 215: Also what do you think to move this to `PLL_Query` which is supposed to include our methods related to `WP_Query`?

Chouby (changes requested):

frontend/frontend.php
Screenfeed: Indeed. Done in [eaf0254](https://github.com/polylang/polylang/pull/1650/commits/eaf02541d7cf7dcbc3c193a88e60b29eb095df7f).

frontend/frontend.php
Screenfeed: Why not. Done in [3385903](https://github.com/polylang/polylang/pull/1650/commits/338590358f861a365f0b500343f550c38bb48d15).

PR #1651, opened 2025-04-08
Run PHPUnit in wp-env containers.

Hug0-Drelon:
## What?
Improve DX by allowing PHPUnit to run in wp-env Docker containers.

## Why?
Benefit from wp-env installation that provides a clean WordPress installation, a DB server and `wordpress-test-utils` library out of the box.
No need to install manually all of this in `tmp/` anymore.

## How?
- Add `wordpress/wp-env` to NPM dependencies.
- Add `.wp-env.json` config file.
- Create a custom `wp-tests-config-docker.php` file for this purpose. It differs in two things from the usual `wordpress-tests-lib/wp-tests-config.php`:
    - It defines `ABSPATH` according to wp-env Docker mapping for the root of the web server (where WordPress is installed, no matter which container you want to use).
    - It dynamically use Docker defined database constant, so connexion to MySQL is seamless.
- Adapt bootstrap files for both integration and unit tests, need `WP_TESTS_CONFIG_FILE_PATH` to point to `wp-tests-config-docker.php`.
- Minor changes in `links-trait.php` to avoid a side effect when generating path with `plugins_url()` in specific links configuration.
- Add some NPM script for ease of use (forwarding PHPUnit commands to desired Docker container is a bit verbose).

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build:dev": "npm run clean && webpack --mode development",
     "dev": "npm run clean && webpack --mode development --watch",
     "clean": "rimraf js/build css/build",
-		"env:start": "wp-env start",
+    "env:start": "wp-env start",
     "env:stop": "wp-env stop",
     "env:clean": "wp-env clean",
     "env:composer": "wp-env run tests-cli --env-cwd=wp-content/plugins/polylang composer",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -11,7 +11,12 @@
     "build:dev": "npm run clean && webpack --mode development",
     "dev": "npm run clean && webpack --mode development --watch",
     "clean": "rimraf js/build css/build",
-    "test": "echo \"Error: no test specified\" && exit 1"
+		"env:start": "wp-env start",
```
manooweb, line 14:
```suggestion
    "env:start": "wp-env start",
```

manooweb (changes requested):

manooweb (approved):

PR #1653, opened 2025-04-08
Enforce transient deletion and creation in options table when using external object cache.

Hug0-Drelon:
## What?
Partially fixes #2125.

## How?
- Ensure `pll_languages_list` transient is deleted from/created in options table even if external object cache is used.
    - In usual process thanks to `set_transient_pll_languages_list` and `delete_transient_pll_languages_list` hooks with new methods in `Languages` model.
    - In upgrade or uninstallation by calling related core functions.
- Add a PHPUnit test to cover the issue, introducing a new drop in for external object cache in tests (`Object_Cache_Annihilator`).

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -735,6 +735,34 @@ class Languages {
 	}
 
 	/**
+	 * Deletes the transient from the options table since WordPress does not do it when using object cache.
+	 *
+	 * @since 3.8
+	 *
+	 * @return void
+	 */
+	public function delete_transient_from_options_table(): void {
+		if ( wp_using_ext_object_cache() || wp_installing() ) {
+			delete_option( '_transient_' . self::TRANSIENT_NAME );
+		}
+	}
+
+	/**
+	 * Sets the transient in the options table since WordPress does not do it when using object cache.
+	 *
+	 * @since 3.8
+	 *
+	 * @param mixed $value Value of the transient.
+	 * @return void
+	 */
+	public function set_transient_to_options_table( $value ): void {
+		if ( wp_using_ext_object_cache() || wp_installing() ) {
+			wp_prime_option_caches( array( '_transient_' . self::TRANSIENT_NAME ) );
+			add_option( '_transient_' . self::TRANSIENT_NAME, $value, '', true );
+		}
+	}
+
+	/**
 	 * Builds the language metas into an array and serializes it, to be stored in the term description.
 	 *
 	 * @since 3.4
--- a/include/model.php
+++ b/include/model.php
@@ -121,6 +121,9 @@ class PLL_Model {
 
 		// Just in case someone would like to display the language description ;).
 		add_filter( 'language_description', '__return_empty_string' );
+
+		add_action( 'delete_transient_pll_languages_list', array( $this->languages, 'delete_transient_from_options_table' ) );
+		add_action( 'set_transient_pll_languages_list', array( $this->languages, 'set_transient_to_options_table' ) );
 	}
 
 	/**
--- a/install/upgrade.php
+++ b/install/upgrade.php
@@ -58,6 +58,7 @@ class PLL_Upgrade {
 		}
 
 		delete_transient( 'pll_languages_list' );
+		delete_option( '_transient_pll_languages_list' ); // Force deletion of the transient from the options table in case external object cache is used.
 		add_action( 'admin_init', array( $this, '_upgrade' ) );
 		return true;
 	}
--- a/tests/phpunit/data/object-cache-annihilator.php
+++ b/tests/phpunit/data/object-cache-annihilator.php
@@ -1,0 +1,633 @@
+<?php
+/**
+ * Object Cache Annihilator.
+ *
+ * @package WP_Syntex\Object_Cache_Annihilator
+ *
+ * A simple file-based object cache implementation for testing purposes.
+ * Stores cache data in the filesystem with support for groups and expiration.
+ */
+
+namespace WP_Syntex;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// Check if WordPress is already loaded
+if ( function_exists( 'wp_cache_add' ) ) {
+	return;
+}
+
+use WP_Object_Cache;
+
+/**
+ * Cache object taking place of `WP_Object_Cache` and allows for easy testing of cache operations.
+ *
+ * @since 0.1.0
+ */
+class Object_Cache_Annihilator {
+	/**
+	 * Directory where cache files are stored.
+	 *
+	 * @var string
+	 */
+	private $cache_dir;
+
+	/**
+	 * Prefix for cache files.
+	 *
+	 * @var string
+	 */
+	private $cache_prefix;
+
+	/**
+	 * In-memory cache array.
+	 *
+	 * @var array
+	 */
+	private $cache = array();
+
+	/**
+	 * Number of successful cache retrievals.
+	 *
+	 * @var int
+	 */
+	private $cache_hits = 0;
+
+	/**
+	 * Number of failed cache retrievals.
+	 *
+	 * @var int
+	 */
+	private $cache_misses = 0;
+
+	/**
+	 * List of global cache groups.
+	 *
+	 * @var array
+	 */
+	private $global_groups = array();
+
+	/**
+	 * List of non-persistent cache groups.
+	 *
+	 * @var array
+	 */
+	private $non_persistent_groups = array();
+
+	/**
+	 * Whether the site is a multisite installation.
+	 *
+	 * @var bool
+	 */
+	private $multisite = false;
+
+	/**
+	 * Blog prefix for multisite installations.
+	 *
+	 * @var string
+	 */
+	private $blog_prefix = '';
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->cache_dir = WP_CONTENT_DIR . '/object-cache/';
+		$this->cache_prefix = 'wp_cache_';
+		$this->multisite = is_multisite();
+		$this->blog_prefix = $this->multisite ? get_current_blog_id() . ':' : '';
+
+		if ( ! file_exists( $this->cache_dir ) ) {
+			wp_mkdir_p( $this->cache_dir );
+		}
+	}
+
+	/**
+	 * Adds data to the cache if it doesn't already exist.
+	 *
+	 * @param string $key    The cache key.
+	 * @param mixed  $data   The data to store.
+	 * @param string $group  Optional. The cache group. Default 'default'.
+	 * @param int    $expire Optional. When to expire the cache contents. Default 0 (no expiration).
+	 * @return bool True if the data was added, false if it already exists.
+	 */
+	public function add( $key, $data, $group = 'default', $expire = 0 ) {
+		if ( empty( $group ) ) {
+			$group = 'default';
+		}
+
+		if ( $this->get( $key, $group ) !== false ) {
+			return false;
+		}
+
+		return $this->set( $key, $data, $group, $expire );
+	}
+
+	/**
+	 * Sets data in the cache.
+	 *
+	 * @param string $key    The cache key.
+	 * @param mixed  $data   The data to store.
+	 * @param string $group  Optional. The cache group. Default 'default'.
+	 * @param int    $expire Optional. When to expire the cache contents. Default 0 (no expiration).
+	 * @return bool True on success, false on failure.
+	 */
+	public function set( $key, $data, $group = 'default', $expire = 0 ) {
+		if ( empty( $group ) ) {
+			$group = 'default';
+		}
+
+		$cache_file = $this->get_cache_file_path( $key, $group );
+		$cache_data = array(
+			'value' => $data,
+			'expires' => $expire ? time() + $expire : 0,
+		);
+
+		$this->cache[ $group ][ $key ] = $data;
+
+		// Ensure directory exists
+		$dir = dirname( $cache_file );
+		if ( ! file_exists( $dir ) ) {
+			wp_mkdir_p( $dir );
+		}
+
+		// Serialize with validation
+		$serialized = serialize( $cache_data ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_serialize
+		if ( false === $serialized ) {
+			return false;
+		}
+
+		return file_put_contents( $cache_file, $serialized );
+	}
+
+	/**
+	 * Retrieves data from the cache.
+	 *
+	 * @param string $key    The cache key.
+	 * @param string $group  Optional. The cache group. Default 'default'.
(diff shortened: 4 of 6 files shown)


tests/phpunit/tests/test-object-cache.php
```diff
@@ -0,0 +1,96 @@
...
+
+		// Annihilate the annihilator.
+		$wp_object_cache->flush();
+		unlink( WP_CONTENT_DIR . '/object-cache.php' );
+		$wp_object_cache = $this->cache_backup;
+
+		parent::tear_down();
+	}
+
+	public function test_object_cache_unavailable() {
```
Screenfeed, line 41: The test doesn't fail if I comment the 2 lines added to `PLL_Model` :s

Screenfeed (changes requested):

include/Model/Languages.php
```diff
@@ -734,6 +734,41 @@ public function clean_cache(): void {
...
+	 * @return void
+	 */
+	public function set_transient_to_options_table( $value ): void {
+		global $wp_version;
+
+		if ( wp_using_ext_object_cache() || wp_installing() ) {
+			// Backward compatibility with WordPress 6.4.
+			if ( version_compare( $wp_version, '6.4', '>' ) ) {
+				wp_prime_option_caches( array( '_transient_' . self::TRANSIENT_NAME ) );
+			}
```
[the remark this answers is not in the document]
Hug0-Drelon: Indeed, I misunderstood the usage of `wp_prime_option_caches()` here.
Removed in d6e9238e6003e2811214e6c4d8268a60dd16affe.

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,633 @@
+<?php
+/**
+ * Object Cache Annihilator.
+ *
+ * @package WP_Syntex\Object_Cache_Annihilator
```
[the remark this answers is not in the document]
Hug0-Drelon: I've cleaned up this file in a75944e8a7b11a13148ee569d1142745d342602b.
This is from the plugin implementation. This has nothing to do in tests.
But you have to know that with namespaces this is the way to use `@package` tags according to PHPDoc.

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,633 @@
...
+ *
+ * A simple file-based object cache implementation for testing purposes.
+ * Stores cache data in the filesystem with support for groups and expiration.
+ */
+
+namespace WP_Syntex;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
```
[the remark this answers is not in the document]
Hug0-Drelon: I've cleaned up the file in a75944e8a7b11a13148ee569d1142745d342602b.
No need to have production-like code here.

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,633 @@
...
+
+// Check if WordPress is already loaded
+if ( function_exists( 'wp_cache_add' ) ) {
+	return;
+}
+
```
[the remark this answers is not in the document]
Hug0-Drelon: Removed in a75944e8a7b11a13148ee569d1142745d342602b.
This line and core function implementations are for production. This is required to implement a cache drop-in.
To put in simply, WordPress won't load its cache related files if a drop-in is found with proper implementations of core functions.

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,633 @@
...
+		}
+
+		$file_contents = file_get_contents( $cache_file );
+		if ( false === $file_contents ) {
+			++$this->cache_misses;
+			$found = false;
+			return false;
+		}
+
+		$data = unserialize( $file_contents ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.serialize_unserialize
```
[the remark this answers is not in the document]
Hug0-Drelon: It's test code, I'd rather not.

tests/phpunit/tests/test-object-cache.php
```diff
@@ -0,0 +1,96 @@
...
+		$GLOBALS['polylang'] = $this->pll_env;
+	}
+
+	public function tear_down() {
+		global $wp_object_cache;
+
+		// Annihilate the annihilator.
+		$wp_object_cache->flush();
+		unlink( WP_CONTENT_DIR . '/object-cache.php' );
+		$wp_object_cache = $this->cache_backup;
```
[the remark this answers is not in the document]
Hug0-Drelon: The previous instance is the WordPress one (`WP_Object_Cache`), and it seems that it holds some global scope information useful for other test classes.

tests/phpunit/tests/test-object-cache.php
Hug0-Drelon: Fixed in dda9d27b371a94a532f7c9d7e54b707f231a116c.
It should fail if the fix is disabled now. I don't know where I fucked things up during my development...

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,633 @@
...
+	 *
+	 * @param string $group Optional. The cache group. Default 'default'.
+	 * @return bool True on success, false on failure.
+	 */
+	public function flush_group( $group ) {
+		if ( empty( $group ) ) {
+			$group = 'default';
+		}
+
+		$group_dir = $this->cache_dir . $group . '/';
```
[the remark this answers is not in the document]
Hug0-Drelon: `$this->cache_dir` is recursively deleted when `Object_Cache_Annihilator::flush()` is called (during `tear_down()` in our case.
See `Object_Cache_Annihilator::cleanup_directory()`for details.

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,633 @@
...
+		}
+
+		if ( ! is_numeric( $value ) ) {
+			$value = 0;
+		}
+
+		$value += $offset;
+		if ( $value < 0 ) {
+			$value = 0;
+		}
```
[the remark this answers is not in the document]
Hug0-Drelon: Why not.

tests/phpunit/data/object-cache-annihilator.php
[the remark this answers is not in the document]
Hug0-Drelon: What production code are you referring to?

tests/phpunit/data/object-cache-annihilator.php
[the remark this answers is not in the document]
Screenfeed: You mean you prefer to trigger an error if something goes wrong? OK, fair.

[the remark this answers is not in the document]
Hug0-Drelon: Exactly.

tests/phpunit/data/object-cache-annihilator.php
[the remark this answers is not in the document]
Screenfeed: Yes, the "plugin implementation" you were referring to.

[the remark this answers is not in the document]
Hug0-Drelon: I see.
Frankly I don't see any benefit of maintaining a mirror of this plugin in our tests.
Also, core function implementation are "almost" copied from core and are not used at all in tests.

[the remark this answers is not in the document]
Hug0-Drelon: In fact, only the class `Object_Cache_Annihilator` is useful in our PHPUnit tests, nothing more. The drop-in is managed directly in `set_up()`.

tests/phpunit/tests/test-object-cache.php
```diff
@@ -0,0 +1,110 @@
...
+		$this->assertSame( 'en', $object_cache_languages[0]['slug'], 'The first language should be en.' );
+		$this->assertSame( 'fr', $object_cache_languages[1]['slug'], 'The second language should be fr.' );
+
+		// Let's mess around and create a new language. We have to manually create the language to use only one instance of the model.
+		$languages        = include POLYLANG_DIR . '/settings/languages.php';
+		$de               = $languages['de_DE'];
+		$de['slug']       = $de['code'];
+		$de['rtl']        = 0;
+		$de['term_group'] = 0;
+		$this->assertGreaterThan( 0, $this->pll_env->model->languages->add( $de ) );
```
Screenfeed, line 73: `Languages::add()` returns `true|WP_Error`.

```suggestion
		$this->assertTrue( $this->pll_env->model->languages->add( $de ) );
```

Screenfeed (changes requested):

Screenfeed (approved):

include/Model/Languages.php
```diff
@@ -734,6 +734,33 @@ public function clean_cache(): void {
...
 
+	/**
+	 * Deletes the transient from the options table since WordPress does not do it when using object cache.
+	 *
+	 * @since 3.8
+	 *
+	 * @return void
+	 */
+	public function delete_transient_from_options_table(): void {
+		if ( wp_using_ext_object_cache() || wp_installing() ) {
```
Chouby, line 745: Why `wp_installing()` too?

include/Model/Languages.php
```diff
@@ -734,6 +734,33 @@ public function clean_cache(): void {
...
+	 * @since 3.8
+	 *
+	 * @param mixed $value Value of the transient.
+	 * @return void
+	 */
+	public function set_transient_to_options_table( $value ): void {
+		if ( wp_using_ext_object_cache() || wp_installing() ) {
+			add_option( '_transient_' . self::TRANSIENT_NAME, $value, '', true );
+		}
+	}
```
Chouby, line 762: Is it useful? I we previously deleted the transient and at some point the external is not used, [the option will be automatically re-created](https://github.com/polylang/polylang/blob/3.7-rc1/include/Model/Languages.php#L536-L538). Am I wrong?

include/model.php
```diff
@@ -121,6 +121,9 @@ public function __construct( Options &$options ) {
 
 		// Just in case someone would like to display the language description ;).
 		add_filter( 'language_description', '__return_empty_string' );
+
+		add_action( 'delete_transient_pll_languages_list', array( $this->languages, 'delete_transient_from_options_table' ) );
+		add_action( 'set_transient_pll_languages_list', array( $this->languages, 'set_transient_to_options_table' ) );
```
Chouby, line 126: Since you are using the constant in `Languages`, it would be better to use it here the action name too.

install/upgrade.php
```diff
@@ -58,6 +58,7 @@ public function upgrade() {
 		}
 
 		delete_transient( 'pll_languages_list' );
+		delete_option( '_transient_pll_languages_list' ); // Force deletion of the transient from the options table in case external object cache is used.
```
Chouby, line 61: Why is it necessary? I suppose that the action added above will do its job. Am I wrong?

tests/phpunit/data/object-cache-annihilator.php
```diff
@@ -0,0 +1,523 @@
+<?php
+/**
+ * Object Cache Annihilator.
```
Chouby, line 3: I guess the source of this file is coming from https://github.com/polylang/object-cache-annihilator.
I suggest to maintain only the plugin and "require-dev" it with composer to re-use it.

uninstall.php
```diff
@@ -186,6 +186,7 @@ public function uninstall() {
 
 		// Delete transients.
 		delete_transient( 'pll_languages_list' );
+		delete_option( '_transient_pll_languages_list' ); // Force deletion of the transient from the options table in case external object cache is used.
```
Chouby, line 189:
```suggestion
		delete_option( '_transient_pll_languages_list' ); // Force deletion of the transient from the options table in case external object cache is used.
```
We don't need to delete the external cache in case of clean uninstall. Do we?

Chouby (changes requested):

include/Model/Languages.php
Hug0-Drelon: As discussed during our meeting, this is the way WordPress "prevent" transient from being deleted in DB. So we have to go against in any cases.

include/Model/Languages.php
Hug0-Drelon: Ass discussed during our meeting, this is safer to avoid race conditions between cache and DB.

tests/phpunit/data/object-cache-annihilator.php
Hug0-Drelon: Done in e7895856639828f9502d63d492905ccf1594b797.

include/model.php
Hug0-Drelon: Done in a0e53137.

install/upgrade.php
Hug0-Drelon: For the added hooks to be effective, we must delay the deletion. Done in b01def54.

uninstall.php
Hug0-Drelon: Done in 3e66676a.

composer.json
```diff
@@ -16,8 +16,16 @@
...
+		"wpsyntex/wp-phpunit": "dev-master",
+		"wpsyntex/object-cache-annihilator": "dev-master"
 	},
+	"repositories": [
+		{
+			"name": "wpsyntex/object-cache-annihilator",
+			"type": "vcs",
+			"url":  "git@github.com:polylang/object-cache-annihilator.git"
+		}
+	],
```
Chouby, line 28: I've added the package to Packagist: https://packagist.org/packages/wpsyntex/object-cache-annihilator
SO you shouldn't need this.

uninstall.php
```diff
@@ -185,7 +185,7 @@ public function uninstall() {
 		delete_option( 'pll_language_from_content_available' );
 
 		// Delete transients.
-		delete_transient( 'pll_languages_list' );
+		delete_option( '_transient_pll_languages_list' ); // Force deletion of the transient from the options table in case external object cache is used.
```
Chouby, line 188:
```suggestion
		delete_option( '_transient_pll_languages_list' ); // Force deletion of the transient from the options table even in case external object cache is used.
```

composer.json
Hug0-Drelon: Nice 👍 
Done in 1fd4beb0.

include/Model/Languages.php
Hug0-Drelon: We eventually chose to relay only on transient deletion to keep the DB up to date.

So I removed `set_transient_to_options_table()` in 9d2816186f2e66c18bc9186c7383fcf5a061cd4a.

I also ensure the test fails correctly with this change (if the fix is removed), see ad1608524f5f8df74bd1bcbf99b6d302db10dc86.

Then I took the opportunity to create a dedicated testcase in 9e461881e869c6a5245ce1cdc20b4f3f9698b0cb, indeed I'm currently working on the same issue in Polylang Pro and I could use a testcase for it.

PR #1658, opened 2025-04-29
Create a `PLL_MO_Trait` for PHPUnit.

Hug0-Drelon:
## What?
All in the title, suggested in https://github.com/polylang/polylang-pro/pull/2598/files#r2063806053.

## Why?
Keep it DRY.

## How?
- Create a dedicated trait, no need to pollute the main trait, also some tests might only need that (and not the whole features from `PLL_UnitTestCase_Trait`)
- We must pass the languages, because some tests use different models to get them (static vs non static...)

### DIFF
--- a/tests/phpunit/includes/pll-mo-trait.php
+++ b/tests/phpunit/includes/pll-mo-trait.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * Trait to handle PLL_MO objects.
+ */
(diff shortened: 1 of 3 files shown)


Marianne380 (approved):

Chouby (changes requested):
I'm not sure it's useful to do this in a separate trait. I would add the introduced method directly to main test trait.

PR #1659, opened 2025-05-06
Allow `null` in the `nav_menus` option 

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2615.

In the `nav_menus` option, if a menu ID is `null` instead of an integer, it is now casted to `0` instead of invalidating the whole option.

Example:
```

'twentytwentyone' => array(
	'primary' => array(
		'en' => 22,
		'fr' => null, // This will become `0`.
	),
	'footer' => array(
		'en' => 24,
		'fr' => 0,
	),
),
```

This is done by implementing `Abstract_Option::prepare()` in the class `Nav_Menus`. It is called before `rest_sanitize_value_from_schema()` (in constructor) and before `rest_validate_value_from_schema()` in the `set()` method.

This new implementation replaces all `null` menu IDs by `0`. 

Also, it removes all non-array values that are not in the current theme "branch". For non-array values that are in the current theme "branch", we let the validation process trigger its error. Note: I don't expect this to be triggered often, even less in the contructor (since the data comes from the DB).

### DIFF
--- a/include/Options/Business/Nav_Menus.php
+++ b/include/Options/Business/Nav_Menus.php
@@ -84,6 +84,57 @@ class Nav_Menus extends Abstract_Option {
 	}
 
 	/**
+	 * Prepares a value before validation.
+	 *
+	 * @since 3.7.2
+	 *
+	 * @param mixed $value Value to format.
+	 * @return mixed
+	 */
+	protected function prepare( $value ) {
+		if ( ! is_array( $value ) ) {
+			return $value;
+		}
+
+		$cur_theme = get_option( 'stylesheet' );
+
+		foreach ( $value as $theme => &$menus_by_loc ) {
+			if ( ! is_array( $menus_by_loc ) ) {
+				if ( $theme !== $cur_theme ) {
+					// Not the current theme: prevent a validation error.
+					unset( $value[ $theme ] );
+				}
+				// Current theme: let the validation process trigger an error.
+				continue;
+			}
+			foreach ( $menus_by_loc as $location => &$menus_by_lang ) {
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-nav-menus.php
```diff
@@ -482,6 +482,54 @@ public function test_set_theme_mod_from_manage_locations_tab() {
...
+		$_REQUEST['_wpnonce'] = wp_create_nonce( 'save-menu-locations' );
+
+		$nav_menu_locations = array(
+			$locations[0] => 4, // `primary`.
+			$locations[1] => 'foo', // `primary___fr`, invalid.
+			$locations[2] => null, // `primary___de`.
+		);
+
+		set_theme_mod( 'nav_menu_locations', $nav_menu_locations );
+		$this->assertSame( array(), $pll_admin->options['nav_menus'][ $cur_theme ], 'The menus should be empty and not be updated.' );
```
Hug0-Drelon, line 514: `$locations[0] => 4,` is valid, right?

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-nav-menus.php
Screenfeed: Yes, but `‘foo’` isn’t.

Hug0-Drelon: Okay, thanks I missed that...
I must say I'm not very fond of checking different cases in a single test but...

Hug0-Drelon (approved):

PR #1664, opened 2025-05-13
Allow to customize the name of the translation term

Screenfeed:
This will be used by WPML to PLL.

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -503,9 +503,17 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 	 * Creates translations groups in mass.
 	 *
 	 * @since 1.6.3
-	 * @since 3.4 Moved from PLL_Admin_Model class.
+	 * @since 3.4 Moved from PLL_Admin_Model class. The `$type` parameter is removed.
+	 * @since 3.8 The name of the translation terms can be customized.
 	 *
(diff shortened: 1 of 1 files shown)


include/translated-object.php
```diff
@@ -518,8 +526,8 @@ public function set_translation_in_mass( $translations ) {
 		$description = array();
 		$count       = array();
 
-		foreach ( $translations as $t ) {
-			$term = uniqid( 'pll_' ); // The term name.
+		foreach ( $translations as $k => $t ) {
+			$term = is_string( $k ) ? "pll_{$k}" : uniqid( 'pll_' ); // The term name.
```
Hug0-Drelon, line 530: Why prefixing it? Can't we assume the calling code have a proper name already?

Hug0-Drelon (changes requested):

include/translated-object.php
Screenfeed: We can assume it has IMHO. I can remove it if you prefer.

Screenfeed: Done in [d41a193](https://github.com/polylang/polylang/pull/1664/commits/d41a1933a8d8f69dc847596d6ad4e0ac549118b9).

Hug0-Drelon (approved):

PR #1665, opened 2025-05-13
Fix warning when querying a non-existing language

### LINKED ISSUE #1661: Requests for unsupported languages generate php warnings
Requests for nonsense language strings produce a PHP warning

```
PHP Warning:  Attempt to read property "slug" on false in /var/www/html/wp-content/plugins/polylang/include/crud-terms.php on line 208
```

(This is happening as a result of an attacker sending many random arguments looking for unexpected behaviour.)

to reproduce

```
curl "http://localhost/?lang=abcd"
````

The following patch will use `_unknown` as the cache key.

```
--- crud-terms.php.dist	2025-05-07 16:19:38.538670990 +0000
+++ crud-terms.php	2025-05-07 16:42:17.193770714 +0000
@@ -206,5 +206,6 @@
 		if ( $lang = $this->get_queried_language( $taxonomies, $args ) ) {
 			$lang = is_string( $lang ) && strpos( $lang, ',' ) ? explode( ',', $lang ) : $lang;
-			$key = '_' . ( is_array( $lang ) ? implode( ',', $lang ) : $this->model->get_language( $lang )->slug );
+			$default_slug = (! empty( $this->model->get_language( $lang ) ) ) ? $this->model->get_language( $lang )->slug : 'unknown';
+			$key = '_' . ( is_array( $lang ) ? implode( ',', $lang ) : $default_slug );
 			$args['cache_domain'] = empty( $args['cache_domain'] ) ? 'pll' . $key : $args['cache_domain'] . $key;
 		}
```

### PR DESCRIPTION
**Pull request by Chouby**
Fix #1661 Although I did not reproduce the issue, it's a valid bug.

The main purpose of the PR is to ensure that we don't attempt to get the slug of an invalid language.
Then, the code is simplified by changing the behavior of `get_queried_language()` - renamed to `get_queried_languages()` - always return an array of `PLL_Language` objects (possibly empty or with a unique item.
Positive side effect: 3 PHPStan errors are fixed.

### DIFF
--- a/include/crud-terms.php
+++ b/include/crud-terms.php
@@ -159,7 +159,7 @@ class PLL_CRUD_Terms {
 	 *
 	 * @param string[] $taxonomies Queried taxonomies.
 	 * @param array    $args       WP_Term_Query arguments.
-	 * @return PLL_Language|string|false The language(s) to use in the filter, false otherwise.
+	 * @return PLL_Language|string|null The language(s) to use in the filter, null otherwise.
 	 */
 	protected function get_queried_language( $taxonomies, $args ) {
 		global $pagenow;
@@ -167,7 +167,7 @@ class PLL_CRUD_Terms {
 		// Does nothing except on taxonomies which are filterable
 		// Since WP 4.7, make sure not to filter wp_get_object_terms()
 		if ( ! $this->model->is_translated_taxonomy( $taxonomies ) || ! empty( $args['object_ids'] ) ) {
-			return false;
+			return null;
 		}
 
 		// If get_terms is queried with a 'lang' parameter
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: include/crud-terms.php [hunk 1]
```diff
@@ -159,28 +159,32 @@ public function save_term( $term_id, $tt_id, $taxonomy ) {
...
+			return array();
 		}
 
-		// If get_terms is queried with a 'lang' parameter
+		// If get_terms() is queried with a 'lang' parameter.
 		if ( isset( $args['lang'] ) ) {
-			return $args['lang'];
+			$languages = is_string( $args['lang'] ) ? explode( ',', $args['lang'] ) : $args['lang'];
+			$languages = is_array( $languages ) ? $languages : array( $languages );
+			return array_filter( array_map( array( $this->model, 'get_language' ), $languages ) );
```
Screenfeed, line 179: This means `$args['lang']` can only be a string, an integer, or an array. Am I right?

Do you think we should trim the slugs?

Also, what do you think of this?

```suggestion
			$languages = is_string( $args['lang'] ) ? explode( ',', $args['lang'] ) : $args['lang'];
			return array_filter( array_map( array( $this->model, 'get_language' ), (array) $languages ) );
```

include/model.php
```diff
@@ -214,14 +214,14 @@ public function get_terms_args( $args ) {
 	 *
 	 * @since 1.2
 	 *
-	 * @param string[]           $clauses The list of sql clauses in terms query.
-	 * @param PLL_Language|false $lang    PLL_Language object.
-	 * @return string[]                   Modified list of clauses.
+	 * @param string[]       $clauses   The list of sql clauses in terms query.
+	 * @param PLL_Language[] $languages PLL_Language object.
```
Screenfeed, line 218: This method is used [in PLL for WC](https://github.com/polylang/polylang-wc/blob/f70ed62313ef8473a98bf556ea7e9bf6e3f4acc5/src/admin/admin-reports.php#L189) and injects a `PLL_Language` object.

```suggestion
	 * @param string[]                    $clauses   The list of sql clauses in terms query.
	 * @param PLL_Language[]|PLL_Language $languages `PLL_Language` object or array of `PLL_Language` objects.
```

Also, since it's public, we should add:

```
	 * @since 3.8 Accepts an object or an array of objects.
```

include/translatable-object.php
```diff
@@ -324,8 +324,6 @@ public function join_clause( $alias = '' ) {
 	 *
 	 * @param PLL_Language|PLL_Language[]|string|string[] $lang A `PLL_Language` object, or a comma separated list of language slugs, or an array of language slugs or objects.
 	 * @return string The WHERE clause.
-	 *
-	 * @phpstan-param PLL_Language|PLL_Language[]|non-empty-string|non-empty-string[] $lang
```
Screenfeed, line 328: Why remove this?

Screenfeed (changes requested):

include/translatable-object.php
Chouby: In fact, it seems to me that the me that the method can work with empty strings.

Screenfeed (changes requested):

include/translatable-object.php
Screenfeed: Indeed.

Screenfeed (approved):

PR #1666, opened 2025-05-14
Reformat Twenty Twenty Four tests

Chouby:
Use the factory for translations, the Polylang context object and pass coding standards.

### DIFF
--- a/tests/phpunit/tests/themes/test-twenty-fourteen.php
+++ b/tests/phpunit/tests/themes/test-twenty-fourteen.php
@@ -2,7 +2,6 @@
 
 class Twenty_Fourteen_Test extends PLL_UnitTestCase {
 	protected static $stylesheet;
-	protected static $tags;
 
 	/**
 	 * @param PLL_UnitTest_Factory $factory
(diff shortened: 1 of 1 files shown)


tests/phpunit/tests/themes/test-twenty-fourteen.php
```diff
@@ -42,52 +33,62 @@ public static function wpTearDownAfterClass() {
...
+				'taxonomy' => 'post_tag',
+				'name'     => 'featured',
+				'lang'     => 'en',
+			),
+			array(
+				'taxonomy' => 'post_tag',
+				'name'     => 'en avant',
+				'lang'     => 'fr',
+			)
+		);
```
Hug0-Drelon, line 87: You can leverage dedicated factory  and use `tag` property directly.

```suggestion
		self::$tags = self::factory()->tag->create_translated(
			array(
				'name'     => 'featured',
				'lang'     => 'en',
			),
			array(
				'name'     => 'en avant',
				'lang'     => 'fr',
			)
		);
```

tests/phpunit/tests/themes/test-twenty-fourteen.php
```diff
@@ -2,23 +2,18 @@
 
 class Twenty_Fourteen_Test extends PLL_UnitTestCase {
 	protected static $stylesheet;
-	protected static $tag_en;
-	protected static $tag_fr;
+	protected static $tags;
```
Hug0-Drelon, line 5: No sure this property is necessary. A local variable should be fine.
Indeed, I don't see it populated in a static method such as `pllSetUpBeforeClass()` (or maybe I'm missing something).

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

PR #1668, opened 2025-05-15
Remove useless `$post_type` param

Chouby:
We introduced the method `get_data_from_new_post_translation_request()` in v3.7 with a `$post_type` but this param is useless. Indeed the method doesn't return anything if `$post_type !== $_GET['post_type']` so we can access the post type directly from `$_GET['post_type']`. However instead of using `$_GET['post_type']`, I chose to use `$GLOBALS['post_type']` which is more convenient as we don't need to redo the validation done by WP (See https://github.com/WordPress/WordPress/blob/6.8.1/wp-admin/post-new.php#L17-L25). This required to adapt some tests where `$GLOBALS['post_type']` was not defined.

### DIFF
--- a/admin/admin-links.php
+++ b/admin/admin-links.php
@@ -242,7 +242,7 @@ class PLL_Admin_Links extends PLL_Links {
 			return $this->get_data_from_new_media_translation_request();
 		}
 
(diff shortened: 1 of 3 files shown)


Screenfeed (approved):

Chouby:
Yes. But it shouldn't have an impact here.

PR #1677, opened 2025-06-03
Document options type in class properties

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2648.

This PR properly documents all the `$options` class properties as an instance of `Options` instead of an array.

This PR also removes all references like `$this->options = &$options;` and adds a lot of type-hints.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -10,6 +10,11 @@
  */
 abstract class PLL_Admin_Base extends PLL_Base {
 	/**
+	 * @var PLL_Admin_Model
+	 */
+	public $model;
+
+	/**
 	 * Current language (used to filter the content).
 	 *
 	 * @var PLL_Language|null
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -10,11 +10,6 @@
  */
 abstract class PLL_Admin_Base extends PLL_Base {
 	/**
-	 * @var PLL_Admin_Model
-	 */
-	public $model;
-
-	/**
 	 * Current language (used to filter the content).
 	 *
 	 * @var PLL_Language|null
--- a/frontend/canonical.php
+++ b/frontend/canonical.php
@@ -33,7 +33,7 @@ class PLL_Canonical {
 	/**
 	 * Current language.
 	 *
-	 * @var PLL_Language
+	 * @var PLL_Language|null
 	 */
 	protected $curlang;
 
@@ -42,12 +42,12 @@ class PLL_Canonical {
 	 *
 	 * @since 3.3
 	 *
-	 * @param object $polylang Main Polylang object.
+	 * @param PLL_Frontend $polylang Main Polylang object.
 	 */
-	public function __construct( &$polylang ) {
+	public function __construct( PLL_Frontend &$polylang ) {
 		$this->links_model = &$polylang->links_model;
 		$this->model       = &$polylang->model;
-		$this->options     = &$polylang->options;
+		$this->options     = $polylang->options;
 		$this->curlang     = &$polylang->curlang;
 	}
 
@@ -126,6 +126,9 @@ class PLL_Canonical {
 		}
 
 		if ( empty( $language ) ) {
+			if ( empty( $this->curlang ) ) {
+				return $requested_url;
+			}
 			$language = $this->curlang;
 			$redirect_url = $requested_url;
 		} else {
--- a/frontend/canonical.php
+++ b/frontend/canonical.php
@@ -126,10 +126,8 @@ class PLL_Canonical {
 		}
 
 		if ( empty( $language ) ) {
-			if ( empty( $this->curlang ) ) {
-				return $requested_url;
-			}
-			$language = $this->curlang;
+			/** @var PLL_Language $language */
+			$language     = $this->curlang;
 			$redirect_url = $requested_url;
 		} else {
 			$redirect_url = $this->redirect_canonical( $requested_url, $language );
--- a/frontend/canonical.php
+++ b/frontend/canonical.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+use WP_Syntex\Polylang\Options\Options;
+
 /**
  * Manages canonical redirect on frontend.
  *
@@ -12,7 +14,7 @@ class PLL_Canonical {
 	/**
 	 * Stores the plugin options.
 	 *
-	 * @var array
+	 * @var Options
 	 */
 	protected $options;
 
--- a/include/language-factory.php
+++ b/include/language-factory.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+use WP_Syntex\Polylang\Options\Options;
+
 /**
  * PLL_Language factory.
  *
@@ -23,7 +25,7 @@ class PLL_Language_Factory {
 	/**
 	 * Polylang's options.
 	 *
-	 * @var array
+	 * @var Options
 	 */
 	private $options;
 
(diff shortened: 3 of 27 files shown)


szepeviktor:
https://github.com/polylang/polylang/blob/9beb167295e57b68c3466f5653fd9311d11a8572/include/Options/Options.php#L29

include/language-factory.php
```diff
@@ -23,19 +25,20 @@ class PLL_Language_Factory {
...
 	 * @since 3.4
+	 * @since 3.7 The `$options` parameter is an instance of `Options`.
 	 *
-	 * @param array $options Array of Poylang's options passed by reference.
+	 * @param Options $options Poylang's options passed by reference.
 	 * @return void
 	 */
-	public function __construct( &$options ) {
+	public function __construct( Options &$options ) {
 		$this->options = &$options;
```
Hug0-Drelon: We should remove reference now we deal with an object.

install/upgrade.php
```diff
@@ -14,18 +14,19 @@ class PLL_Upgrade {
...
 	 *
 	 * @since 1.2
+	 * @since 3.7 The `$options` parameter is an instance of `Options`.
 	 *
-	 * @param array $options Polylang options
+	 * @param Options $options Polylang options.
 	 */
-	public function __construct( &$options ) {
+	public function __construct( Options &$options ) {
 		$this->options = &$options;
```
Hug0-Drelon: Same remark as before.

Hug0-Drelon (changes requested):
We should remove all references to `$options` in constructor now we deal with objects.
i.e. use `$this->options = $options` instead of `$this->options = &$options`.

include/language-factory.php
Screenfeed: Done in [789ce67](https://github.com/polylang/polylang/pull/1677/commits/789ce6720d145b86477af78b200db04362685d23).

install/upgrade.php
Screenfeed: Done in [789ce67](https://github.com/polylang/polylang/pull/1677/commits/789ce6720d145b86477af78b200db04362685d23).

Hug0-Drelon (changes requested):
`$this->options = &$options` remains in pretty much all modified classes constructors.

admin/admin-base.php
```diff
@@ -9,6 +9,11 @@
  * @since 1.8
  */
 abstract class PLL_Admin_Base extends PLL_Base {
+	/**
+	 * @var PLL_Admin_Model
+	 */
+	public $model;
```
Hug0-Drelon: I thought there is no such thing as `PLL_Admin_Model` anymore.
It'd be better to use `PLL_Model` instead (which is already in `PLL_Base`).

szepeviktor:
@Chouby Slevomat has `DisallowReferencePassingSniff` - please consider weeding out pass-by-refs.

frontend/canonical.php
```diff
@@ -40,12 +42,12 @@ class PLL_Canonical {
 	 *
 	 * @since 3.3
 	 *
-	 * @param object $polylang Main Polylang object.
+	 * @param PLL_Frontend $polylang Main Polylang object.
 	 */
-	public function __construct( &$polylang ) {
+	public function __construct( PLL_Frontend &$polylang ) {
 		$this->links_model = &$polylang->links_model;
 		$this->model       = &$polylang->model;
```
Hug0-Drelon, line 49:
```suggestion
		$this->links_model = $polylang->links_model;
		$this->model       = $polylang->model;
```

Screenfeed:
@szepeviktor Unfortunately we still have a lot of things passed by reference. [For example](https://github.com/polylang/polylang/blob/22ed5939736f7510bbc43bd43e2f4052364310d7/include/filters.php#L50-L53).

Hug0-Drelon (changes requested):

Hug0-Drelon (changes requested):
I think references to `$links_model` and `$model` must be removed too, they exists for the same reason as `$options` references.

The only objects that are truely used as references are `$curlang` and other specific classes instantiated after `pll_init`.

Screenfeed:
This goes well beyond the original purpose of this PR.

Hug0-Drelon:
> This goes well beyond the original purpose of this PR.

As you see fit, but next PR will change the same files at the "almost" same lines.

admin/admin-base.php
Screenfeed: Done in [65a6d5b](https://github.com/polylang/polylang/pull/1677/commits/65a6d5b16227f08fb00b5b9ecf0a235dab9dceef).

Hug0-Drelon:
> > As you see fit, but next PR will change the same files at the "almost" same lines.
> 
> Searching for `&$` returns 188 results in 90 files. Some of them may be kept, but I expect very few (in `foreach` loops for example). Fred said it's not our priority right now.

I get that this is for another day, but we *should* remove refs from loops too.

frontend/canonical.php
```diff
@@ -124,6 +126,9 @@ public function check_canonical_url( $requested_url = '', $do_redirect = true )
 		}
 
 		if ( empty( $language ) ) {
+			if ( empty( $this->curlang ) ) {
+				return $requested_url;
+			}
```
Chouby: Why?

settings/settings-browser.php
```diff
@@ -21,11 +21,11 @@ class PLL_Settings_Browser extends PLL_Settings_Module {
 	 *
 	 * @since 1.8
 	 *
-	 * @param object $polylang polylang object
+	 * @param PLL_Admin_Base $polylang polylang object
 	 */
-	public function __construct( &$polylang ) {
+	public function __construct( PLL_Admin_Base &$polylang ) {
```
Chouby: Why not `PLL_Settings`?

settings/settings-module.php
```diff
@@ -118,8 +120,8 @@ class PLL_Settings_Module {
 	 *   active_option?: non-falsy-string
 	 * } $args
 	 */
-	public function __construct( &$polylang, $args ) {
-		$this->options     = &$polylang->options;
+	public function __construct( PLL_Admin_Base &$polylang, $args ) {
```
Chouby: Same as above.

Chouby (changes requested):

frontend/canonical.php
Screenfeed: Because `$this->curlang` can be `null`. Should we continue in this case?

settings/settings-browser.php
Screenfeed: I can't remember why but there as a reason, probably related to something I changed or fixed later. It doesn't seem to be an issue anymore.
Done in [7eff273](https://github.com/polylang/polylang/pull/1677/commits/7eff273252c7fe6bd1bea227ee469813ec114cd3).

frontend/canonical.php
Chouby: I haven't a use case in mind where `$this->curlang` could be `null`. However if there is such case, these lines changes the behavior. Today we can reach the filter (with `null` as language). This wouldn't be possible anymore.
I don't see anything wrong with the current behavior (do you?) so I'd prefer to keep it to avoid any potential issue in case someone is exploiting the filter in this case.

Screenfeed: I added a `@var` comment in [f96b947](https://github.com/polylang/polylang/pull/1677/commits/f96b9478b0d6bb56687c3614e22f04f1d4b81bf5). The other solution would be to document the next filter with `@param PLL_Language|null $language The language detected.`.

PR #1678, opened 2025-06-06
Fix persisted options after `Options::reset()`.

Hug0-Drelon:
## What?
When calling `Options::reset()`, options `Options::$modified` state isn't updated.
So, persisting data in database doesn't work (e.g. `Options::save()`.

## Why?
`Abstract_Option::get()` returns a reference of `Abstract_Option::$value` which is compared to `Options::reset()` returned value. The thing is that the call to `Options::reset()` changes the option.

## How?
- Remove referenced returned value from `Abstract_Option::get()`.
- Add a test proving the issue.


include/Options/Abstract_Option.php
```diff
@@ -119,7 +119,7 @@ public function set( $value, Options $options ): bool {
 	 *
 	 * @return mixed
 	 */
-	public function &get() {
+	public function get() {
```
Chouby, line 122: What about https://github.com/polylang/polylang/commit/2938fa03ba1e5f7ba2e5d5f5bf332448d45d3899 ?

Hug0-Drelon: Seems that this PR https://github.com/polylang/polylang/pull/1595 removed the returned references from `Options::get()` (added in the commit you mentioned).
And for `Abstract_Options::get()` (the code I changed), it seems we expect errors anyway. See this [test](https://github.com/polylang/polylang/blob/a57a8ac2945c1d6b9f5470f03a9650c5dd2dc239/tests/phpunit/tests/Options/Options/test-OffsetUnset.php#L18-L46).

PR #1679, opened 2025-06-06
Allow multiple encodings for metas in `wpml-config.xml`

Screenfeed:
Follows https://github.com/polylang/polylang-pro/pull/2655.

This PR improves support of encodings of metas in `wpml-config.xml`. This allows to decode/encode in more formats (not limiting to JSON anymore) and multiple consecutive formats.

### DIFF
--- a/tests/phpunit/data/wpml-config.xml
+++ b/tests/phpunit/data/wpml-config.xml
@@ -8,6 +8,7 @@
 				<custom-field action="translate">custom-description</custom-field>
 				<custom-field action="ignore">date-added</custom-field>
 				<custom-field action="translate" encoding="json">a_json_meta</custom-field>
+				<custom-field action="translate" encoding="json, urlencode, base64">an_encoded_meta</custom-field>
 		</custom-fields>
 		<custom-fields-texts>
 			<key name="custom|nested">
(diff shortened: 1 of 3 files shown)


tests/phpunit/data/wpml-config.xml
```diff
@@ -8,6 +8,7 @@
 				<custom-field action="translate">custom-description</custom-field>
 				<custom-field action="ignore">date-added</custom-field>
 				<custom-field action="translate" encoding="json">a_json_meta</custom-field>
+				<custom-field action="translate" encoding="json, urlencode, base64">an_encoded_meta</custom-field>
```
Hug0-Drelon: Should we add a case with `json,urlencode,base64`? (meaning no spaces, like WordPress does for lists).

Hug0-Drelon (changes requested):

tests/phpunit/data/wpml-config.xml
Screenfeed: In [1160694](https://github.com/polylang/polylang/pull/1679/commits/1160694a8cc306a82ca82f2504c0c65a0d1c518f) I removed the space characters entirely: it makes no sense to test it with them since WPML doesn't support them.

Hug0-Drelon (approved):

PR #1681, opened 2025-06-10
Increases Polylang comments_clauses filter priority

manooweb:
Fixes https://github.com/polylang/polylang-wc/issues/939

Due to this WooCommerce improvement https://github.com/woocommerce/woocommerce/pull/54984

## Why?

Until now the SQL alias `wp_posts_to_exclude_reviews` would be added only on `edit-comments` screen 
See https://github.com/woocommerce/woocommerce/blob/9.8.5/plugins/woocommerce/src/Internal/Admin/ProductReviews/ReviewsUtil.php#L19-L22

Now, it seems, since the mentioned PR, this alias is also used in other contexts like product edit page.

## How?

Increases the Polylang `comments_clauses` filter priority to run it before the WooCommerce one. It makes useless to know `wp_posts_to_exclude_reviews` to add Polylang join.

An alternative could be:
- keep the `comments_clauses` filter priority as is
- check `wp_posts_to_exclude_reviews` alias already exists
- apply it to Polylang join by replacing this [line](https://github.com/polylang/polylang/blob/3.7.2/include/filters.php#L184) by something like this

```php
if ( ! strpos( $clauses['join'], 'AS wp_posts_to_exclude_reviews' ) ) {
	$clauses['join'] .= $this->model->post->join_clause();
} else {
	$clauses['join'] .= $this->model->post->join_clause( 'wp_posts_to_exclude_reviews' );
}
```

### DIFF
--- a/include/filters.php
+++ b/include/filters.php
@@ -57,7 +57,7 @@ class PLL_Filters {
 
 		// Filters the comments according to the current language
 		add_action( 'parse_comment_query', array( $this, 'parse_comment_query' ) );
-		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 10, 2 );
+		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 9, 2 );
 
 		// Filters the get_pages function according to the current language
 		if ( version_compare( $wp_version, '6.3-alpha', '<' ) ) {


include/filters.php
```diff
@@ -57,7 +57,7 @@ public function __construct( &$polylang ) {
 
 		// Filters the comments according to the current language
 		add_action( 'parse_comment_query', array( $this, 'parse_comment_query' ) );
-		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 10, 2 );
+		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 9, 2 );
```
Marianne380: Maybe add a comment to explain here why.

manooweb: Done in 5339d0f6e61ae1aea5f89ed68cd6d95b9074730b

Marianne380: And what do you think about 5 (to maybe leave space for other parts) ?
```suggestion
		add_filter( 'comments_clauses', array( $this, 'comments_clauses' ), 5, 2 );
```

Marianne380 (approved):

Chouby:
I propose to improve our test determining if we can add our join clause or not.

manooweb:
> Can you add a test?

Done in 0ad3540d10ac2516b951975a0ca862538d30daa0

tests/phpunit/tests/test-filters.php
```diff
@@ -363,4 +363,37 @@ public function test_get_terms_inside_query() {
...
+			function ( $clauses ) use ( $wpdb ) {
+				$clauses['join'] .= " LEFT JOIN {$wpdb->posts} AS wp_posts_to_exclude_reviews ON comment_post_ID = wp_posts_to_exclude_reviews.ID ";
+
+				return $clauses;
+			}
+		);
+
+		$this->frontend->filters = new PLL_Filters( $this->frontend );
+
+		$this->assertSame(
```
Marianne380: Use `assertCount()` ?

Marianne380 (changes requested):

tests/phpunit/tests/test-filters.php
```diff
@@ -363,4 +363,37 @@ public function test_get_terms_inside_query() {
...
+			0,
+			get_comments(
+				array(
+					'count' => true,
+					'lang'  => 'en',
+				)
+			)
+		);
```
Chouby: This assertion doesn't seem necessary.
```suggestion
		get_comments( array( 'lang'  => 'en' ) );
```

tests/phpunit/tests/test-filters.php
```diff
@@ -363,4 +363,37 @@ public function test_get_terms_inside_query() {
...
 		get_posts(); // fires the action and performs the assertions.
 	}
+
+	/**
+	 * @ticket #1682 and Polylang for WooCommerce #939
+	 *
+	 * @see https://github.com/polylang/polylang/issues/1682
+	 * @See https://github.com/polylang/polylang-wc/issues/939
+	 */
+	public function test_comments_clauses() {
```
Chouby:
```suggestion
	public function test_db_error_in_comments_clauses() {
```

tests/phpunit/tests/test-filters.php
manooweb: No! In this case `getComments()` returns an integer not an iterable.

tests/phpunit/tests/test-filters.php
```diff
@@ -363,4 +363,37 @@ public function test_get_terms_inside_query() {
 		add_action( 'pre_get_posts', array( $this, '_action_pre_get_posts' ) );
 		get_posts(); // fires the action and performs the assertions.
 	}
+
+	/**
+	 * @ticket #1682 and Polylang for WooCommerce #939
+	 *
+	 * @see https://github.com/polylang/polylang/issues/1682
+	 * @See https://github.com/polylang/polylang-wc/issues/939
```
Chouby: Let's use the standard format:
```suggestion
	 * @ticket #1682 {@see https://github.com/polylang/polylang/issues/1682}
	 * @ticket #939 {@See https://github.com/polylang/polylang-wc/issues/939}
```

tests/phpunit/tests/test-filters.php
```diff
@@ -363,4 +363,37 @@ public function test_get_terms_inside_query() {
...
+
+		// Simulates WooCommerce adds a join comments clause.
+		add_filter(
+			'comments_clauses',
+			function ( $clauses ) use ( $wpdb ) {
+				$clauses['join'] .= " LEFT JOIN {$wpdb->posts} AS wp_posts_to_exclude_reviews ON comment_post_ID = wp_posts_to_exclude_reviews.ID ";
+
+				return $clauses;
+			}
+		);
```
Chouby: I'd like that we test 2 priorities before and after we add our filter.

Chouby (changes requested):

tests/phpunit/tests/test-filters.php
manooweb: Done in 6c72c0e8bfa5ad5d86369297b3395c41ecf1448d since https://github.com/polylang/polylang/pull/1681#discussion_r2144655231

tests/phpunit/tests/test-filters.php
manooweb: Done in 6c72c0e8bfa5ad5d86369297b3395c41ecf1448d

tests/phpunit/tests/test-filters.php
manooweb: Done in 6c72c0e8bfa5ad5d86369297b3395c41ecf1448d

tests/phpunit/tests/test-filters.php
manooweb: Done in 6c72c0e8bfa5ad5d86369297b3395c41ecf1448d

tests/phpunit/tests/test-filters.php
manooweb: Done in 6c72c0e8bfa5ad5d86369297b3395c41ecf1448d

PR #1683, opened 2025-06-13
Support encoding for block attributes in `wpml-config.xml`

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2612.

Requires #1679 and https://github.com/polylang/polylang-pro/pull/2660.

This PR allows to decode block attributes via the `wpml-config.xml` file.

How to:

```xml
<wpml-config>
	<gutenberg-blocks>
		<gutenberg-block type="my-plugin/my-block" translate="1">
			<key name="my-attribute" encoding="json">
				<key name="text" />
			</key>
		</gutenberg-block>
	</gutenberg-blocks>
</wpml-config>
```

> [!IMPORTANT]
> 1. The `encoding` attribute is supported only on the first level of `<key>` (like WPML).
> 2. The value of `json` means "a URL-encoded JSON": this is a specifity from WPML, while it should be `json,urlencode`. This is different from using the filter `pll_block_attribute_encodings` directly (in the filter we must use `json,urlencode` to get the same result).
> 3. All the other formats (`base64` etc) are not supported in this context.

### DIFF
--- a/.github/workflows/integration-and-unit-tests.yml
+++ b/.github/workflows/integration-and-unit-tests.yml
@@ -10,9 +10,7 @@ on:
       - '**.yml'
       - '**.sh'
   pull_request:
-    branches:
-      - master
-      - wpml-config-multi-encodings
+    branches: master
     paths:
       - '**.php'
       - 'phpunit.xml'
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -4,9 +4,7 @@ on:
   push:
     branches: master
   pull_request:
-    branches:
-      - master
-      - wpml-config-multi-encodings
+    branches: master
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -4,9 +4,7 @@ on:
   push:
     branches: master
   pull_request:
-    branches:
-      - master
-      - wpml-config-multi-encodings
+    branches: master
   workflow_dispatch:
 
 jobs:


modules/wpml/wpml-config.php
```diff
@@ -14,9 +14,12 @@
  *     non-falsy-string,
  *     array{
  *         action: string,
- *         encoding: 'json'|''
+ *         encoding: string
  *     }
  * >
+ * @phpstan-type BlockXPath    array<non-falsy-string, list<non-empty-string>>
+ * @phpstan-type BlockKey      array<non-falsy-string, array<array|true>>
```
Hug0-Drelon, line 21:
```suggestion
 * @phpstan-type BlockKeys      array<non-falsy-string, array<array|true>>
```
Plural if an array. Also, it could be `1` instead of `true`.

modules/wpml/wpml-config.php
```diff
@@ -508,6 +533,19 @@ protected function extract_blocks_parsing_rules() {
...
 								$parsing_rules['key'][ $block_name ] = $rule;
 							}
+
+							$encoding = trim( $this->get_field_attribute( $child, 'encoding' ), ', ' );
+
+							if ( empty( $encoding ) ) {
+								break;
+							}
+
+							if ( 'json' === $encoding ) {
```
Hug0-Drelon: Since `json` is the only supported encoding in the `wpml-config.xml` file, we should bail out early and not add unknown encoding IMHO.

modules/wpml/wpml-config.php
```diff
@@ -508,6 +533,19 @@ protected function extract_blocks_parsing_rules() {
 							} else {
 								$parsing_rules['key'][ $block_name ] = $rule;
 							}
+
+							$encoding = trim( $this->get_field_attribute( $child, 'encoding' ), ', ' );
```
Hug0-Drelon:
```suggestion
							$encoding = trim( $this->get_field_attribute( $child, 'encoding' ), ',' );
```
No spaces.

modules/wpml/wpml-config.php
```diff
@@ -1029,14 +1067,10 @@ private function parse_xml_metas( string $xpath ): array {
 					continue;
 				}
 
-				$data = array(
+				$this->parsed_metas[ $xpath ][ $name ] = array(
 					'action'   => $this->get_field_attribute( $custom_field, 'action' ),
-					'encoding' => $this->get_field_attribute( $custom_field, 'encoding' ),
+					'encoding' => trim( $this->get_field_attribute( $custom_field, 'encoding' ), ', ' ),
```
Hug0-Drelon:
```suggestion
					'encoding' => trim( $this->get_field_attribute( $custom_field, 'encoding' ), ',' ),
```
No spaces.

modules/wpml/wpml-config.php
Screenfeed: I used singular to match the array keys of the property `$parsing_rules`:

```
	/**
	 * @phpstan-var array{
	 *     xpath?: BlockXPath,
	 *     key?: BlockKey,
	 *     encoding?: BlockEncoding
	 * }|null
	 */
```

Screenfeed: I don't think it can be `1`, because of [this line](https://github.com/polylang/polylang/blob/8fa07752accbd73b785281ae524533445099da6e/modules/wpml/wpml-config.php#L671). Do I miss something?

modules/wpml/wpml-config.php
Screenfeed: It's the only format supported by WPML. While I can't imagine any other format being useful here, I don't think we should restrain to `json`, who knows what 3rd party developers can think of.
We can do it and it doesn’t cost anything. What do you think?

modules/wpml/wpml-config.php
Screenfeed: Removed in [be8b199](https://github.com/polylang/polylang/pull/1683/commits/be8b199de60559dd148750bf5832a09ae2e8cd0b).

modules/wpml/wpml-config.php
Screenfeed: Removed in [be8b199](https://github.com/polylang/polylang/pull/1683/commits/be8b199de60559dd148750bf5832a09ae2e8cd0b).

modules/wpml/wpml-config.php
Hug0-Drelon: Indeed you're right, I'm the one who missed something (the code you're pointing out).

modules/wpml/wpml-config.php
```diff
@@ -508,6 +533,19 @@ protected function extract_blocks_parsing_rules() {
...
+
+							$encoding = $this->get_field_attribute( $child, 'encoding' );
+
+							if ( empty( $encoding ) ) {
+								break;
+							}
+
+							if ( 'json' === $encoding ) {
+								// Compatibility with WPML: for WPML, `json` means `json,urlencode`
+								// (and is the only format supported by WPML in this context).
```
Chouby:
```suggestion
								// For WPML, `json` means `json,urlencode`(and is the only format supported in this context).
```

tests/phpunit/tests/test-wpml-config.php
```diff
@@ -415,6 +415,14 @@ public function test_gutenberg_blocks() {
 				'buttonText' => true,
 			),
 		);
+		$encodings_for_attributes     = array(
+			'my-plugin/my-block'   => array(
+				'other' => 'json,urlencode',
+			),
+			'my-plugin/my-block-7' => array(
+				'other' => 'json,urlencode',
+			),
```
Chouby: Could you add a comment explaining this one will be overriden?

modules/wpml/wpml-config.php
```diff
@@ -138,6 +142,7 @@ public function init() {
 		add_filter( 'pll_term_meta_encodings', array( $this, 'add_term_meta_encodings' ), 20 );
 		add_filter( 'pll_blocks_xpath_rules', array( $this, 'translate_blocks' ) );
 		add_filter( 'pll_blocks_rules_for_attributes', array( $this, 'translate_blocks_attributes' ) );
+		add_filter( 'pll_block_attribute_encodings', array( $this, 'decode_blocks_attributes' ) );
```
Chouby: We are not consistent with all filters. Some have the priority 10, others 20. It would make sense to always use 20 (the idea is that a custom wpml-config can override a filter at the default priority). What do you think?

Chouby (changes requested):

modules/wpml/wpml-config.php
Screenfeed: Without thinking specifically to hook priority, that's a question I asked myself at some point indeed: should the file overwrite a filter hook, or the opposite? Now I have the answer.

Screenfeed: Done in [75d4d1d](https://github.com/polylang/polylang/pull/1683/commits/75d4d1d636455e837330caa1a2bfa2df3ca103ae).

modules/wpml/wpml-config.php
Screenfeed: Done in [5368de4](https://github.com/polylang/polylang/pull/1683/commits/5368de4159eaddc2280ca90083217fbee1e2e4f3).

modules/wpml/wpml-config.php
Screenfeed: Restricted to JSON in [5368de4](https://github.com/polylang/polylang/pull/1683/commits/5368de4159eaddc2280ca90083217fbee1e2e4f3), so we are full compatible with WPML.

tests/phpunit/tests/test-wpml-config.php
Screenfeed: Done in [8761bb3](https://github.com/polylang/polylang/pull/1683/commits/8761bb335528fdb15024d67757829a2d02f8c67e).

PR #1684, opened 2025-06-16
Fix some regression in WPML API.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2661
Enhance WPML API facade.

## Why?
Some changes from https://github.com/polylang/polylang/pull/1559/ were introducing regressions regarding WPML API behavior.

## How?
- `icl_object_id()` does honors `$return_original_if_missing` if the taxonomy or post type is not translated as well as if no language has been found. See `SitePress::get_object_id()`.
- `wpml_get_language_information()` returns `WP_Error` when something goes wrong and check `$post_id` is `null` (and not just empty, which slightly changes the behavior).
- `wpml_get_default_language()` and `icl_get_default_language()` return `false` when no language is found. See `SitePress::get_default_language()`.

### DIFF
--- a/modules/wpml/wpml-legacy-api.php
+++ b/modules/wpml/wpml-legacy-api.php
@@ -187,7 +187,7 @@ if ( ! function_exists( 'icl_object_id' ) ) {
 		}
 
 		if ( empty( $ulanguage_code ) ) {
-			return null;
+			return $return_original_if_missing ? $element_id : null;
 		}
 
 		if ( 'nav_menu' === $element_type ) {
@@ -205,10 +205,6 @@ if ( ! function_exists( 'icl_object_id' ) ) {
 			$tr_id = PLL()->model->post->get_translation( $element_id, $ulanguage_code );
 		} elseif ( pll_is_translated_taxonomy( $element_type ) ) {
 			$tr_id = PLL()->model->term->get_translation( $element_id, $ulanguage_code );
-		}
-
-		if ( ! isset( $tr_id ) ) {
-			return $element_id; // WPML doesn't honor $return_original_if_missing if the post type or taxonomy is not translated.
 		}
 
 		if ( empty( $tr_id ) ) {
@@ -247,15 +243,20 @@ if ( ! function_exists( 'wpml_get_language_information' ) ) {
 	 *
 	 * @param null $empty   optional, not used
 	 * @param int  $post_id optional, post id, defaults to current post
-	 * @return array
+	 * @return array|WP_Error
 	 */
 	function wpml_get_language_information( $empty = null, $post_id = null ) {
-		if ( empty( $post_id ) ) {
+		if ( is_null( $post_id ) ) {
 			$post_id = get_the_ID();
 		}
-
 		if ( empty( $post_id ) ) {
-			return array();
+			return new WP_Error( 'missing_id', __( 'Missing post ID', 'polylang' ) );
+		}
+
+		$post = get_post( $post_id );
+		if ( empty( $post ) ) {
+			// translators: Post id.
+			return new WP_Error( 'missing_post', sprintf( __( 'No such post for ID = %d', 'polylang' ), $post_id ) );
 		}
 
 		$lang = PLL()->model->post->get_language( $post_id );
@@ -392,10 +393,10 @@ if ( ! function_exists( 'icl_get_default_language' ) ) {
 	 *
 	 * @since 1.0.5
 	 *
-	 * @return string default language code
+	 * @return string|false default language code
 	 */
 	function icl_get_default_language() {
-		return (string) pll_default_language();
+		return pll_default_language();
 	}
 }
 
@@ -407,10 +408,10 @@ if ( ! function_exists( 'wpml_get_default_language' ) ) {
 	 *
 	 * @since 1.8.2
 	 *
-	 * @return string default language code
+	 * @return string|false default language code
 	 */
 	function wpml_get_default_language() {
-		return (string) pll_default_language();
+		return pll_default_language();
 	}
 }
 
--- a/modules/wpml/wpml-legacy-api.php
+++ b/modules/wpml/wpml-legacy-api.php
@@ -209,7 +209,15 @@ if ( ! function_exists( 'icl_object_id' ) ) {
 			return $element_id; // WPML doesn't honor $return_original_if_missing if the post type or taxonomy is not translated, @see {SitePress::get_object_id()}.
 		}
 
-		return $return_original_if_missing && empty( $tr_id ) ? $element_id : (int) $tr_id;
+		if ( empty( $tr_id ) ) {
+			if ( $return_original_if_missing ) {
+				return $element_id;
+			}
+
+			return null;
+		}
+
+		return (int) $tr_id;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


modules/wpml/wpml-legacy-api.php
```diff
@@ -207,10 +207,6 @@ function icl_object_id( $element_id, $element_type = 'post', $return_original_if
 			$tr_id = PLL()->model->term->get_translation( $element_id, $ulanguage_code );
 		}
 
-		if ( ! isset( $tr_id ) ) {
-			return $element_id; // WPML doesn't honor $return_original_if_missing if the post type or taxonomy is not translated.
-		}
-
```
Chouby: For me this is still true. See `SitePress::get_object_id()`.

modules/wpml/wpml-legacy-api.php
```diff
@@ -247,15 +243,20 @@ function wpml_object_id_filter( $id, $type = 'post', $return_original_if_missing
 	 *
 	 * @param null $empty   optional, not used
 	 * @param int  $post_id optional, post id, defaults to current post
-	 * @return array
+	 * @return array|WP_Error
 	 */
 	function wpml_get_language_information( $empty = null, $post_id = null ) {
-		if ( empty( $post_id ) ) {
+		if ( is_null( $post_id ) ) {
```
Chouby:
```suggestion
		if ( null === $post_id ) {
```

modules/wpml/wpml-legacy-api.php
Hug0-Drelon: Indeed, I don't know what I was thinking...
Fixed in 5748251c and 3e03404f.
I've also honor `$return_original_if_missing` in the returned value for other cases.

modules/wpml/wpml-legacy-api.php
Hug0-Drelon: I'm not sure this is something to care about. See https://onlinephp.io/c/e1fd8.
It was an issue for old PHP versions, not really now.

Chouby: It wasn't for a matter of performance (for one call it's not something I would look to). It's more a matter of some sort of personal coding standard :)

Hug0-Drelon: Oh I see, I'll apply it then.

Hug0-Drelon:
> WE already have [some tests](https://github.com/polylang/polylang/blob/3.7.3/tests/phpunit/tests/test-wpml.php) for this API, but visibly quite incomplete. Could you add more to cover the cases you fixed?

Done in 25c0124221ddb9356d6e17b68846776b6be564e3 and c73e2aa0ec8b8109d91119c478136a121d731b26.

tests/phpunit/tests/test-wpml.php
```diff
@@ -278,6 +291,111 @@ public function test_wpml_object_id() {
...
+	 * @param bool   $return_original_if_missing Whether to return the original id if the translation is missing.
+	 *
+	 * @return void
+	 */
+	public function test_wpml_object_id( $object_kind, $object_type, $curlang, $lang, $return_original_if_missing ) {
+		if ( 'something' === $object_type ) {
+			if ( 'post' === $object_kind ) {
+				register_post_type( 'something' );
+			} else {
+				register_taxonomy( 'something', 'post' );
```
Chouby: Useless if I correctly interpret the `@testWith`.

tests/phpunit/tests/test-wpml.php
```diff
@@ -278,6 +291,111 @@ public function test_wpml_object_id() {
...
+	 * @param string $object_kind                The kind of object to test.
+	 * @param string $object_type                The type of object to test.
+	 * @param string $curlang                    The current language, empty if none.
+	 * @param string $lang                       The language to test, empty if none.
+	 * @param bool   $return_original_if_missing Whether to return the original id if the translation is missing.
+	 *
+	 * @return void
+	 */
+	public function test_wpml_object_id( $object_kind, $object_type, $curlang, $lang, $return_original_if_missing ) {
+		if ( 'something' === $object_type ) {
```
Chouby: I suggest to rename to `untranslated_cpt` for clarification.

tests/phpunit/tests/test-wpml.php
```diff
@@ -278,6 +291,111 @@ public function test_wpml_object_id() {
...
+		$frontend->curlang = empty( $curlang ) ? null : self::$model->get_language( $curlang );
+
+		$args = array(
+			'lang' => 'en',
+		);
+		if ( 'post' === $object_kind ) {
+			$args['post_type'] = $object_type;
+		} else {
+			$args['taxonomy'] = $object_type;
+		}
```
Chouby, line 314: I propose to remove the first param to simplify the readability of the `@testWith` and to change this to:
```suggestion
		if ( post_type_exists( $object_type ) ) {
			$object_kind       = 'post';
			$args['post_type'] = $object_type;
		} else {
			$object_kind      = 'term';
			$args['taxonomy'] = $object_type;
		}
```

tests/phpunit/tests/test-wpml.php
```diff
@@ -278,6 +291,111 @@ public function test_wpml_object_id() {
...
+
+		$args = array(
+			'lang' => $lang,
+		);
+		if ( 'post' === $object_kind ) {
+			$args['post_type'] = $object_type;
+		} else {
+			$args['taxonomy'] = $object_type;
+		}
```
Chouby: Duplicate from above.

tests/phpunit/tests/test-wpml.php
```diff
@@ -278,6 +291,111 @@ public function test_wpml_object_id() {
...
+		$translated_object_id = self::factory()->$object_kind->create( $args );
+		self::$model->$object_kind->save_translations(
+			$translated_object_id,
+			array(
+				'en'  => $object_id,
+				$lang => $translated_object_id,
+			)
+		);
```
Chouby: Would this make sense for more clarity (even if we create a second object in English)?
```suggestion
		list( $object_id, $translated_object_id ) = self::factory()->$object_kind->create_translated(
			array_merge( $args, array( 'lang' => 'en' ) ),
			array_merge( $args, array( 'lang' => $lang ) ),
		);
```

tests/phpunit/tests/test-wpml.php
Hug0-Drelon: I'm not very fond of it. Yes this shorten the test signature, but this is only a test so no code smell here. And for the test code itself is complicated it.
We now have to manually set `$object_kind` variable since it's used bellow.

But if you really prefer that what, I'll let you see fit.

tests/phpunit/tests/test-wpml.php
Hug0-Drelon: Sure, done in 572c0b42 with a twist. Indeed, `list()` is for indexed arrays not associative like we get here.

tests/phpunit/tests/test-wpml.php
Hug0-Drelon: Removed in 572c0b42.

tests/phpunit/tests/test-wpml.php
Hug0-Drelon: Renamed in 708fbeb794203733f3e422c1ef0ae8e702730fc5.

tests/phpunit/tests/test-wpml.php
Hug0-Drelon: Indeed, removed in 455e8e8fc4275efaba8d5a19a552cb05e371a53f.

PR #1686, opened 2025-06-19
Filter discovery links on frontend.

Hug0-Drelon:
## What?
Since an issue raised by a user ([HelpScout](https://secure.helpscout.net/conversation/2972885260/34900?viewId=860976)), it appears that discovery links are not filtered by the current language of the page they're displayed in.
Indeed, Yoast allows to add (translatable) data such as the site tagline to post title rendered in discovery links (oEmbed and REST alternate link). This seems to work for most of social networks because they use `<title/>` tag in `<head/>`. Although there is an issue with LinkedIn which uses oEmbed alternate links to generate preview.

## Why?
Even though these links point to specific content that has a language, external data rendered through these endpoints are not translated (no current language defined in REST unless a `lang` parameter is sent).

## How?
- Filter `oembed_endpoint_url` for oEmbed enpoint URLs.
- ~~Filter `rest_output_link_wp_head()` using `rest_url` filter and a toggle on `wp_head` (no filter in rest_output_link_wp_head()`...).~~ We chose not to, because it's not linked to the primary issue and we want to clarify how the REST API should behave regarding translatable strings before adding any parameters here.
- Add tests.

### DIFF
--- a/frontend/frontend-filters-links.php
+++ b/frontend/frontend-filters-links.php
@@ -74,7 +74,7 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 			add_filter( 'admin_url', array( $this, 'admin_url' ), 10, 2 );
 		}
 
-		add_filter( 'oembed_endpoint_url', array( $this, 'add_current_language_to_oembed_endpoint' ) );
+		add_filter( 'oembed_endpoint_url', array( $this, 'add_current_language_url_query' ) );
 	}
 
 	/**
@@ -352,23 +352,18 @@ class PLL_Frontend_Filters_Links extends PLL_Filters_Links {
 	}
 
 	/**
-	 * Adds the current language to the oEmbed endpoint URL.
+	 * Adds the current language to URL query.
 	 *
 	 * @since 3.7.4
 	 *
 	 * @param string $url The oEmbed endpoint URL.
 	 * @return string The oEmbed endpoint URL with the language.
 	 */
-	public function add_current_language_to_oembed_endpoint( $url ) {
+	public function add_current_language_url_query( $url ) {
 		if ( empty( $this->curlang ) ) {
 			return $url;
 		}
 
-		return add_query_arg(
-			array(
-				'lang' => $this->curlang->slug,
-			),
-			$url
-		);
+		return add_query_arg( 'lang', $this->curlang->slug, $url );
 	}
 }


frontend/frontend-filters-links.php
```diff
@@ -73,6 +73,11 @@ public function __construct( &$polylang ) {
 			// Rewrites ajax url
 			add_filter( 'admin_url', array( $this, 'admin_url' ), 10, 2 );
 		}
+
+		add_filter( 'oembed_endpoint_url', array( $this, 'add_current_language_to_oembed_endpoint' ) );
```
Screenfeed: The only place `get_oembed_endpoint_url()` is used, is in `wp_oembed_add_discovery_links()`, which is hooked to `wp_head` at priority 10.

Which means this filter is currently useless because this is covered by `toggle_rest_output_link_filter()`. However I would keep this hook because plugins or themes (or WP in the future) may use `get_oembed_endpoint_url()` elsewhere.

tests/phpunit/tests/Frontend/test-Discovery_Links.php
```diff
@@ -0,0 +1,88 @@
...
+
+		$expected = '<link rel="alternate" title="oEmbed (JSON)" type="application/json+oembed" href="http://example.org/wp-json/oembed/1.0/embed?url=http%3A%2F%2Fexample.org%2Ffr%2Ftest%2F&#038;lang=fr" /><link rel="alternate" title="oEmbed (XML)" type="text/xml+oembed" href="http://example.org/wp-json/oembed/1.0/embed?url=http%3A%2F%2Fexample.org%2Ffr%2Ftest%2F&#038;format=xml&#038;lang=fr" />';
+
+		if ( version_compare( $GLOBALS['wp_version'], '6.6.0', '<' ) ) {
+			// Title attribute added in 6.6.0, @see {https://core.trac.wordpress.org/ticket/59006/}.
+			$expected = '<link rel="alternate" type="application/json+oembed" href="http://example.org/wp-json/oembed/1.0/embed?url=http%3A%2F%2Fexample.org%2Ffr%2Ftest%2F&#038;lang=fr" /><link rel="alternate" type="text/xml+oembed" href="http://example.org/wp-json/oembed/1.0/embed?url=http%3A%2F%2Fexample.org%2Ffr%2Ftest%2F&#038;format=xml&#038;lang=fr" />';
+		}
+
+		$this->assertSame( $expected, str_replace( "\n", '', get_echo( 'wp_oembed_add_discovery_links' ) ) );
+		$this->assertSame( $expected, str_replace( "\n", '', get_echo( 'wp_oembed_add_discovery_links' ) ) );
```
Screenfeed: Double? 🤔 

```suggestion
		$this->assertSame( $expected, str_replace( "\n", '', get_echo( 'wp_oembed_add_discovery_links' ) ) );
```

Screenfeed (changes requested):

frontend/frontend-filters-links.php
```diff
@@ -348,4 +350,25 @@ public function home_url( $url, $path ) {
...
+		if ( empty( $this->curlang ) ) {
+			return $url;
+		}
+
+		return add_query_arg(
+			array(
+				'lang' => $this->curlang->slug,
+			),
+			$url
+		);
```
Chouby: You can use this way when there's only one key to add:
```suggestion
		return add_query_arg( 'lang', $this->curlang->slug, $url );
```

frontend/frontend-filters-links.php
```diff
@@ -348,4 +350,25 @@ public function home_url( $url, $path ) {
...
+
+	/**
+	 * Adds the current language to the oEmbed endpoint URL.
+	 *
+	 * @since 3.7.4
+	 *
+	 * @param string $url The oEmbed endpoint URL.
+	 * @return string The oEmbed endpoint URL with the language.
+	 */
+	public function add_current_language_to_oembed_endpoint( $url ) {
```
Chouby: This method is generic and could be reused for other urls, so I suggest to rename it and adapt the doc accordingly.
Just some ideas: `add_current_language_url_query()`, `add_curlang_query_arg()`

frontend/frontend-filters-links.php
Hug0-Drelon: 👍 Applied in 3b5754cd.

frontend/frontend-filters-links.php
Hug0-Drelon: Done in 3b5754cd.

PR #1690, opened 2025-06-24
Improve `save_translations()` by requiring full array of translations when linking posts across languages.

Marianne380:
## What

This PR updates the `save_translations()` function to require the **full array of translations** when linking posts across languages. Partial translations array is no longer sufficient to maintain existing links.

## Why

This change makes the translation behavior more predictable and explicit.

## Fix

Previously, when creating or updating a post (see the example via the REST API), it was enough to pass only one existing translation in the `translations` parameter to link them together.

**Example (before):**
```php
post_en => 1 
post_fr => 2
```

```http
POST /wp/v2/posts?title=post_de&lang=de&translations[en]=1
```

This would successfully link the new German post to the English post with ID `1`, and preserve the existing translation structure (e.g., if the English post was already linked to a French one).

### Now:

To ensure all translations remain linked correctly, we **must provide the full set of translations** we want to be linked.

**Example (now):**

```http
POST /wp/v2/posts?title=post_de&lang=de&translations[en]=1&translations[fr]=2
```

This will link the new German post with both the English (`ID 1`) and French (`ID 2`).

If we omit existing translations, they will be unlinked.

**Bad example:**

```http
POST /wp/v2/posts?title=post de&lang=de&translations[en]=1
```

This will **only link** the new German post to the English one, and the French post (ID `2`) will be removed from the translation group.

## Notes

* This affects all endpoints using `save_translations()`.

### DIFF
--- a/include/translated-object.php
+++ b/include/translated-object.php
@@ -169,7 +169,7 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 			$this->delete_translation( $id );
 		}
 
-		// Check ID we need to create or update the translation group.
+		// Check if we need to create or update the translation group.
 		if ( ! $this->should_update_translation_group( $id, $translations ) ) {
 			return $translations;
 		}
@@ -184,10 +184,17 @@ abstract class PLL_Translated_Object extends PLL_Translatable_Object {
 		} else {
 			// Take care not to overwrite extra data stored in the description field, if any.
 			$group = (int) $term->term_id;
-			$descr = maybe_unserialize( $term->description );
-			$descr = is_array( $descr ) ? array_diff_key( $descr, $old_translations ) : array(); // Remove old translations.
-			$descr = array_merge( $descr, $translations ); // Add new one.
-			wp_update_term( $group, $this->tax_translations, array( 'description' => maybe_serialize( $descr ) ) );
+			$existing_translations = maybe_unserialize( $term->description );
+			$existing_translations = is_array( $existing_translations ) ? $existing_translations : array();
+
+			// Find and unlink translations that are part of the old group but not the new one.
+			$removed_translations = array_diff_assoc( $existing_translations, $translations );
+			foreach ( $removed_translations as $id ) {
+				$this->delete_translation( $id );
+			}
+
+			// Update the term description with the new, correct set of translations.
+			wp_update_term( $group, $this->tax_translations, array( 'description' => maybe_serialize( $translations ) ) );
 		}
 
 		// Link all translations to the new term.


Marianne380:
When a translation is removed from the metabox, we end up with a translations array that contains `null` for the removed translation. For example:

```php
array( 'en' => 1, 'fr' => null )
```

After calling `validate_translations()`, see https://github.com/polylang/polylang/blob/3.7.3/include/translated-object.php#L163, this array is sanitized and transformed into  :

```php
array( 'en' => 1 )
```

At this point, the removed translation is no longer in the array.

Since `$old_translations` still contains the previous values, such as:

```php
array( 'en' => 1, 'fr' => 2 )
```

The `array_diff_assoc( $old_translations, $translations )`, see https://github.com/polylang/polylang/blob/3.7.3/include/translated-object.php#L168-L170, correctly identifies `'fr' => 2` as a removed translation and triggers its deletion via `delete_translation()`.

Marianne380:
@Chouby, about deletion: 

The first block that removes translations:

```php
// Unlink removed translations.
$old_translations = $this->get_translations( $id );
foreach ( array_diff_assoc( $old_translations, $translations ) as $id ) {
    $this->delete_translation( $id );
}
````

Will remove the old translations that are linked to the current object (which means none for a post that didn't have any translations before).

Then, the second block handles the translation group:
```php
// Find and unlink translations that are part of the old group but not the new one.
$existing_translations = array_filter( $existing_data, 'is_numeric' );
$removed_translations  = array_diff_key( $existing_translations, $translations );
foreach ( $removed_translations as $id ) {
	$this->delete_translation( (int) $id );
}
```
 it compares `$existing_translations` (all translations in the existing group) with `$translations` (the new translations), and removes the translations that were part of the existing group but are no longer part of the new group.

## To summarize:

Both blocks seem necessary because they handle different cases:

* The **first block** removes the old translations of the current object.
* The **second block** removes translations from the existing group that are no longer part of the new group.


## Example : 

### Initial Situation:

Post A (English) ↔ Post B (French)  
Existing group: `{en: A, fr: B}`

Call `save_translations(C, {en: A, de: C})` where C = German post:

1. **First block:**

**Result:** Nothing is deleted (makes sense, C had no translations).

2. **Second block:**

**Result:** The French post B (which was from A) is removed from the group.

---

include/translated-object.php
```diff
@@ -183,11 +183,24 @@ public function save_translations( $id, array $translations = array() ) {
...
+				$this->delete_translation( (int) $id );
+			}
+
+			// Update the term description with the new translations + preserve sync data.
+			$extra_data = array_diff_key( $existing_data, $existing_translations );
+			wp_update_term(
+				$group,
+				$this->tax_translations,
+				array( 'description' => maybe_serialize( array_merge( $extra_data, $translations ) ) )
+			);
```
manooweb: See my comment here https://github.com/polylang/polylang/pull/1690#issuecomment-3200152535

manooweb:
> 
> I finally understood why @Chouby wondered why we need to delete translation twice [here](https://github.com/polylang/polylang/blob/d549911e8cb9f0e5c665afa72040305f948af684/include/translated-object.php#L169) and [here](https://github.com/polylang/polylang/blob/d549911e8cb9f0e5c665afa72040305f948af684/include/translated-object.php#L194)
> 
> Thanks to your explanations above, I was able to debug and I discovered the code is actually buggy [here](https://github.com/polylang/polylang/blob/d549911e8cb9f0e5c665afa72040305f948af684/include/translated-object.php#L168) Of course I didn't see it immediately 🙄💤
> 
> Indeed, here, `$id` overrides its initial value passed to the `save_translations()` method. It makes `should_update_translation_group()` not work properly because it returns that we should update the translation group while it was already updated before by the previous `delete_translation()`. In your first example, it makes the french translation kept while it shouldn't.
> 
> So I think the fix you proposed seems useless if we correct the line I pointed. It was introduced by #1159 ➡️
> 
> https://github.com/polylang/polylang/blob/a54c83b94ee963b521bcb57ac8355d304ac17619/include/translated-object.php#L154

Never mind, it has nothing to see with what you proposed. In fact, it's going to update the term in a case it shouldn't and makes an extra DB access. It breaks nothing.

I was lost because I discovered another issue which could explain why some translations group could be broken in DB.
It come from [here](https://github.com/polylang/polylang/blob/a54c83b94ee963b521bcb57ac8355d304ac17619/include/translated-object.php#L163-L164) because we expect that the translations group should be unique but I succeeded to have 2 translations group for the same language. It's possible since we have more than two languages.

It isn't the point here but it's also because we use `wp_get_object_terms()` with the other translations where the unwanted language is still in its translations group.

So I think, we can check translations on all translations passed to the `save_translations()` method instead of only the current one in order to find the translations to remove.

include/translated-object.php
```diff
@@ -162,9 +162,13 @@ public function save_translations( $id, array $translations = array() ) {
 		// Sanitize and validate the translations array.
 		$translations = $this->validate_translations( $translations, $id );
 
-		// Unlink removed translations.
-		$old_translations = $this->get_translations( $id );
+		// Search for the existing translations in all the translations we want to link to.
```
Marianne380: What do you think of ?
Or do you have another idea to clarify what we are doing?

```suggestion
	/*
         * Gather all existing translations from objects we're about to link together.
         * This is needed to clean up old translation groups when merging them.
        */
```

Marianne380: Also should we do an `array_unique`after to avoid duplicate ? I mean, can it happen already?
If so, it would optimize the diff operation just below.

tests/phpunit/tests/test-translated-post.php
```diff
@@ -92,6 +92,56 @@ public function test_delete_post_translation() {
 		$this->assertSame( 0, self::$model->post->get_translation( $de, 'fr' ) ); // fails
 	}
 
+	public function test_save_translations() {
```
Marianne380, line 107: Could we add doc explaining the usefulness of adding this test?

tests/phpunit/tests/test-translated-post.php
```diff
@@ -92,6 +92,56 @@ public function test_delete_post_translation() {
...
+		$this->assertSame( self::$model->post->get_translation( $posts['de'], 'fr' ), $posts['fr'] );
+		$this->assertSame( self::$model->post->get_translation( $posts['de'], 'en' ), $posts['en'] );
+
+		// Removes the translations from the group by updating the German post.
+		self::$model->post->save_translations( $posts['de'], array() );
+
+		$terms = wp_get_object_terms( $posts, 'post_translations' );
+		$this->assertCount( 1, $terms );
+
+		$this->assertSame( self::$model->post->get_translation( $posts['de'], 'fr' ), 0 );
```
Marianne380: Isn't the assertion reversed? (same for others)

include/translated-object.php
manooweb: Done in 199619c61e07b209a4be5ac0e4a4bd2b27250c2b 
As it's an associative array with language slug as key, `array_merge()` is sufficient, isn't it?

tests/phpunit/tests/test-translated-post.php
manooweb: 🤔 I don't think so.
We unlink translations from `de` [here](https://github.com/polylang/polylang/blob/8d1269170938b8977dbb4a68be3a4378376fa093/tests/phpunit/tests/test-translated-post.php#L119).
We check everything is good.

tests/phpunit/tests/test-translated-post.php
manooweb: Done in 7b5762d437412ea4d8311aab9d1cf6747dfb99e4

manooweb (changes requested):

tests/phpunit/tests/test-translated-post.php
Marianne380: I mean shouldn't it be : 
```php
$this->assertSame( 0, self::$model->post->get_translation( $posts['de'], 'fr' ) );
```

manooweb: 🥱

tests/phpunit/tests/test-translated-term.php
```diff
@@ -145,4 +145,70 @@ public function test_get_db_infos() {
...
+		$this->assertSame( self::$model->term->get_translation( $terms['en'], 'fr' ), $terms['fr'] );
+		$this->assertSame( self::$model->term->get_translation( $terms['en'], 'de' ), $terms['de'] );
+
+		$this->assertSame( self::$model->term->get_translation( $terms['fr'], 'en' ), $terms['en'] );
+		$this->assertSame( self::$model->term->get_translation( $terms['fr'], 'de' ), $terms['de'] );
+
+		$this->assertSame( self::$model->term->get_translation( $terms['de'], 'fr' ), $terms['fr'] );
+		$this->assertSame( self::$model->term->get_translation( $terms['de'], 'en' ), $terms['en'] );
+
+		// Removes the translations from the group by updating the German post.
```
Marianne380:
```suggestion
		// Removes the translations from the group by updating the German term.
```

tests/phpunit/tests/test-translated-term.php
```diff
@@ -145,4 +145,70 @@ public function test_get_db_infos() {
...
+		$this->assertSame( self::$model->term->get_translation( $terms['fr'], 'de' ), 0 );
+
+		$this->assertSame( self::$model->term->get_translation( $terms['de'], 'fr' ), 0 );
+		$this->assertSame( self::$model->term->get_translation( $terms['de'], 'en' ), 0 );
+
+		// Links again the French and English terms.
+		self::$model->term->save_translations( $terms['fr'], array( 'fr' => $terms['fr'], 'en' => $terms['en'] ) );
+
+		$translations_terms = wp_get_object_terms( $terms, 'term_translations' );
+		$this->assertCount( 2, $translations_terms ); // Is correct at this step because the German post isn't translated into either English or French.
```
Marianne380:
```suggestion
		$this->assertCount( 2, $translations_terms ); // Is correct at this step because the German term isn't translated into either English or French.
```

tests/phpunit/tests/test-translated-term.php
```diff
@@ -145,4 +145,70 @@ public function test_get_db_infos() {
...
+
+		$this->assertSame( self::$model->term->get_translation( $terms['de'], 'fr' ), $terms['fr'] );
+		$this->assertSame( self::$model->term->get_translation( $terms['de'], 'en' ), $terms['en'] );
+
+		// Removes the translations from the group by updating the German post.
+		self::$model->term->save_translations( $terms['de'], array() );
+
+		$translations_terms = wp_get_object_terms( $terms, 'term_translations' );
+		$term_id = $translations_terms[0]->term_id;
+		$this->assertCount( 3, $terms );
```
Marianne380, line 189: Shouldn't we check `$translations_terms` here ?

manooweb: I didn't for post. The behavior is a bit different for terms because term has always a translations group and explains why we have 3 terms here at the opposite of post where only one translations group is kept.
The assertions below checks that translations are unlinked from each other.

tests/phpunit/tests/test-translated-post.php
manooweb: Done in 744074b55cd2e4dce747c4276873bdde64597313 only for new tests.

manooweb (changes requested):

tests/phpunit/tests/test-translated-term.php
Marianne380: Then why doing that here : 
```php
$translations_terms = wp_get_object_terms( $terms, 'term_translations' );
$term_id = $translations_terms[0]->term_id;
```
Can it be done later only when needed ?

manooweb: As we discussed, I pushed a comment in bb8dbc28f8b3e805d567e70f3892632ef96915b9

manooweb (approved):

include/translatable-object.php
```diff
@@ -246,53 +246,83 @@ public function delete_language( $id ) {
...
+		// Checks the cache first.
+		foreach ( $ids as $id ) {
+			$cached = get_object_term_cache( $id, $taxonomy );
+			if ( is_wp_error( $cached ) ) {
+				$terms[ $id ] = false;
+			} elseif ( empty( $cached ) ) {
+				$to_query[] = $id;
+			} else {
+				/** @var WP_Term[] $cached */
+				$terms[ $id ] = reset( $cached );
```
Marianne380: Why not : 
```php
$term = reset( $cached );
if ( $term instanceof WP_Term ) {
    $terms[ $id ] = $first_term;
}
```

include/translatable-object.php
```diff
@@ -246,53 +246,83 @@ public function delete_language( $id ) {
...
 
-		return $term;
+		return $terms;
+	}
+
+	/**
+	 * Wraps `wp_get_object_terms()` to cache it and return only one object.
+	 * Inspired by the WordPress function `get_the_terms()`.
+	 *
+	 * @since 1.2
```
Marianne380, line 342:
```suggestion
	 * @since 1.2
	 * @since 3.8 Now uses `get_object_terms()` for better cache performance.
```

include/translated-object.php
```diff
@@ -422,6 +412,57 @@ protected function should_update_translation_group( $id, $translations ) {
...
+	 * @return int[][] An associative array of translations arrays with language code as key and translation ID as value.
+	 *
+	 * @phpstan-return array<int,array<non-empty-string, positive-int>>
+	 */
+	protected function get_raw_objects_translations( array $ids ) {
+		$terms = $this->get_object_terms( $ids, $this->tax_translations );
+
+		$translations = array();
+		foreach ( $ids as $id ) {
+			if ( empty( $terms[ $id ]->description ) ) {
```
Marianne380:
```suggestion
			if ( empty( $terms[ $id ] ) || empty( $terms[ $id ]->description ) ) {
```

include/translated-object.php
```diff
@@ -499,6 +540,25 @@ protected function validate_translations( $translations, $id = 0, $context = 'sa
 		/** @phpstan-var array<non-empty-string, positive-int> $translations */
 		return array_merge( array( $lang->slug => $id ), $translations );
 	}
+
+	/**
+	 * Validates and sanitizes multiple translations arrays.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int[][] $translations_arrays An array of translations arrays.
```
Marianne380: Isn't it an array of associative translations arrays ? I think `array[]` is more precise for associative arrays isn't it ?
```suggestion
	 * @param array[] An associative array of translations arrays with language code as key and translation ID as value.

File: include/translated-object.php [hunk 12]
```diff
@@ -422,6 +412,57 @@ protected function should_update_translation_group( $id, $translations ) {
...
+	}
+
+	/**
+	 * Returns an unvalidated array of translations for multiple objects.
+	 * It is generally preferable to use `get_objects_translations()`.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int[] $ids Array of object IDs.
+	 * @return int[][] An associative array of translations arrays with language code as key and translation ID as value.
```
Marianne380:
```suggestion
	 * @return array[] An associative array of translations arrays with language code as key and translation ID as value.
```

include/translatable-object.php
manooweb: 🤔 `get_object_term()` already used a cache. Here it just wraps the new `get_object_terms()` to get terms for only one object and avoid code duplication.
It seems useless to me to add this comment.

include/translatable-object.php
manooweb: Done in 3dd921779c896858725512227ff80de00d12b279

include/translated-object.php
manooweb: 🤔 Not sure:
array<int,array<string, int>>
- First level index = object id, (int)
- second level index = language code (string)
- value = translation id (int)

In fact the phpstan annotation precises it.

include/translated-object.php
manooweb: Done in b470dac51d8c61325bcdecb735f1713f83be1a5a

include/translated-object.php
manooweb: Done in b02ce647184fb3a7427955676c151fffa99d9ff5

include/translated-object.php
manooweb: Done in b02ce647184fb3a7427955676c151fffa99d9ff5

manooweb (changes requested):

include/translatable-object.php
```diff
@@ -246,53 +246,93 @@ public function delete_language( $id ) {
...
-				if ( $t->taxonomy === $taxonomy ) {
-					$term = $t;
-				}
+			$term = reset( $cached );
+			if ( ! $term instanceof WP_Term ) {
+				$terms[ $id ] = false;
+				continue;
 			}
+
+			$terms[ $id ] = $term;
```
Marianne380: Sorry, finally this seems better.
```suggestion
			$terms[ $id ] = $term instanceof WP_Term ? $term : false;
```

manooweb: Done in 5fd84cd5093aedb6c8afb7b5166d3ce7aa1bbad9

include/translatable-object.php
```diff
@@ -246,53 +246,88 @@ public function delete_language( $id ) {
...
+			$cached = get_object_term_cache( $id, $taxonomy );
 
-		// Query terms.
-		$terms        = array();
-		$term         = false;
-		$object_terms = wp_get_object_terms( $id, $this->tax_to_cache, array( 'update_term_meta_cache' => false ) );
+			if ( is_wp_error( $cached ) ) {
+				$terms[ $id ] = false;
+				continue;
+			}
```
Chouby: I don't see the interest of storing `false` here. Leaving this item empty when we don't get `WP_Term[]` should be sufficient.

include/translatable-object.php
```diff
@@ -246,53 +246,88 @@ public function delete_language( $id ) {
...
+					$to_cache = array();
+				} else {
+					$to_cache = array( $terms_by_tax[ $tax ]->term_id );
+				}
+				wp_cache_add( $id, $to_cache, "{$tax}_relationships" );
+			}
 		}
 
-		return $term;
+		return $terms;
```
Chouby: Basically we would need something similar to `update_object_term_cache()` which also takes care to cache only non-cached relationships. Unfortunately we can't use it because it expects a post and not a generic object. Could this function be a source of inspiration from (in particular check if its usage of multiple cache would be interesting for readability)?

include/translatable-object.php
```diff
@@ -246,53 +246,88 @@ public function delete_language( $id ) {
...
+	 *
+	 * @since 1.2
+	 *
+	 * @param int    $id       Object ID.
+	 * @param string $taxonomy Polylang taxonomy depending if we are looking for a post (or term, or else) language.
+	 * @return WP_Term|false The term associated to the object in the requested taxonomy if it exists, `false` otherwise.
+	 */
+	public function get_object_term( $id, $taxonomy ) {
+		$terms = $this->get_object_terms( array( $id ), $taxonomy );
+		return isset( $terms[ $id ] ) ? $terms[ $id ] : false;
```
Chouby: Would it impact our code a lot if we return `null` instead of `false`?

include/translated-object.php
```diff
@@ -163,7 +163,7 @@ public function save_translations( $id, array $translations = array() ) {
 		$translations = $this->validate_translations( $translations, $id );
 
 		// Unlink removed translations.
-		$old_translations = $this->get_translations( $id );
+		$old_translations = $this->get_objects_translations( $translations );
```
Chouby, line 166: Isn't this the same as `$terms = $this->get_object_terms( $translations, $this->tax_translations );` found a few lines below?
Why do we need validated translations in one case and not the other?

include/translatable-object.php
```diff
@@ -246,53 +246,88 @@ public function delete_language( $id ) {
...
+			}
 		}
 
-		return $term;
+		return $terms;
+	}
+
+	/**
+	 * Wraps `wp_get_object_terms()` to cache it and return only one object.
+	 * Inspired by the WordPress function `get_the_terms()`.
```
Chouby: Obsolete
```suggestion
```

include/translated-object.php
```diff
@@ -422,6 +412,57 @@ protected function should_update_translation_group( $id, $translations ) {
...
+		foreach ( $ids as $id ) {
+			if ( empty( $terms[ $id ] ) || empty( $terms[ $id ]->description ) ) {
+				$lang = $this->get_language( $id );
+
+				if ( empty( $lang ) ) {
+					$translations[ $id ] = array();
+					continue;
+				}
+
+				$translations[ $id ] = array( $lang->slug => $id );
```
Chouby: Why this change compared to the original `get_raw_translations()`?

manooweb: We need language code as key for static pages to be able to get its translation
See ➡️ https://github.com/polylang/polylang/blob/3.7.3/include/static-pages.php#L118-L122

include/translatable-object.php
manooweb: At first glance, just documentation and what `get_object_terms()` return and what you mean by "empty" in your [comment](https://github.com/polylang/polylang/pull/1690/files#r2335921485)

Done in 2ea33bb3bb3fbbc34dbe8110a442659debd0313c

include/translatable-object.php
manooweb: Removed in 2f60a02b664569a77e5555ee5924a2b87c6be3f7

include/translated-object.php
manooweb: The goal isn't exactly the same.
`get_objects_translations()` gets ids of translations in all the existing translations groups. Passing through the validations ensure that objects themselves are also included as their own translation.

`get_object_terms()` returns translation group term we need to update. Otherwise it will be created.

include/translatable-object.php
manooweb: Done in 89928f814abce541a2b4842d62242fd013eefbc1

include/translatable-object.php
```diff
@@ -246,53 +246,104 @@ public function delete_language( $id ) {
...
+		if ( empty( $ids ) ) {
+			return array();
+		}
 
-		if ( empty( $id ) ) {
-			return false;
+		$cached_data = wp_cache_get_multiple( $ids, "{$taxonomy}_relationships" );
+		$cached_ids  = array_unique( array_keys( array_filter( $cached_data ) ) );
+
+		$non_cached_ids = array_diff( $ids, $cached_ids );
```
Chouby: `array_intersect()` involves less operations. `array_unique` is useless as `array_keys()` can't return duplicates:
```suggestion
		$non_cached_ids = array_keys( array_intersect( $cached_data, array( false ) ) );
```
Another solution could be to use the function `_get_non_cached_ids()`.

include/translatable-object.php
```diff
@@ -246,53 +246,104 @@ public function delete_language( $id ) {
...
+
+		$results = array();
+
+		foreach ( $cached_ids as $id ) {
+			$term = get_term( $cached_data[ $id ][0] );
+
+			$results[ $id ] = $term instanceof WP_Term ? $term : null;
```
Chouby: 1. To simplify and improve the readability, I suggest to split the method to cache relationships in a separate method (as done by `update_object_term_cache()` and handle terms in this method.
2. There are no reasons to handle cached ids and non cached ids separately.
3. Looping over `get_term()` without ensuring that terms are cached may lead to multiple DB queries.

include/translatable-object.php
```diff
@@ -246,53 +246,104 @@ public function delete_language( $id ) {
...
+			return $results;
+		}
+
+		// Query only non-cached terms
+		$_object_terms = wp_get_object_terms(
+			$non_cached_ids,
+			$this->tax_to_cache,
+			array(
+				'fields'                 => 'all_with_object_id',
+				'orderby'                => 'name',
```
Chouby: Why ordering by name?

Chouby (changes requested):

include/translatable-object.php
manooweb: Removed in d227658d019c805b82c673e5ebc6b0a4d4526072

include/translatable-object.php
manooweb: Done in d227658d019c805b82c673e5ebc6b0a4d4526072

include/translatable-object.php
manooweb: Done in d227658d019c805b82c673e5ebc6b0a4d4526072

include/translatable-object.php
```diff
@@ -246,53 +246,107 @@ public function delete_language( $id ) {
...
-
-		if ( is_array( $object_terms ) ) {
-			foreach ( $object_terms as $t ) {
-				$terms[ $t->taxonomy ] = $t;
-				if ( $t->taxonomy === $taxonomy ) {
-					$term = $t;
-				}
-			}
+		if ( empty( $non_cached_ids ) ) {
+			return $cached_data;
```
Chouby: You still handle separately non-cached data and cached data, resulting in sub-optimal performance (2 calls to `_prime_term_caches()`.

Below is a proposal with `update_object_term_cache()` which is inspired by the WP function (I retained some of your ideas to simplify the original function) and `get_object_terms()` which uses this method + `_prime_term_caches` to ensure everything is cached and finally returns the expected terms.

Note there are only one DB query to get the relationships and one to get all the terms.
I just wrote this quickly. This is not tested, but hopefully you get the idea and can fill the holes I left.

```php
 	private update_object_term_cache( $object_ids ) {			
		$non_cached_ids = array();
		foreach ( $this->tax_to_cache as $taxonomy ) {
			$non_cached_ids = array_merge( $non_cached_ids, _get_non_cached_ids( object_ids, "{$taxonomy}_relationships" ) );
		}
		
		if ( empty( $non_cached_ids ) ) {
			return;
		}
		
		$terms = wp_get_object_terms(
			array_unique( $non_cached_ids ),
			$this->tax_to_cache,
			array(
				'fields'                 => 'all_with_object_id',
				'update_term_meta_cache' => false,
			)
		);

		$object_terms = array();
		foreach ( $terms as $term ) {
			$object_terms[ $term->taxonomy ][ $term->object_id ][] = $term->term_id;
		}
		
		foreach ( $non_cached_ids as $id ) {
			foreach ( $this->tax_to_cache as $taxonomy ) {
			if ( ! isset( $object_terms[ $taxonomy ][ $id ] ) ) {
				$object_terms[ $taxonomy ][ $id ] = array();
			}
		}

		foreach ( $object_terms as $tax => $data ) {
			wp_cache_add_multiple( $data, "{$tax}_relationships" );
		}
	}
	
	protected function get_object_terms( array $object_ids, $taxonomy ) {
		$object_ids = $this->sanitize_int_ids_list( $object_ids );
		if ( empty( $object_ids ) ) {
			return array();
		}
	
		$this->update_object_term_cache( $object_ids );
	
		$cached_values = wp_cache_get_multiple( $object_ids, "{$taxonomy}_relationships" ) );
		
		// Flatten the array to prime the terms cache.
		$all_term_ids = array();
		foreach ( $cached_values as $term_ids ) {
			$all_term_ids = array_merge( $all_term_ids, $term_ids );
		}
		_prime_term_caches( $term_ids, false );
		
		$terms = array();
		foreach ( $cached_values as $object_id => $term_ids ) {
			if ( ! empty( $term_ids ) ) {
				$term_id = reset( $term_ids ); // There is only one term for language or translation groups.
				$terms[ $object_id ] = get_term( $term_id );
			}
		}
		
		return $terms;
	}
```

Chouby (changes requested):

include/translatable-object.php
manooweb: Applied in 3a80f3f776311d42fa836205e0a32c407357c688

include/translated-object.php
manooweb: 🤔 It seems that since the latest code modifications, these changes become useless.
I reverted them in 2a2f65adb80146fdbb33958d669d005467611184

include/translatable-object.php
```diff
@@ -246,53 +246,110 @@ public function delete_language( $id ) {
...
-					$term = $t;
-				}
+		if ( ! is_array( $terms ) ) {
+			return;
+		}
+
+		$object_terms = array();
+		foreach ( $terms as $term ) {
+			if ( ! isset( $term->object_id ) ) {
+				continue;
```
Chouby: I guess this is coming from PHPStan. However, we know that this property is always set. Is there a doc alternative to avoid useless operations?

include/translatable-object.php
```diff
@@ -246,53 +246,110 @@ public function delete_language( $id ) {
...
+			$all_term_ids = array_merge( $all_term_ids, $term_ids );
+		}
+		_prime_term_caches( $all_term_ids, false );
+
+		$terms = array();
+		foreach ( $cached_values as $object_id => $term_ids ) {
+			if ( ! empty( $term_ids ) ) {
+				$term_id             = reset( $term_ids ); // There is only one term for language or translation groups.
+				$term                = get_term( $term_id );
+				$terms[ $object_id ] = $term instanceof WP_Term ? $term : null;
```
Chouby: For which usage do we need to store null values? (apart the fact that I don't see how it could occur as I expect `$term_id` to always be associated to a valid term).

include/translatable-object.php
```diff
@@ -246,53 +246,110 @@ public function delete_language( $id ) {
...
+	/**
+	 * Wraps `wp_get_object_terms()` to cache it and return only one object.
+	 *
+	 * @since 1.2
+	 *
+	 * @param int    $id       Object ID.
+	 * @param string $taxonomy Polylang taxonomy depending if we are looking for a post (or term, or else) language.
+	 * @return WP_Term|null The term associated to the object in the requested taxonomy if it exists, `false` otherwise.
+	 */
+	public function get_object_term( $id, $taxonomy ) {
```
Chouby: To improve the readability, I suggest to rename object ids variables to `$object_id` everywhere in the new methods.

include/translated-object.php
```diff
@@ -422,6 +412,50 @@ protected function should_update_translation_group( $id, $translations ) {
...
+	}
+
+	/**
+	 * Returns an unvalidated array of translations for multiple objects.
+	 * It is generally preferable to use `get_objects_translations()`.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int[] $ids Array of object IDs.
+	 * @return array[] An associative array of translations arrays with language code as key and translation ID as value.
```
Chouby: The method returns `int[][]` and the description is not exact. This is not an associative array but an array of associative arrays, and it describes only the second level key. See the phpstan-return.

include/translated-object.php
```diff
@@ -499,6 +533,25 @@ protected function validate_translations( $translations, $id = 0, $context = 'sa
 		/** @phpstan-var array<non-empty-string, positive-int> $translations */
 		return array_merge( array( $lang->slug => $id ), $translations );
 	}
+
+	/**
+	 * Validates and sanitizes multiple translations arrays.
+	 *
+	 * @since 3.8
+	 *
+	 * @param array[] $translations_arrays An associative array of translations arrays with language code as key and translation ID as value.
```
Chouby: Same as above.

include/translated-object.php
```diff
@@ -499,6 +533,25 @@ protected function validate_translations( $translations, $id = 0, $context = 'sa
 		/** @phpstan-var array<non-empty-string, positive-int> $translations */
 		return array_merge( array( $lang->slug => $id ), $translations );
 	}
+
+	/**
+	 * Validates and sanitizes multiple translations arrays.
```
Chouby: It doesn't only validate but also flattens the array, thus the name and description could be updated.

Chouby (changes requested):

include/translatable-object.php
manooweb: IMO, I'm afraid it isn't possible simply with documentation.
It comes form that `WP_Term` which doesn't declare this property
https://github.com/WordPress/WordPress/blob/6.8.2/wp-includes/class-wp-term.php#L18-L115
while it can be populated by de WP Term query
https://github.com/WordPress/WordPress/blob/6.8.2/wp-includes/class-wp-term-query.php#L136-L137

include/translatable-object.php
manooweb: Again it comes from PHPStan because `get_term()` can return `WP_Error` or `null`.

- If we document `@return WP_Term[]`
```shell
Method PLL_Translatable_Object::get_object_terms() should return array<WP_Term> but returns array<WP_Error|WP_Term|null>
```
- If it is documented as in this PR @return array<int,WP_Term|null>, it isn't sufficient too
```shell
Method PLL_Translatable_Object::get_object_terms() should return array<int, WP_Term|null> but returns array<WP_Error|WP_Term|null>
```

What do you think about simply documented 
```php
	 * @return WP_Term[] Array of terms with object ID as key.
```
and check `get_term()` return is a `WP_Term`

```php
$term                = get_term( $term_id );

if ( ! $term instanceof WP_Term ) {
	continue;
}

$terms[ $object_id ] = $term;

```
?

include/translatable-object.php
manooweb: Your comment is on an old method. Do you mean to modify also here ?

New methods concerned are in `PLL_Translated_Object` class.
Done in 7c75eed6b93d86e6090018ef496f0a568210d793

include/translated-object.php
manooweb: Done in 052478dd4cc9fda0e0c909ff163f09151e52ea85

include/translated-object.php
manooweb: An idea for the name ?

include/translated-object.php
manooweb: Done in 052478dd4cc9fda0e0c909ff163f09151e52ea85

include/translatable-object.php
Chouby: OK. If anything can be done, it should be upstream. I propose to ignore the PHPStan error in `phpstan.neon.dist`.

include/translatable-object.php
Chouby: This proposal is better.
However we could simply document `$term` with `@var WP_Term` if trust that we will always get a term. @Hug0-Drelon your opinion?

include/translatable-object.php
Chouby: Yes. It will clarify this method too.

include/translated-object.php
Chouby: I would remove it and inline it in `get_objects_translations()`.

manooweb: Done in d4df22e6f228c867275cee1871b7fc228421b4ea

include/translatable-object.php
manooweb: Tried by documenting with `@var WP_Term $term` in 98dc703a67c0b359f8ae40a48f22a9d7e7dbbb0c

include/translatable-object.php
manooweb: Done in 458fbb4c8f4da8e557ebf7e28a23c02d8bca7c90

include/translatable-object.php
manooweb: Done in a45232313307db7f09194b9a64b6a42ef376a328

include/translatable-object.php
```diff
@@ -246,53 +246,108 @@ public function delete_language( $id ) {
...
+	 *
+	 * @since 1.2
+	 *
+	 * @param int    $object_id Object ID.
+	 * @param string $taxonomy  Polylang taxonomy depending if we are looking for a post (or term, or else) language.
+	 * @return WP_Term|null The term associated to the object in the requested taxonomy if it exists, `false` otherwise.
+	 */
+	public function get_object_term( $object_id, $taxonomy ) {
+		$terms = $this->get_object_terms( array( $object_id ), $taxonomy );
+		return isset( $terms[ $object_id ] ) ? $terms[ $object_id ] : null;
```
Hug0-Drelon:
```suggestion
		return $terms[ $object_id ] ?? null;
```

include/translatable-object.php
```diff
@@ -246,53 +246,108 @@ public function delete_language( $id ) {
...
 
-			wp_cache_add( $id, $to_cache, "{$tax}_relationships" );
+		foreach ( $object_terms as $tax => $data ) {
+			wp_cache_add_multiple( $data, "{$tax}_relationships" );
 		}
+	}
 
-		return $term;
+	/**
+	 * Wraps `wp_get_object_terms()` to cache it and return only one object.
```
Hug0-Drelon: The documentation description doesn't seem accurate.

I suggest:
```suggestion
	 * Return terms associated to the given object in the given taxonomy.
```

include/translatable-object.php
```diff
@@ -246,53 +246,108 @@ public function delete_language( $id ) {
 	}
 
 	/**
-	 * Wraps `wp_get_object_terms()` to cache it and return only one object.
-	 * Inspired by the WordPress function `get_the_terms()`.
+	 * Wraps wp_get_object_terms() to cache it for multiple objects.
+	 * Similar to get_object_term() but for multiple objects.
```
Hug0-Drelon: Do you refer to one of our methods or a WordPress function here?

include/translatable-object.php
```diff
@@ -246,53 +246,108 @@ public function delete_language( $id ) {
...
+	protected function get_object_terms( array $object_ids, string $taxonomy ) {
+		$object_ids = $this->sanitize_int_ids_list( $object_ids );
+		if ( empty( $object_ids ) ) {
+			return array();
 		}
 
-		$term = get_object_term_cache( $id, $taxonomy );
+		$this->update_object_term_cache( $object_ids );
+
+		$cached_values = wp_cache_get_multiple( $object_ids, "{$taxonomy}_relationships" );
```
Hug0-Drelon: I don't quite get this code, it seems unnatural.
It forces to know what key `update_object_term_cache()` uses internally in order to get a value from `wp_cache_get_multiple()`.
I would expect `update_object_term_cache()` to return cached values instead, and be named something else.

More over, we should populate cache only if `wp_cache_get_multiple( $object_ids, "{$taxonomy}_relationships" );` return `false`. Correct me if I'm wrong but here we compute object IDs each time and call `wp_cache_get_multiple()` no matter what.

include/translatable-object.php
```diff
@@ -246,53 +246,108 @@ public function delete_language( $id ) {
...
 
-		if ( is_array( $term ) ) {
-			return ! empty( $term ) ? reset( $term ) : false;
+		// Flatten the array to prime the terms cache.
+		$all_term_ids = array();
+		foreach ( $cached_values as $term_ids ) {
+			$all_term_ids = array_merge( $all_term_ids, $term_ids );
 		}
+		_prime_term_caches( $all_term_ids, false );
```
Hug0-Drelon, line 270: Are we sure we need it? It can overload memory quite easily, am I right?
Also, what query in our code will this benefit to? As far as I see it, `wp_get_object_terms()` called earlier will call `get_terms()` which is already cached.

tests/phpunit/tests/test-translated-post.php
```diff
@@ -99,6 +99,66 @@ public function test_delete_post_translation() {
...
+		// Removes the translations from the group by updating the German post.
+		self::$model->post->save_translations( $posts['de'], array() );
+
+		$terms = wp_get_object_terms( $posts, 'post_translations' );
+		$this->assertCount( 1, $terms );
+
+		$this->assertSame( 0, self::$model->post->get_translation( $posts['de'], 'fr' ) );
+		$this->assertSame( 0, self::$model->post->get_translation( $posts['de'], 'en' ) );
+
+		// Links again the French and English posts.
```
Hug0-Drelon, line 145: EN and FR are unlinked because DE is removed from the translation group, did I understand correctly?

tests/phpunit/tests/test-translated-term.php
```diff
@@ -153,6 +153,76 @@ public function test_get_db_infos() {
 		$this->assertNotSame( $db_infos['table'], $multi_db_infos['table'], 'The table name should be different between blogs.' );
 	}
 
+	/**
+	 * Checks that the translations group are correctly updated when linking several translations together.
+	 *
+	 * @ticket #2717 see {https://github.com/polylang/polylang-pro/issues/2717}.
+	 * @covers PLL_Translated_Object::save_translations()
```
Hug0-Drelon:
```suggestion
```
This is used by PHPUnit internally for test report. I think it could interfere with `coverage`.

include/translatable-object.php
manooweb: It references the Polylang existing `get_object_term()` method before these changes. It doesn't exist any WordPress function with this name.

Hug0-Drelon: Then you must use `self::` or `ClassName::` notation.
Also, please keep code formatting using backticks as done in all our documentation.
But in the end I doubt this description is useful.

tests/phpunit/tests/test-translated-post.php
manooweb: No!
They are unlinked because they are both deleted from the translations group.
No process is done in this case to keep these translations linked.
See https://github.com/polylang/polylang/blob/3.7.3/include/translated-post.php#L141
and https://wpsyntex.slack.com/archives/C017SSTK4KY/p1756393924140449

Hug0-Drelon: Then the code should document it clearly (maybe in `PLL_Translated_Object::save_translations()`.
Here I stubbled upon this behavior only by reading your test...

Hug0-Drelon: And in the end, EN and FR were unlinked because DE was removed (calling with `array()`).

manooweb:
Here the results:
There is one query less with the new code than with the old code (master-copy branch) and only on post translations save test.
Strangely with the new code and WordPress 6.2.x, we have the same queries count with the old and new codes.
So I decided to skip the concerned test with WordPress < 6.3

<table class="tg"><thead>
  <tr>
    <th class="tg-0lax"></th>
    <th class="tg-cly1" colspan="4">Branches</th>
  </tr></thead>
<tbody>
  <tr>
    <td class="tg-0lax"></td>
    <td class="tg-cly1" colspan="2">CI</td>
    <td class="tg-cly1" colspan="2">local</td>
  </tr>
  <tr>
    <td class="tg-0lax"></td>
    <td class="tg-cly1">Master-copy</td>
    <td class="tg-cly1">Enhance-save-translations</td>
    <td class="tg-cly1">Master-copy</td>
    <td class="tg-cly1">Enchance-save-translations</td>
  </tr>
  <tr>
    <td class="tg-cly1">WP 6.2.x + PHP 7.4 + MS 0 + Allowed to fail false</td>
    <td class="tg-cly1" rowspan="6">39</td>
    <td class="tg-cly1">39</td>
    <td class="tg-cly1">39 (WP 6.2.8)</td>
    <td class="tg-0lax"></td>
  </tr>
  <tr>
    <td class="tg-cly1">WP latest + PHP 7.2 + MS 0 + Allowed to fail false</td>
    <td class="tg-cly1" rowspan="5">38</td>
    <td class="tg-0lax"></td>
    <td class="tg-0lax"></td>
  </tr>
  <tr>
    <td class="tg-cly1">WP latest + PHP 8.3 + MS 0 + Allowed to fail false</td>
    <td class="tg-0lax"></td>
    <td class="tg-cly1">38 (WP 6.8.3)</td>
  </tr>
  <tr>
    <td class="tg-cly1">WP latest + PHP 8.0 + MS 1 + Allowed to fail false</td>
    <td class="tg-0lax"></td>
    <td class="tg-0lax"></td>
  </tr>
  <tr>
    <td class="tg-cly1">WP nightly + PHP 7.2 + MS 0 + Allowed to fail true</td>
    <td class="tg-0lax"></td>
    <td class="tg-0lax"></td>
  </tr>
  <tr>
    <td class="tg-cly1">WP nightly + PHP 8.4 + MS 0 + Allowed to fail true</td>
    <td class="tg-0lax"></td>
    <td class="tg-0lax"></td>
  </tr>
</tbody></table>

include/translatable-object.php
```diff
@@ -290,18 +288,19 @@ protected function get_object_terms( array $object_ids, string $taxonomy ) {
...
 	 *
-	 * @param int[] $object_ids Array of object IDs to retrieve terms for.
+	 * @param int[]  $object_ids Array of object IDs to retrieve terms for.
+	 * @param string $taxonomy   Taxonomy name.
 	 *
-	 * @return void
+	 * @return int[][]
 	 */
-	private function update_object_term_cache( array $object_ids ): void {
+	private function get_from_object_term_cache( array $object_ids, string $taxonomy ) {
```
Hug0-Drelon, line 296: I suggest to update the method doc description as well.

include/translatable-object.php
manooweb: Good idea: done in bb7f81c7d8ef6ee0326d5eeb7f89353862c28487

include/translatable-object.php
manooweb: Done in 3f71b14cf8bd51968075031cd3c54d71cd9fb61a

include/translatable-object.php
```diff
@@ -246,53 +246,108 @@ public function delete_language( $id ) {
...
+	}
+
+	/**
+	 * Return terms associated to the given object in the given taxonomy.
+	 *
+	 * @since 1.2
+	 *
+	 * @param int    $object_id Object ID.
+	 * @param string $taxonomy  Polylang taxonomy depending if we are looking for a post (or term, or else) language.
+	 * @return WP_Term|null The term associated to the object in the requested taxonomy if it exists, `false` otherwise.
```
Chouby: Add a changelog for the change of the return value (false -> null)

tests/phpunit/tests/test-translated-post.php
```diff
@@ -99,6 +99,70 @@ public function test_delete_post_translation() {
...
+		.* Translations in English and French are also unlinked.
+		.* See `PLL_Translated_Post::delete_translation()`.
+		 */
+		self::$model->post->save_translations( $posts['de'], array() );
+
+		$terms = wp_get_object_terms( $posts, 'post_translations' );
+		$this->assertCount( 1, $terms );
+
+		$this->assertSame( 0, self::$model->post->get_translation( $posts['de'], 'fr' ) );
+		$this->assertSame( 0, self::$model->post->get_translation( $posts['de'], 'en' ) );
```
Chouby, line 142: I propose to add an assertion for the relationship between en and fr (as you did lower for terms).

tests/phpunit/tests/test-translated-term.php
```diff
@@ -153,6 +153,76 @@ public function test_get_db_infos() {
...
+		$this->assertSame( $terms['fr'], self::$model->term->get_translation( $terms['de'], 'fr' ) );
+		$this->assertSame( $terms['en'], self::$model->term->get_translation( $terms['de'], 'en' ) );
+
+		// Removes the translations from the group by updating the German term.
+		self::$model->term->save_translations( $terms['de'], array() );
+
+		$translations_terms = wp_get_object_terms( $terms, 'term_translations' );
+		
+		/*
+		 * Keeps the the translation group term_id corresponding to the German term which has just been updated.
```
Chouby:
```suggestion
		 * Keeps the translation group term_id corresponding to the German term which has just been updated.
```

tests/phpunit/tests/test-translated-post.php
```diff
@@ -99,6 +99,70 @@ public function test_delete_post_translation() {
...
+		$terms = wp_get_object_terms( $posts, 'post_translations' );
+		$term_id = $terms[0]->term_id;
+		$this->assertCount( 2, $terms ); // Is correct at this step because the German post isn't translated into either English or French.
+
+		// Links again the German and English posts but not with the French one.
+		self::$model->post->save_translations( $posts['de'], array( 'de' => $posts['de'], 'en' => $posts['en'] ) );
+
+		$terms = wp_get_object_terms( $posts, 'post_translations' );
+		$this->assertCount( 1, $terms );
+		$this->assertSame( $term_id, $terms[0]->term_id ); // The term is the same as the latest save_translations().
```
Chouby: This assertion seems risky as you test one of 2 translation groups, and there is no warranty of the order. Moreover I don't see the added value: is it important that we kept the same term across all changes?

tests/phpunit/tests/test-translated-term.php
```diff
@@ -153,6 +153,76 @@ public function test_get_db_infos() {
...
+
+		$translations_terms = wp_get_object_terms( $terms, 'term_translations' );
+		$this->assertCount( 2, $translations_terms ); // Is correct at this step because the German term isn't translated into either English or French.
+
+		// Links again the German and English terms but not with the French one.
+		self::$model->term->save_translations( $terms['de'], array( 'de' => $terms['de'], 'en' => $terms['en'] ) );
+
+		$translations_terms = wp_get_object_terms( $terms, 'term_translations' );
+		$this->assertCount( 2, $translations_terms ); // Is correct because each term has a translations group event if it isn't translated.
+		$this->assertSame( $term_id, $translations_terms[0]->term_id ); // The term is the same as the German term when we unlinked translations.
```
Chouby: Same comment as for posts.

tests/phpunit/tests/Performances/test-DB_Query_Post.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace WP_Syntex\Polylang\Tests\Integration\Performances;
+
+use PLL_UnitTestCase;
+use PLL_Context_Admin;
+use PLL_UnitTest_Factory;
+
+class DB_Query_Post_Test extends PLL_UnitTestCase {
```
Chouby:
```suggestion
class DB_Query_Translations_Test extends PLL_UnitTestCase {
```

tests/phpunit/tests/Performances/test-DB_Query_Translations.php
```diff
@@ -0,0 +1,87 @@
...
+
+		$pll_admin = ( new PLL_Context_Admin() )->get();
+
+		$query_count = $this->count_queries(
+			function () use ( $pll_admin, $posts ) {
+				$pll_admin->model->post->get_translations( $posts['en'] );
+			}
+		);
+
+		$this->assertSame( 2, $query_count, 'Number of queries when getting post translations should be 2.' );
```
Chouby, line 46: Could you add the 2 expected queries in comment?

tests/phpunit/tests/Performances/test-DB_Query_Translations.php
```diff
@@ -0,0 +1,87 @@
...
+
+		$pll_admin = ( new PLL_Context_Admin() )->get();
+
+		$query_count = $this->count_queries(
+			function () use ( $pll_admin, $post_en, $post_fr, $post_de ) {
+				$pll_admin->model->post->save_translations( $post_en, array( 'fr' => $post_fr, 'de' => $post_de ) );
+			}
+		);
+
+		$this->assertSame( 38, $query_count, 'Number of queries when getting post translations should be 2.' );
```
Chouby:
```suggestion
		$this->assertSame( 38, $query_count, 'Number of queries when getting post translations should be 38.' );
```
OMFG 38 queries! 

Could you add all the expected queries in comment? It would be good to categorize them by operation (I mean by WP function).

include/translatable-object.php
manooweb: Done in 342f830b539baa8b84fb14796aed91d6f76b4a00

tests/phpunit/tests/test-translated-post.php
manooweb: Done in e873a026396a491b8cd73b5c3321f5dfb6b91ff7

tests/phpunit/tests/test-translated-term.php
manooweb: Done in e873a026396a491b8cd73b5c3321f5dfb6b91ff7 but partially useless because this comment was removed.

tests/phpunit/tests/test-translated-post.php
manooweb: Removed in a334fc71743f1c6286a6fbf4d365d442400cc55b

tests/phpunit/tests/test-translated-term.php
manooweb: Removed in a334fc71743f1c6286a6fbf4d365d442400cc55b

tests/phpunit/tests/Performances/test-DB_Query_Post.php
manooweb: Done in 93b702b18c24b66c0d51cd460510e64e8f05fa3a

tests/phpunit/tests/Performances/test-DB_Query_Translations.php
manooweb: Done in 0a8f69f6e1ed1092cc7cbddb13290760c2819cd6 & e0899d5c8cc7a2f530d83ca9c114eb6808f62d76

tests/phpunit/tests/Performances/test-DB_Query_Translations.php
manooweb: Done in 0a8f69f6e1ed1092cc7cbddb13290760c2819cd6

PR #1694, opened 2025-07-08
Filter current locale in REST when `lang` parameter is set.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2683

## How?
- Filter the current locale once the current language is available and matches `lang` parameter if any.

> [!NOTE]  
> I could have used `wpseo_og_locale` or `wpseo_locale` to fix the issue, but any other string translated by WordPress (not Polylang) would not benefit from it.

### DIFF
--- a/tests/phpunit/tests/REST/test-Locale.php
+++ b/tests/phpunit/tests/REST/test-Locale.php
@@ -20,8 +20,6 @@ class Locale_Test extends PLL_UnitTestCase {
 	 */
 	private $server;
 
-	protected static $administrator;
-
 	/**
 	 * @param PLL_UnitTest_Factory $factory
 	 * @return void
@@ -30,8 +28,6 @@ class Locale_Test extends PLL_UnitTestCase {
 		parent::pllSetUpBeforeClass( $factory );
 
 		$factory->language->create_many( 2 );
-
-		self::$administrator = $factory->user->create( array( 'role' => 'administrator' ) );
 	}
 
 	/**


include/rest-request.php
```diff
@@ -130,4 +131,24 @@ public function set_language( $result, $server, $request ) {
...
+	 * @return string The current locale.
+	 */
+	public function set_locale( $locale ) {
+		if ( empty( $_REQUEST['lang'] ) || ! is_string( $_REQUEST['lang'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return $locale;
+		}
+
+		if ( empty( $this->curlang ) || $this->curlang->slug !== $_REQUEST['lang'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return $locale;
+		}
```
Marianne380, line 150: Why not combine both `if` ?

include/rest-request.php
```diff
@@ -130,4 +131,24 @@ public function set_language( $result, $server, $request ) {
...
+
+		return $this->curlang->locale;
```
Marianne380, line 152: Should we check for this ?
```suggestion
		return $this->curlang->locale ?? $locale;

```

include/rest-request.php
Hug0-Drelon: I kept them separated for readability, but I can merge them if you prefer.

include/rest-request.php
Hug0-Drelon: At this point `$this->curlang` is surely a `PLL_Language` and `PLL_Language::$locale` is always set if I'm correct.

PR #1696, opened 2025-07-10
Correctly type the return of `PLL_Settings::redirect()`

Screenfeed:
This prevents to do:

```php
PLL_Settings::redirect();
exit;
```

### DIFF
--- a/settings/settings.php
+++ b/settings/settings.php
@@ -168,6 +168,8 @@ class PLL_Settings extends PLL_Admin_Base {
 	 *
 	 * @param string $action The action name.
 	 * @return void
+	 *
+	 * @phpstan-return never
 	 */
 	public function handle_actions( $action ) {
 		switch ( $action ) {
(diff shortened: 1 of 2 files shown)


settings/settings.php
```diff
@@ -168,6 +168,8 @@ public function set_screen_option( $screen_option, $option, $value ) {
 	 *
 	 * @param string $action The action name.
 	 * @return void
+	 *
+	 * @phpstan-return never
 	 */
 	public function handle_actions( $action ) {
```
Marianne380: Can we take this opportunity to add type hints here as well (for the return)?

Screenfeed: I also type-hinted the parameter (after making sure it won't trigger a fatal error) in [fd76a35](https://github.com/polylang/polylang/pull/1696/commits/fd76a35352000124b347096ba9fe92fde5aab35b).

PR #1698, opened 2025-07-11
Hide Polylang's admin bar menu on content edit screens

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/2671.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -452,6 +452,10 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	 * @return void
 	 */
 	public function admin_bar_menu( $wp_admin_bar ) {
+		if ( $this->should_hide_admin_bar_menu() ) {
+			return;
+		}
+
 		$all_item = (object) array(
 			'slug' => 'all',
 			'name' => __( 'Show all languages', 'polylang' ),
@@ -535,5 +539,22 @@ abstract class PLL_Admin_Base extends PLL_Base {
 				}
 			}
 		}
+	}
+
+	/**
+	 * Tells if the Polylang's admin bar menu should be hidden for the current page.
+	 * Conventionally, it should be hidden on edition pages.
+	 *
+	 * @since 3.8
+	 *
+	 * @return bool
+	 */
+	public function should_hide_admin_bar_menu(): bool {
+		foreach ( $this->model->translatable_objects as $object ) {
+			if ( $object->should_hide_admin_bar_menu() ) {
+				return true;
+			}
+		}
+		return false;
 	}
 }
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -550,11 +550,23 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	 * @return bool
 	 */
 	public function should_hide_admin_bar_menu(): bool {
-		foreach ( $this->model->translatable_objects as $object ) {
-			if ( $object->should_hide_admin_bar_menu() ) {
-				return true;
-			}
+		global $pagenow, $typenow, $taxnow;
+
+		if ( in_array( $pagenow, array( 'post.php', 'post-new.php' ), true ) ) {
+			return ! empty( $typenow ) && $this->model->post->is_translated_object_type( $typenow );
 		}
-		return false;
+
+		if ( 'term.php' === $pagenow ) {
+			return ! empty( $taxnow ) && $this->model->term->is_translated_object_type( $taxnow );
+		}
+
+		/**
+		 * Tells if the Polylang's admin bar menu should be hidden.
+		 *
+		 * @since 3.8
+		 *
+		 * @param bool $hide Should the admin bar menu be hidden? Default is `false`.
+		 */
+		return (bool) apply_filters( 'pll_should_hide_admin_bar_menu', false );
 	}
 }
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -475,13 +475,13 @@ abstract class PLL_Admin_Base extends PLL_Base {
 		}
 
 		/**
-		 * Filters the admin languages filter submenu items.
+		 * Filters the admin bar language filter submenu items.
 		 *
 		 * @since 2.6
-		 * @since 3.8 New parameter `$all_items`.
+		 * @since 3.8 Added `$all_items` parameter.
 		 *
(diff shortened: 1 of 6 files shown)


admin/admin-base.php
```diff
@@ -536,4 +540,33 @@ public function remove_customize_submenu() {
...
+		}
+
+		/**
+		 * Tells if the Polylang's admin bar menu should be hidden.
+		 *
+		 * @since 3.8
+		 *
+		 * @param bool $hide Should the admin bar menu be hidden? Default is `false`.
+		 */
+		return (bool) apply_filters( 'pll_should_hide_admin_bar_menu', false );
```
Chouby: We already have an existing filter to achieve the same result: 
`add_filter( 'pll_admin_languages_filter', '__return_empty_array' );`
But, as is, it would not allow to easily re-add the admin languages filter (the same for you) when it's not present. And I believe it could be interesting to allow this.

Thus I propose two alternatives:
1. Keep only the current filter. Add a second param which would always contain all items. The first param would still contain the same value or `array()` when we want to hide the admin languages filter.
2. Use a separate filter as you proposed, but pass the result of the test on `$pagenow` as first param. I would also invert the logic and name it `pll_show_admin_languages_filter` to keep the same vocabulary as the existing one.

admin/admin-base.php
```diff
@@ -536,4 +540,33 @@ public function remove_customize_submenu() {
...
+		global $pagenow, $typenow, $taxnow;
+
+		if ( in_array( $pagenow, array( 'post.php', 'post-new.php' ), true ) ) {
+			return ! empty( $typenow ) && $this->model->post->is_translated_object_type( $typenow );
+		}
+
+		if ( 'term.php' === $pagenow ) {
+			return ! empty( $taxnow ) && $this->model->term->is_translated_object_type( $taxnow );
+		}
+
```
Chouby: I'm not sure that we need  keep the menu when editing non-translated objects.

Screenfeed: I think we can keep it. The switcher can't "mess" with the object's language switcher in the user's head, since it doesn't have any. But it's up to you, I'm not against its removal.

admin/admin-base.php
Screenfeed: Done in [743e975](https://github.com/polylang/polylang/pull/1698/commits/743e975d1f04fe71df0c7fc05efb1e0fc3809c01).

admin/admin-base.php
Screenfeed: After discussing this with everyone, done in [028c823](https://github.com/polylang/polylang/pull/1698/commits/028c823e051c337b62548037c41b97f5cd71cb01).

admin/admin-base.php
```diff
@@ -467,31 +467,39 @@ public function admin_bar_menu( $wp_admin_bar ) {
...
-
-		$menu = array(
-			'id'    => 'languages',
-			'title' => $selected->flag . $title,
-			'href'  => esc_url( add_query_arg( 'lang', $selected->slug, remove_query_arg( 'paged' ) ) ),
-			'meta'  => array(
-				'title' => __( 'Filters content by language', 'polylang' ),
-			),
-		);
+		$items = apply_filters( 'pll_admin_languages_filter', $items, $all_items );
```
Marianne380: Can we take the opportunity to enhance the filter doc, something as : 
```php
/**
* Filters the admin bar language filter submenu items.
*
* @since 2.6
* @since 3.8 Added `$all_items` parameter.
*
* @param array $items     The items of the admin languages filter to display (may be empty if menu hidden).
* @param array $all_items Complete unfiltered list of all available language items.
*/
```

tests/phpunit/tests/test-admin.php
```diff
@@ -49,6 +49,25 @@ public function test_admin_bar_menu() {
 		$this->assertEquals( '/wp-admin/edit.php?lang=fr', $fr->href );
 	}
 
+	public function test_admin_bar_menu_should_hide() {
+		global $wp_admin_bar;
+		add_filter( 'show_admin_bar', '__return_true' ); // Make sure to show admin bar.
+
+		$this->go_to( admin_url( 'post-new.php?post_type=page' ) );
+		$GLOBALS['pagenow'] = 'post-new.php';
+		$GLOBALS['typenow'] = 'page';
```
Marianne380, line 58: Maybe add cleanup for this in `tearDown()` ?

tests/phpunit/tests/test-admin.php
```diff
@@ -49,6 +49,25 @@ public function test_admin_bar_menu() {
...
+
+		$links_model = self::$model->get_links_model();
+		$pll_admin = new PLL_Admin( $links_model );
+		$pll_admin->init();
+
+		_wp_admin_bar_init();
+		do_action_ref_array( 'admin_bar_menu', array( &$wp_admin_bar ) );
+
+		$languages = $wp_admin_bar->get_node( 'languages' );
+		$this->assertEmpty( $languages );
```
Marianne380:
```suggestion
		$this->assertEmpty( $languages, 'Languages admin bar menu should be hidden on post edit pages' );
```

Marianne380 (changes requested):

tests/phpunit/tests/test-admin.php
Screenfeed: This is done [here](https://github.com/polylang/polylang/blob/480fb6cb006679dc87f6ad8fb33708f318700a54/tests/phpunit/includes/testcase-trait.php#L142). Except that `$GLOBALS['typenow']` is missing so I added it in [a1d5e3f](https://github.com/polylang/polylang/pull/1698/commits/a1d5e3f4a356b4d0207def1d02d6798d9bcba998).

tests/phpunit/tests/test-admin.php
Screenfeed: Done in [b9ae37a](https://github.com/polylang/polylang/pull/1698/commits/b9ae37aed1929aee524120a4556f53a2537318c1).

admin/admin-base.php
Screenfeed: Why not. Done in [4c6c00f](https://github.com/polylang/polylang/pull/1698/commits/4c6c00f7cd8e2eb818a4f5937ba94bf76ec408ef).

Marianne380 (approved):

PR #1700, opened 2025-07-18
Fix `Uncaught TypeError: in_array()`

Screenfeed:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2684.

This error happens on multisite, when doing `in_array( 'comment_status', Options::get( 'sync' ) )`  on a switched site where Polylang is not activated. In such case, `Options::get()` returns `null`.

In this PR, `Options::get()` returns the option's default value instead, so its type is always correct.

Bonus: this also removes a useless loop in `register()`.

### DIFF
--- a/include/Options/Inactive_Option.php
+++ b/include/Options/Inactive_Option.php
@@ -43,7 +43,6 @@ class Inactive_Option extends Abstract_Option {
 	 */
 	public function __construct( Abstract_Option $option ) {
 		$this->option = $option;
-		self::$key    = $option::key();
 		$this->errors = new WP_Error();
 
 		// Make sure the option doesn't contain any value.


include/Options/Options.php
Hug0-Drelon, line 1: I thinks all `! empty( $this->inactive[ $this->current_blog_id ] )` occurence could be replaced by `$this->is_plugin_active()`.
Then instead of having around line 171 :
```php
if ( ! $this->is_plugin_active() ) {
	$this->inactive[ $this->current_blog_id ] = true;
}
```
you would have:
```php
if ( ! pll_is_plugin_active( POLYLANG_BASENAME ) && ! doing_action( 'activate_' . POLYLANG_BASENAME ) ) {
	$this->inactive[ $this->current_blog_id ] = true;
}
```

As of today, `is_plugin_active()` is only used in one place, in the contrary my proposal allow for it to be used everywhere. Also, my proposal could be moved into a new method such as `set_inactive_blog()` for instance.

include/Options/Options.php
```diff
@@ -559,24 +575,58 @@ private function get_modified(): array {
...
 		 *
 		 * @since 3.7
+		 * @since 3.7.5 New parameter `$is_active`.
 		 *
 		 * @param Options $options         Instance of the options.
 		 * @param int     $current_blog_id Current blog ID.
+		 * @param bool    $is_active       True if Polylang is activated on the current site.
 		 */
-		do_action( 'pll_init_options_for_blog', $this, $this->current_blog_id );
+		do_action( 'pll_init_options_for_blog', $this, $this->current_blog_id, $is_active );
```
Hug0-Drelon: We should consider to *not* do this action if `false === $is_active`. This would prevent side effect in case a third party is hooked and isn't aware of your new parameter.

Screenfeed: This is required, otherwise the options array (`$this->options[ $this->current_blog_id ]`) will remain empty. Remember how the array is populated:

```php
add_action( 'pll_init_options_for_blog', array( Options_Registry::class, 'register' ) );
$options = new Options();
```

include/Options/Options.php
Screenfeed: You're right but I wanted to avoid calling `! pll_is_plugin_active( POLYLANG_BASENAME ) && ! doing_action( 'activate_' . POLYLANG_BASENAME )` each time. `$this->inactive` is a sort of cache.

Screenfeed:
> Other idea 💡
> 
> Instead of modifying each part of the code that set/modify/delete options value to ensure current blog has Polylang activated or not (at the risk of missing something), I suggest to "reverse" the problem and create a stub in `Options::register()` each time we register options for a blog without Polylang activated. The stub being a decorator for all option instances, such as:

I like the idea. Applied in [2c0592d](https://github.com/polylang/polylang/pull/1700/commits/2c0592d1e649ee71807f79bf641a2bd4ddac45bf).

Hug0-Drelon (approved):

include/Options/Options.php
```diff
@@ -108,9 +123,13 @@ public function __construct() {
...
+		$key = $class_name::key();
+
+		if ( ! $this->is_plugin_active() && ! array_key_exists( $key, $this->default ) ) {
+			$this->default[ $key ] = $class_name::get_default();
+		}
 
+		foreach ( $this->options as &$options ) {
 			if ( ! array_key_exists( $key, $options ) ) {
 				// Option raw value doesn't exist in database, use default instead.
 				$options[ $key ] = new $class_name();
```
Hug0-Drelon: This is what @Chouby wanted to move to `self::get()`.
Along the code you added just before.
This way you won't need to have `Options::$default`.

tests/phpunit/tests/test-multisite-options.php
```diff
@@ -0,0 +1,84 @@
...
+			$result = $pll_admin->options->set( 'media_support', false );
+			$this->assertInstanceOf( WP_Error::class, $result, 'Setting an option should always return an error.' );
+			$this->assertSame( 'pll_not_active', $result->get_error_code(), 'The option should return a blocking error on a site where Polylang is not activated.' );
+
+			$result = $pll_admin->options->set( 'sync', array( 'taxonomies' => array( 'category' ) ) );
+			$this->assertInstanceOf( WP_Error::class, $result, 'Setting an option should always return an error.' );
+			$this->assertSame( 'pll_not_active', $result->get_error_code(), 'The option should return a blocking error on a site where Polylang is not activated.' );
+		}
+
+		public function test_options_class_properties() {
```
Hug0-Drelon: I think we should avoid using `getPropertyValue()` which are testing implementation details, not the proper output.

Hug0-Drelon (changes requested):

include/Options/Inactive_Option.php
```diff
@@ -0,0 +1,137 @@
...
+
+	/**
+	 * Returns the default value.
+	 *
+	 * @since 3.8
+	 *
+	 * @return mixed
+	 */
+	public function get_default() {
+		return $this->option->get_default();
```
Hug0-Drelon: Using `Abstract_Option::reset()` would prevent to make `Abstract_Option::get_default()` public. Is it something we would want?

include/Options/Options.php
```diff
@@ -85,8 +93,9 @@ class Options implements ArrayAccess, IteratorAggregate {
 	 */
 	public function __construct() {
 		// Keep track of the blog ID.
-		$this->blog_id         = (int) get_current_blog_id();
-		$this->current_blog_id = $this->blog_id;
+		$this->blog_id          = (int) get_current_blog_id();
+		$this->current_blog_id  = $this->blog_id;
+		$this->is_plugin_active = array( $this->blog_id => true );
```
Hug0-Drelon, line 98: What happens if a new instance of `Options` is created during a `switch_blog` action toward a blog without Polylang activated?

include/Options/Inactive_Option.php
Screenfeed: I thought about it at some point, then the idea got lost in the multiple changes and reverts.
We could use `reset()` in the constructor as a safety measure (however it shouldn't be needed since the real option is supposed to be instantiated with no value). Then we can simply use `get()`.

Screenfeed: Done in [2e9b969](https://github.com/polylang/polylang/pull/1700/commits/2e9b9698a6bd70a73203e59dc5a8c039336d5590) and [ca41fb2](https://github.com/polylang/polylang/pull/1700/commits/ca41fb2e42f8f2364420f7144925652160110224).

include/Options/Options.php
Screenfeed: That means someone has `switch_to_blog()` before `plugins_loaded` without restoring (or restoring later).
Shouldn't we put this person in jail?
Do you think we still should protect us from this? 🤔

Hug0-Drelon: Jail is too kind of you, they should burn at the stake.
But put that way, maybe we shouldn't handle such case.

include/Options/Inactive_Option.php
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.8
+	 *
+	 * @param Abstract_Option $option The option to wrap.
+	 */
+	public function __construct( Abstract_Option $option ) {
+		$this->option = $option;
+		self::$key    = $option::key();
```
Chouby: As discussed, we can't do that.
```suggestion
```

Chouby (changes requested):

PR #1704, opened 2025-07-25
Rework `PLL_Install`

Screenfeed:
## Aim of this PR

- Improve the use of (de)activation hooks in Polylang Pro.
- Allow to use php 7.2+ in the class `Polylang`.
- Prevent a possible fatal error with php < 7.0.
- The version of php can be tested now in Unit Tests.

Until now, `Polylang` was instantiated before doing the `PLL_Install` checks, so the class had to remain compatible with php < 7.2.
However, `Polylang` is currently compatible php 7.0+ only (`string` type hint added in PLL 3.6).

## How

`PLL_Install` is not used inside the `Polylang` class anymore. This allows to:

- Register activation hooks only if the WP version and php version are hight enough.
- Then create the `Polylang` instance if we're not deactivating the plugin.
- `PLL_Install` and `PLL_Install_Base` are replaced by several classes: `PLL_Usable`, `PLL_Activate`, and `PLL_Deactivate`.

## Edge-case, behavior change

1. If WP's version is too low, there is a scenario where the user keeps Polylang activated, then updates WP. In this case, the activation hooks won't trigger (since PLL is already active).
2. If Polylang doesn't pass the 2 checks (php version, WP version):
    - the constants are still defined,
    - the autoloader is still enabled,
    - but the class `Polylang` isn't instantiated.
    - This means the function `PLL()` exists but returns `null`.
    - This case was already existing if not passing `$_GET['deactivate-polylang']`.


install/install-base.php
```diff
@@ -4,131 +4,169 @@
...
+	 * Plugin activation.
 	 *
 	 * @since 0.5
+	 * @since 3.8 Static method.
 	 *
 	 * @return void
 	 */
-	protected function _activate() {
+	protected static function _activate() {
 		// Can be overridden in child class
```
Marianne380:
```suggestion
		// Can be overridden in child class.
```

install/install-base.php
```diff
@@ -4,131 +4,169 @@
...
+	 * Plugin deactivation.
 	 *
 	 * @since 0.5
+	 * @since 3.8 Static method.
 	 *
 	 * @return void
 	 */
-	protected function _deactivate() {
+	protected static function _deactivate() {
 		// Can be overridden in child class
```
Marianne380:
```suggestion
		// Can be overridden in child class.
```

install/install-base.php
```diff
@@ -4,131 +4,169 @@
...
+	}
+
+	/**
+	 * Returns (and cache) the list of sites on multi-site.
+	 *
+	 * @since 3.8
+	 *
+	 * @return array
+	 */
+	protected static function get_sites() {
```
Marianne380:
```suggestion
	protected static function get_sites(): array {
```

install/install-base.php
```diff
@@ -4,131 +4,169 @@
...
+	 * @param string $what        Either 'activate' or 'deactivate'.
+	 * @param bool   $networkwide Whether the plugin is (de)activated for all sites in the network or just the current site.
+	 * @return void
+	 */
+	protected static function do_for_all_blogs( $what, $networkwide ) {
+		if ( is_multisite() && $networkwide ) {
+			// Network.
+			foreach ( self::get_sites() as $blog_id ) {
+				switch_to_blog( $blog_id );
+				'activate' == $what ? static::_activate() : static::_deactivate();
```
Marianne380: Can we take this opportunity to replace `==` with `===` ?

install/install.php
```diff
@@ -7,11 +7,70 @@
...
+	 * Constructor.
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $plugin_file     Path to the plugin's main file.
+	 * @param string $plugin_version  Version of the plugin.
+	 * @param string $min_wp_version  Minimal WP version required to run the plugin.
+	 * @param string $min_php_version Minimal php version required to run the plugin.
+	 */
+	public function __construct( $plugin_file, $plugin_version, $min_wp_version, $min_php_version ) {
```
Marianne380: Can we type hint ?

Marianne380 (changes requested):

install/install-base.php
Screenfeed: Yes, done in [b380dab](https://github.com/polylang/polylang/pull/1704/commits/b380dab17d8667ce1c54a971564ce5d2373fd72a).

install/install.php
Screenfeed: I can't remember why I didn't in the first place 🤔 
Done in [96a7276](https://github.com/polylang/polylang/pull/1704/commits/96a727606ec08ea014cae2e8bad5d062a5ed9db7).

install/install-base.php
```diff
@@ -4,131 +4,169 @@
...
+	 * @since 3.8
+	 *
+	 * @return array
+	 */
+	protected static function get_sites(): array {
+		global $wpdb;
+
+		if ( empty( self::$sites ) ) {
+			self::$sites = $wpdb->get_col( "SELECT blog_id FROM $wpdb->blogs" );
+		}
```
Chouby: Instead of having our own method, could we now use the function `get_sites()`. It seems to me that `number` set to 0 would retrurn all sites (I did not test). 
We would benefit of the WP cache.

polylang.php
```diff
@@ -36,25 +36,41 @@
...
 	}
-} else {
-	// Go on loading the plugin.
-	define( 'POLYLANG_VERSION', '3.8-dev' );
-	define( 'PLL_MIN_WP_VERSION', '6.2' );
-	define( 'PLL_MIN_PHP_VERSION', '7.2' );
+	return;
+}
+
+// Stopping here if we are going to deactivate the plugin (avoids breaking rewrite rules).
```
Chouby: This comment did not refer to this line. See : https://github.com/polylang/polylang/blob/3.7.3/include/class-polylang.php#L45-L48

polylang.php
```diff
@@ -36,25 +36,41 @@
...
+if ( ! empty( $_GET['deactivate-polylang'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+	return;
+}
+
+require_once __DIR__ . '/install/install-base.php';
+require_once __DIR__ . '/install/install.php';
```
Chouby: Why not using the autoloader here?

polylang.php
```diff
@@ -36,25 +36,41 @@
...
+$install = new PLL_Install(
+	__FILE__, // Plugin file.
+	'3.8-dev', // Plugin version.
+	'6.2', // WP version.
+	'7.2' // PHP version.
+);
```
Chouby: I'm not fan to replace the constants by parameters in this case. I often change their values and it's easier for me as this was earlier. Moreover I'm not comfortable to change the behavior of these constant definitions (they are defined this way for a very long time. Who knows how they are used?)

Although I don't see new tests, I guess your goal is to make the more code in the class `PLL_Install` testable. I suggest to read constants with `pll_get_constant()` to be able to mock constants instead.

polylang.php
```diff
@@ -36,25 +36,41 @@
...
+
+if ( ! $install->is_deactivation() && $install->can_activate() ) {
```
Chouby: All other methods have been made static in this PR. If you do the same tor these 2 methods, this would allow not add `$install` to the global scope (even if you remove it later).

Chouby (changes requested):

polylang.php
Screenfeed: Because of https://github.com/polylang/polylang/pull/1704/files#diff-ae092ac7c3df32deed6d43deddc79a341466290d38b72c6d8ad685d218936238R83-R86.
And also because things like `function_exists( 'PLL' )` will return `false` if we can't activate the plugin.

polylang.php
Screenfeed: >I'm not fan to replace the constants by parameters in this case. I often change their values and it's easier for me as this was earlier.

Even with the comments I placed especially for this?

>I'm not comfortable to change the behavior of these constant definitions

Isn't it something we decided together?

polylang.php
Screenfeed: So all properties become static too?

polylang.php
Screenfeed: Moved in [7b0e18e](https://github.com/polylang/polylang/pull/1704/commits/7b0e18e60880b26f7e5e9d7df5672771b4077f92).

polylang.php
Screenfeed: `pll_*_constant()` functions in [0967346](https://github.com/polylang/polylang/pull/1704/commits/0967346e52fa2a371034713655266f45f4191ebf) and [0934edd](https://github.com/polylang/polylang/pull/1704/commits/0934eddb739c627ee076a61ca55be999fe53e380).

install/install-base.php
Screenfeed: Done in [05aa1d1](https://github.com/polylang/polylang/pull/1704/commits/05aa1d185b46b6f156faf0dcb2cc9e87b07ca238).
You're right about `'number' => 0`, it returns all sites. About the cache: unless another plugin get the same idea, we're just swapping our cache for WP's, there's no gain because `get_sites()` and `get_sites( array( 'number' => 0 ) )` are stored differently. Still, I think it's a good thing.

polylang.php
Screenfeed: Done in [5960897](https://github.com/polylang/polylang/pull/1704/commits/59608970dff0f8b5910ae090a265ced0b9884670).

polylang.php
Screenfeed: Constants are back in [5960897](https://github.com/polylang/polylang/pull/1704/commits/59608970dff0f8b5910ae090a265ced0b9884670).

install/install-base.php
```diff
@@ -1,134 +1,140 @@
 <?php
 /**
  * @package Polylang
+ *
+ * /!\ THE CODE IN THIS FILE MUST BE COMPATIBLE WITH PHP 5.6.
```
Marianne380: Could we briefly explain why here ?

Screenfeed: You mean, why the code that tells if the version of php is new enough should be compatible old versions of php? 😅

Marianne380: Put like that, it sounds like overkill :joy:

install/install.php
```diff
@@ -1,35 +1,128 @@
...
+	 * } $info
+	 */
+	public static function init( array $info ) {
+		static::$plugin_name     = $info['plugin_name'];
+		static::$plugin_basename = $info['plugin_basename'];
+		static::$plugin_version  = $info['plugin_version'];
+		static::$min_wp_version  = $info['min_wp_version'];
+		static::$min_php_version = $info['min_php_version'];
+		static::$cur_php_version = ! empty( $info['cur_php_version'] ) ? $info['cur_php_version'] : PHP_VERSION;
+	}
```
Chouby: I would not pass the constants to the class but use the constants directly. As an improvement to the current situation, I propose to use `pll_get_constant()` to be able to test the class.
Yes this requires to include functions.php sooner and make it compatible with PHP 5.6 but I consider it minor issues compared to the simplification of the class usage in production code and in tests.

install/install-base.php
Chouby: In addition to this comment, I propose to add a check to the CI. This may require a new phpcs config file or we can do it with a direct command line `vendor/bin/phpcs install/install-base.php --standard=PHPCompatibility --runtime-set testVersion 5.6-` as I'm not sure that we cans specify different PHP versions for different files in a single config.

polylang.php
```diff
@@ -36,42 +36,69 @@
...
-	}
+// Whether we are using Polylang or Polylang Pro, get the filename of the plugin in use.
+if ( ! defined( 'POLYLANG_ROOT_FILE' ) ) {
+	define( 'POLYLANG_ROOT_FILE', __FILE__ );
+}
 
-	define( 'POLYLANG', ucwords( str_replace( '-', ' ', dirname( POLYLANG_BASENAME ) ) ) );
+if ( ! defined( 'POLYLANG_BASENAME' ) ) {
+	define( 'POLYLANG_BASENAME', plugin_basename( __FILE__ ) ); // Plugin name as known by WP.
+	require __DIR__ . '/vendor/autoload.php';
```
Chouby, line 70: For me it's not consistent to require the autoloader here and then require install files directly.
If we trust that the autoloader remains compatible with PHP 5.6 (we could add a check in CI to avoid an unnoticed change), we can trust the autoloader to load install files too.
If you fear an unnoticed change in the autoloader and thus want to keep requiring install files manually then the autoloader should be required after the PHP version check. In fact in this scenario, we don't need it before the Polylang instantiation.

Screenfeed: I did that because this autoloader is not required with PLL Pro. But after checking in PLL Pro, PLL Pro's autoloader is required before PLL, so it's fine. Done in [ad2987c](https://github.com/polylang/polylang/pull/1704/commits/ad2987cf9404e29b36821380bed8038d5c3c596e).

install/install.php
Screenfeed: Done in [5a89eb6](https://github.com/polylang/polylang/pull/1704/commits/5a89eb64fb1bab64c3010f0210275f6296dfd444):

1. The `pll_*_constant()` functions have been moved to a separate file `constant-functions.php`: the code in `functions.php` can still use php 7.2.
2. The file `constant-functions.php` needs to be required manually (not by Composer) otherwise the functions cannot be stubbed in tests.
3. The second parameter of `pll_get_constant()` (`$default`) is always provided (as an empty string) because of PHPStan. The type returned by `pll_get_constant()` was `mixed`, which was an issue, so the types of the parameters and return of the function have been modified: the return type now matches the type of the default value. It's cheating but it works for any constant.

install/install-base.php
Screenfeed: Good idea.
We can't specify multiple php versions for different files, I would have done it long ago otherwise. I had to create a dedicated configuration file because we want to run the command on multiple php files.
Done in [e2cb389](https://github.com/polylang/polylang/pull/1704/commits/e2cb38948b44695b6cc98d6ab230b02bd756042f) and fixed in the following commits.

install/activate.php
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		// Avoid 1 query on every pages if no wpml strings is registered.
+		if ( ! get_option( 'polylang_wpml_strings' ) ) {
+			update_option( 'polylang_wpml_strings', array() );
+		}
+
+		// Don't use flush_rewrite_rules at network activation. See #32471.
+		// Thanks to RavanH for the trick. See https://polylang.wordpress.com/2015/06/10/polylang-1-7-6-and-multisite/.
+		// Rewrite rules are created at next page load.
```
Chouby:
```suggestion
		/*
		 * Don't use flush_rewrite_rules at network activation. See #32471.
		 * Thanks to RavanH for the trick. See https://polylang.wordpress.com/2015/06/10/polylang-1-7-6-and-multisite/.
		 * Rewrite rules are created at next page load.
		 */
```

install/abstract-activable.php
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A generic (de)activation class compatible with multisite.
+ *
+ * @since 3.8
+ */
+abstract class PLL_Abstract_Activable {
+	/**
+	 * (De)Activation for all blogs.
+	 *
+	 * @since 3.8
```
Chouby: I like to keep the original `@since` even when the method is moved to another class.
```suggestion
	 * @since 1.2
```
There are other cases in other classes.

install/abstract-activate.php
```diff
@@ -0,0 +1,42 @@
...
+	/**
+	 * Adds the required hooks.
+	 *
+	 * @since 3.8
+	 *
+	 * @return void
+	 */
+	public static function add_hooks(): void {
+		// Plugin activation.
+		register_activation_hook( pll_get_constant( 'POLYLANG_BASENAME', '' ), array( static::class, 'do_for_all_blogs' ) );
```
Chouby: Thinking to re-usability, what do you think to introduce a method `get_plugin_basename()`? This one could be in `PLL_Abstract_Activable`.
For another plugin, we could re-use the class just overriding this method instead of rewriting `add_hooks()`.

install/usable.php
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * /!\ THE CONSTANTS `POLYLANG`, `PLL_MIN_PHP_VERSION`, AND `PLL_MIN_WP_VERSION` MUST BE DEFINED.
+ */
+
+/**
+ * Class tat can tell if Polylang can be activated.
+ *
+ * @since 3.8
+ */
+class PLL_Usable {
```
Chouby, line 15: Similar to above for re-usability, what do you think to introduce 3 methods:
```
get_min_php_version()
get_min_wp_version ()
get_plugin_name()
```

Chouby (changes requested):

install/abstract-activate.php
Screenfeed: I'm not sure it will be useful some day but I'm not against it. Done in [8c09281](https://github.com/polylang/polylang/pull/1704/commits/8c09281f707e4bbb8860c7ca65b45ed3e517dd92). I also added one for the plugin version.

install/usable.php
Screenfeed: Done in [8c09281](https://github.com/polylang/polylang/pull/1704/commits/8c09281f707e4bbb8860c7ca65b45ed3e517dd92).

install/abstract-activable.php
Screenfeed: Fixed in [ee66d86](https://github.com/polylang/polylang/pull/1704/commits/ee66d862f32c9c9b0dce15b0dc478589373ab454).

install/activate.php
Screenfeed: Done in [4e6dc26](https://github.com/polylang/polylang/pull/1704/commits/4e6dc26250fc7d71a1da641a07e6120f4229553f).

PR #1708, opened 2025-08-08
Migrate `set_filters_sanitization()`.

Hug0-Drelon:
## What
Migrate Polylang Pro code from `PLL_REST_Post::set_filters_sanitizations()` to `PLL_REST_Request`.

Originally proposed in https://github.com/polylang/polylang-pro/pull/2608#discussion_r2248079664.

### DIFF
--- a/include/rest-request.php
+++ b/include/rest-request.php
@@ -92,8 +92,6 @@ class PLL_REST_Request extends PLL_Base {
 		}
 
 		add_filter( 'rest_pre_dispatch', array( $this, 'set_language' ), 10, 3 );
-
-		// Use rest_pre_dispatch_filter to get the right language locale and initialize correctly sanitization filters.
 		add_filter( 'rest_request_before_callbacks', array( $this, 'set_filters_sanitization' ) );
 
 		$this->filters_links           = new PLL_Filters_Links( $this );


include/rest-request.php
```diff
@@ -67,6 +72,9 @@ public function __construct( &$links_model ) {
 		}
 
 		$this->model->set_languages_ready();
+
+		// Use rest_pre_dispatch_filter to get the right language locale and initialize correctly sanitization filters.
+		add_filter( 'rest_pre_dispatch', array( $this, 'set_filters_sanitization' ), 10, 3 );
```
Chouby: I would move this to `init()` as other filters.

include/rest-request.php
```diff
@@ -130,4 +138,43 @@ public function set_language( $result, $server, $request ) {
...
+	 * @param WP_REST_Request $request Request used to generate the response.
+	 * @return mixed
+	 *
+	 * @phpstan-template T of array
+	 * @phpstan-param WP_REST_Request<T> $request
+	 */
+	public function set_filters_sanitization( $result, $server, $request ) {
+		if ( ! current_user_can( 'edit_posts' ) ) {
+			return $result;
+		}
```
Chouby: I don't believe it's useful in this context.

include/rest-request.php
```diff
@@ -130,4 +138,43 @@ public function set_language( $result, $server, $request ) {
...
+
+		$id   = $request->get_param( 'id' );
+		$lang = $request->get_param( 'lang' );
+
+		if ( is_string( $lang ) && ! empty( $lang ) ) {
+			$language = $this->model->get_language( sanitize_key( $lang ) );
+		} elseif ( is_numeric( $id ) && ! empty( $id ) ) {
+			// Otherwise we need to get the language from the post itself.
+			$language = $this->model->post->get_language( (int) $id );
+		}
```
Chouby: This won't work in all contexts (for ex when creating terms, so I suggest to check that the request is of the expected type)
While we are at it, we would activate the filters for term routes [as done on admin side](https://github.com/polylang/polylang/blob/3.7.3/admin/admin.php#L170).

Hug0-Drelon: The bug was there from the beginning then.
Also, I think we have a problem here. Unfortunately the object type cannot be retrieved easily from `$request` nor `$server`.
If you do want to handle such case we would have to parse current URL ourselves.

include/rest-request.php
Hug0-Drelon: Applied in e13a0f31.

include/rest-request.php
Hug0-Drelon: Applied in e13a0f31.

include/rest-request.php
Hug0-Drelon: As discussed during our meeting, let's use a better suited hook ; `rest_request_before_callbacks` so the ID can be fetched correctly. Then use the callback to deduce the object type.

See 3f330311.

modules/REST/Request.php
```diff
@@ -0,0 +1,225 @@
...
+		 * @param WP_REST_Request $request Request used to generate the response.
+		 * @param array           $handler Route handler used for the request.
+		 */
+		$type = apply_filters( 'pll_rest_request_object_type', null, $this->request, $this->handler );
+
+		if ( in_array( $type, array( 'post', 'term' ), true ) ) {
+			return $type;
+		}
+
+		$controller = $this->handler['callback'][0] ?? null;
```
Chouby: What do you think to
1. pass the controller to the filter
2. pass the handler before the request (as the probability of usage is higher for the handler).
```suggestion
		$controller = $this->handler['callback'][0] ?? null;

		$type = apply_filters( 'pll_rest_request_object_type', null, $controller, $this->handler, $this->request );

		if ( in_array( $type, array( 'post', 'term' ), true ) ) {
			return $type;
		}
		
```

modules/REST/Request.php
```diff
@@ -0,0 +1,225 @@
...
+			return null;
+		}
+
+		$lang = $this->get_param( 'lang' );
+
+		if ( empty( $lang ) || ! is_string( $lang ) ) {
+			return null;
+		}
+
+		$lang = $this->model->get_language( $lang );
```
Chouby, line 130:
```suggestion
		$lang = $this->model->get_language( sanitize_key( $lang ) );
```

include/rest-request.php
```diff
@@ -130,4 +146,38 @@ public function set_language( $result, $server, $request ) {
...
+	public function set_filters_sanitization( $response, $handler, $request ) {
+		$id   = $request->get_param( 'id' );
+		$lang = $request->get_param( 'lang' );
+
+		if ( is_string( $lang ) && ! empty( $lang ) ) {
+			$language = $this->model->get_language( sanitize_key( $lang ) );
+		} elseif ( is_numeric( $id ) && ! empty( $id ) ) {
+			$type     = $this->request->get_object_type();
+			$language = $type ? $this->model->$type->get_language( (int) $id ) : null;
+		}
```
Chouby: We can use the new Request object to simplify:
```suggestion
	public function set_filters_sanitization( $response ) {
		$language = $this->request->get_language();

		if ( empty( $language ) ) {
			$type     = $this->request->get_object_type();
			$language = $type ? $this->model->$type->get_language( $this->request->get_id() ) : null;
		}
```

Hug0-Drelon: Good idea, done in 0c9b42c0.

modules/REST/Request.php
Hug0-Drelon: Let's pass the`$handler` first as you said. Done in 2d71cc60.

As for the controller itself, not sure it adds any benefit since it can be found easily in `$handler`.

modules/REST/Request.php
Hug0-Drelon: Won't apply as discussed.

PR #1709, opened 2025-08-11
Open language arguments management.

Hug0-Drelon:
## What?
Make language arguments management reusable.
Motivated by https://github.com/polylang/polylang-pro/pull/2653#discussion_r2251762193

## How?
- Create `Languages::get_language_args_from_locale()` with code mostly from `Languages` REST controller.
- Use it in `Languages` REST controller.
- Add proper PHPUnit test for `Model\Languges` CRUD.

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -162,7 +162,7 @@ class Languages {
 	 * } $args
 	 */
 	public function add( $args ) {
-		$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? false;
+		$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? null;
 		$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? null;
 		$args['term_group'] = $args['term_group'] ?? 0;
 
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -162,9 +162,24 @@ class Languages {
 	 * } $args
 	 */
 	public function add( $args ) {
-		$args = $this->validate_lang( $args );
-		if ( is_wp_error( $args ) && $args->has_errors() ) {
-			return $args;
+		if ( ! isset( $args['locale'] ) ) {
+			return new WP_Error( 'pll_missing_locale', __( 'Locale is required', 'polylang' ) );
+		}
+
+		$languages = include POLYLANG_DIR . '/settings/languages.php';
+		if ( ! empty( $languages[ $args['locale'] ] ) ) {
+			$args['name']       = $args['name'] ?? $languages[ $args['locale'] ]['name'];
+			$args['slug']       = $args['slug'] ?? $languages[ $args['locale'] ]['code'];
+			$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? 'rtl' === $languages[ $args['locale'] ]['dir'];
+			$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $languages[ $args['locale'] ]['flag'];
+			$args['term_group'] = $args['term_group'] ?? 0;
+		} elseif ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) || ! isset( $args['rtl'] ) ) {
+			return new WP_Error( 'pll_missing_args', __( 'Missing required arguments', 'polylang' ) );
+		}
+
+		$errors = $this->validate_lang( $args );
+		if ( $errors->has_errors() ) {
+			return $errors;
 		}
 
 		/**
@@ -271,9 +286,16 @@ class Languages {
 			return new WP_Error( 'pll_invalid_language_id', __( 'The language does not seem to exist.', 'polylang' ) );
 		}
 
-		$args = $this->validate_lang( $args, $lang );
-		if ( is_wp_error( $args ) && $args->has_errors() ) {
-			return $args;
+		$args['locale']     = $args['locale'] ?? $lang->locale;
+		$args['name']       = $args['name'] ?? $lang->name;
+		$args['slug']       = $args['slug'] ?? $lang->slug;
+		$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? $lang->is_rtl;
+		$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $lang->flag_code;
+		$args['term_group'] = $args['term_group'] ?? $lang->term_group;
+
+		$errors = $this->validate_lang( $args );
+		if ( $errors->has_errors() ) {
+			return $errors;
 		}
 
 		/**
@@ -850,57 +872,25 @@ class Languages {
 	 *
 	 * @param array             $args Parameters of {@see WP_Syntex\Polylang\Model\Languages::add() or @see WP_Syntex\Polylang\Model\Languages::update()}.
 	 * @param PLL_Language|null $lang Optional the language currently updated, the language is created if not set.
-	 * @return WP_Error|array Error object or array of language arguments.
+	 * @return WP_Error
 	 *
 	 * @phpstan-param array{
-	 *     locale?: string,
-	 *     slug?: string,
-	 *     name?: string,
-	 *     flag?: string,
-	 *     flag_code?: string,
-	 *     is_rtl?: bool,
-	 *     rtl?: bool,
-	 *     term_group?: int|numeric-string,
-	 *     no_default_cat?: bool
+	 *     locale: string,
+	 *     slug: string,
+	 *     name: string,
+	 *     flag?: string
 	 * } $args
 	 */
-	protected function validate_lang( $args, ?PLL_Language $lang = null ) {
+	protected function validate_lang( $args, ?PLL_Language $lang = null ): WP_Error {
 		$errors = new WP_Error();
 
-		if ( empty( $lang ) ) {
-			// Create a new language.
-			if ( ! isset( $args['locale'] ) ) {
-				$errors->add( 'pll_missing_locale', __( 'Locale is required', 'polylang' ) );
-
-				return $errors;
-			} else {
-				$languages = include POLYLANG_DIR . '/settings/languages.php';
-				if ( ! empty( $languages[ $args['locale'] ] ) ) {
-					$args['name']       = $args['name'] ?? $languages[ $args['locale'] ]['name'];
-					$args['slug']       = $args['slug'] ?? $languages[ $args['locale'] ]['code'];
-					$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? 'rtl' === $languages[ $args['locale'] ]['dir'];
-					$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $languages[ $args['locale'] ]['flag'];
-					$args['term_group'] = $args['term_group'] ?? 0;
-				} elseif ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) || ! isset( $args['rtl'] ) ) {
-					$errors->add( 'pll_missing_args', __( 'Missing required arguments', 'polylang' ) );
-				}
-			}
-		} else {
-			// Update an existing language.
-			$args['name']       = $args['name'] ?? $lang->name;
-			$args['slug']       = $args['slug'] ?? $lang->slug;
-			$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? $lang->is_rtl;
-			$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $lang->flag_code;
-			$args['term_group'] = $args['term_group'] ?? $lang->term_group;
-		}
-
 		// Validate locale with the same pattern as WP 4.3. See #28303.
-		if ( isset( $args['locale'] ) && ! preg_match( '#' . self::LOCALE_PATTERN . '#', $args['locale'], $matches ) ) {
+		if ( ! preg_match( '#' . self::LOCALE_PATTERN . '#', $args['locale'], $matches ) ) {
 			$errors->add( 'pll_invalid_locale', __( 'Enter a valid WordPress locale', 'polylang' ) );
 		}
 
 		// Validate slug characters.
-		if ( isset( $args['slug'] ) && ! preg_match( '#' . self::SLUG_PATTERN . '#', $args['slug'] ) ) {
+		if ( ! preg_match( '#' . self::SLUG_PATTERN . '#', $args['slug'] ) ) {
 			$errors->add( 'pll_invalid_slug', __( 'The language code contains invalid characters', 'polylang' ) );
 		}
 
@@ -930,12 +920,7 @@ class Languages {
 			}
 		}
 
-		if ( $errors->has_errors() ) {
-			return $errors;
-		}
-
-		/** @var array $args */
-		return $args;
+		return $errors;
 	}
 
 	/**
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -297,7 +297,7 @@ class Languages {
 		$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $lang->flag_code;
 		$args['term_group'] = $args['term_group'] ?? $lang->term_group;
 
-		$errors = $this->validate_lang( $args );
+		$errors = $this->validate_lang( $args, $lang );
 		if ( $errors->has_errors() ) {
 			return $errors;
 		}
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -757,6 +757,7 @@ class Languages {
 	 * @return array|false Array of language arguments or false if the locale is invalid.
 	 *
 	 * @phpstan-param array{locale: string, name?: string, slug?: string, is_rtl?: bool, flag_code?: string, term_group?: int, no_default_cat?: bool} $args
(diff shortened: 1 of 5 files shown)


include/Model/Languages.php
```diff
@@ -747,6 +747,48 @@ public function delete_transient_from_options_table(): void {
...
+				'term_group'     => $args['term_group'] ?? 0,
+				'no_default_cat' => $args['no_default_cat'] ?? false,
+			);
+		}
+
+		$languages = include POLYLANG_DIR . '/settings/languages.php';
+
+		if ( empty( $languages[ $args['locale'] ] ) ) {
+			return false;
+		}
```
manooweb: If I correctly understand the @chouby [comment](https://github.com/polylang/polylang-pro/pull/2653#discussion_r2251762193), the goal is to allow an unknown locale. But it seems to me that isn't the case here, isn't it?

It seems to me that this is still the case [here](https://github.com/polylang/polylang-pro/blob/72cc51ad452c46798aa1d7dff934e4496e96dea6/modules/CLI/Command/Language.php#L97-L101) too.

Am I wrong?

manooweb (changes requested):

include/Model/Languages.php
Hug0-Drelon: And don't get what you say.
The code you point out in WPCLI is related to this PR. But this PR doesn't care about WPCLI in the end. Also, my PR on WPCLI is not finished yet.

Moreover `get_language_args_from_locale()` allows to get language arguments with an unknown locale.

manooweb (changes requested):

include/Model/Languages.php
Hug0-Drelon: As I said, my PR on Polylang Pro is not finished yet. Also, if unknown locale is used you have to provide valid data along the way. As per the documentation [here](https://github.com/polylang/polylang-pro/blob/72cc51ad452c46798aa1d7dff934e4496e96dea6/modules/CLI/Command/Language.php#L42)

manooweb: Sorry, GH made a mess in my comment and my previous one disappeared.
I understand: it means that we will have to modify this code again later to allow unknown locale when WP CLI PR will be finished.

Hug0-Drelon: No this PR should be merge as it is. No need to modify the code of Polylang Free afterward.

manooweb (approved):

include/Model/Languages.php
```diff
@@ -747,6 +747,48 @@ public function delete_transient_from_options_table(): void {
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param array $args Arguments to create a language.
+	 * @return array|false Array of language arguments or false if the locale is invalid.
+	 *
+	 * @phpstan-param array{locale: string, name?: string, slug?: string, is_rtl?: bool, flag_code?: string, term_group?: int, no_default_cat?: bool} $args
+	 * @phpstan-return array{locale: string, name: string, slug: string, rtl: bool, flag: string, term_group: int|numeric-string, no_default_cat: bool}|false
+	 */
+	public function get_language_args_from_locale( array $args ) {
```
Chouby: Instead of adding a new method to call from outside the class, I propose to inject the feature at the first line of the method `add()` doing something like:
```php
if ( isset( $args['locale'] && ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) || ! isset( $args['rtl'] ) ) ) ) {
	$languages = include POLYLANG_DIR . '/settings/languages.php';
	if ( ! empty( $languages[ $args['locale'] ] ) ) {
		$args['name'] = $args['name'] ?? $languages[ $args['locale'] ]['name'];
		$args['slug'] = $args['slug'] ?? $languages[ $args['locale'] ]['code'];
		$args['rtl']  = $args['is_rtl'] ?? 'rtl' === $languages[ $args['locale'] ]['dir'];
		$args['flag'] = args['flag'] ?? $languages[ $args['locale'] ]['flag'];
	}
}
```
Possibly, we could convert `flag_code` to `flag` inside the method too.
What do you think?

Hug0-Drelon: I've take it a step further and gathered validation for both creation and update. See 066da9e3.

include/Model/Languages.php
```diff
@@ -843,25 +850,55 @@ protected function build_metas( array $args ): string {
...
 	 * } $args
 	 */
-	protected function validate_lang( $args, ?PLL_Language $lang = null ): WP_Error {
+	protected function validate_lang( $args, ?PLL_Language $lang = null ) {
 		$errors = new WP_Error();
 
+		if ( empty( $lang ) ) {
+			// Create a new language.
+			if ( ! isset( $args['locale'] ) ) {
+				$errors->add( 'pll_missing_locale', __( 'Locale is required', 'polylang' ) );
```
Chouby: This new error is quite a duplicate of the invalid locale. It could make sense to keep it if you early return.

include/Model/Languages.php
```diff
@@ -843,25 +850,55 @@ protected function build_metas( array $args ): string {
...
+			} else {
+				$languages = include POLYLANG_DIR . '/settings/languages.php';
+				if ( ! empty( $languages[ $args['locale'] ] ) ) {
+					$args['name']       = $args['name'] ?? $languages[ $args['locale'] ]['name'];
+					$args['slug']       = $args['slug'] ?? $languages[ $args['locale'] ]['code'];
+					$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? 'rtl' === $languages[ $args['locale'] ]['dir'];
+					$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $languages[ $args['locale'] ]['flag'];
+					$args['term_group'] = $args['term_group'] ?? 0;
+				} elseif ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) || ! isset( $args['rtl'] ) ) {
+					$errors->add( 'pll_missing_args', __( 'Missing required arguments', 'polylang' ) );
```
Chouby: The way I propose in https://github.com/polylang/polylang/pull/1709#discussion_r2270032484 had the advantage not to always access to the languages file. I know it's not critical but this is something we can avoid.
The missing args error doesn't add value. The pre-existing validation will duplicate the error and provide more details.

include/Model/Languages.php
```diff
@@ -843,25 +850,55 @@ protected function build_metas( array $args ): string {
...
+				}
+			}
+		} else {
+			// Update an existing language.
+			$args['name']       = $args['name'] ?? $lang->name;
+			$args['slug']       = $args['slug'] ?? $lang->slug;
+			$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? $lang->is_rtl;
+			$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $lang->flag_code;
+			$args['term_group'] = $args['term_group'] ?? $lang->term_group;
```
Chouby: I'm all in to change the behavior of `update()` too. However, the `if ( $lang) ... else` makes me think that `validate_lang()` is not the best place to modify `$args`. I would keep these parts in `add()` and `update()` respectively, keeping `validate_lang()` unchanged.

modules/REST/V1/Languages.php
```diff
@@ -170,33 +170,26 @@ public function create_item( $request ) {
...
+		$args   = $request->get_params();
 		$result = $this->languages->add( $args );
 
 		if ( is_wp_error( $result ) ) {
 			return $this->add_status_to_error( $result );
 		}
 
 		/** @var PLL_Language */
-		$language = $this->languages->get( $args['locale'] );
+		$language = $this->languages->get( $args['locale'] ); // @phpstan-ignore-line $args['locale'] is set at this point.
```
Chouby: I would prefer to ignore the line globally in `phpstan.neon.dist` as we usually do.
Another way could be to keep `locale` instead of `locale?` in `@phpstan-var`. It's not worse than it was before. 
And to be sure it's set, I suggest to set the param as required for creation in `get_endpoint_args_for_item_schema()`

tests/phpunit/tests/test-create-delete-languages.php
```diff
@@ -59,7 +59,7 @@ public function test_add_and_delete_language() {
 		$this->assertEqualSetsWithIndex( array( 'ar', 'en' ), $this->pll_env->model->get_languages_list( array( 'fields' => 'slug' ) ) );
 
 		// Attempt to create a language with the same slug as an existing one.
-		$this->pll_env->model->add_language( array( 'slug' => 'en-gb', 'locale' => 'en_GB' ) );
+		$this->pll_env->model->add_language( array( 'slug' => 'en', 'locale' => 'en_GB' ) );
```
Chouby, line 62: Oh sh.. I reviewed this.

include/Model/Languages.php
Hug0-Drelon: The file is read the same amount of time here.
We read it only during creation (when `$lang` is empty).

Regarding `pll_missing_args`, I don't get your comment. Because as of today there is no proper pre-existing validation, am  I right?

modules/REST/V1/Languages.php
Hug0-Drelon: Error ignored in config file in 8e02967c.
`locale` made mandatory only on creation in bad66994.

include/Model/Languages.php
Hug0-Drelon: Done in d61385c90368b54caf2725e82f2f6974a302a56d.

include/Model/Languages.php
Hug0-Drelon: I get what you say, done in 4710001b.

modules/REST/V1/Languages.php
```diff
@@ -170,31 +170,33 @@ public function create_item( $request ) {
...
 
+		// Ensure locale exists before accessing it
+		if ( ! isset( $args['locale'] ) ) {
+			return new WP_Error(
+				'rest_missing_locale',
+				__( 'Locale is required', 'polylang' ),
+				array( 'status' => 400 )
+			);
+		}
+
```
Chouby: This is dead code as `add()` would return a `WP_Error` in this case. Moreover I believe that WP would return early as this param is declared as required in the schema. 
```suggestion
```

modules/REST/V1/Languages.php
```diff
@@ -233,25 +235,24 @@ public function get_item( $request ) {
...
+		 *     slug?: non-empty-string,
+		 *     name?: non-empty-string,
+		 *     is_rtl?: bool,
+		 *     term_group?: int,
 		 *     flag?: non-empty-string
 		 * } $args
 		 */
-		$args   = (array) $prepared;
+		$args            = $request->get_params();
+		$args['lang_id'] = $args['term_id'];
```
Chouby:
```suggestion
		$args['lang_id'] = $language->term_id;
```

include/Model/Languages.php
```diff
@@ -141,23 +141,42 @@ public function get( $value ) {
...
+		if ( ! empty( $languages[ $args['locale'] ] ) ) {
+			$args['name']       = $args['name'] ?? $languages[ $args['locale'] ]['name'];
+			$args['slug']       = $args['slug'] ?? $languages[ $args['locale'] ]['code'];
+			$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? 'rtl' === $languages[ $args['locale'] ]['dir'];
+			$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? $languages[ $args['locale'] ]['flag'];
+			$args['term_group'] = $args['term_group'] ?? 0;
+		} elseif ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) || ! isset( $args['rtl'] ) ) {
+			return new WP_Error( 'pll_missing_args', __( 'Missing required arguments', 'polylang' ) );
+		}
+
```
Chouby: What do you think of the proposal below?
1. access to `languages.php` only if necessary
2. handle only empty name and slug as empty rtl is already handled in `build_metas()`
2. let `validate_lang()` handle all errors (provides details about missing arguments)

```suggestion
		$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? false;
		$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? '';
		$args['term_group'] = $args['term_group'] ?? 0;
		
		if ( ! empty( $args['locale] ) && ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) ) ) ) {
			$languages = include POLYLANG_DIR . '/settings/languages.php';
			if ( ! empty( $languages[ $args['locale'] ] ) ) {
				$found = $languages[ $args['locale'];
				
				$args['name'] = $args['name'] ?? $found['name'];
				$args['slug'] = $args['slug'] ?? $found['code'];
				$args['rtl']  = $args['rtl'] ?? 'rtl' === found['dir'];
				$args['flag'] = $args['flag'] ?? $found['flag'];
			}
		}
		
```

This requires a slight modification of `validate_lang()`

```php
		if ( empty( $args['locale'] ) || ! preg_match( '#' . self::LOCALE_PATTERN . '#', $args['locale'], $matches ) ) {
			$errors->add( 'pll_invalid_locale', __( 'Enter a valid WordPress locale', 'polylang' ) );
		}

		// Validate slug characters.
		if ( empty( $args['slug'] ) || ! preg_match( '#' . self::SLUG_PATTERN . '#', $args['slug'] ) ) {
			$errors->add( 'pll_invalid_slug', __( 'The language code contains invalid characters', 'polylang' ) );
		}
```

Hug0-Drelon: Let's go, done in 4d8b409a.
I had to make some minore changes, especially regarding argument parsing. It must be done *after* core languages check otherwise they are not taken in account.

modules/REST/V1/Languages.php
Hug0-Drelon: Applied in a570b086e717c6249b62af454db3ab563f84607b.

modules/REST/V1/Languages.php
Hug0-Drelon: Indeed, removed in e20f7e48.

include/Model/Languages.php
```diff
@@ -141,23 +141,46 @@ public function get( $value ) {
...
+	 *     term_group?: int|numeric-string,
 	 *     flag?: string,
+	 *     flag_code?: string,
 	 *     no_default_cat?: bool
 	 * } $args
 	 */
 	public function add( $args ) {
+		if ( ! isset( $args['locale'] ) ) {
+			return new WP_Error( 'pll_missing_locale', __( 'Locale is required', 'polylang' ) );
+		}
```
Chouby: It's useless as `validate_lang()` would do the job.

include/Model/Languages.php
```diff
@@ -141,23 +141,46 @@ public function get( $value ) {
...
+				$found        = $languages[ $args['locale'] ];
+				$args['name'] = $args['name'] ?? $found['name'];
+				$args['slug'] = $args['slug'] ?? $found['code'];
+				$args['rtl']  = $args['rtl'] ?? 'rtl' === $found['dir'];
+				$args['flag'] = $args['flag'] ?? $found['flag'];
+			}
+		}
+
+		$args['rtl']        = $args['rtl'] ?? $args['is_rtl'] ?? false;
+		$args['flag']       = $args['flag'] ?? $args['flag_code'] ?? '';
```
Chouby: The reason why I proposed to move this code above is for example this use case:
```php
add( array( 'locale' => 'fr_FR', 'flag_code' => 'be' ) );
```
Here the default value is set in `$args['flag']` and our `$args['flag_code']` is not used.

include/Model/Languages.php
```diff
@@ -141,23 +141,46 @@ public function get( $value ) {
 	 *   @type string $locale         WordPress locale. If something wrong is used for the locale, the .mo files will
 	 *                                not be loaded...
 	 *   @type bool   $rtl            True if rtl language, false otherwise.
+	 *   @type bool   $is_rtl         True if rtl language, false otherwise. Will be converted to rtl.
```
Chouby: The doc can be improved to specify which params are now optionnal (the same for `update()`).

include/Model/Languages.php
```diff
@@ -141,23 +141,46 @@ public function get( $value ) {
...
 	 *
 	 * @phpstan-param array{
-	 *     name: string,
-	 *     slug: string,
-	 *     locale: string,
-	 *     rtl: bool,
-	 *     term_group: int|numeric-string,
+	 *     name?: string,
+	 *     slug?: string,
+	 *     locale?: string,
```
Chouby, line 155:
```suggestion
	 *     locale: string,
```
I'm not sure of what must be done for PHPStan. Here the locale is documented as optional when in fact it's mandatory but the code is protected against the missing arg.

Chouby (changes requested):

include/Model/Languages.php
Hug0-Drelon: Removed in e6c3b5f4.

include/Model/Languages.php
Hug0-Drelon: I get what you meant now.
I had a problem with `''` as default value, preventing to get core default value. I use `null` instead and move the code above, see e6c3b5f4.

include/Model/Languages.php
Hug0-Drelon: Improved in e6c3b5f4.
Added ``Optional.` where needed as well as proper default value.

include/Model/Languages.php
Hug0-Drelon: IMHO PHPStan is here to prevent code error, not to document how it should be used.
So I'd keep `locale?` because code wise it is not mandatory.

PR #1714, opened 2025-08-13
Fix fatal error in updater.

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/2709.

## Why

The bug happens in the following execution flow:

1. **API Request Fails:** When `get_version_from_remote()` encounters an HTTP error (like 401), it returns `false`
2. In `plugins_api_filter()`, the condition `if ( false !== $api_response )` evaluates to false, so `$_data` retains its initial value of `false` come from applying `plugins_api` by WordPress https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-admin/includes/plugin-install.php#L145

> [!NOTE]
> Triggered when clicking on the link `View version x.y.z details` link on the plugins list page in the notice which warns that the plugin has a new version. Also requires that the transient `edd_sl....` doesn't exist or has been expired. It explains why the issue couldn't be reproduced.

image

3.  Later in the code, we attempt to assign properties to `$_data` which is still `false`:
   ```php
   if ( ! isset( $_data->plugin ) ) {
       $_data->plugin = $this->name; // ← Fatal error: $_data is false
   }
   ```

## Fix

Add a safety check to ensure `$_data` contains valid data before attempting property assignments:

### Also

Should we check `$api_response` before:
```php
$api_response = $this->api_request( 'plugin_information', $to_send );

// Expires in 3 hours
$this->set_version_info_cache( $api_response );
```
Because we'll set a empty string in the cache if `$api_response` is set to `false` ?

~~Moreover, should I also do the same PR for the [updater](https://github.com/polylang/updater/blob/main/src/EDD_SL_Plugin_Updater.php#L378) ?~~ (I'll wait for this one to be merged)

### DIFF
--- a/install/plugin-updater.php
+++ b/install/plugin-updater.php
@@ -398,13 +398,14 @@ class PLL_Plugin_Updater {
 		if ( empty( $edd_api_request_transient ) ) {
 
 			$api_response = $this->api_request( 'plugin_information', $to_send );
+			if ( empty( $api_response ) ) {
+				return $_data;
+			}
 
 			// Expires in 3 hours
 			$this->set_version_info_cache( $api_response );
 
-			if ( false !== $api_response ) {
-				$_data = $api_response;
-			}
+			$_data = $api_response;
 		} else {
 			$_data = $edd_api_request_transient;
 		}
--- a/install/plugin-updater.php
+++ b/install/plugin-updater.php
@@ -410,10 +410,6 @@ class PLL_Plugin_Updater {
 			$_data = $edd_api_request_transient;
 		}
 
-		if ( empty( $_data ) ) {
-			return $_data;
-		}
-
 		// Convert sections into an associative array, since we're getting an object, but Core expects an array.
 		if ( isset( $_data->sections ) && ! is_array( $_data->sections ) ) {
 			$_data->sections = $this->convert_object_to_array( $_data->sections );
--- a/install/plugin-updater.php
+++ b/install/plugin-updater.php
@@ -409,6 +409,10 @@ class PLL_Plugin_Updater {
 			$_data = $edd_api_request_transient;
 		}
 
+		if ( empty( $_data ) ) {
+			return $_data;
+		}
+
 		// Convert sections into an associative array, since we're getting an object, but Core expects an array.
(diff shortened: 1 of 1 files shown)


install/plugin-updater.php
```diff
@@ -409,6 +409,10 @@ public function plugins_api_filter( $_data, $_action = '', $_args = null ) {
 			$_data = $edd_api_request_transient;
 		}
 
+		if ( empty( $_data ) ) {
+			return $_data;
+		}
+
```
manooweb: As you proposed in the PR description, I think it's better to check `$api_response` and make an early return if it's `false`.

Indeed, in addition of `get_version_from_remote()` could return `false` [here](https://github.com/polylang/polylang/blob/master/install/plugin-updater.php#L628) and [here](https://github.com/polylang/polylang/blob/master/install/plugin-updater.php#L636), `api_request()` could also return false [here](https://github.com/polylang/polylang/blob/master/install/plugin-updater.php#L502) and [here](https://github.com/polylang/polylang/blob/master/install/plugin-updater.php#L506) and `NULL` [here](https://github.com/polylang/polylang/blob/master/install/plugin-updater.php#L497).

manooweb:
> Moreover, should I also do the same PR for the [updater](https://github.com/polylang/updater/blob/main/src/EDD_SL_Plugin_Updater.php#L378) ? (I'll wait for this one to be merged)

Of course! It's exactly the same code 😉

manooweb (changes requested):

install/plugin-updater.php
Marianne380: Yep, see https://github.com/polylang/polylang/pull/1714/commits/22353ea1fe458749a45e66ef67d1e7f4c2dd5416.

sebastienserre:
@Chouby I unfortunately can't reproduce the fatal error with Polylang Pro 3.7.3 (before trying the fix).
I've changed the requested URL to our preprod website which is protected by an `.htaccess`.
Checking for an update is adding an option in `wp_options` with name `edd_sl_failed_http_63f6d641b0c46a650bcdaa346a9583d4`  and value the request timestamp.

sebastienserre:
This PR doesn't break the update process

PR #1717, opened 2025-08-26
Replace calls to `glob()` by hard-coded arrays

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2718.
[Similar PR](https://github.com/polylang/polylang-pro/pull/2723) on Polylang Pro.
[Discussion](https://wpsyntex.slack.com/archives/C017SSTK4KY/p1756282217367479) on Slack.

The PR replaces the calls to `glob()` by hard-coded arrays, in hope it will fix the fatal error that some users get upon, what we assume to be, an incomplete removal of old files during an update of the plugin.

`glob()` was used to require integrations and modules, and was used twice:
- 1 to fetch all the `load.php` files for modules,
- 1 to fetch all the `load.php` files for integrations.

The hard-coded arrays are generated when dumping Composer's autoloader.

### DIFF
--- a/bin/Loader_Paths_Finder.php
+++ b/bin/Loader_Paths_Finder.php
@@ -35,7 +35,7 @@ class Loader_Paths_Finder {
 		$base_path   = realpath( $args[0] );
 		$type        = rtrim( basename( $base_path ), 's' );
 		$plugin_name = ucwords( str_replace( '-', ' ', basename( dirname( $base_path ) ) ) );
-		$file_paths  = glob( "{$base_path}/*/load.php", \GLOB_NOSORT );
+		$file_paths  = glob( "{$base_path}/*/load.php" );
 
 		if ( ! is_array( $file_paths ) || empty( $file_paths ) ) {
 			throw new RuntimeException( "Could not retrieve the files in {$base_path}" );


bin/generate-globs.php
Hug0-Drelon, line 1: Shouldn't we have a modern and idiomatic script as done in https://github.com/polylang/updater/blob/main/lib/Script/Namespace_Wrapper.php?

Also it would make our Composer script consistent.

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+	if ( empty( $file_paths ) ) {
+		echo "\e[91mCould not find {$type} files in {$plugin_name}\n";
+	}
+
+	sort( $file_paths, SORT_STRING | SORT_FLAG_CASE );
+
+	$write = "<?php
+/**
+ * @package {$plugin_name}
+ *
```
Hug0-Drelon: Not sure this is the way to use `@package`, even more for generated file during build.
See https://docs.phpdoc.org/guide/references/phpdoc/tags/package.html

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+ * /!\ DO NOT DIRECTLY EDIT THIS FILE, THIS FILE IS AUTO-GENERATED AS PART OF THE BUILD PROCESS.
+ */
+
```
Hug0-Drelon:
```suggestion
```

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+return array(\n";
```
Hug0-Drelon:
```suggestion
return [\n";
```
The shorted the better for build files.

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+
+	foreach ( $file_paths as $path ) {
+		$write .= "\t'" . basename( dirname( $path ) ) . "',\n";
+	}
+	$write .= ");\n";
```
Hug0-Drelon:
```suggestion
	$write .= "];\n";
```

bin/generate-globs.php
Chouby: This makes sense.

bin/generate-globs.php
Chouby: I would rather keep our usual coding standards for the plugin.

bin/generate-globs.php
Screenfeed: What are the advantages, while `glob()` + `file_put_contents()` simply works?

Hug0-Drelon: As I said, more modern, idiomatic, follows Composer best practice (cf [doc](https://getcomposer.org/doc/articles/scripts.md#defining-scripts)) and opens to Composer API.

Hug0-Drelon: You'll use `glob()` and `file_put_contents()` anyway.

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+ * This creates a php file that returns an array folder names.
+ *
+ * @since 3.8
+ *
+ * @param string $base_path   Absolute path to the folder containing the files to load.
+ * @param string $type        Name to use in the messages.
+ * @param string $plugin_name Name of the plugin.
+ * @return void
+ */
+function create_load_data_file( string $base_path, string $type, string $plugin_name ): void {
```
Chouby: We could simplify to pass only one param
The `$type` used to name the file could be replaced by the path basename.
The plugin name if necessary could also be generated from the path with something like (I did not test):
`ucwords( str_replace( '-', ' ', basename( dirname( __FILE__ . "/$base_path" ) ) ) )`

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+		return;
+	}
+
+	echo "\e[32mGenerated {$type}s data file for {$plugin_name}\n";
+}
+
+// Create the integrations loader.
+create_load_data_file( dirname( __DIR__ ) . '/integrations', 'integration', 'Polylang' );
+// Create the modules loader.
+create_load_data_file( dirname( __DIR__ ) . '/modules', 'module', 'Polylang' );
```
Chouby: Following the idea of @Hug0-Drelon of using a similar system as namespace wrapper,  these calls could be made from composer.json.
If we follow the idea to have only one param, It could look as:
```
@generate-globs ./modules
@generate-globs ./integrations
```
or, if you decide to hardcode `modules` and `integrations` in the generation script:  
```
@generate-globs .
```
In Polylang Pro, the first solution would look like:
```
@generate-globs ./modules
@generate-globs ./integrations
@generate-globs ./vendor/wpsyntex/polylang/modules
@generate-globs ./vendor/wpsyntex/polylang/integrations
```
The second:
```
@generate-globs .
@generate-globs ./vendor/wpsyntex/polylang
```

bin/generate-globs.php
```diff
@@ -0,0 +1,52 @@
...
+function create_load_data_file( string $base_path, string $type, string $plugin_name ): void {
+	$file_paths = glob( "{$base_path}/*/load.php", GLOB_NOSORT );
+
+	if ( ! is_array( $file_paths ) ) {
+		echo "\e[91mError while retrieving the {$type} files in {$plugin_name}\n";
+		return;
+	}
+	if ( empty( $file_paths ) ) {
+		echo "\e[91mCould not find {$type} files in {$plugin_name}\n";
+	}
```
Chouby: Shouldn't this be simplified in only one error message?

Chouby (changes requested):

bin/generate-globs.php
Screenfeed: Done in [dffcb65](https://github.com/polylang/polylang/pull/1717/commits/dffcb65da6331ebf0ff4112c9b13f6d5d4d175ca).

bin/generate-globs.php
Screenfeed: As discussed with Fred, no minifying.

bin/generate-globs.php
Screenfeed: I would be OK if modules and integrations we using their own `@package` but they don't. To me it makes more sense to use the same tag as modules and integrations.

bin/generate-globs.php
Screenfeed: Done in [22e4bf9](https://github.com/polylang/polylang/pull/1717/commits/22e4bf9331208968913142eb74db7232568f481c).

bin/generate-globs.php
Screenfeed: Done in [3d2cf50](https://github.com/polylang/polylang/pull/1717/commits/3d2cf50c42f2c8710ddeb71e5e950f8bb8c79247).

bin/generate-globs.php
Screenfeed: Done in [3d2cf50](https://github.com/polylang/polylang/pull/1717/commits/3d2cf50c42f2c8710ddeb71e5e950f8bb8c79247).

bin/generate-globs.php
Hug0-Drelon: I was thinking about removing instead.

bin/Globber.php
```diff
@@ -0,0 +1,66 @@
...
+		$base_path   = realpath( $args[0] );
+		$type        = rtrim( basename( $base_path ), 's' );
+		$plugin_name = ucwords( str_replace( '-', ' ', basename( dirname( $base_path ) ) ) );
+		$file_paths  = glob( "{$base_path}/*/load.php", \GLOB_NOSORT );
+
+		if ( ! is_array( $file_paths ) || empty( $file_paths ) ) {
+			throw new RuntimeException( "Could not retrieve the {$type} files in {$plugin_name}" );
+		}
+
+		sort( $file_paths, \SORT_STRING | \SORT_FLAG_CASE );
```
Hug0-Drelon:
```suggestion
```
`glob()` already sort found files. See the [doc](https://www.php.net/manual/en/function.glob.php#refsect1-function.glob-returnvalues).

bin/Globber.php
```diff
@@ -0,0 +1,66 @@
...
+	public static function hard_code( Event $event ): void {
+		$args = $event->getArguments();
+
+		if ( empty( $args[0] ) ) {
+			throw new RuntimeException( 'No path to folder provided' );
+		}
+
+		$base_path   = realpath( $args[0] );
+		$type        = rtrim( basename( $base_path ), 's' );
+		$plugin_name = ucwords( str_replace( '-', ' ', basename( dirname( $base_path ) ) ) );
```
Hug0-Drelon: What if some day folder tree structure differs between plugins?
Also, it's already the case for PLLWC and even though it's not used in it for the moment it could.

bin/Globber.php
```diff
@@ -0,0 +1,66 @@
...
+
+use RuntimeException;
+use Composer\Script\Event;
+
+/**
+ * Class to hard-code `glob()`s.
+ *
+ * @since 3.8
+ */
+class Globber {
```
Hug0-Drelon: We get that now, maybe not in one year. I suggest a more explicit name.
What about `Loader_Paths_Finder`?

bin/Globber.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * @package Polylang
+ */
+
+namespace WP_Syntex\Polylang\Script;
```
Hug0-Drelon: I like the namespace, so I suggest you rename the folder `/bin` to match it.
Maybe I'm wrong, but `/bin` is usually used for executables.

bin/Globber.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+return array(\n";
+
+		foreach ( $file_paths as $path ) {
+			$write .= "\t'" . basename( dirname( $path ) ) . "',\n";
+		}
+		$write .= ");\n";
+
+		if ( false === file_put_contents( "{$base_path}/{$type}-files.php", $write ) ) {
```
Hug0-Drelon:
```suggestion
		if ( false === file_put_contents( "{$base_path}/{$type}-build.php", $write ) ) {
```
So we know it's autogenerated without opening the file.

Hug0-Drelon (changes requested):

bin/Globber.php
Screenfeed: Yes but it doesn't do it right with uppercase and lowercase:
```
Array(
    Aa.php
    Ap.php
    ac.php
)
```

bin/generate-globs.php
Screenfeed: PHPCS enforces it.
```
Missing @package tag in file comment (Squiz.Commenting.FileComment.MissingPackageTag)
```

bin/Globber.php
Hug0-Drelon: Same for folder names?
Because we do not care about file name here (it's all `load.php`).

Screenfeed: Yes same. I agree that sorting is useless here but it helps debugging.

bin/Globber.php
Screenfeed: I agree ¯\\_(ツ)_/¯

bin/Globber.php
Hug0-Drelon: Debugging what?

bin/Globber.php
Screenfeed: Good idea. Done in [46ec85d](https://github.com/polylang/polylang/pull/1717/commits/46ec85dce641aaba710b8936a8026be995984957).

bin/Globber.php
Screenfeed: You're not fun.
Done in [9494bd0](https://github.com/polylang/polylang/pull/1717/commits/9494bd062776585230ef43d3306640518e435287).

bin/generate-globs.php
Hug0-Drelon: A rule can be ignored. Moreover if we use it incorrectly...

bin/Globber.php
Hug0-Drelon: The you can get rid of it. `$plugin_name` is here for messages. Not the code itself.

bin/generate-globs.php
Screenfeed: Agree but then it should be part of a large "Review `@package` tags" issue, shouldn't it?

bin/Globber.php
Screenfeed: The array, it's much easier to read.

bin/Globber.php
Screenfeed: It was a suggestion from @Chouby, to use less arguments: https://github.com/polylang/polylang/pull/1717#discussion_r2322463915

> We could simplify to pass only one param The `$type` used to name the file could be replaced by the path basename.
> The plugin name if necessary could also be generated from the path with something like (I did not test): `ucwords( str_replace( '-', ' ', basename( dirname( __FILE__ . "/$base_path" ) ) ) )`

bin/Globber.php
Screenfeed: @Chouby what do you think? Does it worth adding a new folder yet?

bin/generate-globs.php
Hug0-Drelon: It can be skipped here.
And improved in an other PR for the others.
But yes, I'd get rid of it. It annoys me each time I create a new file, all of that for nothing.
I add no value and doesn't even respect its purpose (which is helping namespace resolving in generated doc).

bin/Globber.php
Hug0-Drelon: Then do `error_log( sort( $file_paths, \SORT_STRING | \SORT_FLAG_CASE ) );` while debugging.

bin/Globber.php
Hug0-Drelon: I've seen that, but you can still remove it.
As you see fit.

bin/generate-globs.php
Screenfeed: Done in [67136cb](https://github.com/polylang/polylang/pull/1717/commits/67136cb9e2465e57652e0837c52b963ea7873c82).

bin/Globber.php
Screenfeed: Done in [b0f056f](https://github.com/polylang/polylang/pull/1717/commits/b0f056f80135002054d5c5a383bb9d2e21120670).

bin/Globber.php
Screenfeed: Done in [1ff57af](https://github.com/polylang/polylang/pull/1717/commits/1ff57afe4334260795550ab1755a7c6189c0d437).

bin/Loader_Paths_Finder.php
```diff
@@ -0,0 +1,66 @@
...
+		$plugin_name = str_replace( '-', ' ', $args[1] );
+		$type        = rtrim( basename( $base_path ), 's' );
+		$file_paths  = glob( "{$base_path}/*/load.php", \GLOB_NOSORT );
+
+		if ( ! is_array( $file_paths ) || empty( $file_paths ) ) {
+			throw new RuntimeException( "Could not retrieve the {$type} files in {$plugin_name}" );
+		}
+
+		$write = "<?php
+/**
```
Hug0-Drelon, line 45:
```suggestion
/*
```

bin/Loader_Paths_Finder.php
```diff
@@ -0,0 +1,66 @@
...
+
+		if ( empty( $args[0] ) ) {
+			throw new RuntimeException( 'No path to folder provided' );
+		}
+		if ( empty( $args[1] ) ) {
+			throw new RuntimeException( 'No package name provided' );
+		}
+
+		$base_path   = realpath( $args[0] );
+		$plugin_name = str_replace( '-', ' ', $args[1] );
```
Hug0-Drelon: I don't think the variable is useful, it's only used for messages. Only error messages should be returned IMHO, using abspath maybe.
It would simplify the code and remove the extra argument.

composer.json
```diff
@@ -61,7 +64,12 @@
 		],
 		"rector": "vendor/bin/rector process",
 		"build": "bin/build.sh",
-		"dist": "bin/distribute.sh"
+		"dist": "bin/distribute.sh",
+		"glob": "WP_Syntex\\Polylang\\Script\\Loader_Paths_Finder::hard_code",
```
Hug0-Drelon:
```suggestion
		"build-loader": "WP_Syntex\\Polylang\\Script\\Loader_Paths_Finder::hard_code",
```
Juste suggesting, not very fond of the name I propose. That said it should follow the class name and be descriptive.

bin/Loader_Paths_Finder.php
Screenfeed:
```
FILE: /polylang/integrations/integration-build.php
------------------------------------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------
 2 | ERROR | You must use "/**" style comments for a file comment (Squiz.Commenting.FileComment.WrongStyle)
------------------------------------------------------------------------------------------------------------
```

Hug0-Drelon: This is no file comment but just comment. That said, let's keep it simple.

bin/Loader_Paths_Finder.php
Chouby: +1
I also would remove this extra argument to simplify.

bin/Loader_Paths_Finder.php
```diff
@@ -0,0 +1,66 @@
...
+		$type        = rtrim( basename( $base_path ), 's' );
+		$file_paths  = glob( "{$base_path}/*/load.php", \GLOB_NOSORT );
```
Chouby: I don't remember why I initially used this argument but we should be able to remove it without any risk here. 
```suggestion
		$file_paths  = glob( "{$base_path}/*/load.php" );
```

bin/Loader_Paths_Finder.php
```diff
@@ -0,0 +1,66 @@
...
+
+		if ( ! is_array( $file_paths ) || empty( $file_paths ) ) {
+			throw new RuntimeException( "Could not retrieve the {$type} files in {$plugin_name}" );
```
Chouby: Allowing to remove the second argument, usign @Hug0-Drelon's suggestion to use the path:
```suggestion
			throw new RuntimeException( "Could not retrieve the files from {$base_path}" );
```

bin/Loader_Paths_Finder.php
```diff
@@ -0,0 +1,66 @@
...
+
+return array(\n";
+
+		foreach ( $file_paths as $path ) {
+			$write .= "\t'" . basename( dirname( $path ) ) . "',\n";
+		}
+		$write .= ");\n";
+
+		if ( false === file_put_contents( "{$base_path}/{$type}-build.php", $write ) ) {
+			throw new RuntimeException( "Error while writing the data file for {$type}s in {$plugin_name}" );
```
Chouby:
```suggestion
			throw new RuntimeException( "Error while writing the loader file to {$base_path}" );
```

bin/Loader_Paths_Finder.php
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		echo "\e[32mGenerated {$type}s data file for {$plugin_name}\e[0m\n";
```
Chouby:
```suggestion
		echo "\e[32mGenerated loader file in {$base_path}\e[0m\n";
```

composer.json
Chouby: `generate-loader`?

composer.json
```diff
@@ -76,7 +84,8 @@
 		"lint": "Runs both PHPCS and PHPStan.",
 		"rector": "Runs Rector.",
 		"build": "Builds the project.",
-		"dist": "Make the zip file to distribute the project release."
+		"dist": "Make the zip file to distribute the project release.",
+		"glob": "Hard-code calls to glob()."
```
Chouby:
```suggestion
		"dist": "Makes the zip file to distribute the project release.",
		"glob": "Generates a list of files to load."
```

integrations/integrations.php
```diff
@@ -50,9 +50,10 @@ public static function instance(): self {
...
 	 */
 	protected function init(): void {
-		// Loads external integrations.
-		foreach ( glob( __DIR__ . '/*/load.php', GLOB_NOSORT ) as $load_script ) { // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
-			require_once $load_script; // phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.UsingVariable
+		$load_scripts = require __DIR__ . '/integration-build.php';
+
+		foreach ( $load_scripts as $load_script ) {
+			require_once __DIR__ . "/{$load_script}/load.php";
 		}
```
Chouby, line 57: What about moving these lines to the built files, thus avoiding duplication?

phpcs.xml.dist
```diff
@@ -176,6 +180,7 @@
 	<exclude-pattern>rector.php</exclude-pattern>
 
 	<!-- Specific to Polylang -->
+	<exclude-pattern>bin/*.php</exclude-pattern>
```
Chouby, line 183: Why not following CS in this file?

bin/Globber.php
Chouby: I prefer to keep not to multiply the number of folders. If you don't wand to keep `/bin`, this could be renamed to `/lib` as in Polylang Pro, provided that this doesn't require many changes.

bin/Loader_Paths_Finder.php
Screenfeed: Probably for better performances.

phpcs.xml.dist
Screenfeed: Because of https://github.com/polylang/polylang/pull/1717#discussion_r2318106571

integrations/integrations.php
Screenfeed: 1. It won't avoid duplication, just move it to the loader file.
2. We won't be able to use it in `uninstall.php` anymore (Pll Pro).

composer.json
Screenfeed: I used `build-loader` because it makes more sense to me and we already use this term in the final file.
Now I can bring a ring and mud so you can fight over it and decide.

bin/Globber.php
Screenfeed: @Chouby we let you decide if we keep `/bin` or rename to `/lib`. There's not much to modify: `.distignore`, `composer.json`, `phpcs.xml.dist`.

PR #1718, opened 2025-08-26
Prevent a translations group extra update

manooweb:
## What?

The translations group is updated twice when removing all the translations. 

## Why?
- The $id parameter passed to the `save_translations()` method [here](https://github.com/polylang/polylang/blob/3.7.3/include/translated-object.php#L149) is overriden [here](https://github.com/polylang/polylang/blob/3.7.3/include/translated-object.php#L168-L170) when removing all translations
- It makes `should_update_translation_group()` check the wrong `id`, not return `false` as expected and not return early the translations group already updated but run the translations group update process.

## How?
- Revert code introduced by #1159
- Add tests to cover the case for post and term


tests/phpunit/tests/test-translated-post.php
Hug0-Drelon, line 1: I suggest you check for `saved_term` filter calls instead of our own hook `saved_post_translations`.
The reason why is that our own hook is implementation detail and could be moved around without your test noticing there is a regression. Also, `sacve_term` hook has the courtesy of being called both in `wp_inster_term()` and `wp_update_term()`.

Same for `PLL_Translated_Term`.

Also, we have a test namespace for performance (`WP_Syntex\Polylang\Tests\Integration\Performances`), maybe your tests could fit in it.

manooweb: > I suggest you check for `saved_term` filter calls instead of our own hook `saved_post_translations`. The reason why is that our own hook is implementation detail and could be moved around without your test noticing there is a regression. Also, `sacve_term` hook has the courtesy of being called both in `wp_inster_term()` and `wp_update_term()`.
> 

Not sure to understand what you mean 🤔 The goal is to not count action that are not concerned (on other taxonomy) during the process. 

In addition `saved_{$taxonomy}` action is also called both in `wp_inster_term()` and `wp_update_term()`.
https://developer.wordpress.org/reference/hooks/saved_taxonomy/

Moved to Performances folder in ef95c72d22b5771750c4e3629ffade55da9d9a48

Hug0-Drelon: I misread our code and thought `save_post_translations` was one of our hook. My suggestion regarding `save_term` is irrelevant then.

Hug0-Drelon (changes requested):
Please update filename according to new class name.

manooweb:
> Please update filename according to new class name.
Done in d3868ba2e85f337d776942bcc2cf21c63481b798

Hug0-Drelon (approved):

tests/phpunit/tests/Performances/test-Post_Translations_Group.php
```diff
@@ -0,0 +1,49 @@
...
+	public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
+		parent::pllSetUpBeforeClass( $factory );
+
+		$factory->language->create_many( 2 );
+	}
+
+	/**
+	 * Checks that updating a post translations group is done only once when we unlink all translations.
+	 */
+	public function test_should_not_update_translations_group_when_removing_all_translations() {
```
Chouby: OK for the fix. However I would not insert this test in a separate folder.
1. This is not a performance test (because the performance is not measured)
2. It would be separated from related tests and future us would forget its existence.

So I suggest that you move this test to the existing related class. You'll see that you can also simplify it, i.e; remove some assertions as they are already checked in another test. See https://github.com/polylang/polylang/blob/master/tests/phpunit/tests/test-translated-post.php#L65-L69 (this one could be completed as there are only 6/9 assertions).

Chouby (changes requested):

tests/phpunit/tests/Performances/test-Post_Translations_Group.php
Hug0-Drelon: Oops... Sorry @manooweb for pointing in the wrong direction.
That said we do not mesure performance at all in any of the tests from `WP_Syntex\Polylang\Tests\Integration\Performances`.

manooweb: Existing test completed and new ones moved back to existing classes in a97ed3b92b03467cc99b94ab1768653a0f17d51f & f0bab4afda7dc86f8167250663c270d430fa896b

PR #1719, opened 2025-08-29
Improve error management for in `Model\Languages`.

Hug0-Drelon:
## What?
Improve `Model\Languages::add()` and `Model\Languages::update()` API.
Prop @Chouby in Slack: https://wpsyntex.slack.com/archives/C017SSTK4KY/p1756457865061609

## Why?
Low-level errors (from `wp_inster_term()` or `wp_update_term()` for instance) are not managed properly.

## How?
- Make errors from `wp_inster_term()` and `wp_update_term()` bubble in `Model\Languages::add()` and `Model\Languages::update()`.
- Improve `Model\Languages::update_secondary_language_terms()` and `Model\Languages::update_translations()` so they both return `WP_Error` as well and thus benefiting `Model\Languages::add()` and `Model\Languages::update()`.

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -204,13 +204,13 @@ class Languages {
 		);
 		if ( is_wp_error( $result ) ) {
 			/* translators: %s is the detailed error message */
-			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
+			return new WP_Error( 'pll_add_language', __( 'Could not add the language.', 'polylang' ) ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
 		}
 
 		$result = wp_update_term( (int) $result['term_id'], 'language', array( 'term_group' => (int) $args['term_group'] ) ); // Can't set the term group directly in `wp_insert_term()`.
 		if ( is_wp_error( $result ) ) {
 			/* translators: %s is the detailed error message */
-			return new WP_Error( 'pll_add_language', __( 'Impossible to set the language order.', 'polylang' ) );
+			return new WP_Error( 'pll_add_language', __( 'Could not set the language order.', 'polylang' ) );
 		}
 
 		// The other language taxonomies.
@@ -338,8 +338,7 @@ class Languages {
 			)
 		);
 		if ( is_wp_error( $result ) ) {
-			/* translators: %s is the error message from the database */
-			return new WP_Error( 'pll_update_language', __( 'Impossible to update the language, error: %s.', 'polylang' ), $result->get_error_message() );
+			return new WP_Error( 'pll_update_language', __( 'Could not update the language.', 'polylang' ) );
 		}
 
 		if ( $old_slug !== $slug ) {
@@ -1025,7 +1024,6 @@ class Languages {
 				)
 			);
 			if ( false === $result ) {
-				/* translators: %s is the error message from the database */
 				$errors->add( 'pll_delete_relationships', __( 'Could not delete relationships.', 'polylang' ) );
 			}
 		}
@@ -1042,7 +1040,6 @@ class Languages {
 				)
 			);
 			if ( false === $result ) {
-				/* translators: %s is the error message from the database */
 				$errors->add( 'pll_delete_terms', __( 'Could not delete translation groups.', 'polylang' ) );
 			}
 
@@ -1056,7 +1053,6 @@ class Languages {
 				)
 			);
 			if ( false === $result ) {
-				/* translators: %s is the error message from the database */
 				$errors->add( 'pll_delete_term_taxonomy', __( 'Could not delete translation groups.', 'polylang' ) );
 			}
 		}
@@ -1074,7 +1070,6 @@ class Languages {
 				)
 			);
 			if ( false === $result ) {
-				/* translators: %s is the error message from the database */
 				$errors->add( 'pll_update_term_taxonomy', __( 'Could not update translation groups.', 'polylang' ), $wpdb->last_error );
 			}
 		}
@@ -1128,7 +1123,7 @@ class Languages {
 					$errors->add(
 						'pll_add_secondary_language_terms',
 						/* translators: %s is a taxonomy name */
-						sprintf( __( 'Could not add secondary %s language taxonomy term.', 'polylang' ), $object->get_tax_language() )
+						sprintf( __( 'Could not add secondary language term for taxonomy %s.', 'polylang' ), $object->get_tax_language() )
 					);
 				}
 				continue;
@@ -1142,7 +1137,7 @@ class Languages {
 					$errors->add(
 						'pll_update_secondary_language_terms',
 						/* translators: %s is a taxonomy name */
-						sprintf( __( 'Could not update secondary %s language taxonomy term.', 'polylang' ), $object->get_tax_language() )
+						sprintf( __( 'Could not update secondary language term for taxonomy %s.', 'polylang' ), $object->get_tax_language() )
 					);
 				}
 			}
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -203,13 +203,13 @@ class Languages {
 			)
 		);
 		if ( is_wp_error( $result ) ) {
-			/* translators: %s is the error message from the database */
+			/* translators: %s is the detailed error message */
 			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language, database error: %s.', 'polylang' ), $result->get_error_message() ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
 		}
 
 		$result = wp_update_term( (int) $result['term_id'], 'language', array( 'term_group' => (int) $args['term_group'] ) ); // Can't set the term group directly in `wp_insert_term()`.
 		if ( is_wp_error( $result ) ) {
-			/* translators: %s is the error message from the database */
+			/* translators: %s is the detailed error message */
 			return new WP_Error( 'pll_add_language', __( 'Impossible to set the term group, database error: %s.', 'polylang' ), $result->get_error_message() );
 		}
 
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -339,7 +339,7 @@ class Languages {
 		);
 		if ( is_wp_error( $result ) ) {
 			/* translators: %s is the error message from the database */
-			return new WP_Error( 'pll_update_language', __( 'Impossible to update the language, database error: %s.', 'polylang' ), $result->get_error_message() );
+			return new WP_Error( 'pll_update_language', __( 'Impossible to update the language, error: %s.', 'polylang' ), $result->get_error_message() );
 		}
 
 		if ( $old_slug !== $slug ) {
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -213,7 +213,10 @@ class Languages {
 		}
 
 		// The other language taxonomies.
-		$this->update_secondary_language_terms( $args['slug'], $args['name'] );
+		$errors = $this->update_secondary_language_terms( $args['slug'], $args['name'] );
+		if ( $errors->has_errors() ) {
+			return $errors;
+		}
 
 		if ( empty( $this->options['default_lang'] ) ) {
 			// If this is the first language created, set it as default language
@@ -318,7 +321,10 @@ class Languages {
 		$old_slug = $lang->slug;
 
 		// Update the language itself.
-		$this->update_secondary_language_terms( $args['slug'], $args['name'], $lang );
+		$errors = $this->update_secondary_language_terms( $args['slug'], $args['name'], $lang );
+		if ( $errors->has_errors() ) {
+			return $errors;
+		}
 
 		$result = wp_update_term(
 			$lang->get_tax_prop( 'language', 'term_id' ),
@@ -1066,14 +1072,15 @@ class Languages {
 	 * @param PLL_Language|null $language   Optional. A language object. Required to update the existing terms.
 	 * @param string[]          $taxonomies Optional. List of language taxonomies to deal with. An empty value means
 	 *                                      all of them. Defaults to all taxonomies.
-	 * @return void
+	 * @return WP_Error
 	 *
 	 * @phpstan-param non-empty-string $slug
 	 * @phpstan-param non-empty-string $name
 	 * @phpstan-param array<non-empty-string> $taxonomies
 	 */
-	protected function update_secondary_language_terms( $slug, $name, ?PLL_Language $language = null, array $taxonomies = array() ): void {
+	protected function update_secondary_language_terms( $slug, $name, ?PLL_Language $language = null, array $taxonomies = array() ): WP_Error {
 		$slug = 0 === strpos( $slug, 'pll_' ) ? $slug : "pll_$slug";
+		$errors = new WP_Error();
 
 		foreach ( $this->translatable_objects->get_secondary_translatable_objects() as $object ) {
 			if ( ! empty( $taxonomies ) && ! in_array( $object->get_tax_language(), $taxonomies, true ) ) {
@@ -1089,16 +1096,24 @@ class Languages {
 
 			if ( empty( $term_id ) ) {
 				// Attempt to repair the language if a term has been deleted by a database cleaning tool.
-				wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				$result = wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add( 'pll_add_secondary_language_terms', $result->get_error_message() );
+				}
 				continue;
 			}
 
 			/** @var PLL_Language $language */
 			if ( "pll_{$language->slug}" !== $slug || $language->name !== $name ) {
 				// Something has changed.
-				wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
+				$result = wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add( 'pll_update_secondary_language_terms', $result->get_error_message() );
+				}
 			}
 		}
+
+		return $errors;
 	}
 
 	/**
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -204,13 +204,13 @@ class Languages {
 		);
 		if ( is_wp_error( $result ) ) {
 			/* translators: %s is the detailed error message */
-			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language, error: %s.', 'polylang' ), $result->get_error_message() ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
+			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
 		}
 
 		$result = wp_update_term( (int) $result['term_id'], 'language', array( 'term_group' => (int) $args['term_group'] ) ); // Can't set the term group directly in `wp_insert_term()`.
 		if ( is_wp_error( $result ) ) {
 			/* translators: %s is the detailed error message */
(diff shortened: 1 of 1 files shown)


include/Model/Languages.php
```diff
@@ -1039,13 +1066,18 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					array_merge( array_merge( ...$ut['case'] ), $ut['in'] )
 				)
 			);
+			if ( false === $result ) {
+				$errors->add( 'pll_update_term_taxonomy', __( 'Impossible to update the term taxonomy.', 'polylang' ) );
```
Marianne380: Why don't we use `$wpdb->last_error` ?

Marianne380 (changes requested):

include/Model/Languages.php
Hug0-Drelon: I didn't think about it. I like the idea!
See what I propose in 8a587719.

include/Model/Languages.php
```diff
@@ -1082,16 +1121,24 @@ protected function update_secondary_language_terms( $slug, $name, ?PLL_Language
 
 			if ( empty( $term_id ) ) {
 				// Attempt to repair the language if a term has been deleted by a database cleaning tool.
-				wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				$result = wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add( 'pll_add_secondary_language_terms', $result->get_error_message() );
```
Marianne380: Just so I know, why do we sometimes create our own error message and sometimes use `$result->get_error_message()`?

Hug0-Drelon: I think it depends on the scope.

But if we get a proper error message directly from WordPress we can use it.
Sometime we create our own when we want a specific message to display directly to the user.

As of today, `WP_Error` you see here won't be displayed to the end user anyway.

Do you have any example a an error we get and don't use?

Marianne380: Here https://github.com/polylang/polylang/pull/1719/files#diff-f612298d4b72d331be1c8327d35c40836aa6028ca7e4ef3d18bc0ac0a1603371R209-R213 ?

Hug0-Drelon: Then I made a mistake.
Fixed in c2f2753f658ec323b090c008bad13bc56000f252 and 3748591ec5459dca539955a2ada2744b4c56416e.

include/Model/Languages.php
```diff
@@ -194,22 +194,30 @@ public function add( $args ) {
...
 			array(
 				'slug'        => $args['slug'],
 				'description' => $this->build_metas( $args ),
 			)
 		);
-		if ( is_wp_error( $r ) ) {
-			// Avoid an ugly fatal error if something went wrong (reported once in the forum).
-			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) );
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the error message from the database */
```
Marianne380:
```suggestion
			/* translators: %s is the error message from the database */
```
Should we keep from the database ?

Hug0-Drelon: I guess. What would you suggest instead?

Marianne380: I don't know but most of the time it's not coming from the database in `wp_insert/update_term()`.

Marianne380: Maybe : 
`/* translators: %s is the error message returned by WordPress */`
Or : 
`/* translators: %s is the detailed error message */`

Hug0-Drelon: I like your second proposal. Applied in bbdb2f17.

include/Model/Languages.php
```diff
@@ -194,22 +194,30 @@ public function add( $args ) {
...
 				'slug'        => $args['slug'],
 				'description' => $this->build_metas( $args ),
 			)
 		);
-		if ( is_wp_error( $r ) ) {
-			// Avoid an ugly fatal error if something went wrong (reported once in the forum).
-			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) );
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the detailed error message */
+			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language, database error: %s.', 'polylang' ), $result->get_error_message() ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
```
Marianne380: Then I guess we should remove **_database_** from the error message too ?

Marianne380 (changes requested):

include/Model/Languages.php
Hug0-Drelon: 👍 
Done in 5570582c.

include/Model/Languages.php
```diff
@@ -326,10 +337,17 @@ public function update( $args ) {
 				'term_group'  => (int) $args['term_group'],
 			)
 		);
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the error message from the database */
+			return new WP_Error( 'pll_update_language', __( 'Impossible to update the language, database error: %s.', 'polylang' ), $result->get_error_message() );
```
Marianne380:
```suggestion
			return new WP_Error( 'pll_update_language', __( 'Impossible to update the language, error: %s.', 'polylang' ), $result->get_error_message() );
```

Marianne380 (changes requested):

include/Model/Languages.php
```diff
@@ -194,22 +194,30 @@ public function add( $args ) {
...
 				'slug'        => $args['slug'],
 				'description' => $this->build_metas( $args ),
 			)
 		);
-		if ( is_wp_error( $r ) ) {
-			// Avoid an ugly fatal error if something went wrong (reported once in the forum).
-			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) );
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the detailed error message */
+			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language, error: %s.', 'polylang' ), $result->get_error_message() ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
```
Chouby: The placeholders let me think that you forgot `sprintf()` everywhere in your code.
That said, I'm not fan to display internal errors information to the user.
Moreover, in this case, if we wrote correctly our code, the error should always be https://github.com/WordPress/WordPress/blob/6.8.2/wp-includes/taxonomy.php#L2612
I'm however not against adding extra information to the `WP_Error` object (what your code currently does without `sprintf()`. The error code or the error data could be more interesting than the error message though.

include/Model/Languages.php
```diff
@@ -194,22 +194,30 @@ public function add( $args ) {
...
+		}
+
+		$result = wp_update_term( (int) $result['term_id'], 'language', array( 'term_group' => (int) $args['term_group'] ) ); // Can't set the term group directly in `wp_insert_term()`.
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the detailed error message */
+			return new WP_Error( 'pll_add_language', __( 'Impossible to set the term group, error: %s.', 'polylang' ), $result->get_error_message() );
```
Chouby: For the error message:
```suggestion
			return new WP_Error( 'pll_add_language', __( 'Impossible to set the language order', 'polylang' ) );
```
I let you decide if you want to add more info in the error data.

include/Model/Languages.php
```diff
@@ -1003,11 +1024,15 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					array_merge( $dr['id'], $dr['tt'] )
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_delete_relationships', __( 'Impossible to delete the relationships, database error: %s.', 'polylang' ), $wpdb->last_error );
```
Chouby:
```suggestion
				$errors->add( 'pll_delete_relationships', __( 'Could not delete relationships.', 'polylang' ), $wpdb->last_error );
```

include/Model/Languages.php
```diff
@@ -1016,7 +1041,12 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					$dt['t']
 				)
 			);
-			$wpdb->query(
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_delete_terms', __( 'Impossible to delete the terms, database error: %s.', 'polylang' ), $wpdb->last_error );
```
Chouby:
```suggestion
				$errors->add( 'pll_delete_terms', __( 'Could not delete translation groups.', 'polylang' ), $wpdb->last_error );
```

include/Model/Languages.php
```diff
@@ -1025,11 +1055,15 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					$dt['tt']
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_delete_term_taxonomy', __( 'Impossible to delete the term taxonomy, database error: %s.', 'polylang' ), $wpdb->last_error );
```
Chouby: Users don't know `term_taxonomy`:
```suggestion
				$errors->add( 'pll_delete_term_taxonomy', __( 'Could not delete translation groups.', 'polylang' ), $wpdb->last_error );
```

include/Model/Languages.php
```diff
@@ -1039,13 +1073,19 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					array_merge( array_merge( ...$ut['case'] ), $ut['in'] )
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_update_term_taxonomy', __( 'Impossible to update the term taxonomy, database error: %s.', 'polylang' ), $wpdb->last_error );
```
Chouby:
```suggestion
				$errors->add( 'pll_update_term_taxonomy', __( 'Could not update translation groups.', 'polylang' ), $wpdb->last_error );
```

include/Model/Languages.php
```diff
@@ -1082,16 +1123,24 @@ protected function update_secondary_language_terms( $slug, $name, ?PLL_Language
 
 			if ( empty( $term_id ) ) {
 				// Attempt to repair the language if a term has been deleted by a database cleaning tool.
-				wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				$result = wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add( 'pll_add_secondary_language_terms', $result->get_error_message() );
```
Chouby: To copy the error, you can use [`merge_from()`](https://developer.wordpress.org/reference/classes/wp_error/merge_from/). Otherwise here you are losing the error data if any (likely as if our code is well written, this a db insert error.

include/Model/Languages.php
```diff
@@ -1082,16 +1123,24 @@ protected function update_secondary_language_terms( $slug, $name, ?PLL_Language
...
 				continue;
 			}
 
 			/** @var PLL_Language $language */
 			if ( "pll_{$language->slug}" !== $slug || $language->name !== $name ) {
 				// Something has changed.
-				wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
+				$result = wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add( 'pll_update_secondary_language_terms', $result->get_error_message() );
```
Chouby: Same as above.

Chouby (changes requested):

include/Model/Languages.php
Hug0-Drelon: Damn how stupid I was to forgot *all* `sprintf()`...

Anyway, your suggestion about not using internal error message shows how bad `WP_Error` objects are at bubbling and managing errors in our code IMHO.
What message is displayed to the user should depend on the code that call these methods because `Model\Languages` is a low level utility.
For instance, in CLI it would be fine to have internal error information. In REST, maybe not. And in the interface surely not.
You suggest to pass `WP_Error::$error_data` instead, it could be a way of solving the issue I raised above. But it would be displayed to the user anyway in REST for example.

Hug0-Drelon: `$result->get_error_message()` removed in 827b3faf.

include/Model/Languages.php
Hug0-Drelon: Applied in 827b3faf.

include/Model/Languages.php
Hug0-Drelon: Applied in 827b3faf.

include/Model/Languages.php
Hug0-Drelon: Applied in Applied in 827b3faf..

include/Model/Languages.php
Hug0-Drelon: Applied in 827b3faf..

include/Model/Languages.php
Hug0-Drelon: Applied in 827b3faf.

include/Model/Languages.php
Hug0-Drelon: I don't want to copy error here but create a new one.
Even more so now you want message to be generic.
See what I propose in 827b3faf.

include/Model/Languages.php
Hug0-Drelon: Same proposal as above in 827b3faf.

include/Model/Languages.php
```diff
@@ -1039,13 +1073,19 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					array_merge( array_merge( ...$ut['case'] ), $ut['in'] )
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_update_term_taxonomy', __( 'Could not update translation groups.', 'polylang' ), $wpdb->last_error );
```
Chouby: The translators comment must be removed. 
You removed the `last_error` everywhere except here. Intentional?

include/Model/Languages.php
```diff
@@ -194,22 +194,30 @@ public function add( $args ) {
...
 				'slug'        => $args['slug'],
 				'description' => $this->build_metas( $args ),
 			)
 		);
-		if ( is_wp_error( $r ) ) {
-			// Avoid an ugly fatal error if something went wrong (reported once in the forum).
-			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) );
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the detailed error message */
+			return new WP_Error( 'pll_add_language', __( 'Impossible to add the language.', 'polylang' ) ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
```
Chouby: To follow the same pattern as other messages I suggested.
```suggestion
			return new WP_Error( 'pll_add_language', __( 'Could not add the language.', 'polylang' ) ); // Avoid an ugly fatal error if something went wrong (reported once in the forum).
```

include/Model/Languages.php
```diff
@@ -194,22 +194,30 @@ public function add( $args ) {
...
+		}
+
+		$result = wp_update_term( (int) $result['term_id'], 'language', array( 'term_group' => (int) $args['term_group'] ) ); // Can't set the term group directly in `wp_insert_term()`.
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the detailed error message */
+			return new WP_Error( 'pll_add_language', __( 'Impossible to set the language order.', 'polylang' ) );
```
Chouby:
```suggestion
			return new WP_Error( 'pll_add_language', __( 'Could not set the language order.', 'polylang' ) );
```

include/Model/Languages.php
```diff
@@ -326,10 +337,17 @@ public function update( $args ) {
 				'term_group'  => (int) $args['term_group'],
 			)
 		);
+		if ( is_wp_error( $result ) ) {
+			/* translators: %s is the error message from the database */
+			return new WP_Error( 'pll_update_language', __( 'Impossible to update the language, error: %s.', 'polylang' ), $result->get_error_message() );
```
Chouby: You forgot this one:
```suggestion
			return new WP_Error( 'pll_update_language', __( 'Could not update the language.', 'polylang' ) );
```

include/Model/Languages.php
```diff
@@ -1003,11 +1024,15 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					array_merge( $dr['id'], $dr['tt'] )
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_delete_relationships', __( 'Could not delete relationships.', 'polylang' ) );
```
Chouby:
```suggestion
				$errors->add( 'pll_delete_relationships', __( 'Could not delete relationships.', 'polylang' ) );
```

include/Model/Languages.php
```diff
@@ -1016,7 +1041,12 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					$dt['t']
 				)
 			);
-			$wpdb->query(
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_delete_terms', __( 'Could not delete translation groups.', 'polylang' ) );
```
Chouby:
```suggestion
				$errors->add( 'pll_delete_terms', __( 'Could not delete translation groups.', 'polylang' ) );
```

include/Model/Languages.php
```diff
@@ -1025,11 +1055,15 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					$dt['tt']
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
+				$errors->add( 'pll_delete_term_taxonomy', __( 'Could not delete translation groups.', 'polylang' ) );
```
Chouby:
```suggestion
				$errors->add( 'pll_delete_term_taxonomy', __( 'Could not delete translation groups.', 'polylang' ) );
```

include/Model/Languages.php
```diff
@@ -1039,13 +1073,19 @@ protected function update_translations( $old_slug, $new_slug = '' ): void {
 					array_merge( array_merge( ...$ut['case'] ), $ut['in'] )
 				)
 			);
+			if ( false === $result ) {
+				/* translators: %s is the error message from the database */
```
Chouby:
```suggestion
```

include/Model/Languages.php
```diff
@@ -1082,16 +1123,32 @@ protected function update_secondary_language_terms( $slug, $name, ?PLL_Language
 
 			if ( empty( $term_id ) ) {
 				// Attempt to repair the language if a term has been deleted by a database cleaning tool.
-				wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				$result = wp_insert_term( $name, $object->get_tax_language(), array( 'slug' => $slug ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add(
+						'pll_add_secondary_language_terms',
+						/* translators: %s is a taxonomy name */
+						sprintf( __( 'Could not add secondary %s language taxonomy term.', 'polylang' ), $object->get_tax_language() )
```
Chouby: Not sure it's better:
```suggestion
						sprintf( __( 'Could not add secondary language term for taxonomy %s.', 'polylang' ), $object->get_tax_language() )
```

include/Model/Languages.php
```diff
@@ -1082,16 +1123,32 @@ protected function update_secondary_language_terms( $slug, $name, ?PLL_Language
...
 			/** @var PLL_Language $language */
 			if ( "pll_{$language->slug}" !== $slug || $language->name !== $name ) {
 				// Something has changed.
-				wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
+				$result = wp_update_term( $term_id, $object->get_tax_language(), array( 'slug' => $slug, 'name' => $name ) );
+				if ( is_wp_error( $result ) ) {
+					$errors->add(
+						'pll_update_secondary_language_terms',
+						/* translators: %s is a taxonomy name */
+						sprintf( __( 'Could not update secondary %s language taxonomy term.', 'polylang' ), $object->get_tax_language() )
```
Chouby: Same as above:
```suggestion
						sprintf( __( 'Could not update secondary language term for taxonomy %s.', 'polylang' ), $object->get_tax_language() )
```

include/Model/Languages.php
Hug0-Drelon: It's a mistake. Removed in e2ef5eeb.

PR #1725, opened 2025-09-04
Only add translatable content IDs to AJAX filter data.

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/2715.

### What
Currently, `pll_post_id` and `pll_term_id` are automatically added to all admin AJAX requests via `get_ajax_filter_data()`, even when the content isn't translatable.

### Fix
- Check if post type is translatable before adding `pll_post_id`.
- Check if taxonomy is translatable before adding `pll_term_id`.
- Simplify array assignment (direct assignment vs `array_merge()`)

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -320,7 +320,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 
 		if ( ! empty( $tag_ID ) ) {
 			$term = get_term( (int) $tag_ID );
-			if ( $term instanceof WP_Term && $this->model->->taxonomies->is_translated( $term->taxonomy ) ) {
+			if ( $term instanceof WP_Term && $this->model->taxonomies->is_translated( $term->taxonomy ) ) {
 				$params['pll_term_id'] = (int) $tag_ID;
 			}
 		}
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -312,11 +312,17 @@ abstract class PLL_Admin_Base extends PLL_Base {
 
 		$params = array( 'pll_ajax_backend' => 1 );
 		if ( ! empty( $post_ID ) ) {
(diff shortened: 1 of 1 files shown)


admin/admin-base.php
```diff
@@ -312,11 +312,17 @@ public function get_ajax_filter_data(): array {
 
 		$params = array( 'pll_ajax_backend' => 1 );
 		if ( ! empty( $post_ID ) ) {
-			$params = array_merge( $params, array( 'pll_post_id' => (int) $post_ID ) );
+			$post = get_post( (int) $post_ID );
+			if ( $post instanceof WP_Post && $this->model->is_translated_post_type( $post->post_type ) ) {
```
Hug0-Drelon:
```suggestion
			if ( $post instanceof WP_Post && $this->model->post_types->is_translated( $post->post_type ) ) {
```
Follows new `Model\Post_Types`.

admin/admin-base.php
```diff
@@ -312,11 +312,17 @@ public function get_ajax_filter_data(): array {
...
+				$params['pll_post_id'] = (int) $post_ID;
+			}
 		}
 
 		if ( ! empty( $tag_ID ) ) {
-			$params = array_merge( $params, array( 'pll_term_id' => (int) $tag_ID ) );
+			$term = get_term( (int) $tag_ID );
+			if ( $term instanceof WP_Term && $this->model->is_translated_taxonomy( $term->taxonomy ) ) {
```
Hug0-Drelon:
```suggestion
			if ( $term instanceof WP_Term && $this->model->->taxonomies->is_translated( $term->taxonomy ) ) {
```
Follows new `Model\Taxonomies`.

Chouby (changes requested):
Now that we need to access the object, it's simpler to use directly the global `$post` or `$tag` instead of getting them from the ID.

PR #1727, opened 2025-09-09
[Capabilities] String translations screen

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2731.

This allows to customize who can access Polylang' settings screens, more specifically the String Translations screen.

## How

1. The String Translations screen now requires the user capability `manage_translations`, which is mapped to `manage_options` by default.
2. The class `Capabilities` maps `manage_translations` back to `manage_options`: this means nothing changes for the current users, they don't need to have this capability. This makes `manage_translations` an alias to be used by 3rd party plugins or Polylang Pro. It could also be handled as a real capability.
3. The languages list can be filtered by registering an object that implements `Languages_Proxy_Interface`. This will be used in Polylang Pro to filter the list depending on the current user capabilities.
4. The hooks in our settings screens are harmonized: for each hook name containing `toplevel`, a similar hook containing `languages` is also fired. A similar thing is done for the class added to the `<body>` class.
5. The "cleanup translations" checkbox is displayed to admin users only.

> [!NOTE]
> A user can get an empty list of languages: on a site with EN + FR, but the user has the capability `translate_it` (the IT language may have been removed). In such case they are a translator, but no active languages match the capability. It is on purpose, in the such case the person is "still" a translator so they mustn't have access to other languages.


include/Model/Languages.php
```diff
@@ -608,6 +616,11 @@ public function get_list( $args = array() ): array {
 			 */
 			$languages = apply_filters_deprecated( 'pll_after_languages_cache', array( $languages ), '3.4' );
 
+			// Filters that don't depend on arguments can be cached.
+			foreach ( $this->list_filters as $filter ) {
+				$languages = $filter->apply_before_cache( $languages );
+			}
```
Hug0-Drelon: As I understand this is useless for now?
If so, it should be removed.

include/Model/Abstract_Languages_Filter.php
```diff
@@ -0,0 +1,42 @@
...
+	 * Filters a list of languages before putting it into local cache.
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language[] $languages List of languages to filter.
+	 * @return PLL_Language[]
+	 */
+	public function apply_before_cache( array $languages ): array {
+		return $languages;
+	}
```
Hug0-Drelon: As I understand this is useless for now?
If so, it should be removed.

include/Model/Languages.php
```diff
@@ -547,9 +554,10 @@ public function has(): bool {
 	 * @since 3.7 Moved from `PLL_Model::get_languages_list()` to `WP_Syntex\Polylang\Model\Languages::get_list()`.
 	 *
 	 * @param array $args {
-	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`).
-	 *   @type bool   $hide_default Hides default language from the list (default to `false`).
-	 *   @type string $fields       Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 *   @type bool   $hide_empty    Hides languages with no posts if set to `true` (defaults to `false`).
+	 *   @type bool   $hide_default  Hides default language from the list (default to `false`).
+	 *   @type string $fields        Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 *   @type int    $translator_id Optional. ID of the user. `0` for the current user. Used only in Polylang Pro.
```
Hug0-Drelon: I think this should be brought to the team and discussed.

I see several drawbacks:
- This makes `Model\Languages` aware of implementation details it shouldn't be aware of (even more for code in Pro)
- Adding a new argument again, even if it's in an array, is a bad idea. We've seen that having too many arguments in methods leads to problems. This is code smell IMHO.
- It forces you to have things like `$this->model->get_languages_list( array( 'translator_id' = 0 );` which seems really unnatural. It is understandable for someone that has read you code, but not for anyone.

Instead, I suggest you have a real proxy. Fetched with `$this->mode->languages->proxy( 'translator' )->get_list()` or something.

include/Model/Abstract_Languages_Filter.php
Screenfeed: Since it is not used anymore, I suppose we can remove it for now.
Done in [d2edd4a](https://github.com/polylang/polylang/pull/1727/commits/d2edd4a8cdbde6258cf0796f908429b62406f14d).

include/Model/Languages.php
Screenfeed: Done in [d2edd4a](https://github.com/polylang/polylang/pull/1727/commits/d2edd4a8cdbde6258cf0796f908429b62406f14d).

include/Model/Abstract_Languages_Filter.php
Hug0-Drelon, line 1: It should be an interface now it has only one method.

Screenfeed: I think we should keep it as an abstract class:
- With an interface, if we add a method, it will create a fatal error for classes that are not up-to-date with the latest changes.
- With an abstract class it won't, unless the new method is abstract of course.

So here, I'm thinking about 3rd party plugins that would use this. IMHO a class seems more "forward-compatibility"-proof.

Hug0-Drelon: > With an interface, if we add a method, it will create a fatal error for classes that are not up-to-date with the latest changes.

This is what we seek here.
If a third party extends it, I expect it follows the proper interface and not make it "work" by chance.

Hug0-Drelon: Let's @Chouby choose on this one.

admin/admin-base.php
```diff
@@ -111,46 +111,70 @@ public function init() {
...
-		 * @param array $tabs list of tab names
+		 * @see get_plugin_page_hookname()
 		 */
-		$tabs = apply_filters( 'pll_settings_tabs', $tabs );
-
-		$parent = '';
+		foreach ( array( 'load-', 'admin_print_styles-', 'admin_print_scripts-', 'admin_head-', '', 'admin_print_footer_scripts-', 'admin_footer-' ) as $prefix ) {
+			add_action(
+				"{$prefix}toplevel_page_{$parent}",
+				function () use ( $prefix, $page_type, $parent ) {
```
Hug0-Drelon:
```suggestion
				static function () use ( $prefix, $page_type, $parent ) {
```
Since it's called in a loop, let's make it the less impactful possible.

admin/admin-base.php
```diff
@@ -111,46 +111,70 @@ public function init() {
...
+		 * to the body. This adds a class with the `languages` prefix. This ensures we have a common CSS class, even if
+		 * the screen is moved to the 1st position in the menu.
+		 */
+		add_action(
+			// Target the screen in 1st position only.
+			"admin_head-toplevel_page_{$parent}",
+			function () use ( $page_type, $parent ) {
+				add_filter(
+					'admin_body_class',
+					function ( $admin_body_classes ) use ( $page_type, $parent ) {
```
Hug0-Drelon:
```suggestion
					static function ( $admin_body_classes ) use ( $page_type, $parent ) {
```

admin/admin-base.php
```diff
@@ -566,4 +590,84 @@ public function should_hide_admin_bar_menu(): bool {
...
+	 * @phpstan-return array<non-empty-string, array{label: string, capability: non-empty-string}>
+	 */
+	protected function get_sub_menu_items(): array {
+		$tabs = array(
+			'lang' => __( 'Languages', 'polylang' ),
+		);
+
+		// Only if at least one language has been created.
+		if ( $this->model->has_languages() ) {
+			// For better backward compatibility, the new capability is not added here.
```
Hug0-Drelon:
```suggestion
			// For better backward compatibility with Polylang < 3.8, the capability is not added here.
```
New now, not for long.

admin/admin-base.php
```diff
@@ -566,4 +590,84 @@ public function should_hide_admin_bar_menu(): bool {
...
+		 * Filter the list of sub-menu items in Polylang settings.
+		 *
+		 * @since 1.5.1
+		 * @since 3.8 Added the possibility to pass user capabilities.
+		 *
+		 * @param string[]|string[][] $tabs {
+		 *     List of sub-menu items with page slugs as array keys. Can be an array of titles, or an array of arrays as follow:
+		 *
+		 *     @type string $label      Tab title.
+		 *     @type string $capability Optional. User capability. Default is `manage_options`.
```
Hug0-Drelon: I don't think this is useful.

include/Capabilities/Capabilities.php
```diff
@@ -0,0 +1,41 @@
...
+ * @since 3.8
+ */
+class Hooks {
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.8
+	 */
+	public function __construct() {
+		add_filter( 'map_meta_cap', array( $this, 'map_custom_caps' ), 1, 2 );
```
Hug0-Drelon, line 22:
```suggestion
		defined( 'POLYLANG_PRO' ) || add_filter( 'map_meta_cap', array( $this, 'map_custom_caps' ), 1, 2 );
```
Should we have something like that?

include/Model/Languages.php
```diff
@@ -624,6 +632,11 @@ function ( $lang ) use ( $args ) {
 			}
 		);
 
+		// Filters that depend on arguments must be applied after the cache.
```
Hug0-Drelon: Not true, a cache can be primed using given arguments as key. See how a query is cached for instance.

include/Model/Languages.php
```diff
@@ -791,6 +804,19 @@ public function delete_transient_from_options_table(): void {
...
 		}
 	}
+
+	/**
+	 * Registers languages list filters.
+	 *
+	 * @since 3.8
+	 *
+	 * @param Abstract_Languages_Filter $filter Filter.
+	 * @return void
```
Hug0-Drelon:
```suggestion
	 * @return self
```
So we can chain?

settings/table-string.php
```diff
@@ -423,16 +439,21 @@ public function save_translations() {
 				}
 
 				isset( $new_mo ) ? $new_mo->export_to_db( $language ) : $mo->export_to_db( $language );
+				$done = true;
```
Hug0-Drelon: Not entirely true, maybe we should take the opportunity to modify `PLL_MO::export_to_db()` so it returns a boolean. Then you could have something like `$done = isset( $new_mo ) ? $new_mo->export_to_db( $language ) : $mo->export_to_db( $language );`.

That said, here you improve the strings table by returning a proper error to UI when the save process failed (partially as we've seen). This is a bit out of scope, so we may consider to move into another PR. But as you see fit.

settings/table-string.php
```diff
@@ -423,16 +439,21 @@ public function save_translations() {
...
+				pll_add_notice( new WP_Error( 'pll_strings_translations_updated', __( 'Translations updated.', 'polylang' ), 'success' ) );
+
+				/**
+				 * Fires after the strings translations are saved in DB
+				 *
+				 * @since 1.2
+				 */
+				do_action( 'pll_save_strings_translations' );
+			} else {
+				pll_add_notice( new WP_Error( 'pll_strings_translations_not_updated', __( 'No translations updated.', 'polylang' ) ) );
```
Hug0-Drelon: With my proposal you could have a third error like `pll_strings_translations_update_failed`.

Hug0-Drelon (changes requested):

admin/admin-base.php
Screenfeed: This allows 3rd party plugins to decide which capability to use for their sub-menu. And then, depending on this capability, it allows us to decide which capability to use for the main menu. Don't you think it's useful?

include/Model/Languages.php
Screenfeed: Yes but we don't have such cache rn. Do you think we should create one?

include/Capabilities/Capabilities.php
Screenfeed: This makes `Hooks` aware of implementation details it shouldn't be aware of (even more for code in Pro).

:D

Hug0-Drelon: This is not implementation detail, this is a flag for PRO.
Implementation details concern code that is called directly.

But as  you see fit.

include/Model/Languages.php
Hug0-Drelon: I think the filter should not take place here in the first place, as per my previous comment https://github.com/polylang/polylang/pull/1727#discussion_r2341698781

admin/admin-base.php
Hug0-Drelon: I don't think so, as of today this was not possible do filterer added menus by third party by capabilities. So either they already check the capability before adding their menu, or they don't add anything at all.
Either way this makes us responsible of their menu and moreover responsible of the maintenance related.

settings/table-string.php
Screenfeed: For me this is in the scope because with the languages list being filtered, it can be empty, so this message makes sense in this case IMHO.

But good idea for `PLL_MO::export_to_db()`, it could be improved in another PR.

admin/admin-base.php
Screenfeed: Yes, totally forgot that. Done in [d942b8e](https://github.com/polylang/polylang/pull/1727/commits/d942b8e95becaf1422cef3c4a6ed632ca56e6859).

admin/admin-base.php
Screenfeed: Done in [d942b8e](https://github.com/polylang/polylang/pull/1727/commits/d942b8e95becaf1422cef3c4a6ed632ca56e6859).

admin/admin-base.php
Screenfeed: Good idea. Done in [84416a1](https://github.com/polylang/polylang/pull/1727/commits/84416a15ae073f32c6a29f99e66223a7e7201168).

include/Capabilities/Capabilities.php
Screenfeed: I was trolling 😅
OK let's bring this to the daily or to @Chouby.

include/Model/Languages.php
Screenfeed: Yeah why not. Done in [569547d](https://github.com/polylang/polylang/pull/1727/commits/569547d76dba54949e4572db1eca8f27ab87c86c).

include/Capabilities/Capabilities.php
Hug0-Drelon: Bad idea!

include/Model/Abstract_Languages_Filter.php
Screenfeed: Not a problem since [4050720](https://github.com/polylang/polylang/pull/1727/commits/40507203e21385d29b183ee9a9a963d8e3c48833).

include/Model/Languages.php
Screenfeed: Done in [4050720](https://github.com/polylang/polylang/pull/1727/commits/40507203e21385d29b183ee9a9a963d8e3c48833).

include/Model/Languages.php
Screenfeed: Not a problem since [4050720](https://github.com/polylang/polylang/pull/1727/commits/40507203e21385d29b183ee9a9a963d8e3c48833).

admin/admin-base.php
```diff
@@ -111,46 +111,70 @@ public function init() {
 	}
 
 	/**
-	 * Adds the link to the languages panel in the WordPress admin menu
+	 * Adds the links to Polylang's admin panels in the WordPress admin menu.
```
szepeviktor:
```suggestion
	 * Adds links to Polylang's admin panels to the WordPress admin menu.
```

Screenfeed: Thanks 👍

admin/admin-base.php
```diff
@@ -111,46 +111,70 @@ public function init() {
...
-			$tabs['strings'] = __( 'Translations', 'polylang' );
-		}
+		foreach ( $this->get_sub_menu_items() as $tab => $tab_data ) {
+			$page = 'lang' === $tab ? 'mlang' : "mlang_$tab";
 
-		$tabs['settings'] = __( 'Settings', 'polylang' );
+			if ( empty( $parent ) ) {
+				$parent = $page;
+				add_menu_page( $tab_data['label'], __( 'Languages', 'polylang' ), $tab_data['capability'], $page, '__return_null', 'dashicons-translation' );
+				$admin_page_hooks[ $page ] = $page_type; // Hack to avoid the localization of the hook name. See: https://core.trac.wordpress.org/ticket/18857
```
Hug0-Drelon, line 132: Could third parties used that "bug"?
My question being, if the hook was already internationalized before, maybe we should keep it that way for backward compatibility? Even though it's ugly as fuck...

include/Capabilities/Hooks.php
```diff
@@ -0,0 +1,41 @@
...
+namespace WP_Syntex\Polylang\Capabilities;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * A class allowing to determine if a user can translate content.
+ *
+ * @since 3.8
+ */
+class Hooks {
```
Hug0-Drelon:
```suggestion
class Retro_Compatibility {
```
I'm not very fond of the current class name 😅

include/Model/Abstract_Languages_Proxy.php
```diff
@@ -0,0 +1,73 @@
...
+	/**
+	 * Filters the given list of languages, according to this proxy.
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language[] $languages The list of language objects.
+	 * @param array          $args      Optional arguments passed to `get_list()`.
+	 * @return PLL_Language[]
+	 */
+	abstract protected function filter_list( array $languages, array $args ): array;
```
Hug0-Drelon: This method is not useful, child classes must only leverage `get_list()`.

include/Model/Abstract_Languages_Proxy.php
```diff
@@ -0,0 +1,73 @@
...
+
+	/**
+	 * Returns the list of available languages after passing it through this proxy.
+	 *
+	 * @since 3.8
+	 *
+	 * @param array $args Optional arguments to pass to `Languages::get_list()`.
+	 * @return array List of `PLL_Language` objects or `PLL_Language` object properties.
+	 */
+	public function get_list( array $args = array() ): array {
```
Hug0-Drelon: I'd make it abstract.
Here you implement the method and make decisions for other behaviors.

include/Model/Languages.php
```diff
@@ -791,6 +804,58 @@ public function delete_transient_from_options_table(): void {
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language[] $languages The list of language objects.
+	 * @param array          $args {
+	 *   @type string $fields Optional. Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 * }
+	 * @return array List of `PLL_Language` objects or `PLL_Language` object properties.
+	 */
+	public function apply_transformative_args( array $languages, array $args ): array {
```
Hug0-Drelon:
```suggestion
	public function pluck_list( array $languages, array $args ): array {
```
I suggest an easier to understand name.

include/Model/Languages.php
```diff
@@ -791,6 +804,58 @@ public function delete_transient_from_options_table(): void {
...
+	 *
+	 * @param string $key Proxy's key.
+	 * @return Abstract_Languages_Proxy
+	 *
+	 * @phpstan-param non-falsy-string $key
+	 */
+	public function proxy( string $key ): Abstract_Languages_Proxy {
+		if ( isset( $this->proxies[ $key ] ) ) {
+			$proxy = $this->proxies[ $key ];
+			return new $proxy( $this );
```
Hug0-Drelon: As seen, proxies should be instantiated by their own module and passed to `register_proxy()` instead.

include/Model/Languages.php
```diff
@@ -27,8 +27,12 @@ class Languages {
 	public const LOCALE_PATTERN = '^' . self::INNER_LOCALE_PATTERN . '$';
 	public const SLUG_PATTERN   = '^' . self::INNER_SLUG_PATTERN . '$';
 
-	public const TRANSIENT_NAME = 'pll_languages_list';
-	private const CACHE_KEY     = 'languages';
+	public const TRANSIENT_NAME      = 'pll_languages_list';
+	public const TRANSFORMATIVE_ARGS = array(
+		'fields' => false,
+	);
```
Hug0-Drelon: I find it to be more confusing with this constant than using `unset( $args['fields'])` or something.
I don't see any case where this list will grow, this is more a backward compatibility with legacy code.
If you really want a constant, I suggest to rename to something like `FIELDS_LIST_ARG` or something.

include/Capabilities/Hooks.php
Screenfeed: Well, I don't like `Retro_Compatibility` neither 😅 Moreover, this class has nothing to do with ~retro~ backward compatibility.

include/Model/Abstract_Languages_Proxy.php
Screenfeed: It is useful, it avoids duplicating:

```php
$languages = $this->languages->get_list(
	array_diff_key( $args, $this->languages::TRANSFORMATIVE_ARGS )
);
```

and:

```php
$languages = array_values( $languages ); // Re-index.

return $this->languages->apply_transformative_args( $languages, $args );
```

in concrete proxies.

include/Model/Abstract_Languages_Proxy.php
Screenfeed: Which behaviors?

include/Model/Languages.php
Screenfeed: Actually I find `apply_transformative_args()` much easier to understand than `pluck_list()`.

include/Model/Languages.php
Screenfeed: The aim is for the proxies to not know about `$args['fields']`. So you prefer `public const FIELDS_LIST_ARG = 'fields';`?

include/Model/Abstract_Languages_Proxy.php
Hug0-Drelon: Then use `parent::get_list()` instead of creating another method.

include/Model/Abstract_Languages_Proxy.php
Hug0-Drelon: The one you coded in it.

include/Model/Languages.php
Hug0-Drelon: `apply_transformative_args()` doesn't tell what it does at all.
We should find another name.
What about `convert_list()` or something?

include/Model/Languages.php
Screenfeed: You're right, this is not needed anymore. Done in [7c0bea3](https://github.com/polylang/polylang/pull/1727/commits/7c0bea39129ce90b9ca64098e9b4e5c552ada5d3).

include/Model/Languages.php
Hug0-Drelon: I prefer no constant at at all. It doesn't add value IMHO.
The only argument to know is `fields` so we can stick with that.
But as said, if you really want a constant, I suggest you avoid using an array (so your proposal fits, even though it's a bit verbose).

include/Capabilities/Hooks.php
Hug0-Drelon: What I meant by "retro" is for installation that doesn't not have `manage_translations` capability (old install or free install).
I agree `Retro_Compatibility` is not ideal at all.

include/Model/Languages.php
Screenfeed: Much better 👍 
Rename in [7d51cbb](https://github.com/polylang/polylang/pull/1727/commits/7d51cbb1510c9142b3d848fef5e09e5f5641e82f).

include/Model/Languages.php
Screenfeed: Removed in [937d06c](https://github.com/polylang/polylang/pull/1727/commits/937d06cdb9b14eaf4cc4b3c30f8bd70d8f0b6688).

include/Capabilities/Hooks.php
Screenfeed: Renamed into `Capabilities/Capabilities` in [e48a0bd](https://github.com/polylang/polylang/pull/1727/commits/e48a0bdbd491b263b2e15eaf4707772ed0bb17b0).

include/Model/Abstract_Languages_Proxy.php
Screenfeed: Done in [b68176f](https://github.com/polylang/polylang/pull/1727/commits/b68176f2c93fbe6903bdaa9d4f6e830a3eff69e1).

include/Model/Abstract_Languages_Proxy.php
Screenfeed: Done in [b68176f](https://github.com/polylang/polylang/pull/1727/commits/b68176f2c93fbe6903bdaa9d4f6e830a3eff69e1).

include/Model/Languages.php
```diff
@@ -553,7 +562,27 @@ public function has(): bool {
...
+	/**
+	 * Returns the list of available languages.
+	 * - Stores the list in a db transient (except flags), unless `PLL_CACHE_LANGUAGES` is set to false.
+	 * - Caches the list (with flags) in a `PLL_Cache` object.
+	 *
+	 * @since 3.8
+	 *
+	 * @param array $args {
+	 *   @type bool $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`).
+	 *   @type bool $hide_default Hides default language from the list (default to `false`).
```
Hug0-Drelon: To avoid multiplying arguments on `PLL_Model` API, which is already big, I suggest to remove `$args` from `get_objects()` and keep it only in `get_list()`.

include/Capabilities/Capabilities.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @package Polylang
+ */
+
+namespace WP_Syntex\Polylang\Capabilities;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * A class allowing to determine if a user can translate content.
```
Hug0-Drelon:
```suggestion
 * A class allowing to alias `manage_translations` and `manage_options`.
```

include/Model/Abstract_Languages_Proxy.php
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * Returns the proxy's key.
+	 *
+	 * @since 3.8
+	 *
+	 * @return string
+	 *
+	 * @phpstan-return non-falsy-string
+	 */
+	abstract public static function key(): string;
```
Hug0-Drelon: Not sure we need it to be static.

include/Model/Default_Languages_Proxy.php
```diff
@@ -0,0 +1,42 @@
...
+use PLL_Language;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Default proxy the list of languages.
+ *
+ * @since 3.8
+ */
+class Default_Languages_Proxy extends Abstract_Languages_Proxy {
```
Hug0-Drelon: Proposing a better name here, since `Default_Languages_Proxy` suggests it does something or it manages default language.
```suggestion
class Pass_Through_Proxy extends Abstract_Languages_Proxy {
```
or
```suggestion
class No_Op_Proxy extends Abstract_Languages_Proxy {
```

include/Model/Languages.php
```diff
@@ -553,7 +562,7 @@ public function has(): bool {
 	 * }
 	 * @return array List of PLL_Language objects or PLL_Language object properties.
 	 */
-	public function get_list( $args = array() ): array {
+	public function get_list( array $args = array() ): array {
```
Hug0-Drelon: Even though I like it, I wouldn't type hint this extensively used public method. This is part of our public API and I fear other developers are not as strict as us.

include/Model/Default_Languages_Proxy.php
Screenfeed: I see your point. Done in [3273708](https://github.com/polylang/polylang/pull/1727/commits/327370892aaaeaf20bbce0a2d154be0717524da8).
I looked for other classes that act similarly in our code (maybe we could use the same naming) but I could only find `NOOP_Translations`, which is WP's.

include/Model/Languages.php
Screenfeed: Done in [9360c11](https://github.com/polylang/polylang/pull/1727/commits/9360c1150c450200db072d3bda227f1470e89336) >_>

include/Model/Abstract_Languages_Proxy.php
Screenfeed: I fear it will come back and bite our asses soon or later but done in [3c4d854](https://github.com/polylang/polylang/pull/1727/commits/3c4d85420f637ed40460a1819e499b9ab3cd5a08).

include/Capabilities/Capabilities.php
Screenfeed: What do you think of this instead? I don't think we should be too specific.

```
A class allowing to map Polylang's custom user capabilities to WP's native ones.
```

include/Model/Abstract_Languages_Proxy.php
Hug0-Drelon: We'll make it static if we need in the future.

include/Capabilities/Capabilities.php
Hug0-Drelon: I'm fine with your proposal.

include/Model/Languages.php
Hug0-Drelon: The type hint in `convert_list()` can be kept though, just need some check before calling it.

Screenfeed: Done in [8163416](https://github.com/polylang/polylang/pull/1727/commits/81634166ebe9d947dcdce5bc45e7683293048b42).

include/Capabilities/Capabilities.php
Screenfeed: Done in [c32def2](https://github.com/polylang/polylang/pull/1727/commits/c32def2e0eb321be1980050e98e1c6b392ff088b).

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		$languages      = array_values( $languages ); // Re-index.
+		$args['fields'] = $fields;
+
+		return $this->languages->convert_list( $languages, $args );
+	}
+
+	/**
+	 * Returns a proxy that allows to filter `get_list()`.
```
Hug0-Drelon:
```suggestion
	 * Returns proxies chain.
```

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+	 * @var string[]
+	 */
+	protected $stack = array();
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.8
+	 *
+	 * @param Languages $languages The object to proxy.
```
Hug0-Drelon:
```suggestion
	 * @param Languages $languages Languages' model.
```
To be completely accurate, the whole `Languages` object is not proxied, only `get_list()` is.

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+	 * @param array     $proxies   List of proxies.
```
Hug0-Drelon:
```suggestion
	 * @param array     $proxies   List of registered proxies.
```
Minor improvement.

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+	 * @param string    $parent    Key of the first item of the stack.
```
Hug0-Drelon:
```suggestion
	 * @param string    $parent    Key of the first item of the proxies stack to traverse.
```
Minor improvement.

settings/table-string.php
```diff
@@ -19,28 +19,28 @@ class PLL_Table_String extends WP_List_Table {
 	 *
 	 * @var PLL_Language[]
 	 */
-	protected $languages;
+	protected $languages = array();
```
Chouby: All the modifications may be useful, but are not related to to capabilities. I propose to to move them to a separate PR. Our future us will thank us to have proper PR description dedicated to these changes.

include/Model/Languages.php
```diff
@@ -791,6 +799,50 @@ public function delete_transient_from_options_table(): void {
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language[] $languages The list of language objects.
+	 * @param array          $args {
+	 *   @type string $fields Optional. Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 * }
+	 * @return array List of `PLL_Language` objects or `PLL_Language` object properties.
+	 */
+	public function convert_list( array $languages, array $args ): array {
```
Chouby: I suggest `maybe_list_pluck()` (to let us think to `wp_list_pluck()`).

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+	 * Returns the list of available languages after passing it through this proxy.
+	 *
+	 * @since 3.8
+	 *
+	 * @param array $args Optional arguments to pass to `Languages::get_list()`.
+	 * @return array List of `PLL_Language` objects or `PLL_Language` object properties.
+	 */
+	public function get_list( array $args = array() ): array {
+		$fields = $args['fields'] ?? null;
+		unset( $args['fields'] );
```
Chouby: This can be simplified by making a copy:
```suggestion
		$all_args = $args;
		unset( $args['fields'] );
```
And then:
```php
return $this->languages->convert_list( $languages, $all_args );
```

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+namespace WP_Syntex\Polylang\Model;
+
+use PLL_Language;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Abstract class allowing to proxy the list of languages.
```
Chouby: Not abstract

include/Model/Languages_Proxy_Chain.php
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @since 3.8
+ */
+class Languages_Proxy_Chain {
```
Chouby: What do you think?
```suggestion
class Languages_Proxies {
```

include/Model/Languages.php
```diff
@@ -791,6 +799,50 @@ public function delete_transient_from_options_table(): void {
...
+
+	/**
+	 * Returns a proxy that allows to filter `get_list()`.
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $key Proxy's key.
+	 * @return Languages_Proxy_Chain
+	 */
+	public function proxy( string $key ): Languages_Proxy_Chain {
```
Chouby: What do you think of `filter()`?

include/Capabilities/Capabilities.php
```diff
@@ -0,0 +1,41 @@
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param string[] $caps Primitive capabilities required by the user.
+	 * @param string   $cap  Capability being checked.
+	 * @return string[]
+	 */
+	public function map_custom_caps( $caps, $cap ) {
+		if ( 'manage_translations' === $cap ) {
+			return array( 'manage_options' );
```
Chouby: Imagine a previous filter added `do_not_allow`.
```suggestion
			$caps[] = 'manage_options';
```

include/Capabilities/Capabilities.php
```diff
@@ -0,0 +1,41 @@
...
+namespace WP_Syntex\Polylang\Capabilities;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * A class allowing to map Polylang's custom user capabilities to WP's native ones.
+ *
+ * @since 3.8
+ */
+class Capabilities {
```
Chouby, line 15: Isn't this class a bit overkill if it does only that? I would move this hook to `PLL_Admin_Base`.

admin/admin-base.php
```diff
@@ -566,4 +591,52 @@ public function should_hide_admin_bar_menu(): bool {
...
+	 * @return array[] {
+	 *     List of sub-menu items with page slugs as array keys, and arrays as follow as array values:
+	 *
+	 *     @type string $label      Sub-menu title.
+	 *     @type string $capability Optional. User capability. Default is `manage_options`.
+	 * }
+	 *
+	 * @phpstan-return array<non-empty-string, string>
+	 */
+	protected function get_sub_menu_items(): array {
```
Chouby:
```suggestion
	protected function get_menu_items(): array {
```

admin/admin-base.php
```diff
@@ -111,46 +111,71 @@ public function init() {
...
+		$parent    = false;
+		$page_type = 'languages';
 
-		// Only if at least one language has been created
-		if ( $this->model->has_languages() ) {
-			$tabs['strings'] = __( 'Translations', 'polylang' );
-		}
+		foreach ( $this->get_sub_menu_items() as $tab => $label ) {
+			$page = 'lang' === $tab ? 'mlang' : "mlang_$tab";
+			$capa = 'strings' === $tab ? 'manage_translations' : 'manage_options';
```
Chouby, line 128:
```suggestion
			$capa = 'strings' === $tab ? 'manage_translations' : 'manage_options';
			
			if ( ! current_user_can ( $capa ) {
				continue;
			}
```
This should avoid extra (duplicate) code in `get_sub_menu_items()`.

admin/admin-base.php
```diff
@@ -566,4 +591,52 @@ public function should_hide_admin_bar_menu(): bool {
...
+		foreach ( $tabs as $tab => $label ) {
+			$capa = 'strings' === $tab ? 'manage_translations' : 'manage_options';
+
+			if ( current_user_can( $capa ) ) {
+				// Keep only useful tabs to ease `add_menu_page()` in `PLL_Admin_Base::add_menus()`.
+				$return[ $tab ] = $label;
+			}
+		}
+
+		return $return;
```
Chouby:
```suggestion
		return apply_filters( 'pll_settings_tabs', $tabs );
```

admin/admin-base.php
```diff
@@ -111,46 +111,71 @@ public function init() {
...
 	 *
 	 * @return void
 	 */
-	public function add_menus() {
+	public function add_menus(): void {
 		global $admin_page_hooks;
 
-		// Prepare the list of tabs
-		$tabs = array( 'lang' => __( 'Languages', 'polylang' ) );
+		$parent    = false;
```
Chouby: Why changing the type?
```suggestion
		$parent    = '';
```

admin/admin-base.php
```diff
@@ -111,46 +111,71 @@ public function init() {
...
+		$page_type = 'languages';
```
Chouby, line 124: This intermediate variable doesn't add value in my opinion. I Would hardcode.

admin/admin-base.php
```diff
@@ -111,46 +111,71 @@ public function init() {
...
-
-		$parent = '';
+		foreach ( array( 'load-', 'admin_print_styles-', 'admin_print_scripts-', 'admin_head-', '', 'admin_print_footer_scripts-', 'admin_footer-' ) as $prefix ) {
+			add_action(
+				"{$prefix}toplevel_page_{$parent}",
+				static function () use ( $prefix, $page_type, $parent ) {
+					do_action( "{$prefix}{$page_type}_page_{$parent}" );
+				}
+			);
+		}
```
Chouby, line 165: Maybe we can run this only if the parent page is not the languages pages. What do you think?

settings/table-string.php
Screenfeed: It is related: if we don't pass the `if ( empty( $languages ) ) {` check, the properties are not set.

include/Model/Languages.php
Screenfeed: It's a discussion we had with @Hug0-Drelon: "pluck" doesn't talk to me at all, and `wp_list_pluck()` is only the WP version of `array_column()`.
`convert_list()` seemed adapted because the method changes the type of the returned list, and the name matches `get_list()`.
`maybe_convert_list()` could be used if you prefer, but I really don't like `pluck` :(

include/Capabilities/Capabilities.php
Screenfeed: I would wait until we finish the whole feature, we will probably have other things to put in here (and usually ~you~ we don't like to add things in `PLL_Base` and child classes 😏).

admin/admin-base.php
Screenfeed: I used "sub_menu" because we add 1 menu item (with `add_menu_page()`) and several sub-menu items (with `add_submenu_page()`), which are returned by this method. What do you think?

admin/admin-base.php
Screenfeed: ~Removing items that won't be displayed allows to determine which user capability to use for the main menu item.
For example with a user that has `manage_translations` but not `manage_options`:~
- ~With what is done here, only the item with `manage_translations` will be returned, meaning that the main item will use `manage_translations`.~
- ~With what you suggest, all items are returned, meaning that the main item will use the capability of the first item (`lang`), which is `manage_options` => the menu won't be displayed.~

I just learned something:
As long as the main menu has sub-menus, WP won't use the user capability passed to `add_menu_page()`. This means a user with `manage_translations` will still be able to access the Translations page, even if the main menu has `manage_options`.
This doesn't really make sense to me but I think I understand the logic behind it.

Anyway, this allows us to not care about it, and use any user capability coming from the sub-menus, like it was before.

Also, I just saw your previous comment about moving `current_user_can()` to remove duplicated code.

Done in [3c96b3e](https://github.com/polylang/polylang/pull/1727/commits/3c96b3e4f1cdb84d87e5c978c64074118ede2919).

admin/admin-base.php
Screenfeed: It is used in 6 places to show that it is used in hook names: it explains how the name of the hooks are built.

admin/admin-base.php
Screenfeed: Doing it in all cases allows to have the languages page not on first position.
Not currently, but we can make a new PR for that:
- In our CSS, replace all selectors `.toplevel_page_mlang` by `.languages_page_mlang`,
- In our php code, replace all `toplevel_page_mlang` by `languages_page_mlang`.

This way, even if someone puts a new sub-menu item at the 1st position, it will still work and we won't have to worry about `toplevel` anymore.

include/Model/Languages_Proxy_Chain.php
Screenfeed: Done in [89855c1](https://github.com/polylang/polylang/pull/1727/commits/89855c1ed1b0cd3dfef38f8648fa2652447c013f).

include/Model/Languages_Proxy_Chain.php
Screenfeed: Fixed in [c21124a](https://github.com/polylang/polylang/pull/1727/commits/c21124a0da91741aea99b7aa75de60f0edb95299).

include/Model/Languages_Proxy_Chain.php
Screenfeed: Yep. Done in [4a0e917](https://github.com/polylang/polylang/pull/1727/commits/4a0e9174e5a764539d55378b769844e4ecda7c15).

include/Model/Languages.php
Screenfeed: Yes. Done in [9ee2149](https://github.com/polylang/polylang/pull/1727/commits/9ee21491c58cb525d750fa25ea04d8531fc235c3).

include/Capabilities/Capabilities.php
Screenfeed: Indeed. Fixed in [07771fc](https://github.com/polylang/polylang/pull/1727/commits/07771fc60eb7fdbb7a5d4f22c455e1abe282acbf) and [47b7402](https://github.com/polylang/polylang/pull/1727/commits/47b74023a3c57738d000abfd123ef502c2101ba6).

admin/admin-base.php
Screenfeed: OK. Reverted in [838e768](https://github.com/polylang/polylang/pull/1727/commits/838e7686036a6c8bb0690bdeae338bb16287f2cc).

admin/admin-base.php
Screenfeed: Duplicate code removed in 3c96b3e.

admin/admin-base.php
Screenfeed: If we don't find anything that please everybody, may I suggest `maybe_pluck_list()`.

include/Model/Languages.php
Screenfeed: If we don't find anything that please everybody, may I suggest `maybe_pluck_list()`.

admin/admin-base.php
```diff
@@ -566,4 +597,50 @@ public function should_hide_admin_bar_menu(): bool {
...
+		$return = array();
+
+		foreach ( $tabs as $tab => $label ) {
+			$return[ $tab ] = array(
+				'label'      => $label,
+				'capability' => 'strings' === $tab ? 'manage_translations' : 'manage_options',
+			);
+		}
+
+		return $return;
```
Chouby: I'm not fan of this intermediate array which adds a useless extra `foreach` loop. For me it doesn't improve the readability of the system as this can be replaced by a single line in `add_menus()`. 
At worst, if you really want to keep it, I suggest to convert it to a `stdClass` object  before returning it.

admin/admin-base.php
Chouby: My reasoning was that we add one menu with several items (items are sub-menus).

include/Model/Languages.php
Chouby: >Not from the stars do I my judgment pluck

settings/table-string.php
```diff
@@ -57,12 +57,14 @@ public function __construct( $languages ) {
 			)
 		);
 
+		if ( empty( $languages ) ) {
+			return;
+		}
```
Chouby: After a survey based on 2, we prefer not display the `Translations` page if there are no languages. I guess this will simplify a bit the changes made to this class.

settings/view-tab-strings.php
```diff
@@ -16,11 +16,15 @@
...
-		<p><?php esc_html_e( 'Use this to remove unused strings from database, for example after a plugin has been uninstalled.', 'polylang' ); ?></p>
-		<?php
-		submit_button(); // Since WP 3.1
+
+		if ( current_user_can( 'manage_options' ) ) {
+			printf( '<br /><label><input name="clean" type="checkbox" value="1" /> %s</label>', esc_html__( 'Clean strings translation database', 'polylang' ) );
+			?>
+			<p><?php esc_html_e( 'Use this to remove unused strings from database, for example after a plugin has been uninstalled.', 'polylang' ); ?></p>
+			<?php
+		}
```
Chouby, line 25: This requires to protect the ability to clear translations in `PLL_Table_String`: https://github.com/polylang/polylang/blob/542d453479c4ae5d30217849f83917a808bda4b7/settings/table-string.php#L433-L439

Also, similar to this feature, we must protect unregistering WPML strings: https://github.com/polylang/polylang/blob/542d453479c4ae5d30217849f83917a808bda4b7/settings/table-string.php#L460-L464

include/Model/Languages.php
Chouby: ~`maybe_fuck_list()`~

Chouby: Let say `maybe_convert_list()` because I don't find anything better and I want to please you ;)

Screenfeed: Renamed in [96a342b](https://github.com/polylang/polylang/pull/1727/commits/96a342b08996df0b2b5dda4659330b28ed0e4766).

admin/admin-base.php
Screenfeed: Ha! I understand your logic. Yours is "items of the menu", while mine is "items that are sub-menus".
Renamed in [835a665](https://github.com/polylang/polylang/pull/1727/commits/835a665280090a0abcb708a96dff90dd152b4de6).

admin/admin-base.php
Screenfeed: Done in [9f4f57e](https://github.com/polylang/polylang/pull/1727/commits/9f4f57e9a50a2dd0b4021c6ca77361d39747d2e7).

settings/table-string.php
Screenfeed: Done in [4c37c73](https://github.com/polylang/polylang/pull/1727/commits/4c37c73748b6f7eb864ef3ebc137698c9a04f008).
This simplifies it in PLL, not in PLL Pro.

settings/view-tab-strings.php
Screenfeed: Totally missed that 😱 
Fixed in [2a80e24](https://github.com/polylang/polylang/pull/1727/commits/2a80e2420081f417da02c042c963d6fcf56705d1) and [1d0c1e3](https://github.com/polylang/polylang/pull/1727/commits/1d0c1e3e84b674d0dce5b1d7ff81d40ea34b9dfd).

settings/table-string.php
Screenfeed: All reverted in [4c37c73](https://github.com/polylang/polylang/pull/1727/commits/4c37c73748b6f7eb864ef3ebc137698c9a04f008): the page will not be displayed at all to users without the translator capabilities for currently existing languages.

settings/table-string.php
Screenfeed: OK I found a way to keep it simple in [a6d7f73](https://github.com/polylang/polylang/pull/1727/commits/a6d7f7332873a710e08786ff24534714e786decd).

settings/table-string.php
Hug0-Drelon, line 1: I don't see any capability check in `save_translations()` such as `translate_**`. Is it on purpose?

Screenfeed: Good catch, but it's fine.

1. First, `save_translations()` is hooked to `mlang_action_string-translation`. The related `do_action()` is [in `PLL_Settings::handle_actions()`](https://github.com/polylang/polylang/blob/93f6a3d62a4784a4f86c1e6490249774bf5b12ec/settings/settings.php#L264), which is called [by `PLL_Settings::languages_page()`](https://github.com/polylang/polylang/blob/93f6a3d62a4784a4f86c1e6490249774bf5b12ec/settings/settings.php#L299), which is the callback passed [to `add_submenu_page()`](https://github.com/polylang/polylang/blob/93f6a3d62a4784a4f86c1e6490249774bf5b12ec/admin/admin-base.php#L143). So, here, we're fine, the capability check is done.
2. Then about the languages in `save_translations()`. Since [we loop through `$this->languages`](https://github.com/polylang/polylang/blob/93f6a3d62a4784a4f86c1e6490249774bf5b12ec/settings/table-string.php#L386), and this list of languages [contains only the languages that the user can translate](https://github.com/polylang/polylang/blob/93f6a3d62a4784a4f86c1e6490249774bf5b12ec/settings/settings.php#L288), we're fine.

However, this would have been an issue for my next PR, where I must display all languages with disabled inputs.

Hug0-Drelon: Okay, this is a bit convoluted. May I suggest to rename `PLL_Table_String::$languages` for `PLL_Table_String::$authorized_languages` (and updating the property doc as well)?

Because next time I stumble upon `PLL_Table_String::save_translations()`, I'm not sure I'll remember the filtered list of language has been injected.

Screenfeed: In the next PR it will be reverted to the original list of languages, and I'll add another list for the "authorized" ones.

PR #1729, opened 2025-09-11
Fix error in `WP_Syntex\Polylang\REST\Request::get_object_type()`.

Hug0-Drelon:
## What?
Prevent `Error: Cannot use object of type Closure as array` in  `WP_Syntex\Polylang\REST\Request::get_object_type()`.
Raised in PLLWC PHPUnit tests (`Previewed_Emails_Test::test_previewed_email_subject_string`).

## How?
Ensure callback in request handler is an array before checking for controller classes.

### DIFF
--- a/modules/REST/Request.php
+++ b/modules/REST/Request.php
@@ -213,10 +213,14 @@ class Request {
 			return $type;
 		}
 
-		$callback = $this->handler['callback'] ?? null;
-		if ( is_array( $callback ) && reset( $callback ) instanceof WP_REST_Posts_Controller ) {
+		if ( ! is_array( $this->handler['callback'] ) ) {
+			return null;
+		}
+
+		$controller = reset( $this->handler['callback'] );
+		if ( $controller instanceof WP_REST_Posts_Controller ) {
 			return 'post';
-		} elseif ( is_array( $callback ) && reset( $callback ) instanceof WP_REST_Terms_Controller ) {
+		} elseif ( $controller instanceof WP_REST_Terms_Controller ) {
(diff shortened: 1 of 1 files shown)


modules/REST/Request.php
```diff
@@ -213,10 +213,10 @@ public function get_object_type(): ?string {
...
 
-		$controller = $this->handler['callback'][0] ?? null;
-		if ( $controller instanceof WP_REST_Posts_Controller ) {
+		$callback = $this->handler['callback'] ?? null;
+		if ( is_array( $callback ) && reset( $callback ) instanceof WP_REST_Posts_Controller ) {
 			return 'post';
-		} elseif ( $controller instanceof WP_REST_Terms_Controller ) {
+		} elseif ( is_array( $callback ) && reset( $callback ) instanceof WP_REST_Terms_Controller ) {
 			return 'term';
 		}
```
Chouby: Here is a suggestion to decrease the max number of operations (-2):
```php
		if ( ! isset( $this->handler['callback'] ) || ! is_array( $this->handler['callback'] ) ) {
			return null;
		}
		
		$controller = reset( $this->handler['callback'] );
		
		if ( $controller instanceof WP_REST_Posts_Controller ) {
			return 'post';
		} elseif ( $controller instanceof WP_REST_Terms_Controller ) {
			return 'term';
		}
```

Hug0-Drelon: Good idea, done in 5f250ed4.
I avoided to check `$this->handler` once again because it's already done above.

PR #1730, opened 2025-09-19
Removes use of PLL() function that could be not declared

manooweb:
For the moment, the goal of this PR is to share what we study inspired by the @chouby [comment](https://github.com/polylang/polylang-pro/issues/2713#issuecomment-3204839229)

1. Instantiate a new `PLL_Translatable_Objects` instead of accessing the `PLL_Model` property. To make this work, we should use static properties to be able to share them among all instances (we can register translatable objects in one instance and read them in another one).

2. Make `get_tax_language()` static methods should be easy.

Point 1:
It could work. This what the

Point2:
It isn't necessary because if the `PLL_Translatable_Objects::$objects` is now shared for each instance, all translatable object instances are accessible and then `their get_tax_language()` method too.
It is something similar we already done [here](https://github.com/polylang/polylang/blob/master/include/Model/Languages.php#L736-L742) with the `PLL_Translatable_Objects` instance stored in Polylang model.

For the moment, a lot of tests break caused by these changes.

### DIFF
--- a/include/translatable-objects.php
+++ b/include/translatable-objects.php
@@ -28,7 +28,7 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	 *
 	 * @phpstan-var array<non-empty-string, PLL_Translatable_Object>
 	 */
-	private $objects = array();
+	private static $objects = array();
 
 	/**
 	 * Registers a translatable object.
@@ -52,10 +52,10 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 		}
 
 		if ( ! isset( $this->objects[ $object->get_type() ] ) ) {
-			$this->objects[ $object->get_type() ] = $object;
+			self::$objects[ $object->get_type() ] = $object;
 		}
 
-		return $this->objects[ $object->get_type() ];
+		return self::$objects[ $object->get_type() ];
 	}
 
 	/**
@@ -69,7 +69,7 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	 */
 	#[\ReturnTypeWillChange]
 	public function getIterator() {
-		return new ArrayIterator( $this->objects );
+		return new ArrayIterator( self::$objects );
 	}
 
 	/**
@@ -89,11 +89,11 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	 * )
 	 */
 	public function get( $object_type ) {
-		if ( ! isset( $this->objects[ $object_type ] ) ) {
+		if ( ! isset( self::$objects[ $object_type ] ) ) {
 			return null;
 		}
 
-		return $this->objects[ $object_type ];
+		return self::$objects[ $object_type ];
 	}
 
 	/**
@@ -106,7 +106,7 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	 * @phpstan-return array<non-empty-string, PLL_Translatable_Object>
 	 */
 	public function get_secondary_translatable_objects() {
-		return array_diff_key( $this->objects, array( $this->main_type => null ) );
+		return array_diff_key( self::$objects, array( $this->main_type => null ) );
 	}
 
 	/**
@@ -123,7 +123,7 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	public function get_taxonomy_names( $filter = array( 'language', 'translations' ) ) {
 		$taxonomies = array();
 
-		foreach ( $this->objects as $object ) {
+		foreach ( self::$objects as $object ) {
 			if ( in_array( 'language', $filter, true ) ) {
 				$taxonomies[] = $object->get_tax_language();
 			}
--- a/tests/phpunit/includes/testcase-trait.php
+++ b/tests/phpunit/includes/testcase-trait.php
@@ -128,6 +128,7 @@ trait PLL_UnitTestCase_Trait {
 	 */
 	public static function wpTearDownAfterClass() {
 		self::delete_all_languages();
+		self::reset_translatable_objects_registry();
 	}
 
 	/**
@@ -286,5 +287,12 @@ trait PLL_UnitTestCase_Trait {
 		if ( ! file_exists( $path ) ) {
 			self::markTestSkipped( $message );
 		}
+	}
+
+	protected static function reset_translatable_objects_registry() {
+		$reflection = new ReflectionClass( PLL_Translatable_Objects::class );
+		$property   = $reflection->getProperty('objects');
+		$property->setAccessible(true);
+		$property->setValue( null, array() );
 	}
 }
--- a/tests/phpunit/tests/test-admin-filters-post.php
+++ b/tests/phpunit/tests/test-admin-filters-post.php
@@ -19,6 +19,8 @@ class Admin_Filters_Post_Test extends PLL_UnitTestCase {
 		self::create_language( 'de_DE_formal' );
 		self::create_language( 'es_ES' );
 
+		self::require_api();
+
 		self::$editor = $factory->user->create( array( 'role' => 'editor' ) );
 	}
 
--- a/include/Options/Business/Language_Taxonomies.php
+++ b/include/Options/Business/Language_Taxonomies.php
@@ -5,6 +5,8 @@
 
 namespace WP_Syntex\Polylang\Options\Business;
 
+use PLL_Translatable_Objects;
+
 defined( 'ABSPATH' ) || exit;
 
 /**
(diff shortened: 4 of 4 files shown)


install/upgrade.php
```diff
@@ -232,6 +231,20 @@ protected function upgrade_3_7() {
 		$this->empty_duplicated_strings_translations();
 	}
 
+	/**
+	 * Upgrades if the previous version is < 3.8.
+	 * Migrates language taxonomies from polylang option to pll_language_taxonomies option.
```
Hug0-Drelon:
```suggestion
	 * Migrates language taxonomies from `polylang` option to `pll_language_taxonomies` option.
```

include/translatable-objects.php
```diff
@@ -28,7 +28,7 @@ class PLL_Translatable_Objects implements IteratorAggregate {
 	 *
 	 * @phpstan-var array<non-empty-string, PLL_Translatable_Object>
 	 */
-	private $objects = array();
+	private static $objects = array();
```
Hug0-Drelon: Not sure this property needs to be static now you have `pll_language_taxonomies`.

install/upgrade.php
```diff
@@ -232,6 +231,20 @@ protected function upgrade_3_7() {
...
+	 *
+	 * @since 3.8
+	 *
+	 * @return void
+	 */
+	protected function upgrade_3_8() {
+		$options = get_option( Options::OPTION_NAME );
```
Hug0-Drelon:
```suggestion
		$options = get_option( `polylang` );
```
I don't think this class should rely on core classes.

install/upgrade.php
```diff
@@ -232,6 +231,20 @@ protected function upgrade_3_7() {
...
+		$language_taxonomies = is_array( $options ) && isset( $options['language_taxonomies'] ) && is_array( $options['language_taxonomies'] ) ? $options['language_taxonomies'] : array();
+		update_option( 'pll_language_taxonomies', $language_taxonomies );
```
Hug0-Drelon: Given that there is more than one condition in your ternary. I suggest to use an if statement to make the code more readable.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -19,6 +19,8 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 		self::create_language( 'de_DE_formal' );
 		self::create_language( 'es_ES' );
 
+		self::require_api();
```
Hug0-Drelon: Why do you need it?

tests/phpunit/includes/testcase-trait.php
```diff
@@ -287,4 +288,14 @@ protected static function markTestSkippedIfFileNotExists( string $path, string $
 			self::markTestSkipped( $message );
 		}
 	}
+
+	/**
+	 * Resets PLL_Translatable_Objects::$objects static property to avoid conflict in tests.
+	 */
+	protected static function reset_translatable_objects_registry() {
```
Hug0-Drelon: If the property is no longer a static one, it'll become useless.

include/translatable-objects.php
manooweb: Yes! I completely forgot to remove the code from the older proposition 🥱

manooweb: Done in 64cb5c58e8a60cfa8175ec37d150606576324cbf

tests/phpunit/includes/testcase-trait.php
manooweb: Of course! Removed in 64cb5c58e8a60cfa8175ec37d150606576324cbf

install/upgrade.php
manooweb: Done in 11c128406ba70edaa1f71bfcbd9ff7df4abc0d50

tests/phpunit/tests/test-admin-filters-post.php
manooweb: Useful only in one test when running its test class alone on local environment.
Here https://github.com/polylang/polylang/blob/3.7.3/tests/phpunit/tests/test-admin-filters-post.php#L332
Due to https://github.com/polylang/polylang/blob/3.7.3/include/translated-post.php#L351 (call to `PLL()` )

I propose to remove this fix from this PR.

manooweb: Reverted in 1e6a5d7940e000a8d1e2a83305afd1486c691816

install/upgrade.php
```diff
@@ -232,6 +231,25 @@ protected function upgrade_3_7() {
...
+	 */
+	protected function upgrade_3_8() {
+		$options = get_option( 'polylang' );
+
+		$language_taxonomies = array();
+		if ( is_array( $options ) && isset( $options['language_taxonomies'] ) && is_array( $options['language_taxonomies'] ) ) {
+			$language_taxonomies = $options['language_taxonomies'];
+		}
+
+		update_option( 'pll_language_taxonomies', $language_taxonomies );
```
Hug0-Drelon, line 250:
```suggestion
		if ( is_array( $options ) && isset( $options['language_taxonomies'] ) && is_array( $options['language_taxonomies'] ) ) {
			update_option( 'pll_language_taxonomies', $options['language_taxonomies'] );
		}
```
Let's avoid an unnecessary variable assignment.

Hug0-Drelon: One question though, maybe @Chouby will choose whether or not to pursue, but should we check for already existing `language_taxonomies` option value here as well (just in case a rollback has bee made before updating again).

tests/phpunit/tests/plugins/test-translated-table.php
```diff
@@ -0,0 +1,30 @@
...
+				// Register the DB table in Polylang.
+				$foo = ( new PLLTest_Translatable( $model, 'foo1' ) )->init();
+				$model->translatable_objects->register( $foo );
+			}
+		);
+
+		new PLL_Context_Admin();
+
+		$this->assertSame( array( 'foo1_language' ), get_option( 'pll_language_taxonomies' ) );
+	}
```
Chouby: This seems to have a similar intention as this test https://github.com/polylang/polylang/blob/67d462ae9c3170fddefedc6dc83aa9e0c3a9840b/tests/phpunit/tests/test-model.php#L203 
which unfortunately didn't fail when we introduced the bug. Instead of introducing a new test, the best would be to modify the original test to ensure that it fails on the bug and then add the new assertion.

install/upgrade.php
```diff
@@ -232,6 +231,22 @@ protected function upgrade_3_7() {
...
+	 * @since 3.8
+	 *
+	 * @return void
+	 */
+	protected function upgrade_3_8() {
+		$options = get_option( 'polylang' );
+
+		if ( is_array( $options ) && isset( $options['language_taxonomies'] ) && is_array( $options['language_taxonomies'] ) ) {
+			update_option( 'pll_language_taxonomies', $options['language_taxonomies'] );
+		}
```
Chouby, line 248: You must ensure that this option is always created, with an empty array if necessary.
You also must endure that it's the case for new installs as we always do. The goal is to avoid a useless extra query to the options table.

Chouby (changes requested):

install/upgrade.php
manooweb: Reverted in ea49a18d707d57d5c9620d8774b6f8f2260e8656

tests/phpunit/tests/plugins/test-translated-table.php
manooweb: This test doesn't fail because Polylang API is loaded at the beginning.
See https://github.com/polylang/polylang/blob/3.7.3/tests/phpunit/tests/test-model.php#L14
So it can't reproduce what happens really in production even if we trigger the object registration with `do_action( 'pll_model_init' )`.
Indeed `PLL()` is always defined in this case in `Model_Test` test class.
Of course we can add the new assertion to check the new option but we can't here simulate the wrong usage of `PLL()` in the old process.

Done in 941ac51229a05fcc4c013e0f6e1d66755ee225a0

I think the new test is closer of what really happens when the fatal error occurs and should be kept.

Chouby: Is the API necessary in the existing test?

install/upgrade.php
Chouby: You also must ensure that the option is set for new installs.

tests/phpunit/tests/plugins/test-translated-table.php
manooweb: Only in these two tests 
https://github.com/polylang/polylang/blob/3.7.3/tests/phpunit/tests/test-model.php#L140-L184

install/upgrade.php
manooweb: Done in 9d4ba73eeb7f207561be089b0c51f4ae860e8c02

tests/phpunit/tests/plugins/test-translated-table.php
manooweb: Isolate the test in its own class isn't sufficient. It must be run before any other test which runs `self::require_api()` because all of its functions are defined for the whole running PHPUnit process.
So I decided to isolate it in its own test suite as `test-install.php` https://github.com/polylang/polylang/blob/3.7.3/phpunit.xml#L13
Done in 1a5768e9db846f87d1d2b1d70fc5d00e9dfa7b12

install/upgrade.php
Chouby: The best is to keep as is. In case of downgrade, this will allow to account for any change in the option between the 2 upgrades.

PR #1731, opened 2025-09-24
Remove cache domain for `get_terms()`

Chouby:
While working on https://core.trac.wordpress.org/ticket/64038, I noticed that the cache key used for `get_terms()` includes the sql statement instead of only (default) arguments [since WordPress 4.4](https://core.trac.wordpress.org/ticket/21267#comment:19)!
This makes the usage of the cache domain useless so I propose to remove the related code from Polylang.

### DIFF
--- a/include/crud-terms.php
+++ b/include/crud-terms.php
@@ -79,9 +79,6 @@ class PLL_CRUD_Terms {
 		add_filter( 'pre_term_name', array( $this, 'set_pre_term_name' ) );
 		add_filter( 'pre_term_slug', array( $this, 'set_pre_term_slug' ), 10, 2 );
 
-		// Adds cache domain when querying terms
-		add_filter( 'get_terms_args', array( $this, 'get_terms_args' ), 10, 2 );
-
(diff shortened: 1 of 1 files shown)


include/crud-terms.php
```diff
@@ -189,16 +187,14 @@ protected function get_queried_languages( $taxonomies, $args ): array {
...
 	 *
 	 * @since 1.3
 	 *
-	 * @param array    $args       WP_Term_Query arguments.
-	 * @param string[] $taxonomies Queried taxonomies.
+	 * @param array $args WP_Term_Query arguments.
 	 * @return array Modified arguments.
 	 */
-	public function get_terms_args( $args, $taxonomies ) {
+	public function get_terms_args( $args ) {
```
Hug0-Drelon: I suggest we rename this method so it describes what it does.
Some could say it has been removed completely because of bad naming 😅

Chouby:
Reported typos in Spell check action are fixed in #1733

PR #1734, opened 2025-10-01
[Capabilities] Display disabled fields in Strings Translations screen

Screenfeed:
This displays disabled fields in the Strings Translations screen, instead of hiding them.

Tested in https://github.com/polylang/polylang-pro/pull/2752.

### DIFF
--- a/settings/table-string.php
+++ b/settings/table-string.php
@@ -111,7 +111,7 @@ class PLL_Table_String extends WP_List_Table {
 			esc_attr( $item['row'] ),
 			/* translators:  accessibility text, %s is a string potentially in any language */
 			sprintf( __( 'Select %s', 'polylang' ), format_to_edit( $item['string'] ) ),
-			empty( $item['icl'] ) ? 'disabled' : '' // Only strings registered with WPML API can be removed.
+			disabled( empty( $item['icl'] ), true, false ) // Only strings registered with WPML API can be removed.
 		);
 	}
 
@@ -149,12 +149,12 @@ class PLL_Table_String extends WP_List_Table {
 				'<input type="text" name="translation[%1$s][%2$s]" id="%1$s-%2$s" value="%4$s" %5$s/>';
 
 			$out .= sprintf(
-				'<div class="translation"><label for="%1$s-%2$s">%3$s</label>' . $input_type . '</div>' . "\n",
+				'<div class="translation"><label for="%1$s-%2$s">%3$s</label>' . $input_type . "</div>\n",
 				esc_attr( $key ),
 				esc_attr( $item['row'] ),
 				esc_html( $languages[ $key ] ),
 				format_to_edit( $translation ), // Don't interpret special chars.
-				in_array( $key, $this->authorized_language_slugs, true ) ? '' : 'disabled'
+				disabled( in_array( $key, $this->authorized_language_slugs, true ), false, false )
 			);
 		}
 
--- a/settings/table-string.php
+++ b/settings/table-string.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+use WP_Syntex\Polylang\Model\Languages;
+
 if ( ! class_exists( 'WP_List_Table' ) ) {
 	require_once ABSPATH . 'wp-admin/includes/class-wp-list-table.php'; // since WP 3.1
 }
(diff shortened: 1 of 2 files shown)


settings/table-string.php
Hug0-Drelon, line 1: if `User` class lands in Polylang Free, would it make sense to use it instead of `$authorized_language_slugs`?

Hug0-Drelon (changes requested):

settings/table-string.php
Screenfeed: It would, but I would still keep this property because it has an advantage IMHO: where `can_translate()` would check the user capabilities on each field for each language, this property allows to do it only once at the beginning.

Hug0-Drelon (approved):

settings/table-string.php
```diff
@@ -46,20 +55,22 @@ class PLL_Table_String extends WP_List_Table {
...
 				'plural' => 'Strings translations', // Do not translate ( used for css class )
 				'ajax'   => false,
 			)
 		);
 
-		$this->languages = $languages;
-		$this->strings = PLL_Admin_Strings::get_strings();
-		$this->groups = array_unique( wp_list_pluck( $this->strings, 'context' ) );
+		$this->languages                 = $languages->get_list();
+		$this->authorized_language_slugs = $languages->filter( 'translator' )->get_list( array( 'fields' => 'slug' ) );
```
Chouby: I propose you to keep only the languages model as property: 
```suggestion
		$this->languages                 = $languages;
```
Then access to this model to get the list you need when you need it. There are only very few places to modify (see also my other suggestions linked to this one).

settings/table-string.php
```diff
@@ -134,14 +145,16 @@ public function column_translations( $item ) {
...
+				'<input type="text" name="translation[%1$s][%2$s]" id="%1$s-%2$s" value="%4$s" %5$s/>';
+
 			$out .= sprintf(
 				'<div class="translation"><label for="%1$s-%2$s">%3$s</label>' . $input_type . '</div>' . "\n",
 				esc_attr( $key ),
 				esc_attr( $item['row'] ),
 				esc_html( $languages[ $key ] ),
-				format_to_edit( $translation ) // Don't interpret special chars.
+				format_to_edit( $translation ), // Don't interpret special chars.
+				in_array( $key, $this->authorized_language_slugs, true ) ? '' : 'disabled'
```
Chouby: Two suggestions in one:
1. Use the languages model property as proposed above
2. Use WP function `disabled()` as done elsewhere
```suggestion
				disabled( ! in_array( $key, wp_list_pluck( $languages->filter( 'translator' )->get_list(), 'slug' ), true ), true, false )
```

settings/table-string.php
```diff
@@ -388,6 +401,18 @@ public function save_translations() {
...
+						sprintf(
+							/* translators: %1$s is a language name, %2$s is a language locale. */
+							esc_html__( 'Sorry, you are not allowed to translate in %1$s (%2$s).', 'polylang' ),
+							esc_html( $language->name ),
+							sprintf( '<code>%s</code>', esc_html( $language->locale ) )
+						),
+						403
+					);
+				}
+
```
Chouby: This can be avoided as it was previously provided that we loop over the authorized languages at line 399:
```php
foreach ( $this->languages->filter( 'translator' )->get_list() as $language ) {
```
```suggestion
```

Chouby (changes requested):

settings/table-string.php
Screenfeed: Done in [dee12cd](https://github.com/polylang/polylang/pull/1734/commits/dee12cd9bb4a988555165e129cc154bb7f9f0818).

settings/table-string.php
Screenfeed: Partially done in [dee12cd](https://github.com/polylang/polylang/pull/1734/commits/dee12cd9bb4a988555165e129cc154bb7f9f0818).

settings/table-string.php
Screenfeed: As we've seen together:
1. Use `get_list()` in this loop is not efficient, so wee keep the property `$authorized_language_slugs`.
2. `disabled()` used in [3ee1226](https://github.com/polylang/polylang/pull/1734/commits/3ee1226e995dccd5b303118836bd18f6abe0c728).

PR #1737, opened 2025-10-07
Use FQCN for `Options` when the type is used only for PHPStan.

Hug0-Drelon:
## What?
Generated stubs are broken when the `Options` type is used for properties only to please PHPStan ([see](https://github.com/polylang/polylang-stubs/blob/8e1160d23276b4de48f7a2cdcd61ba69426da7c5/polylang-stubs.php#L14006)).

## How?
Use FQCN when a property needs to be typed as `Options` object without it to be coerced in code.

### DIFF
--- a/include/nav-menu.php
+++ b/include/nav-menu.php
@@ -3,8 +3,6 @@
  * @package Polylang
  */
 
-use WP_Syntex\Polylang\Options\Options;
-
 /**
  * Manages custom menus translations
  * Common to admin and frontend for the customizer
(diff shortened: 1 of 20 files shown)


include/nav-menu.php
```diff
@@ -139,7 +137,7 @@ public function explode_location( $loc ) {
 	 *
 	 * @since 0.9.4
 	 *
-	 * @param array $options Options stored in the option nav_menu_options.
+	 * @param array $options \WP_Syntex\Polylang\Options\Options stored in the option nav_menu_options.
```
Chouby: Not our options

Chouby (changes requested):

include/nav-menu.php
Hug0-Drelon: Oops, removed in f28df5ce.

PR #1738, opened 2025-10-08
Get language to assign to new posts and terms

Hug0-Drelon:
## What?
Make language retrieval during entity insertion reusable.
Necessary for https://github.com/polylang/polylang-pro/pull/2744

## How?
- Bring `User` class so the language a translator can managed can be used.
- Move logic from `PLL_CRUD_Posts::set_default_language()` and `PLL_CRUD_Terms::set_default_language()` to new public methods `PLL_CRUD_Posts::get_language_from_context()` and `PLL_CRUD_Terms::get_language_from_context()`.

PR #1740, opened 2025-10-09
Prevent language assignation on entities for unauthorized users.

Hug0-Drelon:
## What?
Check capabilities when trying to assign language in UI action callbacks.
Follows https://github.com/polylang/polylang/pull/1738 (requires the new `User` class).

##  How?
Everywhere `set_language()` is called in an action callback, check the user capabilities for languages.

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -182,6 +182,8 @@ class PLL_Admin_Classic_Editor {
 			wp_die( esc_html( "{$lang_slug} is not a valid language code." ) );
 		}
 
+		( new User() )->can_translate_or_die( $lang );
+
 		$post_type_object = get_post_type_object( $post_type );
 
 		if ( empty( $post_type_object ) ) {
@@ -191,8 +193,6 @@ class PLL_Admin_Classic_Editor {
 		if ( ! current_user_can( $post_type_object->cap->edit_post, $post_ID ) ) {
 			wp_die( 'You are not allowed to edit this post.' );
 		}
-
-		( new User() )->can_translate_or_die( $lang );
 
 		$this->model->post->set_language( $post_ID, $lang );
 


include/Capabilities/User.php
```diff
@@ -116,4 +116,20 @@ public function get_preferred_language( PLL_Model $model ): ?PLL_Language {
...
+	 * Checks if the current user has the rights to assign a language to an object and dies if not.
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language $language    The language to assign.
+	 * @param string       $object_type The type of object (usually `post` or `term`), used to display the error message.
+	 * @return void|never Dies if the user does not have the rights, does nothing otherwise.
+	 */
+	public static function check_assign_rights_or_die( PLL_Language $language, string $object_type ) {
+		if ( ! ( new User() )->can_translate( $language ) ) {
```
Screenfeed: Copy/Paste drama 😏 

```suggestion
		if ( ! $this->can_translate( $language ) ) {
```

include/Capabilities/User.php
```diff
@@ -116,4 +116,20 @@ public function get_preferred_language( PLL_Model $model ): ?PLL_Language {
...
+	/**
+	 * Checks if the current user has the rights to assign a language to an object and dies if not.
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language $language    The language to assign.
+	 * @param string       $object_type The type of object (usually `post` or `term`), used to display the error message.
+	 * @return void|never Dies if the user does not have the rights, does nothing otherwise.
+	 */
+	public static function check_assign_rights_or_die( PLL_Language $language, string $object_type ) {
```
Screenfeed: We can add the return type.

```suggestion
	public static function check_assign_rights_or_die( PLL_Language $language, string $object_type ): void {
```

include/Capabilities/User.php
```diff
@@ -116,4 +116,20 @@ public function get_preferred_language( PLL_Model $model ): ?PLL_Language {
...
+		if ( ! ( new User() )->can_translate( $language ) ) {
+			/* translators: %s: language name */
+			wp_die( esc_html( sprintf( __( 'You are not allowed to assign %1$s to a %2$s.', 'polylang' ), $language->name, $object_type ) ) );
```
Screenfeed: If we want to internationalize this message, we must do it with a translated `$object_type`.

```suggestion
			switch ( $object_type ) {
				case 'post':
					/* translators: %s is a language name. */
					wp_die( esc_html( sprintf( __( 'You are not allowed to assign the %s language to a post.', 'polylang' ), $language->name ) ) );

				case 'media':
					/* translators: %s is a language name. */
					wp_die( esc_html( sprintf( __( 'You are not allowed to assign the %s language to a media.', 'polylang' ), $language->name ) ) );

				case 'term':
					/* translators: %s is a language name. */
					wp_die( esc_html( sprintf( __( 'You are not allowed to assign the %s language to a pterm.', 'polylang' ), $language->name ) ) );

				default:
					/* translators: %s is a language name. */
					wp_die( esc_html( sprintf( __( 'You are not allowed to assign the %s language.', 'polylang' ), $language->name ) ) );
			}
```

include/Capabilities/User.php
```diff
@@ -116,4 +116,20 @@ public function get_preferred_language( PLL_Model $model ): ?PLL_Language {
...
 		return $language ?: null;
 	}
+
+	/**
+	 * Checks if the current user has the rights to assign a language to an object and dies if not.
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language $language    The language to assign.
+	 * @param string       $object_type The type of object (usually `post` or `term`), used to display the error message.
```
Screenfeed:
```suggestion
	 * @param string       $object_type The type of object (usually `post`, `media`, or `term`), used to display the error message.
```

include/Capabilities/User.php
Hug0-Drelon: No, it's not. This is a static method.

include/Capabilities/User.php
Hug0-Drelon: Why not.

include/Capabilities/User.php
Hug0-Drelon: I get why you want to internationalize the entity slugs, but I don't like your proposal. It goes against the idea of passing an argument for the entity kind because you force the method to know what type of entity kind could be passed.

I propose to simplify and remove the entity kind from the string instead.

Hug0-Drelon: See d59d9e75

include/Capabilities/User.php
Screenfeed: Bad reading drama.

Screenfeed: As we saw together: for me, this method being static means it may not be at the right place. However there are not many places where it could make sense if we don't want to create a new class.

So I suggest 2 things:
1. Make it not-static, so `( new User() )` is done outside, and internally it uses `$this`.
2. Rename it to something like `can_assign_language_or_die()`. It would be a sort of `can_translate_or_die()` but dedicated to language assignation because of its error message.

include/Capabilities/User.php
```diff
@@ -124,4 +124,19 @@ private function get_language_caps(): array {
 
 		return $this->language_caps;
 	}
+
+	/**
+	 * Checks if the current user has the rights to assign a language to an object and dies if not.
```
Screenfeed, line 125:
```
Checks if the current user has the rights to assign a language to an object and dies if not (the method, not the user, the user is safe).
```
😂

include/Capabilities/User.php
```diff
@@ -120,4 +120,19 @@ private function get_language_caps(): array {
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param PLL_Language $language The language to assign.
+	 * @return void|never Dies if the user does not have the rights, does nothing otherwise.
+	 */
+	public function can_translate_or_die( PLL_Language $language ): void {
+		if ( ! $this->can_translate( $language ) ) {
+			/* translators: %s: language name */
+			wp_die( esc_html( sprintf( __( 'You are not allowed to assign %1$s to an object.', 'polylang' ), $language->name ) ) );
```
Chouby: Although it seems attractive, I wonder if having a centralized method is the best. The message must be adapted to all cases. Here it's ok for language assignment. But it would not be adapted 
to https://github.com/polylang/polylang-pro/pull/2744/files#diff-380f661f12e6165d70fe385010110023f1adb83237aa282121112694e5ea26fbR79 
or https://github.com/polylang/polylang-pro/blob/f76c3b4275a7c410ace92e7a8673d3d35f4288f4/modules/Machine_Translation/Strings/Action.php#L120-L127

And if we have this centralized method we will want to use it everywhere.

admin/admin-classic-editor.php
```diff
@@ -190,6 +192,8 @@ public function post_lang_choice() {
 			wp_die( 'You are not allowed to edit this post.' );
 		}
 
+		( new User() )->can_translate_or_die( $lang );
+
```
Chouby: I suggest to move the code higher just after we check if `$lang` is empty.

admin/admin-filters-media.php
```diff
@@ -122,6 +124,8 @@ public function save_media( $post, $attachment ) {
 			return $post;
 		}
 
+		( new User() )->can_translate_or_die( $language );
+
```
Chouby: I believe that we are out of the scope of this PR (language change)

admin/admin-filters-term.php
```diff
@@ -284,9 +286,13 @@ protected function save_language( $term_id, $taxonomy ) {
...
 
-			if ( ! empty( $language ) ) {
-				$this->model->term->set_language( $term_id, $language );
+			if ( empty( $language ) ) {
+				return;
 			}
+
+			( new User() )->can_translate_or_die( $language );
+
+			$this->model->term->set_language( $term_id, $language );
```
Chouby: Almost everything in this method comes too late (which assigns language to newly created terms).

admin/admin-classic-editor.php
Hug0-Drelon: Sure, done in d0572fd5.

include/Capabilities/User.php
Hug0-Drelon: As said, let's keep `can_translate_or_die`. I updated the message in 8e5959e3.

PR #1744, opened 2025-10-13
[Capabilities] Custom capa for languages

Screenfeed:
Partially fixes https://github.com/polylang/polylang-pro/issues/2733.

This introduces one new custom capability: `manage_languages`.

1. This capability is used in the REST controller and in the `lang` page.
2. The callbacks to create/edit/delete languages don't need more capability check, it is done by WP when accessing the page.
3. This capability and `manage_translations` are put into constants.
4. These constants are in the class `Capabilities`.

Those capabilities are mapped to `manage_options`, so nothing changes for our users.

### DIFF
--- a/include/Capabilities/Capabilities.php
+++ b/include/Capabilities/Capabilities.php
@@ -35,15 +35,9 @@ class Capabilities {
 	 * @return string[]
 	 */
 	public function map_custom_caps( $caps, $cap ) {
-		$custom_caps = array( self::TRANSLATIONS, self::LANGUAGES );
-
-		foreach ( $custom_caps as $custom_cap ) {
-			if ( $custom_cap !== $cap ) {
-				continue;
-			}
-			$caps   = array_diff( $caps, array( $custom_cap ) );
+		if ( in_array( $cap, array( self::TRANSLATIONS, self::LANGUAGES ), true ) ) {
+			$caps   = array_diff( $caps, array( $cap ) );
 			$caps[] = 'manage_options';
-			return $caps;
 		}
 
 		return $caps;


manooweb (approved):

include/Capabilities/Capabilities.php
```diff
@@ -13,6 +13,9 @@
  * @since 3.8
  */
 class Capabilities {
+	public const LANGUAGES_CAPABILITY    = 'manage_languages';
+	public const TRANSLATIONS_CAPABILITY = 'manage_translations';
```
Chouby: Given the "Capabilities" is already included in the class name, I propose to simplify:
```suggestion
	public const LANGUAGES    = 'manage_languages';
	public const TRANSLATIONS = 'manage_translations';
```

Chouby (changes requested):

include/Capabilities/Capabilities.php
Screenfeed: Done in [f44c792](https://github.com/polylang/polylang/pull/1744/commits/f44c79219ce66ea8d46b3c165db67b62d28948d6).

include/Capabilities/Capabilities.php
```diff
@@ -32,9 +35,15 @@ public function __construct() {
...
-			$caps   = array_diff( $caps, array( 'manage_translations' ) );
+		$custom_caps = array( self::TRANSLATIONS, self::LANGUAGES );
+
+		foreach ( $custom_caps as $custom_cap ) {
+			if ( $custom_cap !== $cap ) {
+				continue;
+			}
+			$caps   = array_diff( $caps, array( $custom_cap ) );
 			$caps[] = 'manage_options';
+			return $caps;
```
Chouby: What do you think of:
```suggestion
		if ( in_array( $cap, array( self::TRANSLATIONS, self::LANGUAGES ), true ) ) {
			$caps   = array_diff( $caps, array( $cap ) );
			$caps[] = 'manage_options';
```

Screenfeed: Done in [6303dbf](https://github.com/polylang/polylang/pull/1744/commits/6303dbf6155efb7d7ae5ee9f988529d387fa9bcf).

PR #1746, opened 2025-10-16
 Prevent language activation when Polylang is not activated on the current blog.

manooweb:
New implementation in #1738 shows that we could have fatal errors on multisite when switching blog.

It was detected by one of our test
https://github.com/polylang/polylang/actions/runs/18344020312/job/52246097892#step:4:648

As this fix needs a deeper analysis (see https://github.com/polylang/polylang-pro/issues/2760), we decided to propose it in its own PR to fix temporarily and avoid test fail on CI.

EDIT:
- For multisite and inside `switch_to_blog()`, check if there are languages before attempting to set a default.
- The logic is included in `save_post()` and `save_term()` to avoid firing the action `pll_save_post` and `pll_save_term` in these cases.
- For `PLL_CRUD_Posts`, the logic is duplicated in `set_default_language()` because the method is public and directly hooked to `add_attachment`.

### DIFF
--- a/include/crud-posts.php
+++ b/include/crud-posts.php
@@ -84,18 +84,25 @@ class PLL_CRUD_Posts {
 	 * @return void
 	 */
 	public function set_default_language( $post_id ) {
-		if ( ! $this->model->post->get_language( $post_id ) && pll_is_plugin_active( POLYLANG_BASENAME ) ) {
-			$post_language = new Create_Post(
-				$this->model,
-				$this->request,
-				$this->pref_lang instanceof PLL_Language ? $this->pref_lang : null, // Can be `false` as well...
-				$this->curlang instanceof PLL_Language ? $this->curlang : null // Can be `false` as well...
-			);
-			$this->model->post->set_language(
-				$post_id,
-				$post_language->get_language( new User(), (int) $post_id )
-			);
+		if ( is_multisite() && ms_is_switched() && ! $this->model->has_languages() ) {
+			return;
 		}
+
+		if ( $this->model->post->get_language( $post_id ) ) {
+			return;
+		}
+
+		$post_language = new Create_Post(
+			$this->model,
+			$this->request,
+			$this->pref_lang instanceof PLL_Language ? $this->pref_lang : null, // Can be `false` as well...
+			$this->curlang instanceof PLL_Language ? $this->curlang : null // Can be `false` as well...
+		);
+
+		$this->model->post->set_language(
+			$post_id,
+			$post_language->get_language( new User(), (int) $post_id )
+		);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Hug0-Drelon (approved):

Chouby (changes requested):
Could you fix the merge conflict? Also early return may avoid a pyramid step.

manooweb:
> Could you fix the merge conflict? Also early return may avoid a pyramid step.

By resolving conflicts, there are no longer pyramid step in the new code. Do you still want an early return ?

include/crud-posts.php
```diff
@@ -109,29 +116,34 @@ public function set_default_language( $post_id ) {
...
 	 */
 	public function save_post( $post_id, $post ) {
-		// Does nothing except on post types which are filterable.
-		if ( $this->model->is_translated_post_type( $post->post_type ) ) {
-			if ( $id = wp_is_post_revision( $post_id ) ) {
-				$post_id = $id;
-			}
+		if ( is_multisite() && ms_is_switched() && ! $this->model->has_languages() ) {
+			return;
+		}
```
Hug0-Drelon, line 121: Is it really necessary since it's already done in `set_default_language()`?

include/crud-terms.php
```diff
@@ -109,18 +109,17 @@ public function __construct( PLL_Base &$polylang ) {
...
+			$this->model,
+			$this->request,
+			$this->pref_lang instanceof PLL_Language ? $this->pref_lang : null, // Can be `false` as well...
+			$this->curlang instanceof PLL_Language ? $this->curlang : null // Can be `false` as well...
+		);
+
+		$this->model->term->set_language(
+			$term_id,
+			$term_language->get_language( new User(), (int) $term_id, (string) $taxonomy )
+		);
```
Hug0-Drelon, line 122: If find weird to check MS in `PLL_CRUD_Post::set_default_language()` but not in `PLL_CRUD_Term::set_default_language()`. Is it a problem?

include/crud-posts.php
Chouby: The reason is to prevent the action `pll_save_post` to be fired. Since we don't want to save anything in this context, I believe we should prevent the usage of this hook.

include/crud-terms.php
Chouby: The check is done in `PLL_CRUD_Post::set_default_language()` because it's called directly for media (without going through `save_post()`.
For terms, `set_default_language()` is always called from `save_term()` so it would be useless to double check.

Hug0-Drelon: I understood that it works like that. I just feel like it should be symmetric, like we usually do for post/term objects.

Chouby: In fact this wasn't symmetric already. One is public, the other is protected.

PR #1748, opened 2025-10-17
Get rid of the remaining `toplevel`

Screenfeed:
Follows the harmonization work done with #1727, about hook names and CSS classes in `PLL_Admin_Base::add_menus()`.
This also fixes a bug that prevented to display our admin notices on PLL's translations and settings screens.

## About the harmonization

1. Hack the current screen's ID and base, to change `toplevel_page_mlang` into `languages_page_mlang`. This simplifies a lot of things for us because we don't need to worry about the `toplevel` prefix anymore. While side effects are not impossible (we're changing the identity of the current screen, its ID is used in hooks), the probability is low IMHO, it would require someone to use `get_plugin_page_hookname()` (which returns the `toplevel` prefix and is not filterable).
2. Create 2 new static methods:
    - `PLL_Admin_Base::get_screen_slug()` returns the slug of a screen, as it is shown in the screen's URL.
    - `PLL_Admin_Base::get_screen_id()` returns the ID of the screen: it can be used in hook names and to change the screen ID.

## About the bug being fixed

`languages` is hard-coded in `PLL_Admin_Base::add_menus()`:

```php
$admin_page_hooks['mlang'] = 'languages';
```

Trace:

- `$GLOBALS['admin_page_hooks']` is used in `get_plugin_page_hookname()`,
- which is used by `get_plugin_page_hook()`,
- which is used to determine `$GLOBALS['hook_suffix']`,
- which is used in `WP_Screen::get()` to determine the screen ID.

### DIFF
--- a/admin/admin-base.php
+++ b/admin/admin-base.php
@@ -625,7 +625,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	}
 
 	/**
-	 * Returns the ID of a Polyang's settings screen.
+	 * Returns the ID of a Polylang's settings screen.
 	 *
 	 * @since 3.8
 	 *
@@ -637,7 +637,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	}
 
 	/**
-	 * Returns the slug of a Polyang's settings screen, as seen in the URL.
+	 * Returns the slug of a Polylang's settings screen, as seen in the URL.
 	 *
 	 * @since 3.8
 	 *


tests/phpunit/tests/test-settings.php
```diff
@@ -72,6 +72,7 @@ public function test_notice_for_objects_with_no_lang() {
 
 		$links_model = self::$model->get_links_model();
 		new PLL_Settings( $links_model );
+		do_action( 'admin_menu' );
 		do_action( 'load-toplevel_page_mlang' );
```
Marianne380: Could we use the new hook here too ?
```suggestion
		do_action( 'load-languages_page_mlang' );
```

Screenfeed: Yes. I didn't use it at first because `do_action( 'admin_menu' );` was missing and it was failing.
Done in [0cbfce6](https://github.com/polylang/polylang/pull/1748/commits/0cbfce6a8473eb424f4d1394549b13a74787f0ae).

admin/admin-base.php
```diff
@@ -599,6 +624,30 @@ public function should_hide_admin_bar_menu(): bool {
 		return false;
 	}
 
+	/**
+	 * Returns the ID of a Polyang's settings screen.
```
Marianne380:
```suggestion
	 * Returns the ID of a Polylang's settings screen.
```

admin/admin-base.php
```diff
@@ -599,6 +624,30 @@ public function should_hide_admin_bar_menu(): bool {
...
+	 *
+	 * @param string $tab The name of the screen (`lang`, `strings`, `settings`).
+	 * @return string
+	 */
+	public static function get_screen_id( string $tab ): string {
+		return sprintf( '%s_page_%s', self::SCREEN_PREFIX, self::get_screen_slug( $tab ) );
+	}
+
+	/**
+	 * Returns the slug of a Polyang's settings screen, as seen in the URL.
```
Marianne380:
```suggestion
	 * Returns the slug of a Polylang's settings screen, as seen in the URL.
```

PR #1750, opened 2025-10-22
Add translator capabilities filtering for inline edit and terms management.

Marianne380:
Add support for translator capabilities filtering in inline/quick edit and terms management interfaces.

For term creation : when creating a new term, if the user's preferred language is not in their list of authorized languages, it now defaults to the first available language in the filtered list. 
This ensures the language dropdown flag displays correctly on initial page load, as the selected language must exist in the filtered languages list for the flag to render properly.

Also prevent modifying terms translations in the save callback.

### DIFF
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -376,16 +376,13 @@ class PLL_Admin_Filters_Term {
 		// Security check as 'wp_update_term' can be called from outside WP admin.
 		check_admin_referer( 'pll_language', '_pll_nonce' );
 
-		$translations = $this->model->term->get_translations( $term_id );
-		if ( ! isset( $_POST['term_tr_lang'] ) || ! is_array( $_POST['term_tr_lang'] ) ) {
-			return $translations;
-		}
+		$translations = array();
 
-		// Make sure a translator won't edit translations they're not allowed to.
-		$sent_translations = array_map( 'absint', $_POST['term_tr_lang'] );
-		foreach ( $this->model->languages->filter( 'translator' )->get_list() as $lang ) {
-			if ( isset( $sent_translations[ $lang->slug ] ) ) {
-				$translations[ $lang->slug ] = $sent_translations[ $lang->slug ];
+		// Save translations after checking the translated term is in the right language ( as well as cast id to int ).
+		if ( isset( $_POST['term_tr_lang'] ) ) {
+			foreach ( array_map( 'absint', $_POST['term_tr_lang'] ) as $lang => $tr_id ) {
+				$tr_lang = $this->model->term->get_language( $tr_id );
+				$translations[ $lang ] = $tr_lang && $tr_lang->slug == $lang ? $tr_id : 0;
 			}
 		}
 
--- a/admin/view-translations-term.php
+++ b/admin/view-translations-term.php
@@ -27,14 +27,10 @@ else {
 ?>
 <table class="widefat term-translations"  id="<?php echo isset( $term_id ) ? 'edit' : 'add'; ?>-term-translations">
 	<?php
-	$authorized_slugs = $this->model->languages->filter( 'translator' )->get_list( array( 'fields' => 'slug' ) );
 	foreach ( $this->model->get_languages_list() as $language ) {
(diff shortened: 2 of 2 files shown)


admin/view-translations-term.php
```diff
@@ -27,7 +27,7 @@
 ?>
 <table class="widefat term-translations"  id="<?php echo isset( $term_id ) ? 'edit' : 'add'; ?>-term-translations">
 	<?php
-	foreach ( $this->model->get_languages_list() as $language ) {
+	foreach ( $this->model->languages->filter( 'translator' )->get_list() as $language ) {
```
Screenfeed: I think the inputs related to forbidden languages should be disabled instead of being hidden. It's what we do in the post metabox.

admin/admin-filters-term.php
```diff
@@ -118,8 +118,15 @@ public function add_term_form() {
 
 		$dropdown = new PLL_Walker_Dropdown();
 
+		$available_languages = $this->model->languages->filter( 'translator' )->get_list();
+
+		// Ensure that the selected language is in the filtered list to display the flag correctly on page load.
+		if ( $lang && ! in_array( $lang, $available_languages, true ) ) {
+			$lang = reset( $available_languages ) ?: $lang;
```
Screenfeed: I don't think that selecting another language makes sense: this case should never happen because the process will die before that, thanks to https://github.com/polylang/polylang-pro/pull/2744. So we can just bail out IMHO.

```suggestion
			return;
```

Screenfeed (changes requested):

admin/view-translations-term.php
Marianne380: You're right, see https://github.com/polylang/polylang/pull/1750/commits/90cabbbb55a773d8c426d2ee3f52f8e47e2c45e5.

admin/admin-filters-term.php
Marianne380: We just want to display the terms page here; we are only displaying the term form, we are not inserting a term.

The issue here is that if the language `pref_lang` is not a authorized language for translation, the selected language (in `‘selected’`) will not be among the available languages (with `$this->model->languages->filter( ‘translator’ ) ->get_list()`), and as a result we won't have the flag that should appear next to the switcher on the terms page with the creation form.

Screenfeed: OK, you're talking about `pref_lang`, good point. I was talking about `$_GET['new_lang']`:

```php
$lang = isset( $_GET['new_lang'] ) ? $this->model->get_language( sanitize_key( $_GET['new_lang'] ) ) : $this->pref_lang;
```

In the case of `$_GET['new_lang']` we are creating a term, but Hugo's PR will `die()` if the new language is not allowed. So we can forget this case. Which leaves us with only your case.
All good then 👍

admin/view-translations-term.php
```diff
@@ -27,11 +27,16 @@
 ?>
 <table class="widefat term-translations"  id="<?php echo isset( $term_id ) ? 'edit' : 'add'; ?>-term-translations">
 	<?php
+	$authorized_languages = $this->model->languages->filter( 'translator' )->get_list();
+	$authorized_slugs     = wp_list_pluck( $authorized_languages, 'slug' );
```
Screenfeed: Since `$authorized_languages` is not used:

```suggestion
	$authorized_slugs = $this->model->languages->filter( 'translator' )->get_list( array( 'fields' => 'slug' ) );
```

Screenfeed (approved):

admin/admin-filters-term.php
```diff
@@ -118,8 +118,15 @@ public function add_term_form() {
...
+		}
```
Chouby: This operation would probably be required elsewhere.
Instead of it, I propose that we ensure that `$this->pref_lang` is an authorized language when this property is created. 
This can be done in a separate PR.

Marianne380: Good point!
I've reverted this check and will open a separate issue to ensure `pref_lang` is always an authorized language when the property is created.
See https://github.com/polylang/polylang/issues/1752.

PR #1751, opened 2025-10-23
[Capabilities] Classic editor

Screenfeed:
Partially closes https://github.com/polylang/polylang-pro/issues/2779.
The "Sync" button must still be handled.

1. Limit the language selector to allowed languages.
2. Translations list: display gray icons when the "Add/Edit" actions are not allowed.
3. Media edition popup: when the medias list is displayed as tiles, disallowed users can still click any tile and access the media data. In the case of a user that is not allowed to edit the media, the language selector is disabled.

### DIFF
--- a/admin/admin-filters-media.php
+++ b/admin/admin-filters-media.php
@@ -54,30 +54,23 @@ class PLL_Admin_Filters_Media extends PLL_Admin_Filters_Post_Base {
 			return $fields; // Don't add anything on edit media panel since we have the metabox.
 		}
 
-		$post_id = $post->ID;
-		$lang    = $this->model->post->get_language( $post_id );
-		$user    = new User();
+		$post_id   = $post->ID;
+		$lang      = $this->model->post->get_language( $post_id );
+		$user      = new User();
+		$languages = $this->model->languages->filter( 'translator' )->get_list();
 
-		if ( ! empty( $lang ) ) {
-			// The item has a language.
-			if ( $user->can_translate( $lang ) ) {
-				// The user is allowed to translate this language: display the list.
-				$languages = $this->model->languages->filter( 'translator' )->get_list();
-				$disabled  = false;
-			} else {
-				// The user is not allowed to translate this language: display only the item's language in a disabled selector.
-				$languages = array( $lang );
-				$disabled  = true;
-			}
-		} elseif ( ! $user->is_translator() ) {
-			// The item has no language and the user has access to all languages: display the full list.
-			$languages = $this->model->languages->get_list();
-			$disabled  = false;
-		} else {
-			// The item has no language and the user is not allowed to assign one: disabled and empty selector.
-			$languages = array();
-			$disabled  = true;
+		if ( empty( $lang ) ) {
+			// The media has no language: prepend an empty option to prevent displaying a wrong information.
+			array_unshift( $languages, (object) array( 'slug' => '', 'name' => '' ) );
+		} elseif ( ! $user->can_translate( $lang ) ) {
+			// The user cannot translate this language, so it isn't in `$languages` (the dropdown will be disabled).
+			array_unshift( $languages, $lang );
 		}
+
+		// Disable the dropdown if:
+		// - the media has no language and the user is a translator,
+		// - or the media has a language but the user is not allowed to translate it.
+		$disabled = empty( $lang ) ? $user->is_translator() : ! $user->can_translate( $lang );
 
 		$dropdown = new PLL_Walker_Dropdown();
 		$fields['language'] = array(


admin/admin-links.php
```diff
@@ -3,60 +3,50 @@
...
 	/**
-	 * Returns the html markup for a translation link.
+	 * Returns the html markup for a new post translation link.
 	 *
-	 * @since 2.6
+	 * @since 1.8
 	 *
-	 * @param string       $link     The translation link.
-	 * @param PLL_Language $language The language of the translation.
+	 * @param int          $post_id  The source post id.
```
Marianne380:
```suggestion
	 * @param int          $post_id  The source post ID.
```

admin/admin-links.php
```diff
@@ -146,14 +164,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
 	}
 
 	/**
-	 * Get the link to create a new term translation.
+	 * Returns the URL to edit a post.
+	 * Returns an empty string if the current user is not allowed to edit the post.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int          $post_id  The source post id.
```
Marianne380:
```suggestion
	 * @param int          $post_id  The source post ID.
```

admin/admin-links.php
```diff
@@ -146,14 +164,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
+			esc_html( sprintf( __( 'Edit the translation in %s', 'polylang' ), $language->name ) )
+		);
+	}
+
+	/**
+	 * Returns the html markup for a new term translation.
+	 *
+	 * @since 1.8
+	 *
+	 * @param int          $term_id   Source term id.
```
Marianne380:
```suggestion
	 * @param int          $term_id   Source term ID.
```

admin/admin-links.php
```diff
@@ -218,12 +294,38 @@ public function new_term_translation_link( $term_id, $taxonomy, $post_type, $lan
...
 		return $this->edit_translation_link( $link, $language );
 	}
 
+	/**
+	 * Returns the URL to edit a term.
+	 * Returns an empty string if the current user is not allowed to edit the term.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int          $term_id   Translation term id.
```
Marianne380:
```suggestion
	 * @param int          $term_id   Translation term ID.
```

tests/phpunit/tests/test-columns.php
```diff
@@ -48,84 +53,129 @@ public function test_post_with_no_language() {
...
+	 *
+	 * @param int $user_id Current user ID.
+	 * @return void
+	 */
+	public function test_post_language( int $user_id ) {
 		$en = self::factory()->post->create( array( 'lang' => 'en' ) );
 
+		if ( ! $user_id ) {
+			wp_set_current_user( 0 );
+		}
```
Chouby, line 68:
```suggestion
		wp_set_current_user( $user_id );
```

admin/admin-links.php
```diff
@@ -125,17 +127,33 @@ public function get_new_post_translation_link( $post_id, $language, $context = '
...
 	 * @param PLL_Language $language The language of the new translation.
 	 * @return string
 	 */
-	public function new_post_translation_link( $post_id, $language ) {
-		$link = $this->get_new_post_translation_link( $post_id, $language );
-		return $this->new_translation_link( $link, $language );
+	protected function new_translation_link( string $link, PLL_Language $language ): string {
+		if ( empty( $link ) ) {
+			/* translators: accessibility text, %s is a native language name */
+			$hint = sprintf( __( 'You are not allowed to add a translation in %s', 'polylang' ), $language->name );
```
Chouby: I would keep this only visual, not adding any hint as no action is possible.

admin/admin-links.php
```diff
@@ -146,14 +164,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
-		$link = get_edit_post_link( $post_id );
+	public function edit_post_translation_link( int $post_id ): string {
 		$language = $this->model->post->get_language( $post_id );
+
+		if ( empty( $language ) ) {
+			// Should not happen.
+			return '';
+		}
+
+		$link = $this->get_edit_post_translation_link( $post_id, $language );
```
Chouby: We should keep using the WP function as it checks capability 'edit_post'. If a user can't edit in the post's language, the WP function won't return any link.

admin/admin-links.php
```diff
@@ -146,14 +164,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
+	 * Returns the URL to edit a post.
+	 * Returns an empty string if the current user is not allowed to edit the post.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int          $post_id  The source post ID.
+	 * @param PLL_Language $language The language of the post.
+	 * @return string
+	 */
+	public function get_edit_post_translation_link( int $post_id, PLL_Language $language ): string {
```
Chouby: See my comment above.

admin/admin-links.php
```diff
@@ -146,14 +164,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
+	 * @since 3.8 Visibility changed from `public` to `protected`.
+	 *
+	 * @param string       $link     The translation link.
+	 * @param PLL_Language $language The language of the translation.
+	 * @return string
+	 */
+	protected function edit_translation_link( string $link, PLL_Language $language ): string {
+		if ( empty( $link ) ) {
+			/* translators: accessibility text, %s is a native language name */
+			$hint = sprintf( __( 'You are not allowed to edit a translation in %s', 'polylang' ), $language->name );
```
Chouby: Same as for post.

admin/admin-links.php
```diff
@@ -218,12 +294,38 @@ public function new_term_translation_link( $term_id, $taxonomy, $post_type, $lan
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param int          $term_id   Translation term ID.
+	 * @param string       $taxonomy  Taxonomy name.
+	 * @param string       $post_type Post type name.
+	 * @param PLL_Language $language  The language of the term.
+	 * @return string
+	 */
+	public function get_edit_term_translation_link( int $term_id, string $taxonomy, string $post_type, PLL_Language $language ): string {
```
Chouby: Same as above.

tests/phpunit/tests/test-columns.php
Screenfeed: No, the `$user_id` parameter is not really a user ID. I tried to do that at first but:
1. `1` is the ID of the admin user, not the editor's created during the setup.
2. The editor is already set as current user [in `set_up()`](https://github.com/polylang/polylang/pull/1751/files#diff-9bba848c2cd3c8ffd25862c8060c2fe72c3b042c1562ca97db212c163d298114R33).
3. Users and user IDs provided by data provider functions cannot be trusted.

Yes the parameter is badly named, because I tried to provide user IDs at first but in the end it was a bad idea so I changed it. I kept what I've done because it will be overwritten [in another PR anyway](https://github.com/polylang/polylang/pull/1754/files#diff-9bba848c2cd3c8ffd25862c8060c2fe72c3b042c1562ca97db212c163d298114R65-R72), so I'd prefer not to touch this file to prevent creating conflicts.

admin/admin-links.php
Screenfeed: The method uses `get_edit_post_link()` internally.

For each type (post/term, add/edit), I tried to keep the same architecture:
1. A public method that returns the full link:
    - `new_post_translation_link()`,
    - `edit_post_translation_link()`,
    - `new_term_translation_link()`,
    - `edit_term_translation_link()`.
2. A public method that returns the link URL:
    - `get_new_post_translation_link()`,
    - `get_edit_post_translation_link()`,
    - `get_new_term_translation_link()`,
    - `get_edit_term_translation_link()`.
3. A private method that returns the markup, given the URL:
    - `new_translation_link()`,
    - `edit_translation_link()`.

That's why `get_edit_post_link()` is not called directly anymore.

PS: I think those methods will have to be renamed some day because "link" can mean the HTML tag or the URL, so for each method we don't know what we get.

admin/admin-links.php
Screenfeed: If non-blind people get the information that they cannot create a translation, blind people should get it too.

admin/admin-links.php
Screenfeed: `get_edit_term_link()` is used internally.

admin/admin-links.php
Screenfeed: Done in [83d3f4f](https://github.com/polylang/polylang/pull/1751/commits/83d3f4f54d315eba10c674d828248ab5beaa5bc1).

admin/admin-links.php
```diff
@@ -125,17 +127,31 @@ public function get_new_post_translation_link( $post_id, $language, $context = '
...
-	public function new_post_translation_link( $post_id, $language ) {
-		$link = $this->get_new_post_translation_link( $post_id, $language );
-		return $this->new_translation_link( $link, $language );
+	protected function new_translation_link( string $link, PLL_Language $language ): string {
+		if ( empty( $link ) ) {
+			return sprintf(
+				'<span title="%s" class="pll_icon_add wp-ui-text-icon"></span>',
+				/* translators: accessibility text, %s is a native language name */
+				esc_attr( sprintf( __( 'You are not allowed to add a translation in %s', 'polylang' ), $language->name ) )
+			);
```
Chouby, line 144: As discussed I would not add any accessibility text when there is no link.

admin/admin-links.php
Chouby: I keep thinking that `get_edit_post_translation_link()` and `get_edit_term_translation_link()` are useless and just add overhead. 
We don't need the language to know if a post/term can be edited. Only the ID is necessary. That's why a direct call to the WP function is sufficient.

admin/admin-links.php
```diff
@@ -146,14 +162,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
+	 * @since 3.8
+	 *
+	 * @param int          $post_id  The source post ID.
+	 * @param PLL_Language $language The language of the post.
+	 * @return string
+	 */
+	public function get_edit_post_translation_link( int $post_id, PLL_Language $language ): string {
+		if ( ! $this->user->can_translate( $language ) ) {
+			return '';
+		}
```
Chouby: As a complement of my comment above, this part is useless because if the use can't translate in the post's language, `get_edit_post_link()` will return an empty string.

admin/admin-links.php
```diff
@@ -146,14 +162,84 @@ public function new_post_translation_link( $post_id, $language ) {
...
+	 * @param PLL_Language $language The language of the translation.
+	 * @return string
+	 */
+	protected function edit_translation_link( string $link, PLL_Language $language ): string {
+		if ( empty( $link ) ) {
+			return sprintf(
+				'<span title="%s" class="pll_icon_edit wp-ui-text-icon"></span>',
+				/* translators: accessibility text, %s is a native language name */
+				esc_attr( sprintf( __( 'You are not allowed to edit a translation in %s', 'polylang' ), $language->name ) )
+			);
```
Chouby, line 193: Same as above for the accessibility text without link.

admin/admin-filters-media.php
```diff
@@ -51,23 +51,34 @@ public function __construct( &$polylang ) {
...
+		$user    = new User();
+
+		if ( ( empty( $lang ) && $user->is_translator() ) || ( ! empty( $lang ) && ! $user->can_translate( $lang ) ) ) {
+			// The user is not allowed to edit this attachment.
+			$languages = $this->model->languages->get_list();
+			$disabled  = true;
+		} else {
+			$languages = $this->model->languages->filter( 'translator' )->get_list();
+			$disabled  = false;
+		}
```
Chouby: For the languages list, we don't need a complete list if the select is disabled.
For the test, I find it clearer like that (see also my comment below):
```suggestion
		$is_allowed = empty( $lang ) ? ! $user->is_translator() : $user->can_translate( $lang );
		$languages  = $is_allowed ? $this->model->languages->filter( 'translator' )->get_list() : $lang;
```

admin/admin-filters-media.php
```diff
@@ -51,23 +51,34 @@ public function __construct( &$polylang ) {
...
 			'input' => 'html',
 			'html'  => $dropdown->walk(
-				$this->model->get_languages_list(),
+				$languages,
 				-1,
 				array(
 					'name'     => sprintf( 'attachments[%d][language]', $post_id ),
 					'class'    => 'media_lang_choice',
 					'selected' => $lang ? $lang->slug : '',
+					'disabled' => $disabled,
```
Chouby, line 86:
```suggestion
					'disabled' => ! $is_allowed,
```

Chouby (changes requested):

admin/admin-links.php
Screenfeed: As we discussed, we keep the `title` attribute for now.

admin/admin-filters-media.php
Screenfeed: Following our conversation, I handled more cases in [ad66137](https://github.com/polylang/polylang/pull/1751/commits/ad66137513b195c50c169604fef29b7a0cbe1323) instead of shortening the code. I had to make the code more clear by adding more `if` statements.
Good for you?

admin/admin-filters-media.php
```diff
@@ -51,23 +51,46 @@ public function __construct( &$polylang ) {
...
+			}
+		} elseif ( ! $user->is_translator() ) {
+			// The item has no language and the user has access to all languages: display the full list.
+			$languages = $this->model->languages->get_list();
+			$disabled  = false;
+		} else {
+			// The item has no language and the user is not allowed to assign one: disabled and empty selector.
+			$languages = array();
+			$disabled  = true;
+		}
```
Chouby: I suggest to separate the 2 logics to clarify the code.

For the languages, I propose:
```php
$languages = $this->model->languages->filter( 'translator' )->get_list(); // The generic case. This automatically returns the full list wen possible.

if ( empty( $lang ) ) {
	array_unshift( $languages, (object) array( '' => '' ) ); // Specific case when the media has no language. I don't set the variable to an empty array to avoid an ugly empty select.
} elseif ( ! $user->can_translate( $lang ) ) {
	array_unshift( $languages, $lang ); // The option will be disabled anyway. We could use $languages = array( $lang ); if you prefer.
}
```
Maybe we could introduce somewhere a new reusable method to handle this.

To disable the select:
```php
$disabled = empty( $lang ) ? $user->is_translator() : ! $user->can_translate( $lang );
```
This could even be inlined if you want.

Chouby (changes requested):

admin/admin-filters-media.php
Screenfeed: Done in [db8c981](https://github.com/polylang/polylang/pull/1751/commits/db8c9818968ba3a693542215fa4f9e0c9d0aea6c) with a ton of comments.

PR #1754, opened 2025-10-27
[Capabilities] Posts/Terms lists

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/2778.

This PR adds capability management to the list of posts and terms, for the "Add item" and "Edit item" links, similarly to [the post metabox](https://github.com/polylang/polylang/pull/1751).

All links are handled by `PLL_Admin_Links`. The "Edit item methods" in `PLL_Admin_Links` have been reworked to be able to handle all these cases:
1. `metabox` for the buttons in the metabox,
2. `icon` for the buttons in the items list,
3. `flag` for the buttons for the current item in the items list.

Since [bbc56c8](https://github.com/polylang/polylang/pull/1754/commits/bbc56c81fce80484a1c7569b82e153f418fe43d5) (we don't check the user can translate here anymore, it's done by the mapper for `edit_post`), the tests have been moved [to Polylang Pro](https://github.com/polylang/polylang-pro/pull/2800).

Example for a EN/FR translator, on a site with EN/FR/ES/IT, and the DE language has been removed (which implies that a post has no language):

image

> [!CAUTION]
> https://github.com/polylang/polylang-pro/pull/2810 should be merged right after this PR.

### DIFF
--- a/admin/view-translations-media.php
+++ b/admin/view-translations-media.php
@@ -1,7 +1,6 @@
 <?php
 /**
  * Displays the translations fields for media
- * Needs WP 3.5+
  *
  * @package Polylang
  *
@@ -40,7 +39,7 @@ defined( 'ABSPATH' ) || exit;
 					);
 					echo $this->links->get_edit_post_link_html( $translation ); // phpcs:ignore WordPress.Security.EscapeOutput
 				} else {
-					// The translation doesn't exist anymore.
+					// The translation doesn't exist.
 					echo $this->links->get_new_post_link_html( $post, $language ); // phpcs:ignore WordPress.Security.EscapeOutput
 				}
 				?>


admin/admin-filters-columns.php
```diff
@@ -92,9 +96,9 @@ protected function add_column( $columns, $before ) {
 	 *
 	 * @since 0.9
 	 *
-	 * @return string first language column name.
+	 * @return string First language column name.
 	 */
-	protected function get_first_language_column() {
+	protected function get_first_language_column(): string {
```
Hug0-Drelon, line 101: Since you took the opportunity to type hint `get_first_language_column()`, what do you think of removing the `foreach` to get the first item of the array?
Also, it could be improved by using `PLL_Model::get_default_language()`, am I right?

admin/admin-filters-columns.php
```diff
@@ -153,91 +162,63 @@ public function post_column( $column, $post_id ) {
 			return;
 		}
 
-		// Hidden field containing the post language for quick edit
-		if ( $column == $this->get_first_language_column() ) {
-			printf( '<div class="hidden" id="lang_%d">%s</div>', intval( $post_id ), esc_html( $lang->slug ) );
+		// Hidden field containing the post language for quick edit.
+		if ( $column === $this->get_first_language_column() ) {
+			printf( '<div class="hidden" id="lang_%d">%s</div>', $post_id, esc_html( $lang->slug ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Hug0-Drelon:
```suggestion
			printf( '<div class="hidden" id="lang_%d">%s</div>', $post_id, esc_html( $lang->slug ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped `$post_id` already converted to integer above.
```

admin/admin-filters-columns.php
```diff
@@ -266,101 +247,80 @@ public function add_term_column( $columns ) {
...
-		if ( false === strpos( $column, 'language_' ) || ! ( $lang = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		if ( false === strpos( $column, 'language_' ) ) {
 			return $out;
 		}
 
-		if ( isset( $_REQUEST['post_type'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$post_type = sanitize_key( $_REQUEST['post_type'] ); // phpcs:ignore WordPress.Security.NonceVerification
+		$term_id = (int) $term_id;
+		$inline  = wp_doing_ajax() && isset( $_REQUEST['action'], $_POST['inline_lang_choice'] ) && 'inline-save-tax' === $_REQUEST['action']; // phpcs:ignore WordPress.Security.NonceVerification
+		$lang    = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id ); // phpcs:ignore WordPress.Security.NonceVerification
```
Hug0-Drelon:
```suggestion
		$lang    = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id ); // phpcs:ignore WordPress.Security.NonceVerification `$term_id` already converted to integer above.
```

admin/admin-filters-columns.php
```diff
@@ -266,101 +247,80 @@ public function add_term_column( $columns ) {
...
 		}
 
 		if ( isset( $GLOBALS['post_type'] ) ) {
 			$post_type = $GLOBALS['post_type'];
-		}
-
-		if ( isset( $_REQUEST['taxonomy'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$taxonomy = sanitize_key( $_REQUEST['taxonomy'] ); // phpcs:ignore WordPress.Security.NonceVerification
+		} elseif ( isset( $_REQUEST['post_type'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			$post_type = sanitize_key( $_REQUEST['post_type'] ); // phpcs:ignore WordPress.Security.NonceVerification
```
Hug0-Drelon, line 267: Here the behavior changes, `$_REQUEST['post_type']` doesn't take precedence anymore. Is it what you want?

admin/admin-filters-columns.php
```diff
@@ -266,101 +247,80 @@ public function add_term_column( $columns ) {
...
 		}
 
 		if ( isset( $GLOBALS['taxonomy'] ) ) {
 			$taxonomy = $GLOBALS['taxonomy'];
+		} elseif ( isset( $_REQUEST['taxonomy'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
Hug0-Drelon, line 272: Same question as above.

admin/admin-filters-columns.php
```diff
@@ -454,7 +415,70 @@ public function ajax_update_term_rows() {
...
+
+			return sprintf(
+				'<a href="%s" class="%s" title="%s"><span class="screen-reader-text">%s</span>%s</a>',
+				esc_url( $url ),
+				esc_attr( $class ),
+				esc_attr( $item_name ),
+				esc_html( $hint ),
+				$flag // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			);
+		}
```
Hug0-Drelon: I think this method should take place in `PLL_Admin_Links`, much like `get_new_post_translation_link` or `get_new_term_translation_link` for instance.
With your proposal, the code that generates HTML button is spread in `PLL_Admin_Filters_Columns` and `PLL_Admin_Links`.

tests/phpunit/tests/test-columns.php
```diff
@@ -18,11 +20,17 @@ public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
 
 		$factory->language->create_many( 2 );
 
-		self::$editor = self::factory()->user->create( array( 'role' => 'editor' ) );
+		self::$editor        = self::factory()->user->create( array( 'role' => 'editor' ) );
+		self::$translator_fr = self::factory()->user->create( array( 'role' => 'editor' ) );
+		self::$translator_es = self::factory()->user->create( array( 'role' => 'editor' ) );
```
Hug0-Drelon:
```suggestion
		self::$editor        = self::factory()->user->create_and_get( array( 'role' => 'editor' ) );
		self::$translator_fr = self::factory()->user->create_and_get( array( 'role' => 'editor' ) );
		self::$translator_es = self::factory()->user->create_and_get( array( 'role' => 'editor' ) );
```
This way you can get rid of the `new User` bellow.

tests/phpunit/tests/test-columns.php
```diff
@@ -104,64 +115,66 @@ public function test_untranslated_post( int $user_id ) {
 
 		$this->assertNotFalse( strpos( $column, 'pll_icon_add' ) );
 
-		if ( $user_id ) {
-			$this->assertNotFalse( strpos( $column, 'from_post' ) );
-			$this->assertSame( 0, strpos( $column, '<a href="' ) );
+		if ( empty( $user ) || 'translator_fr' === $user ) {
+			$this->assertNotFalse( strpos( $column, 'href="' ), 'A non-translator or a French translator should be able to create a translation in French.' );
+			$this->assertNotFalse( strpos( $column, 'from_post' ), 'The URL allowing to create a translation should contain the `from_post` query arg.' );
```
Hug0-Drelon: Since the test discriminates users all over the place, I suggest to split the case and add an `if` statement to test non-translator and French translator seperately.
You'll get the same assertions but different messages, so if something goes wrong you'll know what directly. What do you think?

tests/phpunit/tests/test-columns.php
```diff
@@ -104,64 +115,66 @@ public function test_untranslated_post( int $user_id ) {
...
 		$column = ob_get_clean();
 
 		$this->assertNotFalse( strpos( $column, 'pll_icon_add' ) );
 
-		if ( $user_id ) {
-			$this->assertNotFalse( strpos( $column, 'from_media' ) );
-			$this->assertSame( 0, strpos( $column, '<a href="' ) );
+		if ( empty( $user ) || 'translator_fr' === $user ) {
+			$this->assertNotFalse( strpos( $column, 'href="' ), 'A non-translator or a French translator should be able to create a translation in French.' );
+			$this->assertNotFalse( strpos( $column, 'from_media' ), 'The URL allowing to create a translation should contain the `from_media` query arg.' );
```
Hug0-Drelon: Same suggestion as above.

tests/phpunit/tests/test-columns.php
```diff
@@ -170,11 +183,13 @@ public function test_translated_post( int $user_id ) {
 		$pll_admin->filters_columns->post_column( 'language_fr', $posts['en'] );
 		$column = ob_get_clean();
 
-		if ( $user_id ) {
-			$this->assertNotFalse( strpos( $column, 'pll_icon_edit' ) );
-			$this->assertNotFalse( strpos( $column, 'href="' ) );
+		$this->assertNotFalse( strpos( $column, 'pll_icon_edit' ) );
+
+		if ( empty( $user ) || 'translator_fr' === $user ) {
+			$this->assertNotFalse( strpos( $column, 'href="' ), 'A non-translator or a French translator should be able to edit a translation.' );
```
Hug0-Drelon: And again 😅

tests/phpunit/tests/test-columns.php
```diff
@@ -193,74 +208,77 @@ public function test_term_with_no_language() {
...
 
-		if ( $user_id ) {
-			$column = $pll_admin->filters_columns->term_column( '', 'language_fr', $en );
-			$this->assertNotFalse( strpos( $column, 'pll_icon_add' ) );
-			$this->assertSame( 0, strpos( $column, '<a href="' ) );
+		$this->assertNotFalse( strpos( $column, 'pll_icon_add' ) );
+
+		if ( empty( $user ) || 'translator_fr' === $user ) {
+			$this->assertNotFalse( strpos( $column, 'href="' ), 'A non-translator or a French translator should be able to create a translation in French.' );
+			$this->assertNotFalse( strpos( $column, 'from_tag' ), 'The URL allowing to create a translation should contain the `from_tag` query arg.' );
```
Hug0-Drelon: And.... again, same suggestion.

admin/admin-filters-columns.php
Screenfeed: Unless I'm missing something, the behavior doesn't change.

Previously:

```php
if ( isset( $_REQUEST['post_type'] ) ) {
	$post_type = sanitize_key( $_REQUEST['post_type'] );
}

if ( isset( $GLOBALS['post_type'] ) ) {
	$post_type = $GLOBALS['post_type'];
}
```

If both are set, `$_REQUEST['post_type']` will be used, then `$GLOBALS['post_type']` will overwrite it.

Now:

```php
if ( isset( $GLOBALS['post_type'] ) ) {
	$post_type = $GLOBALS['post_type'];
} elseif ( isset( $_REQUEST['post_type'] ) ) {
	$post_type = sanitize_key( $_REQUEST['post_type'] );
}
```

If both are set, only `$GLOBALS['post_type']` will be used.

admin/admin-filters-columns.php
Screenfeed: I wondered the same actually. So far I kept it here because it used only here, and it would require to move `get_flag_html()` too (which is used in `add_column()` too) 🤔

tests/phpunit/tests/test-columns.php
Screenfeed: Indeed but then it will make the following a bit less readable IMHO:

```php
wp_set_current_user( self::$$user->ID );
```

What do you think? Worth it?

admin/admin-filters-columns.php
Screenfeed: We can't use `reset()`: `Only variables should be passed by reference`. Any other idea?
Why `PLL_Model::get_default_language()`? Are you saying it is supposed to be the 1st of the list? (it's not)

admin/admin-filters-columns.php
Hug0-Drelon: Okay, I missed the fact that you've changed `$GLOBALS` and `$_REQUEST` order.
To be honest I don't see the purpose of the `elseif`. Why did you add that in the first place?

tests/phpunit/tests/test-columns.php
Hug0-Drelon: I don't find `wp_set_current_user( self::$$user->ID );` less readable than `wp_set_current_user( self::$$user );`. Moreover, I think that if something  hurts readability here it's the named variable `self::$$user`.

admin/admin-filters-columns.php
Hug0-Drelon: It'd be cleaner to do so IMHO.

admin/admin-filters-columns.php
Hug0-Drelon: Okay, it needs term order here, got it.
But you can still use `reset()`, just pass a variable to it.
I don't know what you think about it but using a `foreach` to get the first array element is ugly AF.

admin/admin-filters-columns.php
Screenfeed: Because `$post_type` can't get both values (obviously), it's the perfect example about when `elseif` should be used. The question should be "Why not using `elseif` in the first place?".

Hug0-Drelon: > "Why not using elseif in the first place?"

Specific remarks:
- Having only `if` doesn't prevent to have only one value (variable reassignment).
- I don't see why we shouldn't override `$GLOBALS['post_type']` if `$_REQUEST['post_type']` is set.

General remarks:
- `elseif` usually hurst readability
- `elseif` usage usually means that a dedicated function (or method) is needed
- `elseif` makes test coverage harder (more test to reach each branches)

Screenfeed: >     * Having only `if` doesn't prevent to have only one value (variable reassignment).

It does. How do you pass in both `elseif` cases? That's precisely why it should be used instead of a a double `if`: it does prevent double assignment.

>     * I don't see why we shouldn't override `$GLOBALS['post_type']` if `$_REQUEST['post_type']` is set.

Because the first one is useless since the value will be overwritten by the second assignment. It makes us call `sanitize_key()` for nothing.

>     * `elseif` usually hurst readability

If it's a bunch of `elseif` without empty lines, yes. In this case, no, IMHO.
 
>     * `elseif` usage usually means that a dedicated function (or method) is needed

Then the previous double `if` suffers the same issue but it is hidden.

>     * `elseif` makes test coverage harder (more test to reach each branches)

Then that means the coverage was incomplete: if it reaches the first `if` while it shouldn't, then we miss a test because we don't actually use the value from the first `if`.

Hug0-Drelon: I disagree for the most part but as you see fit. In the end it's only to `if/esleif`. Your answer shows the need for dedicated methods IMHO.

admin/admin-filters-columns.php
Screenfeed: I had to revert it in [852c234](https://github.com/polylang/polylang/pull/1754/commits/852c234e571ac3e9c19f68a3b118e8edd302cc8d) because it was making PHPCS to ignore the comment :/

admin/admin-filters-columns.php
Screenfeed: Reverted in [852c234](https://github.com/polylang/polylang/pull/1754/commits/852c234e571ac3e9c19f68a3b118e8edd302cc8d) :(

admin/admin-filters-columns.php
Hug0-Drelon: Fuck... I must say I'm not a fan of `phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped` without justification.
Do you think it worth adding the comment above instead?

admin/admin-filters-columns.php
Screenfeed: I'm not against using `foreach` in such case personally.
Done in [bf4170c](https://github.com/polylang/polylang/pull/1754/commits/bf4170c477a43aece24d400a43d0803f55557540).

tests/phpunit/tests/test-columns.php
Screenfeed: Done in [4cc6831](https://github.com/polylang/polylang/pull/1754/commits/4cc68318e04ba80997562662a6bdc7955383c19b).

tests/phpunit/tests/test-columns.php
Screenfeed: Done in [dad4de6](https://github.com/polylang/polylang/pull/1754/commits/dad4de6099910ec6ff398d18b71f9be36ed5929e).

admin/admin-filters-columns.php
Screenfeed: Done in [dda6117](https://github.com/polylang/polylang/pull/1754/commits/dda6117d7081df569fc2af6f1b721b930d37ca86).

admin/admin-filters-columns.php
Screenfeed: That's a lot of comments. I suppose we have no other choices that going back to a totally useless `(int) $post_id` 🙄

admin/admin-filters-columns.php
Screenfeed: Oh, this comment was misplaced IMHO 😅

admin/admin-filters-columns.php
Screenfeed: Done in [2d1151a](https://github.com/polylang/polylang/pull/1754/commits/2d1151a338fa555316cf01a6aa5d3629e55a7d97).

admin/admin-filters-columns.php
```diff
@@ -454,7 +413,7 @@ public function ajax_update_term_rows() {
 	 * @param PLL_Language $language PLL_Language object.
 	 * @return string
 	 */
-	protected function get_flag_html( $language ) {
+	protected function get_flag_html( PLL_Language $language ): string {
```
Hug0-Drelon: As discussed, we should internalize this method in `PLL_Language`. Some rework might be needed because `PLL_Language::get_flag_html()` already exists but doesn't serve the same prupose.
Let's address is in a separate PR.

admin/admin-filters-columns.php
```diff
@@ -153,91 +160,63 @@ public function post_column( $column, $post_id ) {
...
+
+		if ( ! $tr_post instanceof WP_Post ) {
+			// Link to add a new translation: no translation for this language yet, or it doesn't exist anymore.
+			echo $this->links->new_post_translation_link( $post_id, $language ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			return;
 		}
+
+		// Link to edit (or not) the post or a translation.
+		$url = $this->links->get_edit_post_translation_link( $tr_post->ID, $language );
+		echo $this->links->get_item_edition_link( $url, $tr_post->ID, $tr_post->post_title, $language, $tr_post->ID === $post_id ? 'flag' : 'icon' ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Hug0-Drelon: Sorry, I must have badly explain my thought. I suggested to have only one method called here, such as `new_post_translation_link()`.

admin/admin-filters-columns.php
```diff
@@ -266,101 +245,80 @@ public function add_term_column( $columns ) {
...
-			 * @since 3.7
-			 *
-			 * @param string $output  First language column output.
-			 * @param int    $term_id Term ID.
-			 * @param string $lang    Language code.
-			 */
-			$out = apply_filters( 'pll_first_language_term_column', $out, $term_id, $lang->slug );
+		// Link to edit (or not) the term or a translation.
+		$url  = $this->links->get_edit_term_translation_link( $tr_term->term_id, $taxonomy, $post_type, $language );
+		$out .= $this->links->get_item_edition_link( $url, $tr_term->term_id, $tr_term->name, $language, $tr_term->term_id === $term_id ? 'flag' : 'icon' );
```
Hug0-Drelon: Same as above, I poorly described my suggestion but I was thinking of a single method.

admin/admin-filters-columns.php
Screenfeed: Done in [ce8ea4c](https://github.com/polylang/polylang/pull/1754/commits/ce8ea4ce0a4f57fc3ce2d1de3fffdd8dbb19d86d).

admin/admin-filters-columns.php
```diff
@@ -92,14 +96,12 @@ protected function add_column( $columns, $before ) {
...
-	protected function get_first_language_column() {
-		foreach ( $this->model->get_languages_list() as $language ) {
-			return 'language_' . $language->slug;
-		}
-
-		return '';
+	protected function get_first_language_column(): string {
+		$languages = $this->model->get_languages_list();
+		$language  = reset( $languages );
+		return ! empty( $language ) ? "language_{$language->slug}" : '';
```
Chouby, line 104: I don't see the benefit. In fact, I expect it to be slightly slower (2 operations instead of 1) even if the impact is negligible for such case.

admin/admin-filters-columns.php
```diff
@@ -126,24 +128,29 @@ public function hidden_columns( $hidden ) {
...
 	 *
 	 * @param string $column  Column name.
 	 * @param int    $post_id Post ID.
 	 * @return void
 	 */
-	public function post_column( $column, $post_id ) {
-		$inline = wp_doing_ajax() && isset( $_REQUEST['action'], $_POST['inline_lang_choice'] ) && 'inline-save' === $_REQUEST['action']; // phpcs:ignore WordPress.Security.NonceVerification
-		$lang = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->post->get_language( $post_id ); // phpcs:ignore WordPress.Security.NonceVerification
+	public function post_column( $column, $post_id ): void {
+		if ( false === strpos( $column, 'language_' ) ) {
```
Chouby: I believe we used this elsewhere:
```suggestion
		if ( ! str_starts_with( $column, 'language_' ) ) {
```

admin/admin-filters-columns.php
```diff
@@ -126,24 +128,29 @@ public function hidden_columns( $hidden ) {
...
+			return;
+		}
+
+		$post_id = (int) $post_id;
+		$inline  = wp_doing_ajax() && isset( $_REQUEST['action'], $_POST['inline_lang_choice'] ) && 'inline-save' === $_REQUEST['action']; // phpcs:ignore WordPress.Security.NonceVerification
+		$lang    = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->post->get_language( $post_id ); // phpcs:ignore WordPress.Security.NonceVerification
```
Chouby: I propose not to ignore the nonce verification and to test only required inputs (it should be sufficient to target the quick edit save):
```suggestion
		$inline  = isset( $_POST['inline_lang_choice'], $_REQUEST['_inline_edit'] ) && wp_verify_nonce( $_REQUEST['_inline_edit'], 'inlineeditnonce' );
		$lang    = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->post->get_language( $post_id );
```

admin/admin-filters-columns.php
```diff
@@ -153,91 +160,62 @@ public function post_column( $column, $post_id ) {
...
-					esc_html( sprintf( __( 'This item is in %s', 'polylang' ), $language->name ) ),
-					$this->get_flag_html( $language ) // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-				);
-			}
-		}
-		// Link to add a new translation
-		else {
-			echo $this->links->new_post_translation_link( $post_id, $language ); // phpcs:ignore WordPress.Security.EscapeOutput
+		$tr_id   = $this->model->post->get( $post_id, $language );
+		$tr_post = $tr_id ? get_post( $tr_id ) : false;
```
Chouby:
```suggestion
		$tr_post = $tr_id ? get_post( $tr_id ) : null;
```

admin/admin-filters-columns.php
```diff
@@ -153,91 +160,62 @@ public function post_column( $column, $post_id ) {
...
-				esc_html__( 'Language', 'polylang' ),
-				$dropdown->walk( $elements, -1, array( 'name' => 'inline_lang_choice', 'id' => '' ) ) // phpcs:ignore WordPress.Security.EscapeOutput
-			);
+		$elements = $this->model->languages->filter( 'translator' )->get_list();
+		if ( current_filter() == 'bulk_edit_custom_box' ) {
+			array_unshift( $elements, (object) array( 'slug' => -1, 'name' => __( '&mdash; No Change &mdash;', 'polylang' ) ) );
 		}
+
+		$dropdown = new PLL_Walker_Dropdown();
+		// The hidden field 'old_lang' allows to pass the old language to ajax request
```
Chouby:
```suggestion
```
This comment seems obsolete.

admin/admin-filters-columns.php
```diff
@@ -266,101 +244,79 @@ public function add_term_column( $columns ) {
...
-		if ( false === strpos( $column, 'language_' ) || ! ( $lang = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		if ( false === strpos( $column, 'language_' ) ) {
 			return $out;
 		}
 
-		if ( isset( $_REQUEST['post_type'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$post_type = sanitize_key( $_REQUEST['post_type'] ); // phpcs:ignore WordPress.Security.NonceVerification
+		$term_id = (int) $term_id;
+		$inline  = wp_doing_ajax() && isset( $_REQUEST['action'], $_POST['inline_lang_choice'] ) && 'inline-save-tax' === $_REQUEST['action']; // phpcs:ignore WordPress.Security.NonceVerification
+		$lang    = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id ); // phpcs:ignore WordPress.Security.NonceVerification
```
Chouby: Same as post quick edit:
```suggestion
		$inline  = isset( $_POST['inline_lang_choice'], $_REQUEST['_inline_edit'] ) && wp_verify_nonce( $_REQUEST['_inline_edit'], 'taxinlineeditnonce' );
		$lang    = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id );

```

admin/admin-filters-columns.php
```diff
@@ -266,101 +244,79 @@ public function add_term_column( $columns ) {
 	 * @return string
 	 */
 	public function term_column( $out, $column, $term_id ) {
-		$inline = wp_doing_ajax() && isset( $_REQUEST['action'], $_POST['inline_lang_choice'] ) && 'inline-save-tax' === $_REQUEST['action']; // phpcs:ignore WordPress.Security.NonceVerification
-		if ( false === strpos( $column, 'language_' ) || ! ( $lang = $inline ? $this->model->get_language( sanitize_key( $_POST['inline_lang_choice'] ) ) : $this->model->term->get_language( $term_id ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+		if ( false === strpos( $column, 'language_' ) ) {
```
Chouby:
```suggestion
		if ( ! str_starts_with( $column, 'language_' ) ) {
```

admin/admin-filters-columns.php
```diff
@@ -266,101 +244,79 @@ public function add_term_column( $columns ) {
...
-				$out .= sprintf(
-					'<span class="pll_column_flag"><span class="screen-reader-text">%1$s</span>%2$s</span>',
-					/* translators: accessibility text, %s is a native language name */
-					esc_html( sprintf( __( 'This item is in %s', 'polylang' ), $language->name ) ),
-					$this->get_flag_html( $language )
-				);
-			}
-		}
+		$tr_id   = $this->model->term->get( $term_id, $language );
+		$tr_term = $tr_id ? get_term( $tr_id, $taxonomy ) : false;
```
Chouby:
```suggestion
		$tr_term = $tr_id ? get_term( $tr_id, $taxonomy ) : null;
```

admin/admin-links.php
```diff
@@ -286,6 +286,106 @@ public function edit_term_translation_link( int $term_id, string $taxonomy, stri
...
+	 *
+	 * @param WP_Post      $post     The post.
+	 * @param PLL_Language $language The language of the post.
+	 * @param string       $mode     Optional. How the link should be displayed: with a pen icon or a language's flag.
+	 *                               Possible values are `icon` and `flag`. Default is `icon`.
+	 * @return string
+	 *
+	 * @phpstan-param 'icon'|'flag' $mode
+	 */
+	public function get_edit_post_item_link_tag( WP_Post $post, PLL_Language $language, string $mode = 'icon' ): string {
```
Chouby: I'm not against moving all the code related to links in this class but I would do it differently. It would make sense to reuse the existing method `edit_post_translation_link()` for translations and introduce a new method for the current post.

Notes:
1. Pass the language as a param is useless as it comes with the post.
2. The introduction of a `$mode` param as you did could allow you to differentiate the output as of `edit_post_translation_link()`.

admin/admin-links.php
```diff
@@ -286,6 +286,106 @@ public function edit_term_translation_link( int $term_id, string $taxonomy, stri
...
+	 * @param int          $item_id   ID of the item.
+	 * @param string       $item_name Name of the item.
+	 * @param PLL_Language $language  Language of the item.
+	 * @param string       $mode      Optional. How the link should be displayed: with a pen icon or a language's flag.
+	 *                                Possible values are `icon` and `flag`. Default is `icon`.
+	 * @return string
+	 *
+	 * @phpstan-param 'icon'|'flag' $mode
+	 */
+	public function get_edit_item_link( string $url, int $item_id, string $item_name, PLL_Language $language, string $mode = 'icon' ): string {
```
Chouby: At least for translation, this could be merged with `edit_translation_link()` and be a protected if not a private method. Possibly all edit link could be managed in this method if you this allows to share code.

admin/admin-links.php
```diff
@@ -286,6 +286,106 @@ public function edit_term_translation_link( int $term_id, string $taxonomy, stri
...
+		if ( 'flag' === $mode ) {
+			$flag  = $language->flag ?: sprintf( '<abbr>%s</abbr>', esc_html( $language->slug ) );
```
Chouby: `get_flag_html()` was introduced to avoid to duplicate this code. It would make sense to move this code in this class too, even if it is not a "link" method.

Chouby (changes requested):

admin/admin-filters-columns.php
Screenfeed: Done in [0c88beb](https://github.com/polylang/polylang/pull/1754/commits/0c88bebce5e7e5bc94a9f0f1e043925c6bd38be2).

admin/admin-filters-columns.php
Screenfeed: Done in [0c88beb](https://github.com/polylang/polylang/pull/1754/commits/0c88bebce5e7e5bc94a9f0f1e043925c6bd38be2).

admin/admin-filters-columns.php
Screenfeed: Done in [27b1aac](https://github.com/polylang/polylang/pull/1754/commits/27b1aac14cd4717e16afb4b6d4b597c9c37beeb0).

admin/admin-filters-columns.php
Screenfeed: Done in [27b1aac](https://github.com/polylang/polylang/pull/1754/commits/27b1aac14cd4717e16afb4b6d4b597c9c37beeb0).

admin/admin-filters-columns.php
Screenfeed: Probably yes. Removed in [0017e53](https://github.com/polylang/polylang/pull/1754/commits/0017e53805254837a32d9d3dbb3e2e2fbae7466a).

admin/admin-links.php
Screenfeed: Fair enough. Done in [3e86f8c](https://github.com/polylang/polylang/pull/1754/commits/3e86f8c96afa3292745beeeb2b85e6e1ba4acd20).

admin/admin-filters-columns.php
Screenfeed: In theory that's not the role of the nonce to check the "intension" (action) but OK, in practice it works because it has a custom name. Done in [70e5c37](https://github.com/polylang/polylang/pull/1754/commits/70e5c37cffceaa32f6c620510d83576a16ac2311).

admin/admin-filters-columns.php
Screenfeed: Done in [70e5c37](https://github.com/polylang/polylang/pull/1754/commits/70e5c37cffceaa32f6c620510d83576a16ac2311).

admin/admin-links.php
Screenfeed: Done in [8ba4e30](https://github.com/polylang/polylang/pull/1754/commits/8ba4e30c4aad2af7534a0a1063e7ead837dd2360).

admin/admin-links.php
Screenfeed: Done in [8ba4e30](https://github.com/polylang/polylang/pull/1754/commits/8ba4e30c4aad2af7534a0a1063e7ead837dd2360).

admin/admin-links.php
```diff
@@ -33,13 +33,13 @@ public function __construct( PLL_Base &$polylang ) {
 	/**
 	 * Returns the html markup for a new post translation link.
 	 *
-	 * @since 1.8
+	 * @since 3.8
```
Chouby, line 36: If we keep the same signature as the old method, I suggest:
```suggestion
	 * @since 1.8
	 * @since 3.8 Renamed from `new_post_translation_link()`.
```
Or does it make sense to use `WP_Post` instead of `int` to match `get_edit_post_link_html()`?

admin/admin-links.php
```diff
@@ -157,49 +157,90 @@ protected function new_translation_link( string $link, PLL_Language $language ):
...
+	 * @param string       $item_name Name of the item. Not used in `metabox` mode.
 	 * @return string
+	 *
+	 * @phpstan-param 'metabox'|'icon'|'flag' $mode
 	 */
-	protected function edit_translation_link( string $link, PLL_Language $language ): string {
-		if ( empty( $link ) ) {
-			return sprintf(
-				'<span title="%s" class="pll_icon_edit wp-ui-text-icon"></span>',
+	private function edit_translation_link( string $url, PLL_Language $language, string $mode, int $item_id, string $item_name ): string {
```
Chouby: I suggest to rename to `get_edit_item_link_html()`.
I would move `$mode` at the beginning or the end because it's very different from the 4 other params which are all related to the item.
I'd like other names for modes.

admin/admin-links.php
```diff
@@ -214,7 +255,7 @@ protected function edit_translation_link( string $link, PLL_Language $language )
 	 * @param PLL_Language $language  The language of the new translation.
 	 * @return string
 	 */
-	public function new_term_translation_link( int $term_id, string $taxonomy, string $post_type, PLL_Language $language ): string {
+	public function get_new_term_link_html( int $term_id, string $taxonomy, string $post_type, PLL_Language $language ): string {
```
Chouby: If we keep like this, could you add a changelog "renamed from new_term_translation_link"?
Or does it make sense to replace the term ID + the taxonomy by a `WP_Term` as done for `get_edit_term_link_html()`?

phpstan.neon.dist
```diff
@@ -78,3 +78,9 @@ parameters:
 			identifier: return.type
 			count: 1
 			path: frontend/frontend-auto-translate.php
+
+		# Ignored because PHPStan is drunk ¯\(°_o)/¯
+		-
+			identifier: deadCode.unreachable
```
Chouby, line 84: Where is it?

admin/admin-links.php
```diff
@@ -157,49 +157,90 @@ protected function new_translation_link( string $link, PLL_Language $language ):
...
-	 * @param PLL_Language $language The language of the translation.
+	 * @param string       $url       URL of the edition link.
+	 * @param PLL_Language $language  Language of the item.
+	 * @param string       $mode      How the link should be displayed: with a pen icon or a language's flag.
+	 *                                Possible values are `metabox`, `icon`, and `flag`.
+	 * @param int          $item_id   ID of the item. Used only in `icon` mode.
+	 * @param string       $item_name Name of the item. Not used in `metabox` mode.
 	 * @return string
+	 *
+	 * @phpstan-param 'metabox'|'icon'|'flag' $mode
```
Chouby: See other suggestions to name these params in https://wpsyntex.slack.com/archives/C017SSTK4KY/p1763136243371199

admin/view-translations-media.php
```diff
@@ -25,17 +25,25 @@
...
+						echo $this->links->get_edit_post_link_html( $translation ); // phpcs:ignore WordPress.Security.EscapeOutput
+					} else {
+						// The translation doesn't exist anymore.
+						echo $this->links->get_new_post_link_html( $post_ID, $language ); // phpcs:ignore WordPress.Security.EscapeOutput
+					}
 				} else {
-					// No translation
-					echo $this->links->new_post_translation_link( $post_ID, $language ); // phpcs:ignore WordPress.Security.EscapeOutput
+					// No translation.
+					echo $this->links->get_new_post_link_html( $post_ID, $language ); // phpcs:ignore WordPress.Security.EscapeOutput
```
Chouby: To clarify the code, I suggest to move a part of the logic above.
What do you think of:
```php
		if ( $language->term_id === $lang->term_id ) {
			continue;
		}

		$translation_id = $this->model->post->get_translation( $post_ID, $language );

		if ( $translation_id && $translation_id !== $post_ID ) {
			$translation = get_post( $translation_id );	
		}
		?>
		<tr>
			<td class = "pll-media-language-column"><span class = "pll-translation-flag"><?php echo $language->flag; // phpcs:ignore WordPress.Security.EscapeOutput ?></span><?php echo esc_html( $language->name ); ?></td>
			<td class = "pll-media-edit-column">
				<?php
				if ( $translation instanceof WP_Post ) {
					// The translation exists.
					printf(
						'<input type="hidden" name="media_tr_lang[%s]" value="%d" />',
						esc_attr( $language->slug ),
						(int) $translation_id
					);
					echo $this->links->get_edit_post_link_html( $translation ); // phpcs:ignore WordPress.Security.EscapeOutput
				} else {
					// No translation.
					echo $this->links->get_new_post_link_html( $post_ID, $language ); // phpcs:ignore WordPress.Security.EscapeOutput
				}
				?>
			</td>
		</tr>
```

Chouby (changes requested):

phpstan.neon.dist
Screenfeed:
```
 ------ ------------------------------------------------------- 
  Line   admin-filters-columns.php                              
 ------ ------------------------------------------------------- 
  :69    Unreachable statement - code above always terminates.  
         🪪  deadCode.unreachable                               
 ------ -------------------------------------------------------
```

It's the last line of the constructor. The error disappears if I remove the line 69 (`add_action( 'wp_ajax_pll_update_term_rows', ... )`).

admin/admin-links.php
Screenfeed: Method renamed in [ac4e0a0](https://github.com/polylang/polylang/pull/1754/commits/ac4e0a0960d525692407a821ba7d39e9ba1bd6d9). `$mode` moved in [1205f6d](https://github.com/polylang/polylang/pull/1754/commits/1205f6da4803f0ddf411e5e0e9c377c8b75ec731) (now it makes sense to have it last indeed).

admin/view-translations-media.php
Screenfeed: Indeed, good idea. Done in [36d83eb](https://github.com/polylang/polylang/pull/1754/commits/36d83eb4a692ced2e15657ea588714f51cb93e0d).

admin/admin-links.php
Screenfeed: Done in [e1bd061](https://github.com/polylang/polylang/pull/1754/commits/e1bd0613316bd5421e3893aa05c7125f423ce7ca).

admin/admin-links.php
Screenfeed: I chose to use objects instead of IDs, to harmonize the methods. But it required a few changes in places that call these methods.
Done in [deb77f9](https://github.com/polylang/polylang/pull/1754/commits/deb77f9b08d72f555ce14c4ae73be4783e25f889).

admin/admin-links.php
Screenfeed: Use of objects in [deb77f9](https://github.com/polylang/polylang/pull/1754/commits/deb77f9b08d72f555ce14c4ae73be4783e25f889).

PR #1757, opened 2025-10-29
Fix PHPStan errors

Chouby:
- Remove PHPStan error from baseline as a follow-up of 471d3838b1201361828a5f35899731857f83b12c
- Ignore errors with `wp_slash()` not expecting `mixed` as it actually accepts all types although not documented.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -55,13 +55,4 @@ parameters:
 
 
 		# Ignored because the function wp_slash accepts mixed although not documented.
-		-
-			message: "#^Parameter \\#1 \\$value of function wp_slash expects array\\|string, mixed given\\.$#"
-			count: 1
-			path: include/translated-post.php
-
-		# Ignored because the function wp_slash accepts mixed although not documented.
-		-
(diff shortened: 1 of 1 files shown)


phpstan.neon.dist
```diff
@@ -53,3 +53,15 @@ parameters:
...
+		-
+			message: "#^Parameter \\#1 \\$value of function wp_slash expects array\\|string, mixed given\\.$#"
+			count: 1
+			path: include/translated-post.php
+
+		# Ignored because the function wp_slash accepts mixed although not documented.
+		-
+			message: "#^Parameter \\#1 \\$value of function wp_slash expects array\\|string, mixed given\\.$#"
+			count: 5
+			path: modules/sync/sync-metas.php
```
Screenfeed: If we go this way, we can ignore it globally IMHO.

```suggestion
		# Ignored because the function wp_slash accepts mixed although not documented.
		- "#^Parameter \\#1 \\$value of function wp_slash expects array\\|string, mixed given\\.$#"
```

Screenfeed (approved):

PR #1758, opened 2025-10-29
Create languages filter proxies for `hide_empty` and `hide_default`.

Hug0-Drelon:
## What?
Continue the ongoing work on `Model\Languages` filters and create proxies for `hide_empty` and `hide_default`.

## How?
- Create `Hide_Empty` and `Hide_Default` proxies implementing `Languages_Proxy_Interface`.
- Register them in `PLL_Model::__construct()`.
- Use them in production code right away.
- Use them in `Test_Model` test class.
- Add depreciation notices in `Model\Languages::get_list()` for `hide_empty` and `hide_default` arguments.
- Adapt `pll_get_languages_list`.
- Add tests to cover `Languages_Proxies`, `Hide_Empty` and `Hide_Default`

### DIFF
--- a/include/Model/Languages.php
+++ b/include/Model/Languages.php
@@ -557,8 +557,8 @@ class Languages {
 	 *
 	 * @param array $args {
 	 *   @type string $fields       Returns only that field if set; {@see PLL_Language} for a list of fields.
-	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`). Deprecated, use the `hide_empty` proxy instead.
-	 *   @type bool   $hide_default Hides default language from the list (default to `false`). Deprecated, use the `hide_default` proxy instead.
+	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`). Deprecated, use the `hide_empty` filter instead.
+	 *   @type bool   $hide_default Hides default language from the list (default to `false`). Deprecated, use the `hide_default` filter instead.
 	 * }
 	 * @return array List of PLL_Language objects or PLL_Language object properties.
 	 */


include/Model/Languages.php
```diff
@@ -556,8 +556,8 @@ public function has(): bool {
 	 * @since 3.7 Moved from `PLL_Model::get_languages_list()` to `WP_Syntex\Polylang\Model\Languages::get_list()`.
 	 *
 	 * @param array $args {
-	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`).
-	 *   @type bool   $hide_default Hides default language from the list (default to `false`).
+	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`). Deprecated, use the `hide_empty` proxy instead.
+	 *   @type bool   $hide_default Hides default language from the list (default to `false`). Deprecated, use the `hide_default` proxy instead.
 	 *   @type string $fields       Returns only that field if set; {@see PLL_Language} for a list of fields.
```
Screenfeed: What do you think about putting `$fields` at the first position? So the deprecated parameters are at the end of the list.

include/Model/Languages.php
```diff
@@ -571,6 +571,14 @@ public function get_list( $args = array() ): array {
 			);
 		}
 
+		if ( ! empty( $args['hide_empty'] ) ) {
+			_deprecated_argument( __METHOD__, '3.8', esc_html__( 'Use the `hide_empty` proxy instead.', 'polylang' ) );
+		}
+
+		if ( ! empty( $args['hide_default'] ) ) {
+			_deprecated_argument( __METHOD__, '3.8', esc_html__( 'Use the `hide_default` proxy instead.', 'polylang' ) );
+		}
```
Screenfeed, line 598: We should use `sprintf()` to move the name of the parameters outside the strings.

To prevent duplicating code we could also use a loop:

```php
foreach ( array( 'hide_empty', 'hide_default' ) as $param ) {
	if ( ! empty( $args[ $param ] ) ) {
		/* translators: %s is the name of a proxy. */
		_deprecated_argument( __METHOD__, '3.8', sprintf( esc_html__( 'Use the `%s` proxy instead.', 'polylang' ), $param ) );
	}
}
```
(I'm not sure if we should put the tick in the string or outside.)

It may be too much but when I started working on this, I did this:

```php
_deprecated_argument(
	__FUNCTION__,
	'3.8',
	sprintf(
		/* translators: %1$s is the name of an argument, %2$s is the code replacing it. */
		esc_html__( 'The argument %1$s has been replaced by %2$s.', 'polylang' ),
		'<code>hide_default</code>',
		'<code>PLL()->model->languages->filter( \'hide_default\' )->get_list()</code>'
	)
);
```

I let you decide.

include/api.php
```diff
@@ -348,7 +348,10 @@ function pll_is_translated_taxonomy( $tax ) {
  */
 function pll_languages_list( $args = array() ) {
 	$args = wp_parse_args( $args, array( 'fields' => 'slug' ) );
-	return PLL()->model->get_languages_list( $args );
+
+	return PLL()->model->languages
+		->filter( ! empty( $args['hide_empty'] ) ? 'hide_empty' : '' )
+		->get_list( $args );
```
Screenfeed, line 358: Even if it's not documented, I think we should also handle `hide_default` here. What do you think?

Also, this will trigger the deprecation notices. What do you think of this?

```php
$languages = PLL()->model->languages;

foreach ( array( 'hide_empty', 'hide_default' ) as $param ) {
	if ( ! empty( $args[ $param ] ) ) {
		$languages = $languages->filter( $param );
		unset( $args[ $param ] );
	}
}

return $languages->get_list( $args );
```

include/Model/Languages.php
Hug0-Drelon: Applied in a218b65bb6b5287b18244de2d266d43f32d2adf2.

include/Model/Languages.php
Hug0-Drelon: Partially applied in a218b65bb6b5287b18244de2d266d43f32d2adf2.
I didn't use `foreach` as it hurts readability IMHO and doesn't prevent logic repetition (only string printing).

include/api.php
Hug0-Drelon: Good idea, applied in a218b65bb6b5287b18244de2d266d43f32d2adf2. But no need for `foreach` as proxies can be chained.

include/Model/Languages.php
```diff
@@ -556,9 +556,9 @@ public function has(): bool {
 	 * @since 3.7 Moved from `PLL_Model::get_languages_list()` to `WP_Syntex\Polylang\Model\Languages::get_list()`.
 	 *
 	 * @param array $args {
-	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`).
-	 *   @type bool   $hide_default Hides default language from the list (default to `false`).
 	 *   @type string $fields       Returns only that field if set; {@see PLL_Language} for a list of fields.
+	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`). Deprecated, use the `hide_empty` proxy instead.
+	 *   @type bool   $hide_default Hides default language from the list (default to `false`). Deprecated, use the `hide_default` proxy instead.
```
Chouby: What do you think to document as filter instead of proxy?
```suggestion
	 *   @type bool   $hide_empty   Hides languages with no posts if set to `true` (defaults to `false`). Deprecated, use the `hide_empty` filter instead.
	 *   @type bool   $hide_default Hides default language from the list (default to `false`). Deprecated, use the `hide_default` filter instead.
```

include/links-directory.php
```diff
@@ -81,12 +81,9 @@ public function add_language_to_link( $url, $language ) {
...
 	public function remove_language_from_link( $url ) {
-		$languages = $this->model->get_languages_list(
-			array(
-				'hide_default' => $this->options['hide_default'],
-				'fields'       => 'slug',
-			)
-		);
+		$languages = $this->model->languages
+			->filter( $this->options['hide_default'] ? 'hide_default' : '' )
+			->get_list( array( 'fields' => 'slug' ) );
```
Chouby, line 86: `array( 'fields' => 'slug' )` is the only remaining param for `get_list()`. What do you think to progressively remove its usage and inline the call to `wp_list_pluck()` when we need it?

PR #1759, opened 2025-10-29
Remove Save button if premium are activated on a constant

sebastienserre:
This PR will hide the save/cancel button when Polylang Pro or Polylang for WooCommerce will be activated through a PHP constant
Needed by https://github.com/polylang/updater/pull/20

### DIFF
--- a/settings/settings-module.php
+++ b/settings/settings-module.php
@@ -82,7 +82,7 @@ class PLL_Settings_Module {
 	/**
 	 * Stores html fragment for the buttons.
 	 *
-	 * @var string
+	 * @var array
 	 */
 	protected $buttons;
 


settings/settings-licenses.php
```diff
@@ -144,4 +142,19 @@ public function deactivate_license() {
 			)
 		);
 	}
+
+	/**
+	 * Get the buttons.
```
Hug0-Drelon, line 147:
```suggestion
	 * Returns the HTML strings for buttons. Empty if the license is set with a constant.
```
Coding standard requires to conjugate verbs in the documentation.

settings/table-settings.php
```diff
@@ -73,17 +73,16 @@ public function single_row( $item ) {
...
+						%2s
+						%3s
+
 					</td>
 				</tr>',
 				esc_attr( $item->module ),
 				esc_html( $item->title ),
 				$form, // phpcs:ignore
-				implode( $item->get_buttons() ) // phpcs:ignore
+				$item->get_buttons()// phpcs:ignore
```
Hug0-Drelon:
```suggestion
				$item->get_buttons()
```
If something must be ignored, it should be specify which rule.

settings/settings-licenses.php
```diff
@@ -144,4 +142,19 @@ public function deactivate_license() {
...
+	 * @since 3.8 Returns a string instead of an array.
+	 *
+	 * @return string The HTML fragment for the buttons.
+	 */
+	public function get_buttons(): string {
+		if ( ! defined( 'POLYLANG_LICENSE_KEY' ) ) {
+			return parent::get_buttons();
+		}
+		return '';
+	}
```
Chouby, line 159: We don't need to make any change to this class as it's there only for Backward compat (if a user updates Polylang before PLLWC).

settings/settings-module.php
```diff
@@ -165,9 +165,14 @@ public function __construct( PLL_Settings &$polylang, $args ) {
...
-			'cancel' => sprintf( '<button type="button" class="button button-secondary cancel">%s</button>', esc_html__( 'Cancel', 'polylang' ) ),
-			'save'   => sprintf( '<button type="button" class="button button-primary save">%s</button>', esc_html__( 'Save Changes', 'polylang' ) ),
+		$this->buttons = sprintf(
+			'<p class="submit inline-edit-save">
+				<button type="button" class="button button-secondary cancel">%1s</button>
+				<button type="button" class="button button-primary save">%2s</button>
+
+			</p>',
+			esc_html__( 'Cancel', 'polylang' ),
+			esc_html__( 'Save Changes', 'polylang' )
```
Chouby: I propose to keep this property unchanged.

settings/settings-module.php
```diff
@@ -366,10 +371,11 @@ public function get_upgrade_message() {
...
 	 *
 	 * @since 1.9
+	 * @since 3.8 Returns a string instead of an array.
 	 *
-	 * @return string[] An array of html fragment for the buttons.
+	 * @return string The HTML fragment for the buttons.
 	 */
-	public function get_buttons() {
+	public function get_buttons(): string {
 		return $this->buttons;
```
Chouby: To have minimal changes, this is the method that we should change returning the full markup with the `submit class` as you propose but base on the array `$this->buttons`.

PR #1760, opened 2025-10-30
Harmonize params in "before_post_translation" hooks

Screenfeed:
The hook `pll_before_post_translations` has a `$post_type` parameter, the hook `pll_before_post_translation_{$lang}` should have one too:
- In PLL Pro, `PLL_Sync_Post_Button::add_icon()` is hooked to `pll_before_post_translation_{$lang}` and expects a post type as first parameter.

### DIFF
--- a/admin/admin-classic-editor.php
+++ b/admin/admin-classic-editor.php
@@ -93,16 +93,12 @@ class PLL_Admin_Classic_Editor {
 	 *
 	 * @since 0.1
 	 *
+	 * @param WP_Post $post Current post object.
 	 * @return void
 	 */
-	public function post_language() {
-		global $post_ID;
-		$post_type = get_post_type( $post_ID );
-
-		if ( empty( $post_type ) ) {
-			return;
-		}
-
+	public function post_language( WP_Post $post ): void {
+		$post_ID       = $post->ID;
+		$post_type     = $post->post_type;
 		$from_post_id  = 0;
 		$lang          = $this->model->post->get_language( $post_ID );
 		$new_post_data = $this->links->get_data_from_new_post_translation_request();


admin/admin-classic-editor.php
```diff
@@ -141,9 +141,11 @@ public function post_language() {
 		);
 
 		/**
-		 * Fires before displaying the list of translations in the Languages metabox for posts
+		 * Fires before displaying the list of translations in the Languages metabox for posts.
 		 *
 		 * @since 1.8
+		 *
+		 * @param string $post_type Post's post type.
```
Hug0-Drelon:
```suggestion
		 * @param string $post_type The post type.
```
Seems more natural.

admin/admin-classic-editor.php
```diff
@@ -97,7 +97,7 @@ public function add_meta_boxes( $post_type ) {
 	 */
 	public function post_language() {
 		global $post_ID;
-		$post_type = get_post_type( $post_ID );
+		$post_type = (string) get_post_type( $post_ID );
```
Hug0-Drelon: I get that you cast for type safety. But if be get empty string (or `false` from `get_post_type()`), maybe we should process anything at all and avoid to call `do_action()`. What do you think?

admin/view-translations-post.php
```diff
@@ -47,8 +49,11 @@
 			 * The dynamic portion of the hook name, `$lang`, refers to the language code.
 			 *
 			 * @since 2.1
+			 * @since 3.8 New `$post_type` parameter.
+			 *
+			 * @param string $post_type Post's post type.
```
Hug0-Drelon:
```suggestion
			 * @param string $post_type The post type.
```
Same suggestion as above.

admin/view-translations-post.php
```diff
@@ -14,6 +14,8 @@
 <p><strong><?php esc_html_e( 'Translations', 'polylang' ); ?></strong></p>
 <table>
 	<?php
+	$post_type = (string) get_post_type( $post_ID );
```
Hug0-Drelon: Same question as above.

admin/admin-classic-editor.php
Screenfeed: We know it will never happen but why not. Done in [b3ca24b](https://github.com/polylang/polylang/pull/1760/commits/b3ca24b638e290c95c68597d15afb547f00c6572).

admin/view-translations-post.php
Screenfeed: Done in [b3ca24b](https://github.com/polylang/polylang/pull/1760/commits/b3ca24b638e290c95c68597d15afb547f00c6572).

admin/view-translations-post.php
```diff
@@ -10,6 +10,12 @@
  */
 
 defined( 'ABSPATH' ) || exit;
+
+$post_type = get_post_type( $post_ID );
+
+if ( empty( $post_type ) ) {
+	return;
+}
```
Chouby: This part seems useless as it's already done in the [calling method](https://github.com/polylang/polylang/pull/1760/files#diff-06762224082eaa1ae3f00d3d8670fcc2d674c155189fb96c4b6d34bcb3fcac86R100-R104):
```suggestion
```

Chouby (changes requested):

admin/view-translations-post.php
Screenfeed: @Chouby It's for PHPStan.
Do you prefer `$post_type = (string) get_post_type( $post_ID );`?

Screenfeed: @Chouby OK I didn't understand what you meant, no I do with our conversation.
Changes done in [f5c02a2](https://github.com/polylang/polylang/pull/1760/commits/f5c02a28438bdbd4a308100ed45ab716ee1f928f).

admin/admin-classic-editor.php
```diff
@@ -99,6 +99,10 @@ public function post_language() {
 		global $post_ID;
 		$post_type = get_post_type( $post_ID );
 
+		if ( empty( $post_type ) ) {
+			return;
+		}
```
Chouby: It's not possible that we get an empty post type here (WP would have stopped before).
What I propose is to modernize a bit the method and use the parameter `$post` sent by WP instead of the global `$post_ID` to access the post information. What do you think?

Screenfeed: Great idea. Done in [84b4c3e](https://github.com/polylang/polylang/pull/1760/commits/84b4c3e2e24e114d4e7289c54d60405504d2e2f6).

PR #1762, opened 2025-11-02
Replace check_admin_referer() by wp_verify_nonce()

Chouby:
This PR replaces calls to `check_admin_referer()` or `check_ajax_referer()` by `wp_verify_nonce()`.

The reasoning is that we should not call `check_admin_referer()` or `check_ajax_referer()` for nonces not belonging to Polylang as they may illegitimately die in conflicting situations (see the conflict with the Vantage theme foxed in v1.5.5). 

In such situation, `wp_verify_nonce()` is better as, if the nonce is not verified, we just don't do anything.

Bonus: Introduce `PLL_Admin_Filters_Term::maybe_set_language()` to avoid duplicate code.

### DIFF
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -272,6 +272,8 @@ class PLL_Admin_Filters_Term {
 	 * @return void|never
 	 */
 	protected function save_language( $term_id, $taxonomy ) {
+		$term_id = (int) $term_id;
+
 		/*
 		 * Category metabox.
 		 */
@@ -337,7 +339,7 @@ class PLL_Admin_Filters_Term {
 		 * Quick edit.
 		 */
 		if ( isset( $_POST['inline_lang_choice'], $_REQUEST['_inline_edit'] ) ) {
-			if ( ! wp_verify_nonce( $_REQUEST['_inline_edit'], 'inlineeditnonce' ) && ! wp_verify_nonce( $_REQUEST['_inline_edit'], 'taxinlineeditnonce' ) ) { // Post quick edit or tag quick edit ?
+			if ( ! wp_verify_nonce( $_REQUEST['_inline_edit'], 'inlineeditnonce' ) && ! wp_verify_nonce( $_REQUEST['_inline_edit'], 'taxinlineeditnonce' ) ) { // Post quick edit or tag quick edit?
 				return;
 			}
 
@@ -348,7 +350,7 @@ class PLL_Admin_Filters_Term {
 		/*
 		 * Edit post.
 		 */
-		if ( isset( $_POST['post_lang_choice'] ) ) { // FIXME should be useless now
+		if ( isset( $_POST['post_lang_choice'] ) ) { // FIXME should be useless now.
 			check_admin_referer( 'pll_language', '_pll_nonce' );
 			$this->maybe_set_language( $term_id, sanitize_key( $_POST['post_lang_choice'] ) );
 		}


admin/admin-filters-post.php
```diff
@@ -124,106 +124,134 @@ public function parse_query( $query ) {
...
 
-			if ( ! current_user_can( $post_type_object->cap->edit_post, $post_id ) ) {
-				return;
-			}
+		if ( empty( $post_type_object ) ) {
+			return;
+		}
 
-			$language = $this->model->get_language( sanitize_key( $_POST['post_lang_choice'] ) );
+		if ( ! current_user_can( $post_type_object->cap->edit_post, $post_id ) ) {
```
Hug0-Drelon: You could create a `$user` object right here and do `$user->has_cap( $post_type_object->cap->edit_post, $post_id )`, then re-user the same object around line 163 with `( new User() )->can_translate_or_die( $language );`.
It avoids too many computing and a second `WP_User` instance.

admin/admin-filters-post.php
```diff
@@ -124,106 +124,134 @@ public function parse_query( $query ) {
...
 
-				$post_ids = array_map( 'intval', (array) $_REQUEST['post'] );
-				foreach ( $post_ids as $post_id ) {
-					if ( current_user_can( 'edit_post', $post_id ) ) {
-						$this->model->post->set_language( $post_id, $lang );
-					}
-				}
+		if ( -1 === $_GET['inline_lang_choice'] ) {
+			return;
+		}
```
Hug0-Drelon, line 193: Is it necessary? It would pass `if ( empty( $language ) )` anyway?

admin/admin-filters-post.php
```diff
@@ -124,106 +124,134 @@ public function parse_query( $query ) {
...
+
+		if ( empty( $language ) ) {
+			return;
+		}
+
+		( new User() )->can_translate_or_die( $language );
+
+		$post_ids = array_map( 'intval', (array) $_REQUEST['post'] );
+		foreach ( $post_ids as $post_id ) {
+			if ( current_user_can( 'edit_post', $post_id ) ) {
```
Hug0-Drelon: Same suggestion as in `admin/admin-filters-post.php`, you could user a single `WP_User` instance.

admin/admin-filters-post.php
```diff
@@ -124,106 +124,134 @@ public function parse_query( $query ) {
...
+		if ( empty( $language ) ) {
+			return;
 		}
+
+		( new User() )->can_translate_or_die( $language );
+
+
+		$post_id = (int) $_POST['post_ID'];
+
+		if ( ! $post_id || ! current_user_can( 'edit_post', $post_id ) ) {
```
Hug0-Drelon: Same suggestion :) One `WP_User` object to rule them all!

admin/admin-filters-term.php
```diff
@@ -342,53 +341,55 @@ protected function save_language( $term_id, $taxonomy ) {
...
+	/**
+	 * Sets the language of a term if the user is allowed to.
+	 *
+	 * @since 3.8
+	 *
+	 * @param int    $term_id  The term ID.
+	 * @param string $language Language slug.
+	 * @return void|never
+	 */
+	private function maybe_set_language( $term_id, $language ) {
```
Hug0-Drelon:
```suggestion
	private function maybe_set_language( int $term_id, string $language ): void {
```
It follows what we usually do for new methods. It'll require casting the arguments in the code above though.

Hug0-Drelon (changes requested):

admin/admin-filters-term.php
```diff
@@ -267,40 +267,37 @@ public function pre_post_update( $post_id ) {
 	 *
 	 * @since 1.5
 	 *
-	 * @param int    $term_id  Term ID.
+	 * @param int    $term_id  The term ID.
 	 * @param string $taxonomy Taxonomy name.
 	 * @return void|never
 	 */
 	protected function save_language( $term_id, $taxonomy ) {
 		global $wpdb;
```
Hug0-Drelon:
```suggestion
		global $wpdb;
		$term_id = (int) $term_id;
```
Since `maybe_set_language` is type hinted.

Hug0-Drelon (changes requested):

PR #1763, opened 2025-11-02
Replace direct DB query by get_terms()

Chouby:
In the past, WP did not provide a function to get all terms with the same name. This is no longer an issue . This PR replaces a direct DB query by a call to `get_terms()`.

NB: this part of the code is already tested by [`test_create_term_from_post_bulk_edit()`](https://github.com/polylang/polylang/blob/3.7.4/tests/phpunit/tests/test-admin-filters-term.php#L124).

### DIFF
--- a/admin/admin-filters-term.php
+++ b/admin/admin-filters-term.php
@@ -315,7 +315,7 @@ class PLL_Admin_Filters_Term {
 
 				// Get all terms with the same name.
 				if ( $term instanceof WP_Term ) {
-					$term_ids = get_terms( array( 'taconomy' => $taxonomy, 'name' => $term->name, 'hide_empty' => false, 'fields' => 'ids' ) );
+					$term_ids = get_terms( array( 'taxonomy' => $taxonomy, 'name' => $term->name, 'hide_empty' => false, 'fields' => 'ids' ) );
 
 					// If we have several terms with the same name, they are translations of each other.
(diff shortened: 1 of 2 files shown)


admin/admin-filters-term.php
```diff
@@ -312,35 +311,22 @@ protected function save_language( $term_id, $taxonomy ) {
...
-						$wpdb->prepare(
-							"SELECT t.term_id FROM $wpdb->terms AS t
-							INNER JOIN $wpdb->term_taxonomy AS tt ON t.term_id = tt.term_id
-							WHERE tt.taxonomy = %s AND t.name = %s",
-							$taxonomy,
-							$term->name
-						)
-					);
-				}
+					$term_ids = get_terms( array( 'taconomy' => $taxonomy, 'name' => $term->name, 'hide_empty' => false, 'fields' => 'ids' ) );
```
Hug0-Drelon:
```suggestion
					$term_ids = get_terms( array( 'taxonomy' => $taxonomy, 'name' => $term->name, 'hide_empty' => false, 'fields' => 'ids' ) );
```
If this doesn't fail in test, it means this is not covered?

szepeviktor: https://github.com/crate-ci/typos/issues/1405#issuecomment-3487757736

Hug0-Drelon: Thanks @szepeviktor!

Hug0-Drelon (approved):

PR #1766, opened 2025-11-10
Drop support for WP < 6.5

Chouby:
This PR is a proposal to drop the support of WP versions older than 6.5 (released in April 2024).
It thus removes the relevant backward compatibility code, including in tests.
As a consequence, files loaded when the PHP version < 7.2 now require PHP 7.0 instead of 5.6.
See https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/


composer.json
Hug0-Drelon, line 1: I thinks the PHP 5.6 should stay as it is. It is required to run this for bootstrap files introduced recently (prevent fatal when activating the plugin on a server with old PHP versions).

include/constant-functions.php
Hug0-Drelon, line 1: This file should not be modified.

install/usable.php
Hug0-Drelon, line 1: This file should not be modified.

phpcs-phpcompatibility-70.xml.dist
Hug0-Drelon, line 1: As per my previous comment, this file should also be untouched.

tests/phpunit/includes/testcase.php
Hug0-Drelon, line 1: Does PLLWC follows Polylang Core on this one?
Also, could it break other projects that test with lower version using our testcase?

composer.json
Chouby: Reverted.

include/constant-functions.php
Chouby: Reverted.

install/usable.php
Chouby: Reverted.

phpcs-phpcompatibility-70.xml.dist
Chouby: Reverted too.

tests/phpunit/includes/testcase.php
Chouby: We indeed need to drop the usage of the class in PLLWC. See https://github.com/polylang/polylang-wc/pull/987

PR #1767, opened 2025-11-12
Run PHPUnit with PHP 8.5

Hug0-Drelon:
## What?
Let's run PHPUnit bleeding edge job with PHP 8.5 and fix depreciation notices.

## How?
- PHP 8.5 workflow added by https://github.com/polylang/.github/pull/5.
- Fix PHP 8.5 depreciations (fix null for array key and `Reflection*::setAccessible()` usage), only found in tests.
- Change Composer config to get latest Behat version (use `"behat/behat": "^3"` and remove `"prefere-stable: true"`) to prevent error with PHP 8.5 and Behat (otherwise Composer uses an old version).
- Use specific version for `wp-coding-standards/wpcs` to ensure we get a recent version (3), and fix new found errors.
- Fix new PHPCS errors: `E_USER_ERROR` deprecated since PHP 8.4, `mixed` type hint

### DIFF
--- a/include/language-deprecated.php
+++ b/include/language-deprecated.php
@@ -82,8 +82,6 @@ abstract class PLL_Language_Deprecated {
 	 *
 	 * @since 3.4
 	 *
-	 * @throws DomainException If the property is not accessible.
-	 *
 	 * @param string $property Property to get.
 	 * @return mixed Required property value.
 	 */
@@ -126,7 +124,7 @@ abstract class PLL_Language_Deprecated {
 		$trace      = debug_backtrace(); // phpcs:ignore PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue.NeedsInspection, WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace
 		$file       = $trace[0]['file'] ?? '';
 		$line       = $trace[0]['line'] ?? 0;
-		throw new DomainException(
+		trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
 			esc_html(
 				sprintf(
 					"Cannot access %s property %s::$%s in %s on line %d.\nError handler",


tests/phpunit/tests/test-settings-list-tables.php
```diff
@@ -13,6 +13,10 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 	}
 
 	public function init( $pagename ) {
+		global $pagenow;
+
+		$pagenow = $pagename; // Set $pagenow so `get_admin_page_parent()` doesn't throw PHP deprecation notice with PHP 8.5.
```
Screenfeed: That's the real value. Does this work in tests? (I can't test myself.)
```suggestion
		$pagenow = 'admin.php'; // Set $pagenow so `get_admin_page_parent()` doesn't throw PHP deprecation notice with PHP 8.5.
```

tests/phpunit/tests/test-settings.php
```diff
@@ -18,8 +18,12 @@ public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
 	}
 
 	public function set_up() {
+		global $pagenow;
+
 		parent::set_up();
 
+		$pagenow = 'settings_page_mlang'; // Set $pagenow so `get_admin_page_parent()` doesn't throw PHP deprecation notice with PHP 8.5.
```
Screenfeed: Same as my previous comment.
```suggestion
		$pagenow = 'admin.php'; // Set $pagenow so `get_admin_page_parent()` doesn't throw PHP deprecation notice with PHP 8.5.
```

tests/phpunit/tests/test-settings-list-tables.php
Hug0-Drelon: I doesn't matter what value is stored, it's only to prevent `null` later in core code. But you're on point, we should use the correct value anyway.

include/language-deprecated.php
```diff
@@ -124,7 +126,7 @@ public function __get( $property ) {
 		$trace      = debug_backtrace(); // phpcs:ignore PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue.NeedsInspection, WordPress.PHP.DevelopmentFunctions.error_log_debug_backtrace
 		$file       = $trace[0]['file'] ?? '';
 		$line       = $trace[0]['line'] ?? 0;
-		trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+		throw new DomainException(
```
Chouby: In my opinion, `E_USER_ERROR` was a mistake. If we want to follow the usual PHP behavior, we should trigger an error of type `E_USER_NOTICE` in PHP 7, `E_USER_WARNING` in PHP 8. We shouldn't stop the execution.

include/model.php
```diff
@@ -167,14 +169,13 @@ public function __call( string $name, array $arguments ) {
 		}
 
 		$debug = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions
-		trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions
+		throw new BadMethodCallException(
 			sprintf(
 				'Call to undefined function PLL()->model->%1$s() in %2$s on line %3$s' . "\nError handler",
```
Chouby:
```suggestion
				'Call to undefined method PLL()->model->%1$s() in %2$s on line %3$s' . "\nError handler",
```

modules/sync/admin-sync.php
```diff
@@ -234,14 +236,13 @@ public function __call( $func, $args ) {
 		}
 
 		$debug = debug_backtrace( DEBUG_BACKTRACE_IGNORE_ARGS ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions
-		trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions
+		throw new BadMethodCallException(
 			sprintf(
 				'Call to undefined function PLL()->sync->%1$s() in %2$s on line %3$s' . "\nError handler",
```
Chouby:
```suggestion
				'Call to undefined method PLL()->sync->%1$s() in %2$s on line %3$s' . "\nError handler",
```

Chouby (changes requested):

include/language-deprecated.php
Hug0-Drelon: It makes sense, applied in 67ce3907
Though I didn't find any ressources about `E_USER_NOTICE` in PHP 7 and `E_USER_WARNING` in PHP 8.
According to the [doc](https://www.php.net/manual/en/function.trigger-error.php), the default value is always `E_USER_NOTICE` so I stuck with that.

include/language-deprecated.php
```diff
@@ -82,6 +82,8 @@ abstract public function get_search_url();
 	 *
 	 * @since 3.4
 	 *
+	 * @throws DomainException If the property is not accessible.
+	 *
```
Chouby:
```suggestion
```

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
Chouby: Is the introduction of this file intentional?

.wp-env.override.json
```diff
@@ -0,0 +1,13 @@
+{
```
Chouby: Same as above

Chouby (changes requested):

.vscode/launch.json
Hug0-Drelon: Oops... Sorry, I must have used the wrong Git install... These files are usually ignored globally in my config. Fixed in d3b99fee.

.wp-env.override.json
Hug0-Drelon: Same as above, fixed in d3b99fee.

include/language-deprecated.php
Hug0-Drelon: Applied in d3b99fee.

PR #1769, opened 2025-11-19
Migrate Behat tests to Playwright.

Hug0-Drelon:
## What?
Get rid of Behat now that Playwright is adopted in all Polylang plugins repositories.
Depends on https://github.com/polylang/e2e-test-utils/pull/4

## Why?
No tests other than those for the browser’s preferred language have been added to Behat. To ease maintenance, let's use the less testing frameworks as possible.

## How?
- Set up e2e tests suite using `@playwright/test`, `@wordpress/e2e-test-utils-playwright` and `@wpsyntex/e2e-test-utils`
- Set up permalink structure (and reset) in custom global setup and teardown configs.
- Write tests for browser preferred languages in `tests/e2e/specs/browser-preferred-language.spec.js`, should be same the specs as in `tests/features/browser-preferred-language.feature`.
- Remove Behat tests and config.
- Activate e2e CI
- Take the opportunity to add ESLint CI and run it against e2e tests code.

> [!WARNING]
> b54ef352 must be reverted before merge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "homepage": "https://github.com/polylang/polylang#readme",
   "devDependencies": {
-    "@wpsyntex/e2e-test-utils": "github:polylang/e2e-test-utils#better-lang-creation",
+    "@wpsyntex/e2e-test-utils": "github:polylang/e2e-test-utils",
     "@playwright/test": "^1",
     "@wordpress/e2e-test-utils-playwright": "^1",
     "@wordpress/env": "^10",
--- a/tests/e2e/package.json
+++ b/tests/e2e/package.json
@@ -4,7 +4,7 @@
 	"private": true,
 	"type": "module",
 	"dependencies": {
-		"@wpsyntex/e2e-test-utils": "github:polylang/e2e-test-utils#better-lang-creation",
+		"@wpsyntex/e2e-test-utils": "github:polylang/e2e-test-utils",
 		"@wordpress/e2e-test-utils-playwright": "^1",
 		"@playwright/test": "^1"
 	}


.github/workflows/end-to-end-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+      - '.wp-env.json'
+
+  pull_request:
+    branches: master
+    paths:
+      - '**.php'
+      - '!tests/**.php'
+      - '**.js'
+      - '**.mjs'
+      - '**.cjs'
```
Marianne380, line 28: Why not on push on master ?

.github/workflows/end-to-end-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  testing:
+    name: Playwright Tests
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out the source code
+      uses: actions/checkout@v3
```
Marianne380:
```suggestion
      uses: actions/checkout@v4
```

.github/workflows/static-analysis.yml
```diff
@@ -23,3 +23,28 @@ jobs:
...
+          js:
+            - 'tests/e2e/**/*.{js,mjs,cjs}'
+
+    - name: Run ESLint only on e2e tests files for now...
+      uses: polylang/actions/eslint@main
+      if: ${{ steps.filter.outputs.js == 'true' }}
+      with:
+        do-js: true
+        js-path: './tests/e2e'
+        do-style: false
```
Marianne380, line 50: What's this ?

tests/e2e/specs/browser-preferred-language.spec.js
```diff
@@ -0,0 +1,213 @@
...
+
+		await setAcceptLanguageHeader( context, 'en-GB;q=1.0' );
+
+		await page.goto( '/' );
+
+		await expect( page ).toHaveURL( /\/en-gb\/?$/ );
+
+		await expectLanguageCookie( page, 'en-gb' );
+
+		await context.close();
```
Marianne380: Why do we have to do this ?

tests/e2e/specs/browser-preferred-language.spec.js
```diff
@@ -0,0 +1,213 @@
...
+		context,
+	} ) => {
+		await requestUtils.rest( {
+			path: '/pll/v1/languages',
+			method: 'POST',
+			data: {
+				locale: 'zh_HK',
+				slug: 'zh-hk',
+			},
+		} );
```
Marianne380: Why not using `createLanguage` ?

tests/e2e/specs/browser-preferred-language.spec.js
```diff
@@ -0,0 +1,213 @@
...
+	 *     - Clear cookies.
+	 */
+	test.beforeEach( async ( { page } ) => {
+		await page.context().clearCookies();
+	} );
+
+	test.afterAll( async ( { requestUtils } ) => {
+		await resetAllSettings( requestUtils );
+		await deleteAllLanguages( requestUtils );
+		await requestUtils.deleteAllPosts();
```
Marianne380: Is this really necessary? Isn't it already done in `afterEach()`?

Marianne380 (changes requested):

.github/workflows/static-analysis.yml
Hug0-Drelon: You mean `do-style: false`?
Because there is no CSS/SCSS file to test in `tests/e2e`.
This is how our ESLint action works, [see](https://github.com/polylang/actions/blob/8cd1dac6dd7c737bb4e6f7a9e232ec774c1d9c8f/eslint/action.yml#L15-L18).

tests/e2e/specs/browser-preferred-language.spec.js
Hug0-Drelon: During my first iteration I created a new `context` ([see](https://playwright.dev/docs/api/class-browser#browser-new-context)).
I don't think it's mandatory anymore, even though I change the default `context ` in `setAcceptLanguageHeader`, I do not create a new one.

.github/workflows/end-to-end-tests.yml
Hug0-Drelon: It's a mistake 🙈 
Fixed in 9efa382cd216bfb12cfaf10bceeaf3bf56e0c55a.

tests/e2e/specs/browser-preferred-language.spec.js
Hug0-Drelon: Not mandatory indeed, removed in 19da83746ae7d60f6f6526b72e1e83b7599eebed.

tests/e2e/specs/browser-preferred-language.spec.js
Hug0-Drelon: Removed in 4df573c87105b374338c8941695c8ee06ece1e4f.

tests/e2e/specs/browser-preferred-language.spec.js
Hug0-Drelon: My bad, fixed in 2789e1a3128d5322d3d5a899656746cabdae16c9.
Even more sad that I made a PR exactly to do that in `@wpsyntex/e2e-test-utiles` 😫

.github/workflows/static-analysis.yml
Marianne380: Ok thanks.

PR #1771, opened 2025-11-24
Remove obsolete JS code

Screenfeed:
Partially fixes https://github.com/polylang/polylang-pro/issues/2802.

This PR removes backward compatibility (WP < 6.5) JS code.

### DIFF
--- a/js/src/settings.js
+++ b/js/src/settings.js
@@ -36,9 +36,6 @@ jQuery(
 		 * Common functions and variables for overriding languages and flags dropdown list by a jQuery UI selectmenu widget.
 		 */
 
-		// Add a boolean variable to be able to check jQuery UI >= 1.12 which is introduced in WP 5.6.
-		// Backward compatibility WP < 5.6
-		var isJqueryUImin112 = $.ui.version >= '1.12.0';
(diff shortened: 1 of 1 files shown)


js/src/settings.js
```diff
@@ -48,9 +45,8 @@ jQuery(
 		var wizardSelectmenuWidth = '100%';
 
 		// Inject flag image when jQuery UI selectmenu is created or an item is selected.
-		// jQuery UI 1.12 introduce a wrapper inside de li tag which is necessary to selectmenu widget to work correctly.
+		// jQuery UI 1.12 introduces a wrapper inside de li tag which is necessary to selectmenu widget to work correctly.
```
Hug0-Drelon, line 48: The comment could be deleted since there is no backward compatibility, am I right?

Hug0-Drelon (approved):
LGTM 👍 
Tested language creation/edition with WP 6.8.3, flags and lists seem to work as usual.

js/src/settings.js
Screenfeed: I'm not against removing it if you want but I think we can keep it, it explains a behavior.

PR #1772, opened 2025-11-25
Test the global $wp_customize is set before using it!

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/2820
Follow up of #1762

## Why ?
Before #1762, the global `$wp_customize` was used after a test which more or less guarantees it's correctly set. See https://github.com/polylang/polylang/pull/1762/files#diff-21d6255b92c1bffc2d68a6cd5a459e49b280fc7cb1bddba613804dac0433c498L225-L227
In #1762, it's now used outside this test.

## How ?
Test `$wp_customize` is of the right instance before using it.

## Bonus
Fix All PHStan errors in the file `admin-nav-menu.php`.

### DIFF
--- a/admin/admin-nav-menu.php
+++ b/admin/admin-nav-menu.php
@@ -147,7 +147,7 @@ class PLL_Admin_Nav_Menu extends PLL_Nav_Menu {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


szepeviktor (approved):

sebastienserre:
This PR is fixing the issue I've found yesterday.

PR #1774, opened 2025-11-26
Returns a language object for `Languages::add()` and `Languages::update()`

Chouby:
Closes https://github.com/polylang/polylang-pro/issues/2726

Very often, we need to call `Languages::get()` after `Languages::add()` or `Languages::update()`. So this PR simplifies the usage of these 2 methods by returning directly the object language instead of `true`.

### To be discussed:
The PR includes modifications to return a `WP_Error` when `Languages::get()` returns `false` after the creation or the update.
See https://github.com/polylang/polylang/pull/1774/files#diff-f612298d4b72d331be1c8327d35c40836aa6028ca7e4ef3d18bc0ac0a1603371R241-R243 and https://github.com/polylang/polylang/pull/1774/files#diff-f612298d4b72d331be1c8327d35c40836aa6028ca7e4ef3d18bc0ac0a1603371R241-R243

This pleases PHPStan but may be a dead code as I'm not sure how it could happen.
Previously, we used to document this as not possible, so we might replicate the same here as well.
See for example https://github.com/polylang/polylang/pull/1774/files#diff-8c31521867cfd1b3643b3ef5d9d0a74ae82e7ec401b2da6ffd5b904d84f3a804L192-L193 and https://github.com/polylang/polylang/pull/1774/files#diff-8c31521867cfd1b3643b3ef5d9d0a74ae82e7ec401b2da6ffd5b904d84f3a804L254-L255

### DIFF
--- a/tests/phpunit/includes/factory/unittest-factory-for-language.php
+++ b/tests/phpunit/includes/factory/unittest-factory-for-language.php
@@ -84,16 +84,10 @@ class PLL_UnitTest_Factory_For_Language extends WP_UnitTest_Factory_For_Thing {
 	 * @return int|WP_Error Language term ID on success, `WP_Error` on failure.
 	 */
 	public function create_object( $args ) {
-		$errors = $this->pll_model->add_language( $args );
+		$language = $this->pll_model->languages->add( $args );
 
-		if ( is_wp_error( $errors ) ) {
-			return $errors;
-		}
-
-		$language = $this->pll_model->get_language( $args['slug'] );
-
-		if ( ! $language instanceof PLL_Language ) {
-			return new WP_Error( 'Could not get the created language.' );
+		if ( is_wp_error( $language ) ) {
+			return $language;
 		}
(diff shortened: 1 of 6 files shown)


include/Model/Languages.php
```diff
@@ -233,7 +236,11 @@ public function add( $args ) {
 
 		// Refresh languages.
 		$this->clean_cache();
-		$this->get_list();
+		$new_language = $this->get( $id );
```
Hug0-Drelon, line 239: There is one drawback here.
Calling `Languages::get()` will warm up the cache right after cleaning it.
Usually a good idea, but it can hurt performances if several languages are created all at once (usually happen in tests, not sure in production).
Is it ok for you?

Chouby: True. We add some operations. However `PLL_Cache` is very fast as it's just an array in memory. So it doesn't seem significant compared to the creation of the language terms in DB.

PR #1775, opened 2025-11-26
Introduce `pll_sanitize_id()`

Chouby:
... together with `pll_sanitize_ids()` in replacement of `PLL_Translatable_Object` equivalent methods.

Closes https://github.com/polylang/polylang-pro/issues/2804

- These functions will be easier to use outside the class of the replaced methods.
- `pll_sanitize_id` is added to WPCS config as an unslashing + sanitizing function.
- Uses internal PHP filter instead of a custom algorithm.

### To be discussed 
The proposed filter results in a different behavior in some cases
| Input | Before | This PR |
|-|-|-|
| `true` | 0 | 1 |
| `array()` | 0 | `false` WTF?|
| `10.1` | 10 | 0 |

Alternative 1: We can eliminate the type issue with arrays with:
```php
	$options = array(
		'options' => array(
			'min_range' => 1,
		),
	);
	return filter_var( $id, FILTER_VALIDATE_INT, $options ) ?: 0;
```
Alternative 2: We can eliminate floats including `10.0` with:
```php
	return is_numeric( $id ) && ! is_float( $id ) && $id >= 1 ? (int) $id : 0;
```

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -162,7 +162,7 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
  * @param mixed $id A supposedly numeric ID.
  * @return int A positive integer. `0` for non numeric values and negative integers.
  *
- * @phpstan-return int<0,max>
+ * @phpstan-return positive-int
  */
 function pll_sanitize_id( $id ): int {
 	return is_numeric( $id ) && $id >= 1 ? abs( (int) $id ) : 0;
--- a/src/functions.php
+++ b/src/functions.php
@@ -185,7 +185,7 @@ function pll_sanitize_id( $id ) {
  *
  * @phpstan-return array<positive-int>
  */
-function pll_sanitize_ids_list( $ids ) {
+function pll_sanitize_ids( $ids ) {
 	if ( empty( $ids ) || ! is_array( $ids ) ) {
 		return array();
 	}
--- a/src/functions.php
+++ b/src/functions.php
@@ -152,3 +152,43 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
 
 	return 'GET' === $request->get_method() && 'edit' === $request->get_param( 'context' );
 }
+
+/**
+ * Sanitizes an ID as positive integer.
+ * Kind of similar to `absint()`, but rejects negative integers instead of making them positive.
+ *
+ * @since 3.8
+ *
+ * @param mixed $id A supposedly numeric ID.
+ * @return int A positive integer. `0` for non numeric values and negative integers.
+ *
+ * @phpstan-return int<0,max>
+ */
+function pll_sanitize_id( $id ) {
+	$options = array(
+		'options' => array(
+			'min_range' => 1,
+			'default'   => 0,
+		),
+	);
+	return filter_var( $id, FILTER_VALIDATE_INT, $options );
+}
+
+/**
+ * Sanitizes an array of IDs as positive integers.
+ * `0` values are removed.
+ *
+ * @since 3.8
+ *
+ * @param mixed $ids A supposedly array of numeric IDs.
+ * @return int[]
+ *
+ * @phpstan-return array<positive-int>
+ */
+function pll_sanitize_ids_list( $ids ) {
+	if ( empty( $ids ) || ! is_array( $ids ) ) {
+		return array();
+	}
+
+	return array_filter( array_map( 'pll_sanitize_id', $ids ) );
+}
--- a/src/functions.php
+++ b/src/functions.php
@@ -157,35 +157,29 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
  * Sanitizes an ID as positive integer.
  * Kind of similar to `absint()`, but rejects negative integers instead of making them positive.
  *
- * @since 3.8
+ * @since 3.9
  *
  * @param mixed $id A supposedly numeric ID.
  * @return int A positive integer. `0` for non numeric values and negative integers.
  *
  * @phpstan-return int<0,max>
  */
-function pll_sanitize_id( $id ) {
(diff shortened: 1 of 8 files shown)


include/functions.php
```diff
@@ -154,3 +154,43 @@ function pll_add_notice( WP_Error $error ) {
...
+ * @phpstan-return int<0,max>
+ */
+function pll_sanitize_id( $id ) {
+	$options = array(
+		'options' => array(
+			'min_range' => 1,
+			'default'   => 0,
+		),
+	);
+	return filter_var( $id, FILTER_VALIDATE_INT, $options );
```
Hug0-Drelon:
```suggestion
	return (int) filter_var(
		$id,
		FILTER_VALIDATE_INT,
		array(
			'options' => array(
				'min_range' => 1,
				'default'   => 0,
			),
		)
	);
```
Shorter proposal with a cast to integer.

src/functions.php
```diff
@@ -154,3 +154,43 @@ function pll_add_notice( WP_Error $error ) {
...
+ * @param mixed $ids A supposedly array of numeric IDs.
+ * @return int[]
+ *
+ * @phpstan-return array<positive-int>
+ */
+function pll_sanitize_ids( $ids ) {
+	if ( empty( $ids ) || ! is_array( $ids ) ) {
+		return array();
+	}
+
```
Hug0-Drelon, line 186:
```suggestion
```
Not needed now we chose to cast the returned value of `filter_var()` in `pll_sanitize_id()`.

include/functions.php
```diff
@@ -154,3 +154,43 @@ function pll_add_notice( WP_Error $error ) {
...
+ * Kind of similar to `absint()`, but rejects negative integers instead of making them positive.
+ *
+ * @since 3.8
+ *
+ * @param mixed $id A supposedly numeric ID.
+ * @return int A positive integer. `0` for non numeric values and negative integers.
+ *
+ * @phpstan-return int<0,max>
+ */
+function pll_sanitize_id( $id ) {
```
Hug0-Drelon: Juste a reminder.
We chose to return `0` for boolean input.

Hug0-Drelon (changes requested):

src/functions.php
Chouby: We could avoid `empty()`. But `is_array()` is necessary for `array_mapr)`

include/functions.php
Chouby: Finally I propose to come back to the original method.

src/functions.php
```diff
@@ -152,3 +152,37 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
...
+/**
+ * Sanitizes an ID as positive integer.
+ * Kind of similar to `absint()`, but rejects negative integers instead of making them positive.
+ *
+ * @since 3.9
+ *
+ * @param mixed $id A supposedly numeric ID.
+ * @return int A positive integer. `0` for non numeric values and negative integers.
+ *
+ * @phpstan-return int<0,max>
```
Hug0-Drelon:
```suggestion
 * @phpstan-return positive-int
```
Let's use the same doc as in `pll_sanitize_ids`.

tests/phpunit/tests/test-functions.php
```diff
@@ -0,0 +1,60 @@
...
+		$this->assertSame( 100, pll_sanitize_id( 100.0 ) );
+		$this->assertSame( 100, pll_sanitize_id( 100.1 ) );
+		$this->assertSame( 100, pll_sanitize_id( '100' ) );
+		$this->assertSame( 0, pll_sanitize_id( '-100' ) );
+		$this->assertSame( 100, pll_sanitize_id( '100.0' ) );
+		$this->assertSame( 100, pll_sanitize_id( '100.1' ) );
+		$this->assertSame( 0, pll_sanitize_id( 'true' ) );
+		$this->assertSame( 0, pll_sanitize_id( 'false' ) );
+		$this->assertSame( 0, pll_sanitize_id( true ) );
+		$this->assertSame( 0, pll_sanitize_id( false ) );
```
Hug0-Drelon, line 25:
```suggestion
		$this->assertSame( 0, pll_sanitize_id( false ) );
		$this->assertSame( 0, pll_sanitize_id( fn( $x ) => $x ) );
```
There was a missing built in PHP type 🥹

tests/phpunit/tests/test-functions.php
```diff
@@ -0,0 +1,60 @@
...
+			113.0,
+			114.1,
+			'121',
+			'-122',
+			'123.0',
+			'124.1',
+			'true',
+			'false',
+			true,
+			false,
```
Hug0-Drelon, line 47:
```suggestion
			false,
			fn( $x ) => $x,
```
Following my previous suggestion.

Hug0-Drelon (changes requested):

src/functions.php
```diff
@@ -152,3 +152,37 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
...
+/**
+ * Sanitizes an ID as positive integer.
+ * Kind of similar to `absint()`, but rejects negative integers instead of making them positive.
+ *
+ * @since 3.9
+ *
+ * @param mixed $id A supposedly numeric ID.
+ * @return int A positive integer. `0` for non numeric values and negative integers.
+ *
+ * @phpstan-return positive-int
```
Hug0-Drelon:
```suggestion
 * @phpstan-return int<0, max>
```
My bad, I thought `positive-int` includes `0`...
See https://github.com/polylang/polylang/actions/runs/27937087816/job/82661253489#step:3:288

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

PR #1776, opened 2025-11-27
Improve cache management in  `PLL_Translatable_Object` and `PLL_WP_Import`.

Hug0-Drelon:
## What?
With WordPress 6.9 comes a new cache API (`wp_cache_*_slated`), [see](https://make.wordpress.org/core/2025/11/17/consistent-cache-keys-for-query-groups-in-wordpress-6-9/).

## How?
- Use `wp_cache_*_slated` when applicable (i.e. `PLL_Translatable_Object::query_objects_with_no_lang()`).
- Take the opportunity to use `wp_cache_set_last_changed` in `PLL_Translatable_Object::set_language/set_language_in_mass()` and  (available since WordPress 6.3, now accessible since https://github.com/polylang/polylang/pull/1766).
- Take the opportunity to use `wp_cache_set_terms_last_changed` in `PLL_WP_Import::process_terms` (available since WordPRess 5.0).
- Add tests to ensure the cache is correctly populated when calling `PLL_Translatable_Object::query_objects_with_no_lang()`.
- Add tests to ensure the lasted changed value in object cache is updated when calling `PLL_Translatable_Object::set_language/set_language_in_mass()`.

> [!NOTE]
> `PLL_Translatable_Object::get_from_object_term_cache()` uses `wp_cache_get_multiple` but it doesn't seem necessary to migrate to `wp_cache_get_multiple_salted` because the key is not tied to the last changed value.
> Same in `PLL_Model::count_posts`, `PLL_Widget_Calendar::get_calendar` and `PLL_Frontend_Filters::option_sticky_posts` that use `wp_cache_get/set`.

### DIFF
--- a/tests/phpunit/tests/Cache/test-Object_With_No_Lang.php
+++ b/tests/phpunit/tests/Cache/test-Object_With_No_Lang.php
@@ -72,7 +72,7 @@ class Test_Object_With_No_Lang extends PLL_Object_Cache_TestCase {
 
 		$object_ids_with_no_lang = $this->pll_env->model->$type->get_objects_with_no_lang( $limit );
 
-		$this->assertLessThan( count( $this->object_ids[ $type ] ), count( $object_ids_with_no_lang ),  'The number of objects with no lang should be less than the total number of objects.' );
+		$this->assertLessThanOrEqual( count( $object_ids_with_no_lang ), count( $this->object_ids[ $type ] ),  'The total number of fixtures should be less than or equal to the number of objects with no lang.' );
 
 		foreach ( $this->object_ids [ $type ] as $object_id ) {
 			$this->assertContains( $object_id, $object_ids_with_no_lang, "The object id {$object_id} should be in the list of objects with no language." );


include/translatable-object.php
```diff
@@ -616,6 +614,47 @@ public function get_rest_description(): string {
...
 
+	/**
+	 * Fetches the value from the cache. Handles backward compatibility with WordPress < 6.9.
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $key The cache key.
+	 * @return mixed|false The cached value, false if not found.
+	 */
+	private function get_from_cache( $key ) {
```
Marianne380:
```suggestion
	private function get_from_cache( string $key ) {
```

include/translatable-object.php
```diff
@@ -616,6 +614,47 @@ public function get_rest_description(): string {
...
+	/**
+	 * Stores the value in the cache. Handles backward compatibility with WordPress < 6.9.
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $key   The cache key.
+	 * @param mixed  $value The value to store in the cache.
+	 * @return bool True if the value has been stored, false otherwise.
+	 */
+	private function set_to_cache( $key, $value ): bool {
```
Marianne380:
```suggestion
	private function set_to_cache( string $key, $value ): bool {
```

tests/phpunit/tests/Cache/test-Object_With_No_Lang.php
```diff
@@ -0,0 +1,103 @@
...
+			$this->pll_env->model->languages->get_list()
+		);
+
+		$this->assertInstanceOf( Object_Cache_Annihilator::class, $wp_object_cache, 'The custom object cache should be enabled.' );
+
+		$wp_object_cache->flush();
+
+		$object_ids_with_no_lang = $this->pll_env->model->$type->get_objects_with_no_lang( $limit );
+
+		$this->assertLessThan( $object_ids_with_no_lang, count( $this->object_ids[ $type ] ), 'The number of objects with no lang should be less than the total number of objects.' );
```
Marianne380: Shouldn't it be ?
```suggestion
		$this->assertLessThan( count( $this->object_ids[ $type ] ), count( $object_ids_with_no_lang ),  'The number of objects with no lang should be less than the total number of objects.' );
```

tests/phpunit/tests/Cache/test-Last_Changed.php
```diff
@@ -0,0 +1,83 @@
...
+		$last_changed = wp_cache_get_last_changed( $cache_type );
+
+		$this->assertIsString( $last_changed, 'The last changed time should be a string.' );
+
+		$this->pll_env->model->$type->set_language( $object_id, 'fr' );
+
+		$last_changed_after = wp_cache_get_last_changed( $cache_type );
+
+		$this->assertGreaterThan(
+			number_format( (float) $last_changed, 8, '.', '' ),
```
Marianne380, line 49: Why adding `number_format` ?

include/translatable-object.php
```diff
@@ -616,6 +614,47 @@ public function get_rest_description(): string {
...
+	 * @param string $key The cache key.
+	 * @return mixed|false The cached value, false if not found.
+	 */
+	private function get_from_cache( $key ) {
+		$last_changed = wp_cache_get_last_changed( $this->cache_type );
+
+		if ( ! function_exists( 'wp_cache_get_salted' ) ) {
+			// Backward compatibility with WordPress < 6.9.
+			$cache_key = "{$key}:{$last_changed}";
+			return wp_cache_get( $cache_key, $this->cache_type );
```
Marianne380, line 631: Shouldn't it be ?
```suggestion
			return wp_cache_get( $key, $this->cache_type );
```
In `wp_cache_get_salted()`, WP gets the cache directly with the key, see https://github.com/WordPress/WordPress/blob/6.9/wp-includes/cache-compat.php#L214.

tests/phpunit/tests/Cache/test-Last_Changed.php
Hug0-Drelon: PHP 7.2 doesn't handle large float like newer version (it uses scientific notation...). Using `number_format` normalizes the value before comparison.

tests/phpunit/tests/Cache/test-Object_With_No_Lang.php
Hug0-Drelon: Ouch

include/translatable-object.php
Hug0-Drelon: As discussed, we should kept the same behavior as before.

PR #1777, opened 2025-11-27
Rewrite language terms order in PHP

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/2698

We want to order languages terms by 'language' taxonomy first, then by term group, then by term ID.
We used to hack the SQL query generated by the `WP_Term_Query` to allow this multiple orderby clause.
Although we could not reproduce, using SQL for ordering proved, at least one time, not to be stable.

This PR rewrites the order of the languages terms, using the classic `usort()` function resulting in a simpler code and hopefully a stable sort.

### DIFF
--- a/tests/phpunit/tests/test-create-delete-languages.php
+++ b/tests/phpunit/tests/test-create-delete-languages.php
@@ -183,7 +183,7 @@ class Create_Delete_Languages_Test extends PLL_UnitTestCase {
 		$this->assertTrue( $this->pll_env->model->add_language( $args ) );
 	}
 
-	public function test_default_language_order() {
+	public function test_languages_order() {
 		$args = array(
 			'name'       => 'English',
 			'slug'       => 'en',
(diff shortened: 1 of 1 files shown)


include/Model/Languages.php
```diff
@@ -1309,51 +1309,31 @@ function ( $language ) {
...
+					return $a->term_id < $b->term_id ? -1 : 1;
+				}
+				return $a->term_group < $b->term_group ? -1 : 1;
+			}
 
-		if ( ! preg_match( '@^(?<alias>[^.]+)\.term_group$@', $orderby, $matches ) ) {
-			return $orderby;
-		}
+			return 'language' === $a->taxonomy ? -1 : 1;
+		};
```
Screenfeed: A suggestion to make this a bit simpler:

```php
		$callback = static function ( $a, $b ) {
			if ( $a->taxonomy !== $b->taxonomy ) {
				// Sort terms by 'language' taxonomy first.
				return 'language' === $a->taxonomy ? -1 : 1;
			}
			if ( $a->term_group !== $b->term_group ) {
				// Then by term_group.
				return $a->term_group < $b->term_group ? -1 : 1;
			}
			// Then by term_id.
			return $a->term_id < $b->term_id ? -1 : 1;
		};
```

PR #1779, opened 2025-12-11
Handle backward compatibility in `PLL_UnitTest_Factory_For_Language`.

Hug0-Drelon:
## What?
Fixes warnings in PLLWC CI ([see](https://github.com/polylang/polylang-wc/actions/runs/20112168292/job/57712169534?pr=994#step:5:782)).

The Polylang test suite should be backward compatible.

## Why?
PLLWC CI runs on several Polylang versions, but it uses the Polylang latest test suite (to benefit from new PHPUnit test features). So when tests run with Polylang 3.7.x and the test suite is in 3.8-dev, a warning appears since https://github.com/polylang/polylang/pull/1774.

## How?
- Handle backward compatibility in the test suite and more specifically in `PLL_UnitTest_Factory_For_Language`.

### DIFF
--- a/tests/phpunit/includes/factory/unittest-factory-for-language.php
+++ b/tests/phpunit/includes/factory/unittest-factory-for-language.php
@@ -92,7 +92,7 @@ class PLL_UnitTest_Factory_For_Language extends WP_UnitTest_Factory_For_Thing {
 
 		// Backward compatibility with Polylang < 3.8.
 		if ( ! $language instanceof PLL_Language ) {
-			$language = $this->pll_model->languages->get( $args['locale'] );
+			$language = $this->pll_model->languages->get( $args['slug'] ?? $args['locale'] );
 		}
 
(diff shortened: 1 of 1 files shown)


tests/phpunit/includes/factory/unittest-factory-for-language.php
```diff
@@ -90,7 +90,12 @@ public function create_object( $args ) {
 			return $language;
 		}
 
-		return $language->term_id;
+		// Backward compatibility with Polylang < 3.8.
+		if ( ! $language instanceof PLL_Language ) {
+			$language = $this->pll_model->languages->get( $args['locale'] );
```
Chouby: I propose to use the slug when possible:
```suggestion
			$language = $this->pll_model->languages->get( $args['slug'] ?? $args['locale'] );
```
to handle correctly cases where we would add two languages with the same locale.

Chouby (changes requested):

PR #1780, opened 2025-12-15
Introduce "automatic" language proxies

Screenfeed:
Required by https://github.com/polylang/polylang-pro/pull/2791.

## What this PR does

1. Allow to register language proxies that are always applied, without requiring to use `Languages::filter()`.
2. Introduce `Languages::clean_local_cache()`, which cleans the local `PLL_Cache` instance only (without busting the transient): this is useful when the current user changes and we want to make sure the list of languages is up-to-date for this user.
3. Rework `PLL_Choose_Lang::set_language()` to allow us to share the filter `pll_get_current_language`.
4. Fix `PLL_Sitemaps::set_language_from_query()`: the type of `$lang` was wrong because the filter `pll_set_language_from_query` expects a language object, not a language slug. This fix is in this PR because it is related to sitemaps, which is related to inactive languages in PLL Pro.
5. The change in `PLL_Frontend_Trait::go_to()` is to prevent a "doing it wrong" notice in PLL Pro:
    ```
    Unexpected incorrect usage notice for is_feed.
    Conditional query tags do not work before the query is run. Before then, they always return false. (This message was added in version 3.1.0.)
    ```

### DIFF
--- a/frontend/choose-lang-url.php
+++ b/frontend/choose-lang-url.php
@@ -63,15 +63,10 @@ class PLL_Choose_Lang_Url extends PLL_Choose_Lang {
 		// Take care to post & page preview http://wordpress.org/support/topic/static-frontpage-url-parameter-url-language-information.
 		if ( isset( $_GET['preview'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 			foreach ( array( 'p', 'page_id' ) as $var ) {
-				if ( ! isset( $_GET[ $var ] ) || ! is_numeric( $_GET[ $var ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+				if ( empty( $_GET[ $var ] ) || ! is_numeric( $_GET[ $var ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 					continue;
 				}
-				$id = (int) $_GET[ $var ]; // phpcs:ignore WordPress.Security.NonceVerification
-
-				if ( empty( $id ) ) {
-					continue;
-				}
-				$lang = $this->model->post->get_language( $id );
+				$lang = $this->model->post->get_language( (int) $_GET[ $var ] ); // phpcs:ignore WordPress.Security.NonceVerification
 				return $lang ?: $this->model->get_default_language();
 			}
 		}


include/Model/Languages.php
```diff
@@ -89,6 +98,8 @@ public function __construct( Options $options, PLL_Translatable_Objects $transla
 		$this->options              = $options;
 		$this->translatable_objects = $translatable_objects;
 		$this->cache                = $cache;
+
+		add_action( 'set_current_user', array( $this, 'maybe_clean_local_cache' ), -10000 );
```
Marianne380: Why `-1000` ? Also could we add a comment for the priority ?

Screenfeed: It's only a way to say "before everybody", I don't think this requires a comment, does it?

modules/sitemaps/sitemaps.php
```diff
@@ -59,14 +59,15 @@ public function init() {
...
-	 * @param WP_Query    $query Main WP query object.
-	 * @return string|bool
+	 * @param PLL_Language|false $lang  Current language code, false if not set yet.
+	 * @param WP_Query           $query Main WP query object.
+	 * @return PLL_Language|false
 	 */
 	public function set_language_from_query( $lang, $query ) {
 		if ( isset( $query->query['sitemap'] ) && empty( $query->query['lang'] ) ) {
-			$lang = $this->options['default_lang'];
+			return $this->model->languages->get_default();
```
Chouby, line 70: It looks like a bug that no test detected?

include/Model/Languages.php
Chouby: ~In my opinion, this should be moved to Polylang Pro. In our current case, the filter depends on the user but we may have filters which could depend on something different.~

Let's discuss this. I'm not sure anymore.

frontend/choose-lang.php
```diff
@@ -84,6 +84,15 @@ protected function set_language( $curlang ) {
...
 		}
 
+		/**
+		 * Filters the current language to define on frontend.
+		 *
+		 * @since 3.8
+		 *
+		 * @param PLL_Language|false $curlang Instance of the current language.
+		 */
+		$curlang = apply_filters( 'pll_frontend_current_language', $curlang );
```
Chouby: We already have a filter `pll_get_current_language` which could be moved here (currently only available in `PLL_Choose_Lang_Content`). This would avoid to introduce a new filter. 

Here is a proposal to rewrite this method:
```php
	protected function set_language( $curlang  = null ) {
		// Don't set the language a second time.
		if ( isset( $this->curlang ) ) {
			return;
		}
		
		if ( ! $curlang instanceof PLL_Language ) {
			$curlang = $this->get_current_language();
		}

		if ( ! $curlang instanceof PLL_Language ) {
			$curlang = $this->get_preferred_language();
		}
		
		/**
		 * Filters the language before it is set.
		 *
		 * @since 0.9
		 *
		 * @param PLL_Language|false $lang Language object or false if none was found.
		 */
		$curlang = apply_filters( 'pll_get_current_language', $lang ?? false );
		
		if ( ! $curlang instanceof PLL_Language ) {
			$curlang = $this->model->get_default_language();
		}
		
		if ( ! $curlang instanceof PLL_Language ) {
			return;
		}

		$this->curlang = $curlang;

		$GLOBALS['text_direction'] = $this->curlang->is_rtl ? 'rtl' : 'ltr';
		if ( did_action( 'wp_default_styles' ) ) {
			wp_styles()->text_direction = $GLOBALS['text_direction'];
		}

		/**
		 * Fires when the current language is defined.
		 *
		 * @since 0.9.5
		 *
		 * @param string       $slug    Current language code.
		 * @param PLL_Language $curlang Current language object.
		 */
		do_action( 'pll_language_defined', $this->curlang->slug, $this->curlang );
	}
```
This requires to introduce an new (abstract) method `get_current_language()`.
In child classes, this will replace `get_language_from_content()`, remove `pll_get_current_language()` and the associated `add_filter()`, replace `set_language_from_url()` (its call will be replaced by a call to `set_language()`.

What do you think?

modules/sitemaps/sitemaps.php
Screenfeed: Indeed.

frontend/choose-lang.php
Screenfeed: Why not. Done in [39cb908](https://github.com/polylang/polylang/pull/1780/commits/39cb908e85b23c031f0d5eb16588756fc0605a3a). At the same time I reworked the methods to remove the variable assignations in `if` statements.

include/Model/Languages.php
Screenfeed: Removed in [692245f](https://github.com/polylang/polylang/pull/1780/commits/692245fb3b1fcdae1a5b6367776b289fb66baa22).

frontend/choose-lang.php
```diff
@@ -70,22 +70,37 @@ public function init() {
...
+		}
+
+		/**
+		 * Filters the language before it is set.
+		 *
+		 * @since 0.9
+		 *
+		 * @param PLL_Language|false $curlang Language object or false if none was found.
+		 */
+		$curlang = apply_filters( 'pll_get_current_language', $curlang ?? false );
```
Marianne380, line 103: Should we add a changelog for moving this filter?

frontend/choose-lang-url.php
```diff
@@ -55,35 +56,41 @@ public function set_language_from_url() {
...
+			return $this->get_home_language();
 		}
 
 		// Take care to post & page preview http://wordpress.org/support/topic/static-frontpage-url-parameter-url-language-information.
-		elseif ( isset( $_GET['preview'] ) && ( ( isset( $_GET['p'] ) && $id = (int) $_GET['p'] ) || ( isset( $_GET['page_id'] ) && $id = (int) $_GET['page_id'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$lang    = $this->model->post->get_language( $id );
-			$curlang = $lang ?: $this->model->get_default_language();
+		if ( isset( $_GET['preview'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			foreach ( array( 'p', 'page_id' ) as $var ) {
+				if ( ! isset( $_GET[ $var ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
Marianne380: Should we check `is_numeric` ?

frontend/choose-lang.php
Screenfeed: Good idea. Done in [081cbda](https://github.com/polylang/polylang/pull/1780/commits/081cbda64e33f2923a771ae284b1b1ad7f23e5e3). I had a hard time figuring out a proper sentence 😅

frontend/choose-lang-url.php
Screenfeed: It didn't seem to bother PHPStan but I added it anyway in [bfee01e](https://github.com/polylang/polylang/pull/1780/commits/bfee01e83278ae37418e08178cbd0227261c11dd).

frontend/choose-lang-url.php
```diff
@@ -55,35 +56,41 @@ public function set_language_from_url() {
...
+			return $this->get_home_language();
 		}
 
 		// Take care to post & page preview http://wordpress.org/support/topic/static-frontpage-url-parameter-url-language-information.
-		elseif ( isset( $_GET['preview'] ) && ( ( isset( $_GET['p'] ) && $id = (int) $_GET['p'] ) || ( isset( $_GET['page_id'] ) && $id = (int) $_GET['page_id'] ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$lang    = $this->model->post->get_language( $id );
-			$curlang = $lang ?: $this->model->get_default_language();
+		if ( isset( $_GET['preview'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			foreach ( array( 'p', 'page_id' ) as $var ) {
+				if ( ! isset( $_GET[ $var ] ) || ! is_numeric( $_GET[ $var ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
Chouby: To avoid testing empty later.
```suggestion
				if ( empty( $_GET[ $var ] ) || ! is_numeric( $_GET[ $var ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```

frontend/choose-lang-url.php
```diff
@@ -55,35 +56,41 @@ public function set_language_from_url() {
...
+					continue;
+				}
+				$id = (int) $_GET[ $var ]; // phpcs:ignore WordPress.Security.NonceVerification
+
+				if ( ! empty( $id ) ) {
+					return $this->model->post->get_language( $id );
+				}
+			}
+			return $this->model->get_default_language();
```
Chouby: This changes the behavior. The test was about a post without language which is not handled anymore.
```suggestion
				$lang = $this->model->post->get_language( (int) $_GET[ $var ] ); // phpcs:ignore WordPress.Security.NonceVerification
				return $lang ?: $this->model->get_default_language();
			}
```

frontend/choose-lang-url.php
```diff
@@ -55,35 +56,41 @@ public function set_language_from_url() {
...
 		}
 
-		// Take care to ( unattached ) attachments.
-		elseif ( isset( $_GET['attachment_id'] ) && $id = (int) $_GET['attachment_id'] ) { // phpcs:ignore WordPress.Security.NonceVerification
-			$lang    = $this->model->post->get_language( $id );
-			$curlang = $lang ?: $this->get_preferred_language();
+		// Take care to (unattached) attachments.
+		if ( ! empty( $_GET['attachment_id'] ) && is_numeric( $_GET['attachment_id'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			return $this->model->post->get_language( (int) $_GET['attachment_id'] ); // phpcs:ignore WordPress.Security.NonceVerification
```
Chouby, line 76: The preferred language will be used if the attachment has no language.

frontend/choose-lang.php
```diff
@@ -112,6 +128,15 @@ protected function set_language( $curlang ) {
...
 	}
 
+	/**
+	 * Returns the language to assign as the current one.
+	 *
+	 * @since 3.8
+	 *
+	 * @return PLL_Language|false
+	 */
+	abstract protected function get_current_language();
```
Chouby:
```suggestion
	abstract protected function get_current_language(): ?PLL_Language;
```

frontend/choose-lang.php
```diff
@@ -210,12 +235,23 @@ public function get_preferred_language() {
 	 * @return void
 	 */
 	protected function home_language() {
+		$this->set_language( $this->get_home_language() );
```
Chouby: I propose to remove this method (now used only once) and inline it in https://github.com/polylang/polylang/pull/1780/files#diff-58eb8f3eaa669c343525d387033b7d241015bd96bcb51b774fba1b1caafef53bR130

frontend/choose-lang.php
```diff
@@ -210,12 +235,23 @@ public function get_preferred_language() {
...
+	/**
+	 * Returns the language when home page is requested.
+	 *
+	 * @since 3.8
+	 *
+	 * @return PLL_Language|false
+	 */
+	protected function get_home_language() {
 		// Test referer in case PLL_COOKIE is set to false. Since WP 3.6.1, wp_get_referer() validates the host which is exactly what we want
 		// Thanks to Ov3rfly http://wordpress.org/support/topic/enhance-feature-when-front-page-is-visited-set-language-according-to-browser
```
Chouby:
```suggestion
		/*
		 * Test thereferer in case PLL_COOKIE is set to false. Since WP 3.6.1, wp_get_referer() validates the host which is exactly what we want.
		 * Thanks to Ov3rfly http://wordpress.org/support/topic/enhance-feature-when-front-page-is-visited-set-language-according-to-browser
		 */
```

modules/sitemaps/sitemaps.php
Chouby: Could you plan to add a test? (maybe in a separate not to delay this one).

include/Model/Languages.php
```diff
@@ -821,13 +834,40 @@ public function maybe_create_terms(): void {
...
+	 * @since 3.8
+	 * @see Languages::get_list()
+	 *
+	 * @return void
+	 */
+	public function maybe_clean_local_cache(): void {
+		if ( ! empty( $this->automatic_proxies ) ) {
+			$this->clean_local_cache();
+		}
+	}
```
Chouby: I propose to remove this method and call directly `clean_local_cache()` in Polylang Pro.

frontend/choose-lang-url.php
Screenfeed: This won't prevent testing for empty later. Ex:

```php
$var = ' ';
var_dump( empty( $var ) ); // bool(false)
var_dump( (int) $var ); // int(0)
```

frontend/choose-lang-url.php
Screenfeed: Yes, in `PLL_Choose_Lang::set_language()`, this is the case each time this method returns `false`.

frontend/choose-lang-url.php
Screenfeed: Done in [5b69efd](https://github.com/polylang/polylang/pull/1780/commits/5b69efd83f21eca1a06ad4e2f9e782f253b010b7).

frontend/choose-lang.php
Screenfeed: Indeed. I kept it because I thought it was still used in several places. I must have got my wires crossed.
Done in [daac02d](https://github.com/polylang/polylang/pull/1780/commits/daac02d0938e09ecaa7d849a3d2609e9425b6da6).

frontend/choose-lang.php
Screenfeed: Done in [de3ac66](https://github.com/polylang/polylang/pull/1780/commits/de3ac66f1a4416ed2a0fec94ac1e8fc85a7bb2a0) for both lines.

include/Model/Languages.php
Screenfeed: Done in [dc180bd](https://github.com/polylang/polylang/pull/1780/commits/dc180bdc235353eff0d12ede0c0a9eb796cfb86b).

frontend/choose-lang-url.php
Chouby: `$var = ' ';` will fail to pass `is_numeric()`.

Screenfeed: OK, you win this one 😆
Done in [38ad3ba](https://github.com/polylang/polylang/pull/1780/commits/38ad3ba81a55160099fa419d28a8d0805c4c4599).

modules/sitemaps/sitemaps.php
Screenfeed: [Issue](https://github.com/polylang/polylang/issues/1800) created.

PR #1782, opened 2025-12-15
Enforce and enhance string sanitization.

Hug0-Drelon:
## What
Fixes https://github.com/polylang/polylang-pro/issues/2831

## Why?
Currently, some strings can be saved without any sanitization at all.

## How?
- Create a new class `WP_Syntex\Polylang\Strings\Translator` to register/sanitize strings
    - Uses the same arguments as `PLL_Admin_Strings::register_string`
    - Adds a default sanitization (`wp_kses_post()` is the current user doesn't have `unfiltered_html` capability).
- Use the new class in `PLL_Admin_Strings` and `PLL_Translate_Option`
- Add tests for `WP_Syntex\Polylang\Strings\Translator`
- Add tests for option with multi-dimensional array and custom sanitization

### Exemple:
How to use the new class:
```php
new Translator(
	'My String', // The string that'll be translated.
	'my_string', // The name of the string as it'll appear in the strings translations table.
	'FooContext', // The context of the string (usually a plugin name).
	false, // Whether the string is multiline.
	'sanitize_text_field' // The sanitization callable to use.
);
```

### DIFF
--- a/include/api.php
+++ b/include/api.php
@@ -5,6 +5,8 @@
  * @package Polylang
  */
 
+use WP_Syntex\Polylang\Strings\Translator;
+
 /**
  * Template tag: displays the language switcher.
  * The function does nothing if used outside the frontend.
@@ -178,18 +180,29 @@ function pll_home_url( $lang = '' ) {
  *
  * @api
  * @since 0.6
+ * @since 3.8 Added $sanitize_callback parameter.
  *
- * @param string $name      A unique name for the string.
- * @param string $string    The string to register.
- * @param string $context   Optional, the group in which the string is registered, defaults to 'polylang'.
- * @param bool   $multiline Optional, true if the string table should display a multiline textarea,
- *                          false if should display a single line input, defaults to false.
+ * @param string        $name              A unique name for the string.
+ * @param string        $string            The string to register.
+ * @param string        $context           Optional, the group in which the string is registered, defaults to 'polylang'.
+ * @param bool          $multiline         Optional, true if the string table should display a multiline textarea,
+ *                                         false if should display a single line input, defaults to false.
+ * @param callable|null $sanitize_callback Optional, a callback function that sanitizes the string. Defaults to null.
  * @return void
  */
-function pll_register_string( $name, $string, $context = 'Polylang', $multiline = false ) {
-	if ( PLL() instanceof PLL_Admin_Base ) {
-		PLL_Admin_Strings::register_string( $name, $string, $context, $multiline );
+function pll_register_string( $name, $string, $context = 'Polylang', $multiline = false, $sanitize_callback = null ) {
+	if (
+		! is_scalar( $string )
+		|| ! is_string( $name )
+		|| ! is_string( $context )
+		|| ! is_bool( $multiline )
+		|| ! ( null === $sanitize_callback || is_callable( $sanitize_callback ) )
+	) {
+		_doing_it_wrong( __FUNCTION__, esc_html__( 'Invalid arguments.', 'polylang' ), '3.8' );
+		return;
 	}
+
+	new Translator( (string) $string, $name, $context, $multiline, $sanitize_callback );
 }
 
 /**


include/Strings/Translator.php
```diff
@@ -0,0 +1,139 @@
...
+	 *
+	 * @param string        $string            The string to translate.
+	 * @param string        $name              The name as defined in pll_register_string.
+	 * @param string        $context           The context of the string.
+	 * @param bool          $multiline         Whether the string is multiline.
+	 * @param callable|null $sanitize_callback The sanitization callback for the string.
+	 */
+	public function __construct(
+		string $string,
+		string $name,
```
Screenfeed, line 65: >Uses the same arguments as `PLL_Admin_Strings::register_string`

But the first 2 are in reverse order. I wonder if we should keep the same order (but we must keep this order in `sanitize()` because of the filter, so it would make it inconsistent, but on the other hand we don't call it directly) 🤔

include/Strings/Translator.php
```diff
@@ -0,0 +1,139 @@
...
+		string $context = 'Polylang',
+		bool $multiline = false,
+		?callable $sanitize_callback = null
+	) {
+		$this->string            = $string;
+		$this->name              = $name;
+		$this->context           = $context;
+		$this->multiline         = $multiline;
+		$this->sanitize_callback = $sanitize_callback ?? Closure::fromCallable( array( self::class, 'default_sanitization' ) );
```
Screenfeed: Why the closure? I suppose it's to keep the method private?

include/Strings/Translator.php
```diff
@@ -0,0 +1,139 @@
...
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $string The string to sanitize.
+	 * @return string The sanitized string.
+	 */
+	private static function default_sanitization( $string ) {
+		if ( ! current_user_can( 'unfiltered_html' ) ) {
+			return wp_kses_post( $string );
+		}
```
Screenfeed, line 140: Maybe we should put this `current_user_can()` outside the sanitization method, something like this:

```php
class Translator {
	public function __construct( ... ) {
		...
		$this->sanitize_callback = $sanitize_callback ?? 'wp_kses_post';
		...
	}

	public function sanitize( $translation, $name, $context, $original ): string {
		if ( ! $this->is_matching( $name, $context ) || current_user_can( 'unfiltered_html' ) ) {
			return $translation;
		}

		return call_user_func( $this->sanitize_callback, $translation, $name, $context, $original );
	}
}
```

Of course it depends on what we expect about the callback: should it follow the "`unfiltered_html` rule" or is it included in the callback?
Do you think it would make sense? (this would also remove the lines you added to `phpstan.neon.dist` I think)

include/Strings/Translator.php
```diff
@@ -0,0 +1,139 @@
...
+
+use Closure;
+use PLL_Admin_Strings;
+
+/**
+ * Class to translate strings.
+ *
+ * @since 3.8
+ */
+class Translator {
```
Screenfeed, line 16: I'm not sure about the name of this class, especially since we have `Capabilities\User::is_translator()`, it may be confusing.
The most disturbing part being that... it doesn't translate: it registers and sanitizes.
Unfortunately we can't use `String` as a class name 😅

include/translate-option.php
```diff
@@ -414,22 +402,13 @@ protected function check_value_recursive( $old_values, $values, $key, $mo ) {
...
-	public function sanitize_option( $value, $name, $context, $original ) {
-		if ( ! in_array( md5( "$original|$name|$context" ), $this->hashes, true ) ) {
-			return $value;
-		}
-
-		if ( is_callable( $this->sanitize_callback ) ) {
-			return call_user_func( $this->sanitize_callback, $value, $name, $context, $original );
-		}
-
+	public static function default_sanitize_option( $value, $name ) {
```
Screenfeed: Since it isn't a hook callback anymore, should we type-hint arguments and return?

include/Strings/Translator.php
Hug0-Drelon: Yes, no need to have more public method here.
Also, it'll create a closure (which is small) at runtime if needed.

include/Strings/Translator.php
Hug0-Drelon: The whole purpose is to have `current_user_can( 'unfiltered_html' )` next to `wp_kses_post()` to ensure a default sanitization when no callback is registered.
Your proposal goes against that, it'll bypass registered sanitization callback from third parties (thinks of PLLWC for instance, we do not call `current_user_can( 'unfiltered_html' )` and we don't want to when sanitization Woo strings).
Moreover, I don't get why it would be preferable (I don't see any benefit as you presented it).

include/Strings/Translator.php
Hug0-Drelon: I get inspired from `PLL_Translate_Option`, but I agree the name is not ideal. It's opened to suggestion.

include/translate-option.php
Hug0-Drelon: Maybe the return, but not the arguments because it comes from a filter (which is not trustworthy in WordPress).

include/Strings/Translator.php
Hug0-Drelon: Let’s put it on the table with the team.

include/Strings/Translator.php
Screenfeed: I suggested this because currently it means that sometimes we sanitize, and sometimes we don't, depending on `unfiltered_html`: it doesn't make sense to me for a default behavior. My example was clumsy but I think your answer proves that something is wrong. Let me explain.

>The whole purpose is to have `current_user_can( 'unfiltered_html' )` next to `wp_kses_post()` to ensure a default sanitization when no callback is registered.

The presence of `current_user_can( 'unfiltered_html' )` in the sanitization callback proves that there are no default sanitization, since `wp_kses_post()` will NOT RUN if it returns `true`, it defeats the purpose of this PR.

So, either we say:
1. The user is trustworthy and we use `current_user_can( 'unfiltered_html' )` on ALL sanitization callbacks, or,
2. The user is not trustworthy, and we don't use `current_user_can( 'unfiltered_html' )` anywhere, or,
3. The issue is not the user but the source of the data, and we must apply `current_user_can( 'unfiltered_html' )` on a case-by-case basis, not by default.

>thinks of PLLWC for instance, we do not call `current_user_can( 'unfiltered_html' )` and we don't want to when sanitization Woo strings

Exactly, so if we forget to sanitize something from PLLWC, `current_user_can( 'unfiltered_html' )` will be applied and the string won't be sanitized.

Hug0-Drelon: You don't get it. The required specification for this feature **is** to have default sanitization for user without `unfiltered_html` capability.
If a `sanitize_callback` is registered, it must run even if the current user has `unfiltered_html` capability, this is not our business to chose whether or not we must call the provided callback.
The default case, in the contrary, is in our hands. And the required behavior is to have a default sanitization (`wp_kses_post()`) **only** if the current user doesn't have `unfiltered_html` capability. In the UI, it means that only the admin can insert unfiltered strings in the table.



> Exactly, so if we forget to sanitize something from PLLWC, current_user_can( 'unfiltered_html' ) will be applied and the string won't be sanitized.

PLLWC already implement a default sanitization (`wc_clean()`), even though a refactor will be require in the add-on (but this is for another day).

Screenfeed: We'll discuss all this with the team, it will be easier to understand each other. For me it's a matter of the real purpose of the PR.

include/Strings/Translator.php
Screenfeed: `Translation_String`? So it relates to the translations strings.

Hug0-Drelon: I don't think so, "String" term being already used in the namespace (makes it redundant).

PR #1788, opened 2026-01-02
Harmonize capabilities error message.

Marianne380:
Fixes partially https://github.com/polylang/polylang-pro/issues/2780.

## What

This PR harmonizes error message in the Capabilities feature.

The main issue is in `User::can_translate_or_die()` which displays:
> "You are not allowed to do action in English."

This message has weird grammar ("do action") and lacks the "Sorry," prefix used in other similar messages.

---

## Notes

I went through all the capabilities related error messages I could find. If you spot any others that need updating, just let me know!

### DIFF
--- a/include/Capabilities/User.php
+++ b/include/Capabilities/User.php
@@ -181,7 +181,7 @@ class User {
 	public function can_translate_or_die( PLL_Language $language ): void {
 		if ( ! $this->can_translate( $language ) ) {
 			/* translators: %s is a language name */
-			wp_die( esc_html( sprintf( __( 'Sorry, you are not allowed to create or edit content in %s.', 'polylang' ), $language->name ) ) );
+			wp_die( esc_html( sprintf( __( 'Sorry, you are not allowed to edit content in %s.', 'polylang' ), $language->name ) ) );
 		}
 	}
 }


include/Capabilities/User.php
```diff
@@ -180,8 +180,8 @@ private function get_language_caps(): array {
 	 */
 	public function can_translate_or_die( PLL_Language $language ): void {
 		if ( ! $this->can_translate( $language ) ) {
-			/* translators: %s: language name */
-			wp_die( esc_html( sprintf( __( 'You are not allowed to do action in %s.', 'polylang' ), $language->name ) ) );
+			/* translators: %s is a language name */
+			wp_die( esc_html( sprintf( __( 'Sorry, you are not allowed to create or edit content in %s.', 'polylang' ), $language->name ) ) );
```
Hug0-Drelon: I agree that "do action" is an awkward wording, but this message could be also displayed in reading mode I think?

Marianne380: We could use a more generic message like:

 **_"Sorry, you are not allowed to work with content in %s."_**

Or do you have other suggestions that would cover all cases?

Hug0-Drelon: I don't know to be honest 😅 
Let's move this to final review.

Chouby: I would simplify to
`Sorry, you are not allowed to edit content in %s.`

Marianne380:
```suggestion
			wp_die( esc_html( sprintf( __( 'Sorry, you are not allowed to edit content in %s.', 'polylang' ), $language->name ) ) );
```

PR #1790, opened 2026-01-06
Fix term slug suffix added incorrectly when changing language.

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/2857.

## What

When changing a term's language in Polylang, an unwanted language suffix was added to the slug.

#### 1. Missing term_id tracking

The code couldn't distinguish between creating and editing.

#### 2. Incorrect slug conflict detection

WordPress `term_exists()` searches by **slug OR name**:

```php
// Term "Dog" exists with slug "cat"
term_exists('dog', 'category'); // Returns TRUE (finds by name "Dog")
```

This caused false positive conflicts, adding suffixes when the slug was actually available.

#### 3. Insufficient language specific checks

The old `term_exists_by_slug()` didn't clearly separate:
- Checking if slug exists globally (any language)
- Checking if slug exists in a specific language

#### 2. Check only slug conflicts

Use directly `get_terms()` and exclude the term being edited.

Also added tests.

### DIFF
--- a/include/term-slug.php
+++ b/include/term-slug.php
@@ -124,17 +124,9 @@ class PLL_Term_Slug {
 		);
 
 		$terms = get_terms( $args );
-		if ( is_wp_error( $terms ) || empty( $terms ) ) {
-			// Slug doesn't exist anywhere: no suffix needed.
-			return false;
-		}
 
-		// Exclude current term if we're editing.
-		if ( $this->term_id > 0 ) {
-			$terms = array_diff( $terms, array( $this->term_id ) );
-		}
-
-		return ! empty( $terms );
+		// Return true if the slug is already used by another term (excluding the term being edited).
+		return is_array( $terms ) && ! empty( array_diff( $terms, array( $this->term_id ) ) );
 	}
 
 	/**


include/crud-terms.php
```diff
@@ -290,6 +299,34 @@ public function set_pre_term_name( $name ) {
...
+	/**
+	 * Stores the term ID to use in `pre_term_slug`.
+	 *
+	 * @since 3.7.7
+	 *
+	 * @param int $term_id Term ID.
+	 * @return int Unmodified term ID.
+	 */
+	public function set_pre_term_id( $term_id ) {
+		$this->pre_term_id = is_int( $term_id ) ? $term_id : 0;
```
Screenfeed: I think this would be safer:

```suggestion
		$this->pre_term_id = is_numeric( $term_id ) ? (int) $term_id : 0;
```

What do you think?

include/crud-terms.php
```diff
@@ -290,6 +299,34 @@ public function set_pre_term_name( $name ) {
...
+	 * @since 3.7.7
+	 *
+	 * @param string|WP_Error $term Term name or WP_Error object.
+	 * @return string|WP_Error Unmodified term.
+	 */
+	public function reset_pre_term_id( $term ) {
+		$this->pre_term_id = 0;
+
+		return $term;
+	}
```
Screenfeed: I can't see the point of this one 🤔. If I follow the logic:

```php
function wp_insert_term( $term, $taxonomy, $args = array() ) {
	// ...
	$term = apply_filters( 'pre_insert_term', $term, $taxonomy, $args );
	// ...
	$args = sanitize_term( $args, $taxonomy, 'db' );
	// ...
}

function sanitize_term( $term, $taxonomy, $context = 'display' ) {
	$fields = array( 'term_id', 'name', 'description', 'slug', 'count', 'parent', 'term_group', 'term_taxonomy_id', 'object_id' );
	// ...
	foreach ( (array) $fields as $field ) {
		// ...
		sanitize_term_field( $field, /* ... */ );
		// ...
	}
	// ...
}
```

So:

1. `pre_insert_term` is applied: `PLL_CRUD_Terms::$pre_term_id` is set to `0`.
2. `sanitize_term()` is called.
3. Seeing the order of the field names in the array in `sanitize_term()`, `sanitize_term_field( 'term_id' )` is called: `PLL_CRUD_Terms::$pre_term_id` is set to the term ID.
4. `sanitize_term_field( 'sug' )` is called: `PLL_CRUD_Terms::$pre_term_id` is used.

So, unless I'm missing something, the 1st step that resets `PLL_CRUD_Terms::$pre_term_id` is useless, since `sanitize_term_field( 'term_id' )` will always be called first.

include/model.php
```diff
@@ -285,6 +315,22 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
...
+	 *
+	 * @since 3.7.7
+	 *
+	 * @param string              $slug     The term slug to test.
+	 * @param string|PLL_Language $language The language slug or object.
+	 * @param string              $taxonomy Optional taxonomy name.
+	 * @param int                 $parent   Optional parent term id.
+	 * @return int The `term_id` of the found term. 0 otherwise.
+	 */
+	public function term_exists_by_slug_and_language( $slug, $language, $taxonomy = '', $parent = 0 ): int {
```
Screenfeed: What about `PLL_Model::term_exists()`? If we follow this logic we should rename it `term_exists_by_name_and_language()`.

We also could create a common method `term_exists_by_field_and_language( $field, $value, $language, $taxonomy = '', $parent = 0 )`, inspired from [`get_term_by()`](https://developer.wordpress.org/reference/functions/get_term_by/).

Some arguments also should be time-hinted IMHO.

include/model.php
```diff
@@ -285,6 +315,22 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
...
+	/**
+	 * Checks if a term slug exists in a given language, taxonomy, hierarchy.
+	 *
+	 * @since 3.7.7
+	 *
+	 * @param string              $slug     The term slug to test.
+	 * @param string|PLL_Language $language The language slug or object.
+	 * @param string              $taxonomy Optional taxonomy name.
+	 * @param int                 $parent   Optional parent term id.
+	 * @return int The `term_id` of the found term. 0 otherwise.
```
Screenfeed, line 285: Like `PLL_Model::term_exists()`, I think we should ensure that we return a term ID >= 0. It may not be very useful for now but it costs nothing to be more precise IMHO. Same goes for `term_exists_by_slug_globally()`.

```suggestion
	 * @return int The `term_id` of the found term. 0 otherwise.
	 *
	 * @phpstan-return int<0, max>
```

Then:

```php
		// PHPCS:ignore WordPress.DB.PreparedSQL.NotPrepared
		$term_id = $wpdb->get_var( $select . $join . $where );
		return max( 0, (int) $term_id );
```

include/model.php
```diff
@@ -272,11 +272,41 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
...
+	 * Only checks by slug, not by name (unlike `term_exists`).
+	 *
+	 * @since 3.7.7
+	 *
+	 * @param string $slug            The term slug to test.
+	 * @param string $taxonomy        Optional taxonomy name.
+	 * @param int    $exclude_term_id Optional term ID to exclude from the search.
+	 * @return bool True if slug exists, false otherwise.
+	 */
+	public function term_exists_by_slug_globally( string $slug, string $taxonomy = '', int $exclude_term_id = 0 ): bool {
```
Screenfeed: No `$parent` argument?

tests/phpunit/tests/test-slugs.php
```diff
@@ -23,24 +23,70 @@ public function set_up() {
...
 	}
 
-	public function test_term_slugs() {
-		$term_id = self::factory()->category->create(
+	/**
+	 * Test creating a translated term with same slug adds suffix.
+	 *
+	 * When creating a translation of an existing term with the same slug, a language suffix should be added to avoid conflicts.
+	 */
+	public function test_translated_terms_with_same_slug_get_suffix() {
```
Screenfeed, line 31: Is it necessary to create 2 tests: one for translations and one for unrelated terms? I would have thought that they use the same mechanic, that the fact that they are translations of each other wouldn't change anything.

tests/phpunit/tests/test-slugs.php
```diff
@@ -205,4 +251,142 @@ function ( $found_language ) use ( $fr_lang ) {
...
+		// Create category "Dog" in French.
+		$term_id = self::factory()->category->create(
+			array(
+				'name' => 'Dog',
+				'slug' => 'dog',
+				'lang' => 'fr',
+			)
+		);
+
+		$term = get_term( $term_id, 'category' );
```
Screenfeed: Idea for here and other tests:
```suggestion
		$term = self::factory()->category->create_and_get(
			array(
				'name' => 'Dog',
				'slug' => 'dog',
				'lang' => 'fr',
			)
		);
```

Then use `$term->term_id` instead of `$term_id`. What do you think?

tests/phpunit/tests/test-slugs.php
```diff
@@ -205,4 +251,142 @@ function ( $found_language ) use ( $fr_lang ) {
...
+
+		wp_update_term( $term_id, 'category' );
+
+		// Verify language changed and slug remained unchanged.
+		$term = get_term( $term_id, 'category' );
+		$this->assertSame( 'en', self::$model->term->get_language( $term_id )->slug, 'Language should be changed to English' );
+		$this->assertSame( 'dog', $term->slug, 'Slug should remain "dog" without suffix when changing language' );
+
+		// Clean up.
+		unset( $_REQUEST, $_POST );
```
Screenfeed: I don't think this is useful, this is done in [`set_up()`](https://github.com/WordPress/wordpress-develop/blob/e9058eea67aea1526b06c67886abba2a5b11593b/tests/phpunit/includes/abstract-testcase.php#L117).

Screenfeed (changes requested):

include/model.php
Marianne380: It doesn't seem useful. The method checks for global existence across all hierarchy levels.

At the point this method is called in `can_add_suffix()`, the parent suffix has already been applied to `$this->slug` via `maybe_get_parent_suffix()`, so we're checking the full hierarchical slug (e.g., _"test-test"_) not just _"test"_.

Do you see any reason to add it anyway?

include/model.php
Marianne380: Renaming `term_exists()` to `term_exists_by_name_and_language()`might be a good idea, but it seems a little out of the scope, since we don't touch it and there doesn't seem to be any name conflict.

Regarding creating a generic `term_exists_by_field_and_language()` method inspired by `get_term_by()`, how would that help us? Do we need to search by anything other than the slug?

We could address both of these improvements in a follow-up PR if there's interest.

include/model.php
Marianne380: See https://github.com/polylang/polylang/pull/1790/changes/6b7e8f7bcb6e6e97bc5327945e7780f777361256.

include/crud-terms.php
Marianne380: Right, see https://github.com/polylang/polylang/pull/1790/changes/2c97b4438e4f6085388a7c31c0d6c4d0858dbb23.

tests/phpunit/tests/test-slugs.php
Marianne380: Personally, I find it useful, even if it behaves the same way today.
It also allows us to check that nothing is broken in either case.
Afterwards, we can remove it if you think it's really overkill.

tests/phpunit/tests/test-slugs.php
Marianne380: See https://github.com/polylang/polylang/pull/1790/commits/b3f0ed0fd11860aa392f11e88a2a369bfd470371.

include/model.php
Screenfeed: `PLL_Model::term_exists()` searches by name, that's why I suggested the generic method. But yes, it could be done in another PR.

include/model.php
Screenfeed: You're right, the parent's slug is included.

include/model.php
```diff
@@ -272,11 +272,42 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
...
+	/**
+	 * Checks if a term slug exists globally (any language).
+	 * Only checks by slug, not by name (unlike `term_exists`).
+	 *
+	 * @since 3.7.7
+	 *
+	 * @param string $slug            The term slug to test.
+	 * @param string $taxonomy        Optional taxonomy name.
+	 * @param int    $exclude_term_id Optional term ID to exclude from the search.
+	 * @return bool True if slug exists, false otherwise.
```
Screenfeed: This one has been forgotten.

```suggestion
	 * @return bool True if slug exists, false otherwise.
	 *
	 * @phpstan-return int<0, max>
```

Screenfeed (changes requested):

include/model.php
Screenfeed: Or not... What did I do?! XD

Screenfeed: Han! Type-hinted as `bool` but we return an integer x)

Marianne380: Oops, I'll revert that.

Marianne380: And no, we do return a Boolean.

Screenfeed: Han! I missed the `> 0`

Screenfeed: So:
- `max( 0, (int) $term_id ) > 0`: `max()` is useless then.
- But I think this method should return the term ID, like `term_exists_by_slug_and_language()` and `term_exists()`. Don't you think?

Marianne380: Simplified in https://github.com/polylang/polylang/pull/1790/commits/d60489c9a72a5ddface852d22f8c9f0306940af0.

At first, I thought about returning an `int` like the others for consistency, but I didn't see the point.

Besides, which ID should we return if we find multiple slugs? An array?

In our use case, the boolean seems to be enough.

Screenfeed: You get a point. `term_exists_by_slug_and_language()` is not supposed to return more than one term, but `term_exists_by_slug_globally()` can (one in each language for example).
OK let's keep the boolean for now and let's see what @Chouby thinks about it.

Screenfeed (approved):

include/crud-terms.php
```diff
@@ -290,6 +298,20 @@ public function set_pre_term_name( $name ) {
...
+	 * @since 3.7.7
+	 *
+	 * @param int $term_id Term ID.
+	 * @return int Unmodified term ID.
+	 */
+	public function set_pre_term_id( $term_id ) {
+		$this->pre_term_id = is_numeric( $term_id ) ? (int) $term_id : 0;
+
+		return $term_id;
+	}
```
Chouby: I suggest to keep the same order for `add_filter()` and method definitions.

include/model.php
```diff
@@ -285,6 +316,23 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
 	 * @return int The `term_id` of the found term. 0 otherwise.
 	 */
 	public function term_exists_by_slug( $slug, $language, $taxonomy = '', $parent = 0 ): int {
+		_deprecated_function( __METHOD__, '3.7.7', 'PLL_Model::term_exists_by_slug_and_language()' );
+		return $this->term_exists_by_slug_and_language( $slug, $language, $taxonomy, $parent );
```
Chouby: Although your name is better, I prefer not to rename this method: It's old, used elsewhere (at least in Polylang Pro), this is a minor version and it's not necessary.

include/term-slug.php
```diff
@@ -113,10 +113,13 @@ private function can_add_suffix() {
 			}
 		}
 
-		if ( ! term_exists( $this->slug, $this->taxonomy ) ) {
+		// Check if the slug exists globally, excluding the current term if we're editing.
+		if ( ! $this->model->term_exists_by_slug_globally( $this->slug, $this->taxonomy, $this->term_id ) ) {
```
Chouby: Instead of creating a new method with a direct SQL query, I propose to keep using WP methods. I foresee 2 possibilities.
1. Use directly `get_terms()` to get what you want. It's noticeable that `term_exists()` now uses `get_terms()` internally.
2. I did not test but the filter `term_exists_default_query_args` could allow to keep using `term_exists()`.

As a bonus, we could now rewrite the `PLL_Model::term_exists_by_slug()` to use WP functions (in a separate PR) which was not possible in older WP versions.

Chouby (changes requested):

include/term-slug.php
Marianne380: Done in https://github.com/polylang/polylang/pull/1790/commits/88f9c122b46ecce7d048e63bdab127786e968284.

And okay to rewrite `PLL_Model::term_exists_by_slug()` in a follow-up PR.

include/model.php
```diff
@@ -272,6 +272,44 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
...
+	 * Only checks by slug, not by name (unlike `term_exists`).
+	 *
+	 * @since 3.7.7
+	 *
+	 * @param string $slug            The term slug to test.
+	 * @param string $taxonomy        Optional taxonomy name.
+	 * @param int    $exclude_term_id Optional term ID to exclude from the search.
+	 * @return bool True if slug exists, false otherwise.
+	 */
+	public function term_exists_by_slug_globally( string $slug, string $taxonomy = '', int $exclude_term_id = 0 ): bool {
```
Chouby: I would inline this method where it's used and simplifying the code to exactly match your needs.

include/model.php
```diff
@@ -272,6 +272,44 @@ public function term_exists( $term_name, $taxonomy, $parent, $language ): int {
...
+		$terms = get_terms( $args );
+
+		if ( is_wp_error( $terms ) || empty( $terms ) ) {
+			return false;
+		}
+
+		// Exclude current term if specified.
+		if ( $exclude_term_id > 0 ) {
+			$terms = array_diff( $terms, array( $exclude_term_id ) );
+		}
```
Chouby: Why not using the `exclude` param of `get_terms()`? See https://developer.wordpress.org/reference/classes/wp_term_query/__construct/#parameters

include/term-slug.php
```diff
@@ -113,10 +113,24 @@ private function can_add_suffix() {
...
+		);
+
+		$terms = get_terms( $args );
+		if ( is_wp_error( $terms ) || empty( $terms ) ) {
+			// Slug doesn't exist anywhere (or only exists for the current term): no suffix needed.
 			return false;
 		}
 
+		// Slug exists for other terms: suffix needed.
 		return true;
```
Chouby: Just a matter of taste
```suggestion
		return is_array( $terms ) && ! empty( $terms );
```

Chouby (changes requested):

include/term-slug.php
```diff
@@ -113,19 +113,36 @@ private function can_add_suffix() {
...
 			return false;
 		}
 
-		return true;
+		// Exclude current term if we're editing.
+		if ( $this->term_id > 0 ) {
+			$terms = array_diff( $terms, array( $this->term_id ) );
+		}
+
+		return ! empty( $terms );
```
Chouby: What do you think of ( 2 tests instead of 4)?
```php
return is_array( $terms ) && ! empty( array_diff( $terms, array( $this->term_id ) ) );
```

Marianne380: I find this condensed version a bit less readable at first glance.

Bug I guess it can work with a nice comment, see https://github.com/polylang/polylang/pull/1790/commits/7150d3eca9913c253a06866634ade642a58c2c9d.

Chrystll:
@Marianne380 I noticed an issue with the slugs in your PR. Translating 'Dog' to French results in 'dog-fr', but I think it should be 'dog-2'.

PR #1791, opened 2026-01-07
Enforce strings sanitization for users without `unfiltered_html` capability.

Hug0-Drelon:
## What?
Now that the Strings panel is opened to users without `manage_options` or `unfiltered_html` capabilities, we must enforce strings sanitization.

## How?
- Check user `unfiltered_html` capability in `PLL_Admin_Strings::sanitize_string_translation()` and call `wp_kses_post()` if applicable. Also take care to **not** overwrite existing strings.
- Add a new parameter `$previous_` to `pll_sanitize_string_translation` filter, used by `PLL_Admin_Strings::sanitize_string_translation()`.
- Add some PHPUnit tests for safe and unsafe strings with both types of users.

### DIFF
--- a/tests/phpunit/tests/Strings/test-Filtered_Sanitization.php
+++ b/tests/phpunit/tests/Strings/test-Filtered_Sanitization.php
@@ -88,4 +88,17 @@ class Test_Filtered_Sanitization extends PLL_UnitTestCase {
 	public function test_sanitize_widget_title( $user, $string, $expected ) {
 		$this->assertSame( $expected, PLL_Admin_Strings::sanitize_string_translation( $string, 'widget_title', 'test', 'test', 'test' ) );
 	}
+
+	/**
+	 * @testWith ["the_boss"]
+	 *           ["not_the_boss"]
+	 *
+	 * @param string $user The user to set as current user.
+	 * @return void
+	 */
+	public function test_sanitize_already_existing_unsafe_translation( $user ) {
+		wp_set_current_user( self::${$user}->ID );
+
+		$this->assertSame( '<script>alert(\'heck\');</script>', PLL_Admin_Strings::sanitize_string_translation( '<script>alert(\'heck\');</script>', 'widget_title', 'test', 'test', '<script>alert(\'heck\');</script>' ) );
+	}
 }


admin/admin-strings.php
```diff
@@ -126,16 +126,24 @@ public static function &get_strings() {
...
+	 * @param string $name        The name as defined in pll_register_string. Unused.
+	 * @param string $context     The context as defined in pll_register_string. Unused.
+	 * @param string $original    The original string to translate. Unused.
+	 * @param string $previous    The previous string translation.
 	 * @return string
 	 */
-	public static function sanitize_string_translation( $translation, $name ) {
+	public static function sanitize_string_translation( $translation, $name, $context, $original, $previous ) {
+		if ( trim( $previous ) === trim( $translation ) ) {
+			// Don't overwrite the translation to prevent to break string.
```
Marianne380:
```suggestion
			// Don't overwrite the translation to prevent breaking the string.
```

admin/admin-strings.php
```diff
@@ -126,16 +126,24 @@ public static function &get_strings() {
...
+	 * @param string $translation The string translation.
+	 * @param string $name        The name as defined in pll_register_string. Unused.
+	 * @param string $context     The context as defined in pll_register_string. Unused.
+	 * @param string $original    The original string to translate. Unused.
+	 * @param string $previous    The previous string translation.
 	 * @return string
 	 */
-	public static function sanitize_string_translation( $translation, $name ) {
+	public static function sanitize_string_translation( $translation, $name, $context, $original, $previous ) {
+		if ( trim( $previous ) === trim( $translation ) ) {
```
Marianne380, line 137: Doesn't it create a security bypass ?
If a malicious translations was previously saved, and a user saves again without modifying it, sanitization would be completely bypassed, right ?

tests/phpunit/tests/Strings/test-Filtered_Sanitization.php
```diff
@@ -0,0 +1,78 @@
+<?php
```
Marianne380, line 1: Should we add a specific test for `widget_title` ?

settings/table-string.php
```diff
@@ -419,8 +419,9 @@ public function save_translations() {
 					 * @param string $name        The name as defined in pll_register_string.
 					 * @param string $context     The context as defined in pll_register_string.
 					 * @param string $original    The original string to translate.
+					 * @param string $previous    The previous string translation.
 					 */
-					$translation = apply_filters( 'pll_sanitize_string_translation', $translation, $this->strings[ $key ]['name'], $this->strings[ $key ]['context'], $this->strings[ $key ]['string'] );
+					$translation = apply_filters( 'pll_sanitize_string_translation', $translation, $this->strings[ $key ]['name'], $this->strings[ $key ]['context'], $this->strings[ $key ]['string'], $mo->translate_if_any( $this->strings[ $key ]['string'] ) );
```
Marianne380, line 425: Should we add a changelog ?

admin/admin-strings.php
```diff
@@ -126,16 +126,24 @@ public static function &get_strings() {
...
+			// Don't overwrite the translation to prevent to break string.
+			return $translation;
+		}
+
 		if ( $name == self::$default_strings['widget_title'] ) {
```
Marianne380: Could we take the opportunity to replace `==` with `===` ?

Also `sanitize_text_field()` already strips HTML, so the subsequent `wp_kses_post()` for widget titles is redundant, right ? 

Is it a choice to behave differently for `widget_title` and for the rest? 
If so, we could consider something like:
```php
if ( $name === self::$default_strings['widget_title'] ) {
    $translation = sanitize_text_field( $translation );
} elseif ( ! current_user_can( 'unfiltered_html' ) ) {
    $translation = wp_kses_post( $translation );
}
```

tests/phpunit/tests/Strings/test-Filtered_Sanitization.php
Hug0-Drelon: Added in b747c52d.

settings/table-string.php
Hug0-Drelon: Added in b747c52d.

admin/admin-strings.php
Hug0-Drelon: Suggestion applied in b747c52d.

admin/admin-strings.php
Hug0-Drelon: Applied in b747c52d.

admin/admin-strings.php
Hug0-Drelon: The requirement for the PR is to avoid string overwrite. Here is the scenario:
- The admin (with `unfitlered_html`) translates a string `foo` with some HTML.
- Another user without `unfiltered_html` translates another string `bar`. But while saving the whole array of strings, `foo` get saved too, even if not modified
- `foo` gets sanitized (and potentially broken).

So I think this is an acceptable trad-off here. What do you think?

tests/phpunit/tests/Strings/test-Filtered_Sanitization.php
```diff
@@ -0,0 +1,91 @@
...
+	 *
+	 * @param string $user   The user to set as current user.
+	 * @param string $string The string to sanitize.
+	 * @param string $expected The expected string.
+	 * @return void
+	 */
+	public function test_sanitize_unsafe_string( $user, $string, $expected ) {
+		wp_set_current_user( self::${$user}->ID );
+
+		$this->assertSame( $expected, apply_filters( 'pll_sanitize_string_translation', $string, 'test', 'test', 'test', 'test' ) );
```
Chouby, line 76: Could you add a test with an unchanged unsafe string?

Hug0-Drelon: Done in 6523a4a9.

PR #1793, opened 2026-01-12
Support new DeepL languages.

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/2812.

## What

This PR adds support for languages recently promoted by DeepL and completes existing language stubs.

## Fix

  - Added DeepL language codes for many languages that were recently promoted from beta to production status by DeepL. 
  - Added some languages that were available in WordPress (but not yet in Polylang and some not in DeepL, I just took the opportunity to add them).
  - Completed some stub entries with missing fields (now that the languages are available on translate WordPress).

#### Regional Flags (currently use national flags)

**bre (Breton)** : Uses 'fr'
**scn (Sicilian)** : Uses 'it'
**lmo (Lombard)** : Uses 'it'

Users can add custom regional flags if desired.
However, for consistency with our existing regional flags (catalonia, basque, occitania, veneto, wales, scotland...), should we consider adding official support for brittany, sicily, and lombardy flags or is it ok this way ?

#### National Flag Choices

**hau (Hausa)** : 'ng' (Nigeria) over 'ne' (Niger).
It follows Facebook locale `ha_NG` and larger speaker population, though Niger made Hausa sole official language in 2025 (replacing French).

**tir (Tigrinya)** : 'et' (Ethiopia) over 'er' (Eritrea). Seems to be a larger speaker population.

**ba (Bashkir)** : 'ru' for consistency with `sah` (Yakut) and `tt_RU` (Tatar).

**lin (Lingala)** : 'cd' (DR Congo) per Facebook locale `ln_CD`.

---

## Known Issue (to be fixed in follow-up PR)

There is an issue in `Services/Deepl.php` with the `get_source_code()` method:

```php
return substr( static::get_target_code( $language ), 0, 2 );
```

Problem: Codes with 3+ letters without hyphens are incorrectly truncated:
- `CEB` (Cebuano) : returns `CE` instead of `CEB` ❌
- `CKB` (Kurdish Sorani) : returns `CK` instead of `CKB` ❌

Translations FROM Cebuano or Kurdish Sorani will fail because DeepL receives an invalid source code.

This will be addressed in a follow-up PR that fixes the `get_source_code()` logic to handle 3-letter codes properly.

### DIFF
--- a/settings/languages.php
+++ b/settings/languages.php
@@ -635,7 +635,7 @@ return array(
 		'name'     => 'Afaan Oromoo',
 		'dir'      => 'ltr',
 		'flag'     => 'et',
-		'w3c'      => 'om',
+		'w3c'      => 'gax',
 		'facebook' => 'om_ET',
 		'deepl'    => 'OM',
 	),


settings/languages.php
```diff
@@ -1259,10 +1425,26 @@
...
 	'xho' => array(
 		'facebook' => 'xh_ZA',
 	),
 	'yor' => array(
+		'code'     => 'yo',
+		'locale'   => 'yor',
+		'name'     => 'Yorùbá',
+		'dir'      => 'ltr',
+		'flag'     => 'ng',
+		'w3c'      => 'yo',
```
Hug0-Drelon, line 1510: Do we agree there is no DeepL code for this one?

settings/languages.php
```diff
@@ -1151,6 +1303,7 @@
 		'dir'      => 'ltr',
 		'flag'     => 'lk',
 		'facebook' => 'ta_IN',
+		'deepl'    => 'TA',
```
Hug0-Drelon, line 1355: The same DeepL code is used for two different locale (though they seems really close). I guess it's one purpose, am I right?

Marianne380: Yes, it's intentional! DeepL only offers one Tamil code ('TA') for all regional variants, while WP distinguishes `ta_IN` (India) and `ta_LK` (Sri Lanka). Both locales point to the same DeepL code since DeepL doesn't make regional distinctions for this language.

Good catch on the potential quality issue though, the DeepL Tamil translation is probably optimized for `ta_IN` and might not be perfectly natural for `ta_LK` speakers.

Do you think we should still keep the 'TA' code for `ta_LK` despite this, or remove it? Users can always review/correct translations afterward, but I'm not sure what's best here.

Hug0-Drelon: I think your code is right, let's keep it that way. I was just asking to be sure.

Hug0-Drelon (approved):

settings/languages.php
```diff
@@ -47,6 +49,21 @@
  * 've_ZA' (Venda)
  * 'yi_DE' (Yiddish)
  * 'zz_TR' (Zazaki)
+ *
+ * DeepL locales without equivalent WordPress locale:
+ * 'ACE' (Acehnese)
+ * 'AY'  (Aymara)
+ * 'GN'  (Guarani)
+ * 'GOM' (Konkani)
+ * 'OM'  (Oromo)
```
Marianne380: I just noticed that `‘OM’` is listed under `‘gax’` in the WordPress languages. So maybe we should add it too?
```php
'gax' => array(
    'code'     => 'om',
    'locale'   => 'gax',
    'name'     => 'Afaan Oromoo',
    'dir'      => 'ltr',
    'flag'     => 'et',
    'w3c'      => 'om',
    'facebook' => 'om_ET',
    'deepl'    => 'OM',
),
```

It's hard to see everything... :dizzy_face:

Hug0-Drelon: Indeed, I think you're right. Let's add it.

Hug0-Drelon (approved):

settings/languages.php
```diff
@@ -568,10 +620,24 @@
...
 		'facebook' => 'ga_IE',
+		'deepl'    => 'GA',
 	),
 	'gax' => array(
+		'code'     => 'om',
+		'locale'   => 'gax',
+		'name'     => 'Afaan Oromoo',
+		'dir'      => 'ltr',
+		'flag'     => 'et',
+		'w3c'      => 'om',
```
Chouby: `gax` is a valid W3C locale, so I would change to gax. I'm unsure for the code `om` is the macro language of `gax`. This one can of course be changed by the user. Or given that the translation is at 0, we could simply keep only the Facebook and DeepL code.

Marianne380: `gax` is now used as the W3C locale, see https://github.com/polylang/polylang/pull/1793/commits/0434653d91f46be28813d82637084a127bc21ce3.
I kept `om` as code since it is the ISO 639-1 macrolanguage.

PR #1794, opened 2026-01-12
Store DeepL locale in PLL_Language object

Chouby:
Two goals for this PR
1. introduce a new `PLL_Language` property for the DeepL locale.
-> to match the other W3C and Facebook locales. 
-> to allow to simplify [`Deepl::get_target_code()`](https://github.com/polylang/polylang-pro/blob/3.7.6/modules/Machine_Translation/Services/Deepl.php#L217-L227).
2. introduce a new filter `pll_additional_locales` to allow developers to add W3C, Facebook or DeepL locales which would not be already included in our predefined list. Only these properties are accepted  in this filter. 

Bonus: `PLL_Language_Factory` PHPStan errors are fixed.

### DIFF
--- a/tests/phpunit/tests/test-create-delete-languages.php
+++ b/tests/phpunit/tests/test-create-delete-languages.php
@@ -101,6 +101,7 @@ class Create_Delete_Languages_Test extends PLL_UnitTestCase {
 			'is_rtl',
 			'w3c',
 			'facebook',
+			'deepl',
 			'home_url',
 			'search_url',
 			'host',


include/language-factory.php
```diff
@@ -95,37 +94,16 @@ public function get_from_terms( array $terms ) {
...
-				$data['w3c'] = $languages[ $data['locale'] ]['w3c'];
-			} else {
-				$data['w3c'] = str_replace( '_', '-', $data['locale'] );
-			}
-
-			if ( isset( $languages[ $data['locale'] ]['facebook'] ) ) {
-				$data['facebook'] = $languages[ $data['locale'] ]['facebook'];
-			}
+		if ( ! is_array( $description ) || empty( $description['locale'] ) || ! is_string( $description['locale'] ) ) {
+			return null;
```
Screenfeed, line 98: This changes the current behavior. Previously we were still returning a `WP_Language` object in case of broken language, while now we return `null`. Is this something that we want?

src/language-factory.php
```diff
@@ -169,43 +152,55 @@ private function sanitize_data( array $data ) {
...
 		foreach ( $positive_fields as $field ) {
 			$data[ $field ] = ! empty( $data[ $field ] ) ? absint( $data[ $field ] ) : 0;
 		}
 
-		$data['active'] = isset( $data['active'] ) ? (bool) $data['active'] : true;
-
-		if ( array_key_exists( 'fallbacks', $data ) && ! is_array( $data['fallbacks'] ) ) {
-			unset( $data['fallbacks'] );
-		}
-
```
Screenfeed, line 185: We shouldn't remove these, this method is used by `PLL_Language_Factory::get()`, which is used by `PLL_Languages::get_list()` to build the `PLL_Language` object, based on the transient value.

If the language term's description needs some sanitization (which is done earlier now), then the transient's value too IMHO.

include/language-factory.php
```diff
@@ -169,43 +152,55 @@ private function sanitize_data( array $data ) {
...
 	 *
-	 * @return array[]
+	 * @since 3.8
 	 *
-	 * @phpstan-return array<string, array<string, string>>
+	 * @param string $locale WordPress locale.
+	 * @return string[]
 	 */
-	private function get_languages() {
+	private function get_additional_locales( $locale ) {
```
Screenfeed, line 175: Types are safe here.

```suggestion
	private function get_additional_locales( string $locale ): array {
```

PR #1798, opened 2026-01-15
Fix quick edit language dropdown for terms.

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/2873.

## What

Quick edit on taxonomy terms displays the wrong language in the dropdown when the term has no translations yet (shows default language instead of the term's current language).

## Why

PR #1754 refactored `PLL_Admin_Filters_Columns::term_column()`. 

However, this refactoring introduced an early `return` when a term has no translation for a given language.

This early return prevented the creation of the hidden field `<div id="lang_{term_id}">` which is required by the JS (`term.js`) to pre populate the language dropdown in the quick edit form.

## Fix

Move the hidden field creation to occur before the translation check, ensuring it's always present in the first language column regardless of whether translations exist.

### DIFF
--- a/admin/admin-filters-columns.php
+++ b/admin/admin-filters-columns.php
@@ -290,13 +290,6 @@ class PLL_Admin_Filters_Columns {
 			return $out;
 		}
 
-		$is_first_column = $this->get_first_language_column() === $column;
-
-		if ( $is_first_column ) {
-			// Hidden field containing the term language for quick edit.
-			$out .= sprintf( '<div class="hidden" id="lang_%d">%s</div>', $term->term_id, esc_html( $lang->slug ) );
-		}
-
 		$tr_id   = $this->model->term->get( $term->term_id, $language );
 		$tr_term = $tr_id ? get_term( $tr_id, $taxonomy ) : null;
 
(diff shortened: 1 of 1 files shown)


admin/admin-filters-columns.php
```diff
@@ -290,33 +290,38 @@ public function term_column( $out, $column, $term_id ) {
...
+			// Link to edit (or not) the term or a translation.
+			$out .= $this->links->get_edit_term_link_html( $tr_term, $post_type, $tr_term->term_id === $term->term_id ? 'list_current' : 'list_translation' );
 		}
 
-		// Link to edit (or not) the term or a translation.
-		$out .= $this->links->get_edit_term_link_html( $tr_term, $post_type, $tr_term->term_id === $term->term_id ? 'list_current' : 'list_translation' );
-
-		if ( $this->get_first_language_column() !== $column ) {
-			return $out;
+		if ( $is_first_column ) {
```
Chouby: Now that there is no more early return isn't it possible to keep the hidden field hare and have only one test?
```suggestion
		if ( $this->get_first_language_column() === $column )  {
			// Hidden field containing the term language for quick edit.
			$out .= sprintf( '<div class="hidden" id="lang_%d">%s</div>', $term->term_id, esc_html( $lang->slug ) );
			
```

Marianne380: Yep, see https://github.com/polylang/polylang/pull/1798/commits/a9a02856ba229ff6cf793a041c037c693a52d704.

PR #1799, opened 2026-01-19
Improve the message when activating a second instance of Polylang

Chouby:
When Polylang is already activated and we activate Polylang Pro, Polylang is automatically activated and the message "Plugin activated is displayed".

When Polylang Pro is already activated and we attempt to activate Polylang, the plugin is not activated. We used to display the message "Plugin deactivated" taken from the list of WordPress available messages.

This PR improves this message to display "You can't activate Polylang and Polylang Pro at the same time." (exact message to be discussed). This is now an error message (to be discussed if a warning is more appropriate).

### DIFF
--- a/polylang.php
+++ b/polylang.php
@@ -45,12 +45,29 @@ if ( defined( 'POLYLANG_VERSION' ) ) {
 
 	if ( is_plugin_active( plugin_basename( __FILE__ ) ) ) {
 		deactivate_plugins( plugin_basename( __FILE__ ) ); // Deactivate this plugin.
-		// WP does not allow us to send a custom meaningful message, so just tell the plugin has been deactivated.
-		wp_safe_redirect( add_query_arg( 'deactivate', 'true', remove_query_arg( 'activate' ) ) );
+		wp_safe_redirect( add_query_arg( 'pll_two_activations', 'true', remove_query_arg( 'activate' ) ) );
 		exit;
 	}
 	return;
 }
+
+add_action(
+	'admin_head',
+	function () {
+		if ( 'plugins.php' !== $GLOBALS['pagenow'] || ! isset( $_GET['pll_two_activations'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return;
+		}
+
+		$_SERVER['REQUEST_URI'] = remove_query_arg( 'pll_two_activations', $_SERVER['REQUEST_URI'] ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput
+		$updated_notice_args = array(
+			'id'                 => 'message',
+			'additional_classes' => array( 'error' ),
+			'dismissible'        => true,
+		);
+		wp_admin_notice( __( 'You can\'t activate Polylang and Polylang Pro at the same time.', 'polylang' ), $updated_notice_args );
+	}
+);
+
 
 // Go on loading the plugin.
 define( 'POLYLANG_VERSION', '3.8-dev' );
--- a/polylang.php
+++ b/polylang.php
@@ -54,7 +54,7 @@ if ( defined( 'POLYLANG_VERSION' ) ) {
 					'additional_classes' => array( 'error' ),
 					'dismissible'        => true,
 				);
-				wp_admin_notice( __( 'Polylang and Polylang Pro cannot be activated at the same time.', 'polylang' ), $updated_notice_args );
+				wp_admin_notice( __( 'Polylang hasn\'t been activated: Polylang and Polylang Pro cannot be activated at the same time.', 'polylang' ), $updated_notice_args );
 			}
 		);
 	}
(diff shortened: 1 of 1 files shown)


polylang.php
```diff
@@ -45,13 +45,30 @@
...
+		wp_safe_redirect( add_query_arg( 'pll_two_activations', 'true', remove_query_arg( 'activate' ) ) );
 		exit;
 	}
 	return;
 }
 
+add_action(
+	'admin_head',
+	function () {
+		if ( 'plugins.php' !== $GLOBALS['pagenow'] || ! isset( $_GET['pll_two_activations'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
Screenfeed: `static` keyword + to please PHPStan.

```suggestion
	static function () {
		if ( 'plugins.php' !== $GLOBALS['pagenow'] || ! isset( $_GET['pll_two_activations'] ) || ! is_string( $_SERVER['REQUEST_URI'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

Note: `$_SERVER['REQUEST_URI']` is "fixed" by WP with [`wp_fix_server_vars()`](https://developer.wordpress.org/reference/functions/wp_fix_server_vars/). However, in theory it could still be not a string (😏).

polylang.php
```diff
@@ -45,13 +45,30 @@
...
-		// WP does not allow us to send a custom meaningful message, so just tell the plugin has been deactivated.
-		wp_safe_redirect( add_query_arg( 'deactivate', 'true', remove_query_arg( 'activate' ) ) );
+		wp_safe_redirect( add_query_arg( 'pll_two_activations', 'true', remove_query_arg( 'activate' ) ) );
 		exit;
 	}
 	return;
 }
 
+add_action(
+	'admin_head',
```
Screenfeed: This will print the notice in the `<head>` tag.

```suggestion
	'all_admin_notices',
```

polylang.php
```diff
@@ -45,13 +45,30 @@
...
+	function () {
+		if ( 'plugins.php' !== $GLOBALS['pagenow'] || ! isset( $_GET['pll_two_activations'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return;
+		}
+
+		$_SERVER['REQUEST_URI'] = remove_query_arg( 'pll_two_activations', $_SERVER['REQUEST_URI'] ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput
```
Screenfeed: Do you think we should also `unset( $_GET['pll_two_activations'] );` ? 🤔

Screenfeed (changes requested):
I would love to see all of this moved to a dedicated file but I'm afraid it complicates things (`__FILE__` must be transmitted).

Maybe:

```php
require __DIR__ . '/whatever.php';

if ( ! pll_can_execute( __FILE__ ) ) {
	return;
}
```

What do you think?

polylang.php
Chouby: Check the new approach which doesn't use this anymore.

polylang.php
Chouby: I love to `load-plugins.php` which avoids one test. The position of the notice is the same as WP notices.

Chouby:
> I would love to see all of this moved to a dedicated file but I'm afraid it complicates things (`__FILE__` must be transmitted).
> 
> Maybe:
> 
> ```
> require __DIR__ . '/whatever.php';
> 
> if ( ! pll_can_execute( __FILE__ ) ) {
> 	return;
> }
> ```
> 
> What do you think?

I also believe it would complicate things.

polylang.php
```diff
@@ -41,19 +41,28 @@
...
+		add_action(
+			'load-plugins.php',
+			static function () {
+				unset( $_GET['activate'] ); // Prevent WP to display its 'Plugin activated.' message.
+				$updated_notice_args = array(
+					'id'                 => 'message',
+					'additional_classes' => array( 'error' ),
+					'dismissible'        => true,
+				);
+				wp_admin_notice( __( 'Polylang and Polylang Pro cannot be activated at the same time.', 'polylang' ), $updated_notice_args );
```
Screenfeed: What do you think about this?


```suggestion
				wp_admin_notice( __( 'Polylang has been deactivated: Polylang and Polylang Pro cannot be activated at the same time.', 'polylang' ), $updated_notice_args );
```

Screenfeed (changes requested):

polylang.php
Chouby: Following your idea, I suggest:
```suggestion
				wp_admin_notice( __( 'Polylang hasn't been activated: Polylang and Polylang Pro cannot be activated at the same time.', 'polylang' ), $updated_notice_args );
```
Indeed, the message is displayed when the user attempts to activate Polylang. Visually the user is not aware that it was activated during the execution of a few line of codes.

What do you think?

Screenfeed: Indeed. In that case, the `'` must be escaped.

```suggestion
				wp_admin_notice( __( 'Polylang hasn\'t been activated: Polylang and Polylang Pro cannot be activated at the same time.', 'polylang' ), $updated_notice_args );
```

Screenfeed (approved):

PR #1803, opened 2026-01-21
Allow to add/remove option sub-values.

Hug0-Drelon:
## What?
Allow to add/remove values from options.
Props https://github.com/polylang/polylang-pro/pull/2653/files#r2705692820

## Why?
Offer a clear interface to add/remove sub values from array-like options.

## How?
- Create `Options::add/remove()`
- Create `Abstract_Map`
- Implement sub-value reset in `Abstract_Map::remove()` (that's the main "feature")
- Cover with unit tests

### DIFF
--- a/include/Options/Options.php
+++ b/include/Options/Options.php
@@ -513,25 +513,9 @@ class Options implements ArrayAccess, IteratorAggregate {
 			);
 		}
 
-		if ( $this->options[ $this->current_blog_id ][ $key ] instanceof Abstract_List ) {
-			/** @var array $updated_value */
-			$updated_value   = $this->options[ $this->current_blog_id ][ $key ]->get();
-			$updated_value[] = $value;
-			$done            = $this->options[ $this->current_blog_id ][ $key ]->set(
-				$updated_value,
-				$this
-			);
-		} elseif ( $this->options[ $this->current_blog_id ][ $key ] instanceof Abstract_Map && is_array( $value ) ) {
-			/** @var array $old_value */
-			$old_value = $this->options[ $this->current_blog_id ][ $key ]->get();
-			$done      = $this->options[ $this->current_blog_id ][ $key ]->set(
-				array_merge(
-					$old_value,
-					$value
-				),
-				$this
-			);
-		} else {
+		$option = $this->options[ $this->current_blog_id ][ $key ];
+
+		if ( ! $option instanceof Abstract_List && ! $option instanceof Abstract_Map ) {
 			return new WP_Error(
 				'pll_invalid_option_type',
 				/* translators: %s is the name of an option. */
@@ -539,11 +523,16 @@ class Options implements ArrayAccess, IteratorAggregate {
 			);
 		}
 
-		if ( $done ) {
+		if ( $option->add( $value, $this ) ) {
 			$this->modified[ $this->current_blog_id ] = true;
+			return new WP_Error();
 		}
 
-		return $this->options[ $this->current_blog_id ][ $key ]->get_errors();
+		return new WP_Error(
+			'pll_add_failed',
+			/* translators: %1$s is the value to add. %2$s is the name of an option. */
+			sprintf( __( 'Failed to add %1$s to %2$s.', 'polylang' ), print_r( $value, true ), "'$key'" ) // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_print_r
+		);
 	}
 
 	/**
--- a/include/Options/Primitive/Abstract_Map.php
+++ b/include/Options/Primitive/Abstract_Map.php
@@ -5,6 +5,7 @@
 
 namespace WP_Syntex\Polylang\Options\Primitive;
 
+use WP_Syntex\Polylang\Options\Options;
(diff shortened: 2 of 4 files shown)


include/Options/Primitive/Abstract_Map.php
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	public function remove( string $key ): bool {
+		if ( ! array_key_exists( $key, $this->value ) ) {
+			return false;
+		}
+
+		$this->value = array_merge(
+			$this->value,
+			array( $key => $this->get_reset_value( $key ) )
+		);
```
Chouby: Isn't it simpler?
```suggestion
		$this->value[ $key ] = $this->get_reset_value( $key );
```

include/Options/Primitive/Abstract_Map.php
```diff
@@ -0,0 +1,80 @@
...
+
+	/**
+	 * Returns the reset value for a key.
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $key The key to reset.
+	 * @return mixed The reset value.
+	 */
+	abstract protected function get_reset_value( string $key );
```
Chouby: This is how I would have named this method. No worry if you prefer to keep `get...`
```suggestion
	abstract protected function reset_value( string $key );
```

include/Options/Options.php
```diff
@@ -452,6 +454,100 @@ public function reset( string $key ) {
...
+				'pll_invalid_option_type',
+				/* translators: %s is the name of an option. */
+				sprintf( __( 'Option %s is not a list or map.', 'polylang' ), "'$key'" )
+			);
+		}
+
+		if ( $done ) {
+			$this->modified[ $this->current_blog_id ] = true;
+			return new WP_Error();
+		}
```
Chouby: I suggest to simplify. The intermediate variable improves the readability. I suggest not to verify `is_string()`. If you need to do it, I suggest to delegate this to `Abstract_Map`:
```suggestion
		$option = $this->options[ $this->current_blog_id ][ $key ];
		
		if ( ! $option instanceof Abstract_List || ! $option instanceof Abstract_Map ) {
			return new WP_Error(
				'pll_invalid_option_type',
				/* translators: %s is the name of an option. */
				sprintf( __( 'Option %s is not a list or map.', 'polylang' ), "'$key'" )
			);
		}
		
		if ( $option->remove( $value ) ) {
			$this->modified[ $this->current_blog_id ] = true;
			return new WP_Error(); 
		}
```

include/Options/Options.php
```diff
@@ -452,6 +454,100 @@ public function reset( string $key ) {
...
+			return new WP_Error(
+				'pll_invalid_option_type',
+				/* translators: %s is the name of an option. */
+				sprintf( __( 'Option %s is not a list or map.', 'polylang' ), "'$key'" )
+			);
+		}
+
+		if ( $done ) {
+			$this->modified[ $this->current_blog_id ] = true;
+		}
```
Chouby: I suggest to simplify the same way I proposed for `remove()`. Introducing an `add()` method in list an map class  should help.

include/Options/Options.php
Hug0-Drelon: I've already tried and it doesn't work.
I think `ArrayAcces` or other array-like interface messes up PHP runtime because we get things like:
```php
$option = $this->options[ $this->current_blog_id ][ $key ]; // \Busines\Domain class
if ( $option instanceof Abstract_Map ) { // FALSE WTF?!
    ...
}
if ( is_a( $option, Abstract_Map::class ) { // FALSE WTF?!
    ...
}
```

include/Options/Options.php
Hug0-Drelon: I avoided to implement `Abstract_Map|Abstract_List::add()` because I'd have to pass `Options` as second argument to call `Abstract_Option::set()` properly (required for sanitization).

Knowing that, do you still want `Abstract_Map|Abstract_List::add( $value, Options $this )`?

include/Options/Primitive/Abstract_Map.php
Hug0-Drelon: Applied in 535aff2f.

include/Options/Primitive/Abstract_Map.php
Hug0-Drelon: Indeed, applied in 535aff2f.

include/Options/Options.php
Chouby: Found. The issue is coming from my test :/
```php
if ( ! $option instanceof Abstract_List || ! $option instanceof Abstract_Map ) { // Wrong
if ( ! $option instanceof Abstract_List && ! $option instanceof Abstract_Map ) { // OK
```

include/Options/Options.php
Chouby: It's the same as `set()` so it doesn't bother me.

Hug0-Drelon: Done in e4e51b3a.

include/Options/Options.php
Hug0-Drelon: Done in 39b9a1707500e2a3bdcd9926bb27031cc34c899a.

PR #1804, opened 2026-01-22
NOOP user language capabilities.

Hug0-Drelon:
## What?
Polylang core has feature ready code for language capabilities, but the way it is added makes it visible to end users (if he/she has `translate_*` capabilities).

## How?
- Create `Capabilities::get_user()` method to retrieve the current user decorator, suitable for capabilities check.
- Create `User_Interface` for third-parties contract.
- Create `NOOP_User`, used by default in Polylang core.
- User `Capabilities::get_user()` instead of `new User()` everywhere in the code.
- Add PHPUnit tests to cover new classes.

> [!NOTE]
> This proposal is inspired by the prototype pattern.
> It allows to easily override the user decorator at runtime.
> And bonus point, the implementation allows for better user instance caching.

### DIFF
--- a/tests/phpunit/tests/Capabilities/test-Get_User.php
+++ b/tests/phpunit/tests/Capabilities/test-Get_User.php
@@ -49,7 +49,7 @@ class Test_Get_User extends PLL_UnitTestCase {
 		$this->assertSame( $user_id, $user_1->get_id() );
 	}
 
-	public function test_get_user_prototype_pattern_with_user_switching() {
+	public function test_get_user_with_user_switching() {
 		$user_id_1 = self::factory()->user->create();
 		$user_id_2 = self::factory()->user->create();
 
@@ -79,9 +79,9 @@ class Test_Get_User extends PLL_UnitTestCase {
 		$user_id = self::factory()->user->create();
 		wp_set_current_user( $user_id );
 
-		// Set a custom prototype.
-		$prototype_stub = $this->createStub( Creator_Interface::class );
-		Capabilities::set_user_creator( $prototype_stub );
+		// Set a custom creator.
+		$creator_stub = $this->createStub( Creator_Interface::class );
+		Capabilities::set_user_creator( $creator_stub );
 
 		$user = Capabilities::get_user();
 
@@ -92,12 +92,12 @@ class Test_Get_User extends PLL_UnitTestCase {
 		$user_id = self::factory()->user->create();
 		wp_set_current_user( $user_id );
 
-		// Set a custom prototype.
-		$custom_prototype = new Creator();
-		Capabilities::set_user_creator( $custom_prototype );
+		// Set a custom creator.
+		$custom_creator = new Creator();
+		Capabilities::set_user_creator( $custom_creator );
 
-		$user_1 = $custom_prototype->get( wp_get_current_user() );
-		$user_2 = $custom_prototype->get( wp_get_current_user() );
+		$user_1 = $custom_creator->get( wp_get_current_user() );
+		$user_2 = $custom_creator->get( wp_get_current_user() );
 
 		$this->assertSame( $user_id, $user_1->get_id() );
 		$this->assertSame( $user_id, $user_2->get_id() );
--- a/tests/phpunit/tests/Capabilities/test-NOOP_User.php
+++ b/tests/phpunit/tests/Capabilities/test-NOOP_User.php
@@ -202,10 +202,10 @@ class Test_NOOP extends PLL_UnitTestCase {
 
 		$language = $this->createMock( PLL_Language::class );
 
-		// This should not throw any exception or call wp_die()
+		// This should not throw any exception or call wp_die().
 		$user->can_translate_or_die( $language );
 
-		// If we reach this point, the test passes
+		// If we reach this point, the test passes.
 		$this->assertTrue( true );
 	}
 
@@ -217,29 +217,29 @@ class Test_NOOP extends PLL_UnitTestCase {
 	public function test_can_translate_or_die_with_real_language( $user_type ) {
 		$user = $this->get_user( $user_type );
 
-		// This should not throw any exception or call wp_die()
+		// This should not throw any exception or call wp_die().
 		$user->can_translate_or_die( self::$language_fr );
 
-		// If we reach this point, the test passes
+		// If we reach this point, the test passes.
 		$this->assertTrue( true );
 	}
 
 	public function test_creator_returns_self_for_same_user() {
-		$prototype = new Creator();
+		$creator = new Creator();
 		$wp_user     = self::factory()->user->create_and_get();
-		$user        = $prototype->get( $wp_user );
-		$creatord_user = $prototype->get( $wp_user );
+		$user        = $creator->get( $wp_user );
+		$creatord_user = $creator->get( $wp_user );
 
 		$this->assertSame( $user->get_id(), $creatord_user->get_id() );
 	}
 
 	public function test_creator_returns_new_instance_for_different_user() {
-		$prototype = new Creator();
+		$creator = new Creator();
 		$wp_user_1 = self::factory()->user->create_and_get();
 		$wp_user_2 = self::factory()->user->create_and_get();
 
-		$user_1 = $prototype->get( $wp_user_1 );
-		$user_2 = $prototype->get( $wp_user_2 );
+		$user_1 = $creator->get( $wp_user_1 );
+		$user_2 = $creator->get( $wp_user_2 );
 
 		$this->assertNotSame( $user_1->get_id(), $user_2->get_id() );
 		$this->assertInstanceOf( NOOP::class, $user_2 );
@@ -250,11 +250,11 @@ class Test_NOOP extends PLL_UnitTestCase {
 		$wp_user_1 = self::factory()->user->create_and_get();
 		$wp_user_2 = self::factory()->user->create_and_get();
 
-		$prototype = new Creator();
-		$user_1 = $prototype->get( $wp_user_1 );
-		$user_2 = $prototype->get( $wp_user_2 );
+		$creator = new Creator();
+		$user_1 = $creator->get( $wp_user_1 );
+		$user_2 = $creator->get( $wp_user_2 );
 
-		// Both should be NOOP instances
+		// Both should be NOOP instances.
 		$this->assertInstanceOf( NOOP::class, $user_1 );
 		$this->assertInstanceOf( NOOP::class, $user_2 );
 	}


include/Capabilities/User/User_Interface.php
```diff
@@ -0,0 +1,98 @@
...
+
+use WP_User;
+use PLL_Language;
+
+/**
+ * An interface for user with translation management feature.
+ *
+ * @since 3.8
+ */
+interface User_Interface {
```
Screenfeed, line 17: `NOOP_User` and `Translator` have a lot of code in common (the `$user` property, constructor, `clone()`, `get_id()`, `has_cap()`), don't you think we should have an abstract class instead of an interface? Yes that means we impose some things like the property and the constructor but, practically, it makes sense IMHO.

tests/phpunit/tests/Capabilities/test-Get_User.php
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Reset the user_prototype static property in the Capabilities class.
+	 *
+	 * @return void
+	 */
+	private function reset_prototype() {
+		$reflection = new \ReflectionClass( Capabilities::class );
+		$property   = $reflection->getProperty( 'user_prototype' );
+		version_compare( PHP_VERSION, '8.1', '<' ) && $property->setAccessible( true );
+		$property->setValue( null, null );
```
Screenfeed: We could `use WP_Syntex\Polylang_Phpunit\TestCaseTrait;` and then:

```suggestion
		$this->resetPropertyValue( Capabilities::class, 'user_prototype' );
```

tests/phpunit/tests/Capabilities/test-Get_User.php
```diff
@@ -0,0 +1,144 @@
...
+		$filter_call_count = 0;
+
+		add_filter(
+			'pll_user_prototype',
+			function ( $user ) use ( &$filter_call_count ) {
+				$filter_call_count++;
+
+				return $user;
+			}
+		);
```
Screenfeed: Tool idea that could go in our PHPUnit package (for another PR):

```php
trait Hook_Counter {
	protected array $hooks = array();

	protected function expect_hook_count( string $hook, int $expected_count ): void {
		if ( ! isset( $this->hooks[ $hook ] ) {
			$this->hooks[ $hook ] = array( 'expected' => $expected_count, 'count' => 0 );
		}

		add_action(
			$hook,
			function ( $v ) use ( $hook ) {
				$this->hooks[ $hook ]['count']++;
				return $v;
			}
		);
	}

	public function assert_hook_count() {
		foreach ( $this->hooks as $hook => $counts ) {
			$this->assertSame( $counts['expected'], $counts['count'], "Expected the hook '{$hook}' to be fired {$counts['expected']} time(s)." );
		}
	}
}
```

tests/phpunit/tests/Capabilities/test-NOOP_User.php
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	/**
+	 * @dataProvider user_provider
+	 *
+	 * @param string $user_type User type to test.
+	 */
+	public function test_get_id_returns_user_id( $user_type ) {
+		$wp_user = 'administrator' === $user_type ? self::$administrator : self::$translator_fr;
+		$user    = new NOOP_User( $wp_user );
```
Screenfeed: We could factorize, but I let you decide, no big deal.

```php
private function get_user( string $user_type ): NOOP_User {
	$wp_user = 'administrator' === $user_type ? self::$administrator : self::$translator_fr;
	return new NOOP_User( $wp_user );
}
```

tests/phpunit/tests/Capabilities/test-NOOP_User.php
```diff
@@ -0,0 +1,281 @@
...
+	/**
+	 * @dataProvider user_provider
+	 *
+	 * @param string $user_type User type to test.
+	 */
+	public function test_can_translate_always_returns_true( $user_type ) {
+		$wp_user = 'administrator' === $user_type ? self::$administrator : self::$translator_fr;
+		$user    = new NOOP_User( $wp_user );
+
+		$language = $this->createMock( PLL_Language::class );
```
Screenfeed, line 77: Why testing a mock? For testing an unknown language? 🤔

Hug0-Drelon: Here we don't tests the mock (maybe I should have used a stub instead).
We are testing `NOOP_User::can_translate()`, which should **always** return `true` no matter what language it gets.

tests/phpunit/tests/Capabilities/test-Get_User.php
Hug0-Drelon: I'm not against but in a separate PR:
- `WP_Syntex\Polylang_Phpunit\TestCaseTrait::resetPropertyValue` doesn't seem to manage backward compatibility yet.
- Using `version_compare() ...` is already done everywhere in the test suite, so it should be addressed in one shot (and in our other repos too).
- This PR is big enough IMHO

include/Capabilities/User/User_Interface.php
Hug0-Drelon: I foresaw that but kept full implementation so both reviews are easier.
Also, it didn't bothered me that much since the duplication is for low-level code and not business logic per se.
But why not, once both PRs are OK if that suits you.

tests/phpunit/tests/Capabilities/test-NOOP_User.php
Hug0-Drelon: Applied in 06c18aa4.

tests/phpunit/tests/Capabilities/test-Get_User.php
Screenfeed: The backward compatibility with php < 8.1 [is handled](https://github.com/polylang/wp-phpunit/blob/ff81a590c0aa4d168ab64073376826b93bf5d328/UnitTests/TestCaseTrait.php#L287) but OK to do this in a dedicated PR if you prefer.

Hug0-Drelon: Indeed, I missed that.

tests/phpunit/tests/Capabilities/test-NOOP_User.php
Screenfeed: Yes that's what I was saying 👍
By "testing a mock" I meant "using a mock" or "testing with a mock", sorry, bad wording.

include/Capabilities/Capabilities.php
```diff
@@ -16,6 +19,13 @@ class Capabilities {
 	public const LANGUAGES    = 'manage_languages';
 	public const TRANSLATIONS = 'manage_translations';
 
+	/**
+	 * The user prototype to be used for capability checks.
+	 *
+	 * @var Prototype_Interface|null
+	 */
+	private static ?Prototype_Interface $user_prototype = null;
```
Chouby: I propose to use _creator_ instead of _prototype_ as vocabulary (because it's more self-explanatory).
What do you think?
```suggestion
	 * The user creator to be used for capability checks.
	 *
	 * @var Creator_Interface|null
	 */
	private static ?Creator_Interface $creator = null;
```

include/Capabilities/Capabilities.php
```diff
@@ -42,4 +52,34 @@ public function map_custom_caps( $caps, $cap ) {
...
+	 *
+	 * @param WP_User|null $user The user to decorate. If null, the current user is used.
+	 * @return User_Interface The user instance.
+	 */
+	public static function get_user( ?WP_User $user = null ): User_Interface {
+		if ( ! self::$user_prototype ) {
+			self::$user_prototype = new Prototype();
+		}
+
+		return self::$user_prototype->get( $user ?? wp_get_current_user() );
```
Chouby:
```suggestion
		if ( ! self::$creator ) {
			self::$creator = new Creator();
		}

		return self::$creator->get( $user ?? wp_get_current_user() );
```

include/Capabilities/Capabilities.php
```diff
@@ -42,4 +52,34 @@ public function map_custom_caps( $caps, $cap ) {
...
+	 * That way, the prototype doesn't depend on `WP_User`.
+	 *
+	 * @since 3.8
+	 *
+	 * @param Prototype_Interface $user_prototype The user prototype to be used for capability checks.
+	 * @return void
+	 */
+	public static function set_user_prototype( Prototype_Interface $user_prototype ): void {
+		self::$user_prototype = $user_prototype;
+	}
```
Chouby:
```suggestion
	 * Sets the user creator to be used for capability checks.
	 *
	 * Having a separate class to create the decorated user allows for better decoupling.
	 * This allows to set a creator object without dependence to a `WP_User`.
	 *
	 * @since 3.8
	 *
	 * @param Creator_Interface $creator The user creator to be used for capability checks.
	 * @return void
	 */
	public static function set_user_creator( Creator_Interface $creator ): void {
		self::$creator = $creator
	}
```

include/Capabilities/Create/Abstract_Object.php
```diff
@@ -58,9 +58,9 @@ public function __construct( PLL_Model $model, Request $request, ?PLL_Language $
...
 	 * @since 3.8
 	 *
-	 * @param User    $user The user object.
-	 * @param integer $id   The object ID.
+	 * @param User_Interface $user The user object.
+	 * @param int            $id   The object ID.
 	 * @return PLL_Language The language defined from the global context.
 	 */
-	abstract public function get_language( User $user, int $id = 0 ): PLL_Language;
+	abstract public function get_language( User_Interface $user, int $id = 0 ): PLL_Language;
```
Chouby: Finally, to inject or not to inject ? That's the question.

include/Capabilities/User/Prototype.php
```diff
@@ -0,0 +1,41 @@
...
+
+use WP_User;
+
+/**
+ * A class for user prototypes.
+ * Always returns a `NOOP` instance so capabilities features are disabled by default.
+ *
+ * @since 3.8
+ */
+class Prototype implements Prototype_Interface {
```
Chouby:
```suggestion
 * A class to create a decorated user.
 * Always returns a `NOOP` instance so capabilities features are disabled by default.
 *
 * @since 3.8
 */
class Creator implements Creator_Interface {
```

include/Capabilities/User/Prototype_Interface.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @package Polylang
+ */
+
+namespace WP_Syntex\Polylang\Capabilities\User;
+
+use WP_User;
+
+/**
+ * An interface for user prototypes.
```
Chouby:
```suggestion
 * An interface for user creators.
```

include/Capabilities/User/Prototype_Interface.php
```diff
@@ -0,0 +1,26 @@
...
+ * Allows to change the user decoration behavior at runtime.
+ *
+ * @since 3.8
+ */
+interface Prototype_Interface {
+	/**
+	 * Returns the user instance to be used for capability checks using prototype pattern.
```
Chouby:
```suggestion
interface Creator_Interface {
	/**
	 * Returns the user instance to be used for capability checks.
```

include/Capabilities/Capabilities.php
```diff
@@ -5,7 +5,10 @@
 
 namespace WP_Syntex\Polylang\Capabilities;
 
-defined( 'ABSPATH' ) || exit;
+use WP_User;
+use WP_Syntex\Polylang\Capabilities\User\Prototype;
+use WP_Syntex\Polylang\Capabilities\User\User_Interface;
+use WP_Syntex\Polylang\Capabilities\User\Prototype_Interface;
```
Chouby:
```suggestion
use WP_User;
use WP_Syntex\Polylang\Capabilities\User\Creator;
use WP_Syntex\Polylang\Capabilities\User\User_Interface;
use WP_Syntex\Polylang\Capabilities\User\Creator_Interface;
```

include/Capabilities/Capabilities.php
```diff
@@ -42,4 +52,34 @@ public function map_custom_caps( $caps, $cap ) {
 
 		return $caps;
 	}
+
+	/**
+	 * Returns the user instance to be used for capability checks using prototype pattern.
```
Chouby:
```suggestion
	 * Returns the user instance to be used for capability checks.
```

tests/phpunit/tests/Capabilities/test-NOOP_User.php
```diff
@@ -0,0 +1,285 @@
...
+		$user = $this->get_user( $user_type );
+
+		// This should not throw any exception or call wp_die()
+		$user->can_translate_or_die( self::$language_fr );
+
+		// If we reach this point, the test passes
+		$this->assertTrue( true );
+	}
+
+	public function test_clone_returns_self_for_same_user() {
```
Chouby: I would also remove references to "clone" as we don't clone anything.

include/Capabilities/Create/Abstract_Object.php
Hug0-Drelon: Let's do it rn, see 9dae8574b199961ab6865024feda4ab1930f4ec5

tests/phpunit/tests/Capabilities/test-NOOP_User.php
Hug0-Drelon: Done in e3a615067bc2ae9377b558f8c1b17d4a6236d0cc.

PR #1806, opened 2026-01-28
Refactor strings management.

Hug0-Drelon:
## What?
Goes a bit further than https://github.com/polylang/polylang/pull/1782.
Bonus: allow string translation per context.

## Why?
As of today, strings are managed and saved in `table-string.php`. No encapsulation and no separation of concern what so ever.
Also, sanitization is possible only through `pll_sanitize_string_translation` hook. The need of a unified and encapsulated way of sanitizing strings emerges now.
It will open the gate to proper string management through different channels (CLI, REST...).

## How?
- New DDD inspired architecture for strings using repository pattern.
    -  `WP_Syntex\Polylang\Strings\Translatable`: base entity
    - `WP_Syntex\Polylang\Strings\Collection`: aggregate
    -  `WP_Syntex\Polylang\Strings\Database_Repository`: database access layer
    - `WP_Syntex\Polylang\Strings\Query`: query builder for strings
    
Usability and readability can be mostly judged in `settings/table-string.php`.
    
## Exemple:
For the most part, here how to use it:
```php
// Read.
$repository                 = new Database_Repository( $languages_model );
$all_strings_collection     = $repository->find_all();
$queried_strings_collection = $repository->query()
	->by_context( 'some context' )
	->by_fragment( 'fuzzy search' )
	->order_by( 'name', 'desc' )
	->paginate( 10, 1 )
	->get();

// Write.
$some_collection->get( 'some_string_id_hash')
	->set_translation( $language, 'some translation' );
$repository->save( $all_strings_collection );
```


src/settings/table-string.php
```diff
@@ -60,15 +55,15 @@ public function __construct( Languages $languages ) {
...
+		$this->items      = array(); // `WP_List_Table::$items` doesn't have a default value, let's be explicit.
+		$this->languages  = $languages;
+		$this->repository = new Database_Repository( $languages );
 
 		$this->selected_group = -1;
 
 		if ( ! empty( $_GET['group'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
 			$group = sanitize_text_field( wp_unslash( $_GET['group'] ) ); // phpcs:ignore WordPress.Security.NonceVerification
-			if ( in_array( $group, $this->groups ) ) {
+			if ( in_array( $group, $this->repository->get_groups() ) ) {
```
Screenfeed: Let's be strict!


```suggestion
			if ( in_array( $group, $this->repository->get_groups(), true ) ) {
```

src/settings/table-string.php
```diff
@@ -258,42 +253,25 @@ public function prepare_items() {
...
-			}
-		}
-
-		// Sorting.
-		uasort( $data, array( $this, 'usort_reorder' ) );
-
-		// Paging.
-		$per_page = $this->get_items_per_page( 'pll_strings_per_page' );
+		$per_page              = $this->get_items_per_page( 'pll_strings_per_page' );
+		$paged                 = ! empty( $_GET['paged'] ) && is_numeric( $_GET['paged'] ) ? absint( $_GET['paged'] ) : 1; // phpcs:ignore WordPress.Security.NonceVerification
```
Screenfeed: Any reason not to use `$this->get_pagenum()`?

src/settings/table-string.php
```diff
@@ -258,42 +253,25 @@ public function prepare_items() {
...
 		$this->_column_headers = array( $this->get_columns(), array(), $this->get_sortable_columns() );
 
-		$total_items = count( $data );
-		$this->items = array_slice( $data, ( $this->get_pagenum() - 1 ) * $per_page, $per_page, true );
+		$items = $this->repository->query()
+			->by_context( -1 !== $this->selected_group && is_string( $this->selected_group ) ? $this->selected_group : null )
```
Screenfeed: Those 2 checks exclude themselves: if it's `-1` it can't be a string, and vice-versa. I think this is enough:


```suggestion
			->by_context( is_string( $this->selected_group ) ? $this->selected_group : null )
```

src/settings/table-string.php
```diff
@@ -396,51 +376,38 @@ public function save_translations() {
...
+					},
+					array_keys( $translations ),
+					array_values( $translations )
+				);
+			}
 
-				// Clean database (removes all strings which were registered some day but are no more).
-				if ( ! empty( $_POST['clean'] ) && current_user_can( 'manage_options' ) ) {
-					$new_mo = new PLL_MO();
+			// Save all languages at once
```
Screenfeed: Can't see your point.

```suggestion
			// Save all languages at once.
```

src/settings/table-string.php
```diff
@@ -455,9 +422,12 @@ public function save_translations() {
...
-			foreach ( array_map( 'sanitize_key', $_POST['strings'] ) as $key ) {
-				icl_unregister_string( $this->strings[ $key ]['context'], $this->strings[ $key ]['name'] );
-			}
+			array_map(
+				function ( $key ) {
+					$this->repository->remove_wpml_string( sanitize_key( $key ) );
+				},
+				$_POST['strings'] // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			);
 		}
```
Screenfeed, line 434: This `function_exists( 'icl_unregister_string' )` bugs me, it shouldn't be outside the repository.

Idea: instead of a method that removes 1 string, have a method that removes a list of strings.

```php
if ( $this->current_action() === 'delete' && ! empty( $_POST['strings'] ) && current_user_can( 'manage_options' ) ) {
	$this->repository->remove_wpml_strings( array_map( 'sanitize_key', $_POST['strings'] ) ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
}
```

This way, `function_exists( 'icl_unregister_string' )` can be moved in the method.

src/Strings/Database_Repository.php
```diff
@@ -0,0 +1,308 @@
...
+	 * @since 3.8
+	 *
+	 * @param string $id The identifier.
+	 * @return void
+	 */
+	public function remove_wpml_string( string $id ): void {
+		$string_to_remove = array_find(
+			self::get_strings(),
+			static function ( $string_data ) use ( $id ) {
+				return md5( $string_data['string'] . $string_data['context'] ) === $id && function_exists( 'icl_unregister_string' );
```
Screenfeed: With my previous suggestion about this method being able to remove several strings instead of one, this `function_exists( 'icl_unregister_string' )` will be called only once before the loop. Also, this `array_find()` may be more efficient: `in_array( md5( $string_data['string'] . $string_data['context'] ), $ids, true )`.

src/Strings/Collection.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @param Translatable[] $translatables Initial translatables to add to the collection.
+	 */
+	public function __construct( array $translatables = array() ) {
+		array_map(
+			function ( Translatable $translatable ) {
+				$this->add( $translatable );
+			},
+			$translatables
+		);
```
Screenfeed: Shorter:

```suggestion
		array_map( array( $this, 'add' ), $translatables );
```

src/Strings/Collection.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @param string $id The identifier.
+	 * @return void
+	 */
+	public function remove( string $id ): void {
+		unset( $this->translatables[ $id ] );
+	}
+
+	/**
+	 * Gets a translatable by ID.
```
Screenfeed, line 70: Several `Gets` in this class.

```suggestion
	 * Returns a translatable by ID.
```

src/Strings/Collection.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @param string $id The identifier.
+	 * @return bool
+	 */
+	public function has( string $id ): bool {
+		return isset( $this->translatables[ $id ] );
+	}
+
+	/**
+	 * Gets the count of translatables in the collection.
```
Screenfeed:
```suggestion
	 * Returns the count of translatables in the collection.
```

src/Strings/Collection.php
```diff
@@ -0,0 +1,142 @@
...
+
+	/**
+	 * Gets an iterator for the translatables.
+	 *
+	 * @since 3.8
+	 *
+	 * @return \ArrayIterator<string, Translatable>
+	 */
+	public function getIterator(): \ArrayIterator {
+		return new ArrayIterator( $this->translatables );
```
Screenfeed:
```suggestion
	 * @return ArrayIterator<string, Translatable>
	 */
	public function getIterator(): ArrayIterator {
		return new ArrayIterator( $this->translatables );
```

src/Strings/Database_Repository.php
```diff
@@ -0,0 +1,308 @@
...
+				 * @param string $context     The context as defined in pll_register_string.
+				 * @param string $original    The original string to translate.
+				 * @param string $previous    The previous translation if any.
+				 */
+				$sanitized_translation = apply_filters(
+					'pll_sanitize_string_translation',
+					$translatable->get_translation( $language ),
+					$translatable->get_name(),
+					$translatable->get_context(),
+					$mo->translate_if_any( $translatable->get_source(), $translatable->get_context() ),
```
Screenfeed: This is supposed to be the original string.

```suggestion
					$translatable->get_source(),
```

src/Strings/Database_Repository.php
```diff
@@ -0,0 +1,308 @@
...
+	 *
+	 * @since 3.8
+	 */
+	public static function reset(): void {
+		self::$registered_strings = array();
+		self::$read               = false;
+	}
+
+	/**
+	 * Gets the groups of registered strings.
```
Screenfeed, line 202: Several `Gets` in this class too.

```suggestion
	 * Returns the groups of registered strings.
```

src/Strings/Database_Repository.php
```diff
@@ -0,0 +1,308 @@
...
+		}
+
+		$default_strings = array(
+			'widget_title' => __( 'Widget title', 'polylang' ),
+			'widget_text'  => __( 'Widget text', 'polylang' ),
+		);
+
+		global $wp_registered_widgets;
+		$sidebars = wp_get_sidebars_widgets();
+		foreach ( $sidebars as $sidebar => $widgets ) {
```
Screenfeed: No need of the intermediate var.

```suggestion
		foreach ( wp_get_sidebars_widgets() as $sidebar => $widgets ) {
```

src/Strings/Database_Repository.php
```diff
@@ -0,0 +1,308 @@
...
+			if ( 'wp_inactive_widgets' == $sidebar || empty( $widgets ) ) {
```
Screenfeed: Strict comparison, and the `empty( $widgets )` check is useless. (Yes I know it's copied from old code)

```suggestion
			if ( 'wp_inactive_widgets' === $sidebar ) {
```

src/Strings/Database_Repository.php
```diff
@@ -0,0 +1,308 @@
...
+			}
+		}
+
+		/**
+		 * Filter the list of strings registered for translation
+		 * Mainly for use by our PLL_WPML_Compat class
+		 *
+		 * @since 1.0.2
+		 *
+		 * @param array $strings list of strings
```
Screenfeed:
```suggestion
		 * Filter the list of strings registered for translation.
		 * Mainly for use by our `PLL_WPML_Compat` class.
		 *
		 * @since 1.0.2
		 *
		 * @param array $strings List of strings.
```

src/Strings/Query.php
```diff
@@ -0,0 +1,292 @@
...
+
+		if ( $this->fragment ) {
+			$translatables = $this->filter_by_fragment( $translatables );
+		}
+
+		if ( 'none' !== $this->order_by ) {
+			$translatables = $this->sort( $translatables );
+		}
+
+		// Store total count before pagination
```
Screenfeed:
```suggestion
		// Store total count before pagination.
```

src/Strings/Query.php
```diff
@@ -0,0 +1,292 @@
...
+		$total_count = count( $translatables );
+
+		if ( $this->per_page ) {
```
Screenfeed, line 217: So instead of having `$per_page` being `null|int<1, max>` we could do `int<0, max>`.

Then:

```php
	/**
	 * Number of items per page (0 = no pagination).
	 *
	 * @var int
	 */
	private int $per_page = 0;

	public function paginate( int $per_page, int $page = 1 ): self {
		$this->per_page = max( 0, $per_page ); // <== 0 instead of 1.
		$this->page     = max( 1, $page );

		return $this;
	}
```

What do you think?

src/Strings/Query.php
```diff
@@ -0,0 +1,292 @@
...
+			function ( Translatable $translatable ) use ( $fragment ) {
+				if (
+					false !== stripos( $translatable->get_source(), $fragment ) ||
+					false !== stripos( $translatable->get_name(), $fragment )
+				) {
+					return true;
+				}
+
+				foreach ( $this->languages->get_list() as $language ) {
+					if ( false !== stripos( $translatable->get_translation( $language ), $fragment ) ) {
```
Screenfeed, line 298: I think we can do better:

```suggestion
		$fragment  = strtolower( $this->fragment );
		$languages = $this->languages->get_list();
		return array_filter(
			$translatables,
			static function ( Translatable $translatable ) use ( string $fragment, array $languages ) {
				if (
					false !== strpos( strtolower( $translatable->get_source() ), $fragment ) ||
					false !== strpos( strtolower( $translatable->get_name() ), $fragment )
				) {
					return true;
				}

				foreach ( $languages as $language ) {
					if ( false !== strpos( strtolower( $translatable->get_translation( $language ) ), $fragment ) ) {
```

That would be better IMHO because it will prevent to call `$this->languages->get_list()` on each row and it makes the callback static.
Also, maybe to be confirmed, but years ago, `stripos()` was slower than `strpos()` (I expect it to still be true), so `strpos( strtolower() )` may be faster.

src/Strings/Translatable.php
```diff
@@ -0,0 +1,276 @@
...
+		$this->name              = $name;
+		$this->context           = $context;
+		$this->multiline         = $multiline;
+		$this->sanitize_callback = $sanitize_callback ?? Closure::fromCallable( array( $this, 'default_sanitization' ) );
+
+		add_filter( 'pll_sanitize_string_translation', array( $this, 'sanitize' ), 10, 5 );
+	}
+
+	/**
+	 * Gets the unique identifier.
```
Screenfeed, line 99: Several `Gets` in this class too x)

src/Strings/Translatable.php
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * @since 3.8
+ */
+class Translatable {
+	/**
+	 * The unique identifier for the string.
+	 *
+	 * @var string
+	 */
+	private $id;
```
Screenfeed:
```suggestion
	private string $id;
```

src/Strings/Translatable.php
```diff
@@ -0,0 +1,276 @@
...
+	 * @since 3.8
+	 *
+	 * @return string
+	 */
+	public function get_name(): string {
+		return $this->name;
+	}
+
+	/**
+	 * Gets the translation for a specific language.
```
Screenfeed, line 121: I think it's worth noting:

```suggestion
	 * Returns the translation for a specific language.
	 * Returns the source string if the translation doesn't exist.
```

src/Strings/Translatable.php
```diff
@@ -0,0 +1,276 @@
...
+	/**
+	 * Checks if the string matches the given name and context.
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $name    The name to compare.
+	 * @param string $context The context to compare.
+	 * @return bool Whether the strings match.
+	 */
+	private function is_matching( string $name, string $context ): bool {
```
Screenfeed, line 260: It's used only once so it can be removed.

src/api.php
```diff
@@ -179,16 +181,17 @@ function pll_home_url( $lang = '' ) {
  * @api
  * @since 0.6
```
Screenfeed, line 182:
```suggestion
 * @since 0.6
 * @since 3.9 Parameter `$sanitize_callback` added.
```

src/mo.php
```diff
@@ -83,15 +94,28 @@ public function import_from_db( $lang ) {
 		}
 
 		foreach ( $strings as $msg ) {
+			if ( ! is_array( $msg ) ) {
```
Screenfeed: Just to be sure because... you know.

```suggestion
			if ( ! is_array( $msg ) || ! isset( $msg[0], $msg[1] ) ) {
```

src/mo.php
```diff
@@ -115,11 +139,17 @@ public function delete_entry( $string ) {
 	 *
 	 * @since 3.7
```
Screenfeed, line 221:
```suggestion
	 * @since 3.7
	 * @since 3.9 Parameter `$context` added.
```

src/mo.php
```diff
@@ -115,11 +139,17 @@ public function delete_entry( $string ) {
...
-	 * @param string $source The source string to translate.
+	 * @param string      $source  The source string to translate.
+	 * @param string|null $context Optional. The context to translate.
 	 * @return string The translated string or empty string if not found.
 	 */
-	public function translate_if_any( string $source ) {
-		$entry = new Translation_Entry( array( 'singular' => $source ) );
+	public function translate_if_any( string $source, ?string $context = null ) {
+		$entry_args = array( 'singular' => $source );
+		if ( $context ) {
```
Screenfeed: If we don't differentiate `null` and an empty string (or even `'0'`), maybe we should simply type `$context` as `string` instead of `string|null`. What do you think?

tests/phpunit/tests/plugins/test-wp-importer.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use WP_Syntex\Polylang\Strings\Database_Repository;
+
```
Screenfeed: 🤔 

```suggestion
```

tests/phpunit/tests/Strings/Table/test-Prepare_Items.php
```diff
@@ -0,0 +1,376 @@
...
+
+	public function test_prepare_items_combined_group_and_search() {
+		$this->register_test_strings();
+		$_GET['group'] = 'Animals';
+		$_GET['s']     = 'First';
+
+		$table = new PLL_Table_String( $this->pll_env->model->languages );
+		$table->prepare_items();
+
+		$this->assertCount( 1, $table->items, 'Only 1 string should match both filters.' );
```
Screenfeed: This would also work without the group, since there is only 1 string containing `First`.
Idea:

```php
Database_Repository::register( 'alpha', 'First string foobar', 'Animals', null, false );
Database_Repository::register( 'beta', 'Second string', 'Animals', null, false );
Database_Repository::register( 'gamma', 'Third string foobar', 'Plants', null, false );
```

Then search for `foobar`.

tests/phpunit/tests/Strings/Table/test-Save.php
```diff
@@ -0,0 +1,451 @@
...
+	}
+
+	/**
+	 * Creates a table instance with registered strings.
+	 *
+	 * @return PLL_Table_String
+	 */
+	private function create_table_instance(): PLL_Table_String {
+		$table = new PLL_Table_String( $this->pll_env->model->languages );
+		return $table;
```
Screenfeed: No need for intermediate var.

```suggestion
		return new PLL_Table_String( $this->pll_env->model->languages );
```

tests/phpunit/tests/Strings/Table/test-Save.php
```diff
@@ -0,0 +1,451 @@
...
+		$mo = new PLL_MO();
+		$mo->import_from_db( $fr_language );
+
+		// Empty translations might be saved as empty string or entry might be removed.
+		// The behavior depends on PLL_MO implementation.
+		if ( isset( $mo->entries['Foo'] ) ) {
+			$this->assertSame( '', $mo->translate( 'Foo', 'Group1' ), 'Empty translation should be saved as empty string.' );
+		} else {
+			$this->assertEmpty( $mo->entries, 'Empty translation entry should be removed.' );
+		}
```
Screenfeed: This behavior changed [in 3.4](https://github.com/polylang/polylang/pull/1273), so I don't think this is useful.


```suggestion
		$this->assertEmpty( $mo->entries, 'Empty translation entry should be removed.' );
```

src/Strings/Database_Repository.php
Screenfeed: Also I don't think doing `md5( $string_data['string'] . $string_data['context'] )` is a good idea. For example the following values will collide:

```php
array(
	'string'  => 'foo',
	'context' => 'barbaz',
)
array(
	'string'  => 'foobar',
	'context' => 'baz',
)
```

Maybe something like this? `md5( "{$string_data['string']}|{$string_data['context']}" )`. We already use pipes in [other situations](https://github.com/polylang/polylang/blob/9d4c52a9be9cd16e473bc1ed42a36308368fe561/src/translate-option.php#L234).

tests/phpunit/tests/Strings/test-String_Query.php
```diff
@@ -0,0 +1,393 @@
...
+
+		$result = $this->query->by_context( 'context_a' )->get();
+
+		$this->assertCount( 2, $result );
+		$this->assertTrue( $result->has( md5( 'source_1context_a' ) ) );
+		$this->assertTrue( $result->has( md5( 'source_3context_a' ) ) );
+		$this->assertFalse( $result->has( md5( 'source_2context_b' ) ) );
+	}
+
+	public function test_by_context_returns_self_for_chaining() {
```
Screenfeed: Maybe all the tests that use `by_context()` should depend on this one. Same goes for all the methods that have a "returns_self_for_chaining" test. Kind of overkill IMHO so I let you decide.

tests/phpunit/tests/Strings/test-String_Query.php
```diff
@@ -0,0 +1,393 @@
...
+		$result = $this->query->get();
+
+		$this->assertCount( 2, $result );
+	}
+
+	public function test_multiple_fragments_combined() {
+		Database_Repository::register( 'test_name', 'apple orange', 'test_context' );
+		Database_Repository::register( 'other_name', 'banana', 'test_context' );
+
+		$result = $this->query->by_fragment( 'apple' )->get();
```
Screenfeed, line 273: Multiple fragments? 🤔

Screenfeed (changes requested):
Nice 👍

src/Strings/Query.php
Hug0-Drelon: It'd force to do `if ( 0 === $this->per_page )` instead of just `if ( $this->per_page )`. Moreover, other disabled props use `null` as inactive state, so for simplicity I suggest to keep it that way.

src/Strings/Database_Repository.php
Hug0-Drelon: > Also I don't think doing md5( $string_data['string'] . $string_data['context'] ) is a good idea.

The edge case you've found is statistically irrelevant I think. That said, we could use you suggestion, but only once the PR validated. The reason being that all occurence have been kept as they are (straight to the point) waiting for a decision on the format/api to move this logic.

src/settings/table-string.php
Hug0-Drelon: See a proposal in 7003e7a22377b277a823db56ae2f18520dde04a0. Bonus point, it introduce a handy new query builder `Query::by_ids()`, which I found useful.

src/Strings/Query.php
Hug0-Drelon: `stripos` is fine now, https://github.com/php/php-src/issues/7847.
I don't think using `strpos( strtolower( ... ) )` is a good idea as it hurts readability given that `stripos` perf issue is fixed now.

src/Strings/Translatable.php
Hug0-Drelon: It was written for readability, not duplication. `is_matching` is self documented and self explanatory IMHO.

tests/phpunit/tests/Strings/test-String_Query.php
Hug0-Drelon: The last call to `by_fragment()` will get precedence over the previous ones, as it is written today.
If the need emerges, we'll be able to change it easily (all is encapsulated in `Query`). But I don't think this is required right now.

src/mo.php
Hug0-Drelon: I don't see the benefit. Do you want to have an empty string as default value?

src/settings/table-string.php
Hug0-Drelon: I wasn't aware of this method. Used in 08b89b89cda1a9a265e1f0390541df828201a6da.

src/Strings/Collection.php
Hug0-Drelon: What's the problem with "gets"? This is fine English if I'm correct?

src/Strings/Query.php
Hug0-Drelon: Added `static` keyword and injected `$languages` array in callback in 7b535de3.

tests/phpunit/tests/Strings/Table/test-Save.php
Hug0-Drelon: Removed in d4de849b.

tests/phpunit/tests/Strings/test-String_Query.php
Hug0-Drelon: Indeed, these tests are no longer useful. Even more so with the returned type hint in corresponding methods. It's a TDD artifact. Removed in 65a3dd99.

tests/phpunit/tests/Strings/Table/test-Save.php
Hug0-Drelon: Applied in eb565ba4.

tests/phpunit/tests/Strings/Table/test-Prepare_Items.php
Hug0-Drelon: Indeed, see a proposition in 5247d154 to make a real test.

src/Strings/Collection.php
Screenfeed: We decided long ago that we should prefer `Returns` over `Gets`.

src/mo.php
Screenfeed: Yes, I can't see the point of allowing `null` for the context, it always leads to:
```php
if ( null !== $context && '' !== $context ) {
```

So instead of having `?string $context = null` (or `$context = null` in `delete_entry()` and `translate()`) I think we should simplify it by allowing only strings everywhere: `string $context = ''`.

src/settings/table-string.php
```diff
@@ -304,14 +282,16 @@ public function prepare_items() {
...
-			foreach ( $this->items as $key => $row ) {
-				$this->items[ $key ]['translations'][ $language->slug ] = $mo->translate_if_any( $row['string'] );
-				$this->items[ $key ]['row']                             = $key; // Store the row number for convenience.
-				$this->items[ $key ]['disabled'][ $language->slug ]     = $disabled;
+			foreach ( $items as $translatable ) {
+				$this->items[ $translatable->get_id() ]['string']                          = $translatable->get_source();
+				$this->items[ $translatable->get_id() ]['name']                            = $translatable->get_name();
+				$this->items[ $translatable->get_id() ]['context']                         = $translatable->get_context();
+				$this->items[ $translatable->get_id() ]['translations'][ $language->slug ] = $mo->translate_if_any( $translatable->get_source(), $translatable->get_context() );
+				$this->items[ $translatable->get_id() ]['row']                             = $translatable->get_id();                                                          // Store the row number for convenience.
```
Screenfeed, line 292: Did you put the comment this far on purpose?


```suggestion
				$this->items[ $translatable->get_id() ]['row']                             = $translatable->get_id(); // Store the row number for convenience.
```

Screenfeed (commented):
Some conflicts to solve.

PR #1642, opened 2025-03-21
Move PHP source code to `src/` folder

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/2554.

PR #1808, opened 2026-02-03
Move PHP source code in src/ folder.

Marianne380:
Fixes https://github.com/polylang/polylang-pro/issues/2554.

### Changes

- Move PHP source code to `src/` folder.
- Remove the `include/` subdirectory (move it directly to `src/`).
- Add `/src/` prefix in all file paths throughout the code.
- Update configuration files: `composer.json`, `phpstan.neon.dist`, `phpcs.xml.dist`, `rector.php`, `.typos.toml`.
- Update all documentation annotations (`@see`) and cross-references.

### DIFF
--- a/src/settings/settings.php
+++ b/src/settings/settings.php
@@ -394,6 +394,7 @@ class PLL_Settings extends PLL_Admin_Base {
 
 		/**
 		 * Filters the list of predefined languages.
+		 *
 		 * @see src/settings/languages.php the list of predefined languages.
 		 *
 		 * @since 1.7.10


composer.json
```diff
@@ -22,13 +22,13 @@
...
-			"install/",
-			"modules/",
-			"settings/"
+            "src/admin/",
+			"src/frontend/",
+			"src/include/",
+			"src/integrations/",
+			"src/install/",
+			"src/modules/",
+			"src/settings/"
```
Hug0-Drelon:
```suggestion
            "src/"
```
Shorter and also easier to maintain (any file/folder added to `src/` would benefit from classmap right away).

phpstan.neon.dist
```diff
@@ -5,16 +5,16 @@ parameters:
...
-		- include/
-		- install/
-		- modules/
-		- settings/
+		- src/admin/
+		- src/frontend/
+		- src/include/
+		- src/install/
+		- src/modules/
+		- src/settings/
```
Hug0-Drelon:
```suggestion
		- src/
```
Let's run PHPStan on the whole directory.

phpunit.xml
```diff
@@ -22,17 +22,17 @@
...
-			<directory suffix=".php">./integrations</directory>
-			<directory suffix=".php">./modules</directory>
-			<directory suffix=".php">./settings</directory>
+			<directory suffix=".php">./src/admin</directory>
+			<directory suffix=".php">./src/frontend</directory>
+			<directory suffix=".php">./src/include</directory>
+			<directory suffix=".php">./src/install</directory>
+			<directory suffix=".php">./src/integrations</directory>
+			<directory suffix=".php">./src/modules</directory>
+			<directory suffix=".php">./src/settings</directory>
```
Hug0-Drelon: Could we simplify using only `src/` here?

phpstan.neon.dist
```diff
@@ -5,16 +5,12 @@ parameters:
...
-		- modules/
-		- settings/
+		- src/
 	excludePaths:
 		- **/load.php
-		- include/widget-calendar.php
-		- install/plugin-updater.php
+		- src/include/widget-calendar.php
+		- src/install/plugin-updater.php
+		- src/integrations/
```
Hug0-Drelon: Just a proposal, maybe we should wait for @Chouby review.
But I would suggest to write down all integration folders one by one.
So next integration will have a proper PHPStan analysis. I fear we'll forgot to update the config.

Marianne380: Yes, I quite agree with that.

Chouby: No objection

src/Model/Languages.php
```diff
@@ -186,7 +186,7 @@ public function add( $args ) {
 		$args['term_group'] = $args['term_group'] ?? 0;
 
 		if ( ! empty( $args['locale'] ) && ( ! isset( $args['name'] ) || ! isset( $args['slug'] ) ) ) {
-			$languages = include POLYLANG_DIR . '/settings/languages.php';
+			$languages = include POLYLANG_DIR . '/src/settings/languages.php';
```
Chouby, line 189: Check also `@see` annotations for example line 165 (and several others).

Chouby (changes requested):
It seems that changes in `phpcs.xml.dist`, `rector.php` files are missing. See #1642
I suggest to remove the include folder and keep its content directly in src as was done in #1642
It seems the PR description is not up to date.

PR #1811, opened 2026-02-05
Migrate Polylang blocks.

Hug0-Drelon:
## What?
Bring `polylang/navigation-language-switcher` and `polylang/language-switcher` to Polylang Free
Related to https://github.com/polylang/polylang-pro/pull/2902

## How?
- copy/paste source code
- Add namespaces to PHP code
- Use shared webpack build scripts from `@wpsyntex/polylang-build-scripts`
- Use shared React library `@wpsyntex/polylang-react-library` for some icons and hooks
- Minor tweaks in `Abstract_Block::get_path()` for it to work when used from Polylang Pro
- `useCurrentLanguage` hook relies on a new JS SSR variable `pllEditorCurrentLanguageSlug` because `currentPost.lang` is available only with Polylang Pro (see `Javascript_SSR`)
- Change text domain from `polylang-pro` to `polylang`

> [!WARNING]
> PHPUnit and e2e tests are not moved yet. So the tests can run in Polylang Pro while building blocks from Polylang Core.
> Once all validated, I'll move tests in a separate PR.

### DIFF
--- a/src/modules/Blocks/Language_Switcher/Abstract_Block.php
+++ b/src/modules/Blocks/Language_Switcher/Abstract_Block.php
@@ -127,10 +127,19 @@ abstract class Abstract_Block {
 			return;
 		}
 
-		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+		if ( ! register_block_type(
+			$this->get_path(),
+			array(
+				'render_callback' => array( $this, 'render' ),
+			)
+		) ) {
+			return;
+		}
 
+		$script_handle   = 'pll_blocks';
+		$suffix          = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
 		$script_filename = 'js/build/blocks' . $suffix . '.js';
-		$script_handle = 'pll_blocks';
+
 		wp_register_script(
 			$script_handle,
 			plugins_url( $script_filename, POLYLANG_ROOT_FILE ),
@@ -148,15 +157,6 @@ abstract class Abstract_Block {
 		);
 
 		wp_localize_script( $script_handle, 'pll_block_editor_blocks_settings', PLL_Switcher::get_switcher_options( 'block', 'string' ) );
-
-		if ( ! register_block_type(
-			$this->get_path(),
-			array(
-				'render_callback' => array( $this, 'render' ),
-			)
-		) ) {
-			return;
-		}
 
 		// Translated strings used in JS code
 		wp_set_script_translations( $script_handle, 'polylang' );
@@ -177,8 +177,6 @@ abstract class Abstract_Block {
 			// Should not happen since the module is loaded only if there are languages.
 			return;
 		}
-
-		$script_handle = 'pll_blocks'; // Script handles matches the one for Polylang blocks.
 
 		if ( str_contains( wp_scripts()->get_inline_script_data( $script_handle, 'after' ), 'pllEditorCurrentLanguageSlug' ) ) {
 			return;


src/modules/blocks/load.php
```diff
@@ -0,0 +1,28 @@
+<?php
+/**
+ * @package Polylang-Pro
```
Marianne380: We should change that, and some others.

Hug0-Drelon: Oops, fixed in 75755b4d.

src/modules/Blocks/Language_Switcher/Standard/block.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "https://schemas.wp.org/trunk/block.json",
+    "apiVersion": 3,
+    "name": "polylang/language-switcher",
+    "version": "3.8",
+    "title": "Language Switcher",
+    "category": "widgets",
+    "description": "Language switcher to insert in content or as widget.",
```
Chouby:
```suggestion
    "description": "Language switcher to insert in content or as a widget.",
```

src/modules/Blocks/Javascript_SSR.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @return void
+	 */
+	public function render_js_variable() {
+		// Fallback to default language if current language is not set, usually happens in Site Editor.
+		$current_language_slug = $this->current_language ? $this->current_language->slug : $this->default_language->slug;
+
+		$pll_settings_script = 'let pllEditorCurrentLanguageSlug = ' . wp_json_encode( $current_language_slug );
+
+		// Script handles matches the one for Polylang blocks.
+		wp_add_inline_script( 'pll_blocks', $pll_settings_script, 'after' );
```
Chouby: I suggest to move the logic of this class to `Abstract_Block` as it is strongly linked to block registration.

To avoid adding the inline script several times, you can test something like:
```php
in_array( $pll_settings_script, get_inline_script_data( 'pll_blocks', 'after' ) )
```

Hug0-Drelon: See 6eecd193.

Hug0-Drelon: I've improved it in 74affe08.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
```diff
@@ -0,0 +1,246 @@
...
+	 * @return self
+	 */
+	public function init() {
+		// Use rest_pre_dispatch_filter to get additional parameters for language switcher block.
+		add_filter( 'rest_pre_dispatch', array( $this, 'get_rest_query_params' ), 10, 3 );
+
+		// Register language switcher block.
+		add_action( 'init', array( $this, 'register' ) );
+
+		add_action( 'admin_enqueue_scripts', array( $this, 'render_js_variable' ) );
```
Chouby: I don't believe it's necessary to use this hook. You should be able to call `add_inline_script()` just after the script is registered.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	/**
+	 * Adds current language slug JavaScript variable to the editors.
+	 *
+	 * @since 3.8
+	 *
+	 * @return void
+	 */
+	public function render_js_variable() {
```
Chouby: I suggest to merge the method with `register()`. See my comment above about the hook.

Also as you suggested, you should add:
```php
		if ( class_exists( 'WP_Syntex\Polylang_Pro\REST\Translated\Post' ) ) {
			return;
		}
```
Maybe this new test requires a comment.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
```diff
@@ -0,0 +1,246 @@
...
+		if ( ! $current_language ) {
+			// Should not happen since the module is loaded only if there are languages.
+			return;
+		}
+
+		$script_handle = 'pll_blocks'; // Script handles matches the one for Polylang blocks.
+
+		if ( str_contains( wp_scripts()->get_inline_script_data( $script_handle, 'after' ), 'pllEditorCurrentLanguageSlug' ) ) {
+			return;
+		}
```
Chouby: I didn't pay attention. My initial suggestion should have used `get_data()` to get an array.This way seems as good.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
Hug0-Drelon: Done in 80a7ee77.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
Hug0-Drelon: Done in 80a7ee77.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
Hug0-Drelon: I thought that checking a shorter string against another string was better here.

PR #1812, opened 2026-02-05
Added scripts for code coverage

Screenfeed:
Monosite:

```
npm run test:php:coverage
```

Multisite:

```
npm run test:php:coverage:multisite
```


package.json
Hug0-Drelon, line 1: I suggest to add this composer [script](https://github.com/polylang/polylang-pro/blob/3.7.7/composer.json#L76-L81) and reuse it in the npm scripts you added.

Hug0-Drelon (approved):

package.json
Screenfeed: Done in [0cc2165](https://github.com/polylang/polylang/pull/1812/commits/0cc2165f17e39733aace593a89fea2c757025f53).
However it's much simpler than in PLL Pro.

Hug0-Drelon: Yup, we've only got one test suite here.

PR #1813, opened 2026-02-05
Fix test library backward compatibility with /src/ directory restructuring.

Marianne380:
## What
The test library (`polylang-test-lib`) used by PLLWC fails when testing with Polylang < 3.8 because it hardcodes paths to the new `/src/` directory structure introduced in https://github.com/polylang/polylang/pull/1808.

This breaks PLLWC CI which tests both:
- PLL dev (master with new `/src/` structure)
- PLL 3.7.x (old structure without `/src/`)

### New trait for centralized path resolution

Created `PLL_File_Path_Helper_Trait` with a single `get_pll_file_path()` method that uses `file_exists()` to detect which architecture is present:
- New (3.8+): `src/settings/languages.php`, `src/api.php`
- Old (<3.8): `settings/languages.php`, `include/api.php`

### Why a new trait?

A dedicated trait was necessary because:
1. `PLL_Multisites_TestCase` already has `tear_down()` and `create_options()` methods that conflict with `PLL_UnitTestCase_Trait`.
2. `PLL_UnitTest_Factory_For_Language` is a factory class, not a test case, so using `PLL_UnitTestCase_Trait` would be inappropriate.
3. Only the file path resolution logic needs to be shared, not the entire test case.

### DIFF
--- a/tests/phpunit/includes/file-path-helper-trait.php
+++ b/tests/phpunit/includes/file-path-helper-trait.php
@@ -1,7 +1,9 @@
 <?php
 
 /**
- * Helper trait for backward compatibility with Polylang /src/ restructuring.
+ * Helper trait for backward compatibility with Polylang < 3.8.
+ * Required as Polylang for WooCommerce always uses the latest test library,
+ * even when testing with older versions of Polylang. 
  */
 trait PLL_File_Path_Helper_Trait {
 


Screenfeed (commented):
What do you think about this?

```php
	protected static function get_pll_file_path( string $path ) {
		$path = ltrim( $path, '/' );
		// Backward compatibility with Polylang < 3.8 (/src/ directory restructuring). Required for PLLWC.
		if ( file_exists( POLYLANG_DIR . "/src/{$path}" ) ) {
			return POLYLANG_DIR . "/src/{$path}";
		}
		if ( false === strpos( $path, '/' ) ) {
			return POLYLANG_DIR . "/include/{$path}";
		}
		return POLYLANG_DIR . "/{$path}";
	}
```

`get_pll_file_path( 'api.php' )` would return `/foobar/src/api.php` or `/foobar/include/api.php`.
`get_pll_file_path( 'settings/languages.php' )` would return `/foobar/src/settings/languages.php` or `/foobar/settings/languages.php`.

Marianne380:
> What do you think about this?
> 
> ```
> 	protected static function get_pll_file_path( string $path ) {
> 		$path = ltrim( $path, '/' );
> 		// Backward compatibility with Polylang < 3.8 (/src/ directory restructuring). Required for PLLWC.
> 		if ( file_exists( POLYLANG_DIR . "/src/{$path}" ) ) {
> 			return POLYLANG_DIR . "/src/{$path}";
> 		}
> 		if ( false === strpos( $path, '/' ) ) {
> 			return POLYLANG_DIR . "/include/{$path}";
> 		}
> 		return POLYLANG_DIR . "/{$path}";
> 	}
> ```
> 
> `get_pll_file_path( 'api.php' )` would return `/foobar/src/api.php` or `/foobar/include/api.php`. `get_pll_file_path( 'settings/languages.php' )` would return `/foobar/src/settings/languages.php` or `/foobar/settings/languages.php`.

Good idea, see https://github.com/polylang/polylang/pull/1813/commits/af70aca319d44f05ea3c6439c8526d2427cfe9eb and the updated description of the PR for details.

tests/phpunit/includes/factory/unittest-factory-for-language.php
```diff
@@ -36,7 +38,7 @@ public function create( $args = array(), $generation_definitions = null ) {
 			throw new InvalidArgumentException( 'A locale is required to create a language.' );
 		}
 
-		$languages = include POLYLANG_DIR . '/src/settings/languages.php';
+		$languages = include self::get_pll_file_path( 'settings/languages.php' ); // phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.NotAbsolutePath
```
Screenfeed, line 41: Damn, PHPCS.
Maybe we can avoid it if we add a `/`: `self::get_pll_file_path( '/settings/languages.php' );`. And since we use `ltrim()`, no need to change the method.

tests/phpunit/includes/file-path-helper-trait.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+/**
+ * Helper trait for backward compatibility with Polylang /src/ restructuring.
+ */
+trait PLL_File_Path_Helper_Trait {
+
+	/**
+	 * Gets the path to a Polylang file, with backward compatibility for the /src/ restructuring.
```
Screenfeed:
```suggestion
	 * Returns the path to a Polylang file, with backward compatibility for the /src/ restructuring.
```

tests/phpunit/includes/testcase-trait.php
```diff
@@ -244,7 +245,7 @@ protected function require_wp_menus( $trigger_hooks = true ) {
 	 * @return void
 	 */
 	protected static function require_api(): void {
-		require_once POLYLANG_DIR . '/src/api.php';
+		require_once self::get_pll_file_path( 'api.php' ); // phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.NotAbsolutePath
 	}
```
Screenfeed: Idea: move this method to `PLL_File_Path_Helper_Trait`, and then we can use it in `PLL_Multisites_TestCase` instead of `get_pll_file_path()`.
What do you think?

tests/phpunit/includes/factory/unittest-factory-for-language.php
Marianne380: It doesn't change anything for PHPCS.

Screenfeed: pepehands

tests/phpunit/includes/file-path-helper-trait.php
```diff
@@ -0,0 +1,38 @@
...
+		return POLYLANG_DIR . "/{$path}";
+	}
+
+	/**
+	 * Requires the API functions.
+	 *
+	 * @return void
+	 */
+	protected static function require_api(): void {
+		require_once self::get_pll_file_path( '/api.php' ); // phpcs:ignore WordPressVIPMinimum.Files.IncludingFile.NotAbsolutePath
```
Screenfeed: Useless `/` but... meh 😅

tests/phpunit/includes/file-path-helper-trait.php
```diff
@@ -0,0 +1,38 @@
...
+	protected static function get_pll_file_path( string $path ): string {
+		$path = ltrim( $path, '/' );
+
+		// New structure (Polylang 3.8+): files in /src/ directory.
+		if ( file_exists( POLYLANG_DIR . "/src/{$path}" ) ) {
+			return POLYLANG_DIR . "/src/{$path}";
+		}
+
+		// Backward compatibility with Polylang < 3.8: files in root or /include/ directory.
+		if ( false === strpos( $path, '/' ) ) {
```
Chouby:
```suggestion
		if ( file_exists( POLYLANG_DIR . "/include/{$path}" ) ) {
```

Chouby (changes requested):

PR #1814, opened 2026-02-09
Add Brittany flag

YoruNoHikage:
Hello, I just stumbled on #1793 and the Breton support in DeepL.

Here's a PR to add the widely used flag all over Brittany and even worldwide.


Chouby:
Thank you for your contribution. 

In version 3.9, we plan to replace PNG flags by SVG flags. See #1864.
You'll be happy to learn that [the Brittany flag is already included](https://github.com/polylang/flags/blob/main/fr-bre.svg).

However there is still a lot of work before a WordPress language pack becomes available. See https://translate.wordpress.org/locale/br/default/wp/dev/

PR #1815, opened 2026-02-10
Test capabilities features in PHPUnit.

Hug0-Drelon:
## What?
Moves tests from https://github.com/polylang/polylang-pro/pull/2861/changes#diff-f88c67acfaf94515bfb19ede45c17257ef25c2712405d80c79d72f6eb979d8d2
Fixes partially https://github.com/polylang/polylang-pro/issues/2815

## How?
- Cover `PLL_Admin_Filters_Media`
- Cover `User_interface`, `Creator_Interface` and `Capabilities`
- Cover `PLL_Admin_links`
- Cover `Create\Post` and `Create\Term`

Try to cover methods more thoroughly (i.e. not only run code but intentionally test branches).

### DIFF
--- a/tests/phpunit/tests/Capabilities/UI/test-Media.php
+++ b/tests/phpunit/tests/Capabilities/UI/test-Media.php
@@ -146,7 +146,7 @@ class Test_Media extends PLL_UnitTestCase {
 		$fields = get_attachment_fields_to_edit( get_post( $attachment ) );
 		$html   = $fields['language']['html'];
 
-		$this->assert_languages_in_dropdown( array( 'en-US', 'en-US', 'fr-FR', 'de-DE' ), $html ); // Two times "en-US" because the empty option is prepend to a fuul list (no Pro feature active).
+		$this->assert_languages_in_dropdown( array( 'en-US', 'fr-FR', 'de-DE' ), $html ); // Two times "en-US" because the empty option is prepend to a fuul list (no Pro feature active).
 		$this->assert_selected_language( 'en', $html );
 		$this->assertTrue( $this->is_dropdown_disabled( $html ) );
 	}


tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		self::$contributor = self::factory()->user->create_and_get( array( 'role' => 'contributor' ) );
+	}
+
+	public static function wpTearDownAfterClass() {
+		self::delete_user( self::$editor->ID );
+		self::delete_user( self::$translator_fr->ID );
+		self::delete_user( self::$contributor->ID );
+
+		parent::wpTearDownAfterClass();
+	}
```
Chouby: This shouldn't be useful.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		Capabilities::set_user_creator( new Creator() );
+
+		parent::tear_down();
+	}
+
+	/**
+	 * Tests that a translator can only create translations in their allowed language.
+	 */
+	public function test_translator_cannot_translate_to_unauthorized_language() {
+		wp_set_current_user( self::$translator_fr->ID );
```
Chouby: Useless as `User` is mocked.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+	 * @testWith ["en", "fr"]
+	 *           ["fr", "en"]
+	 *
+	 * @param string $post_lang   Slug of the lang of the post.
+	 * @param string $target_lang Slug of the lang to assign to the post.
+	 * @return void
+	 */
+	public function test_editor_can_translate_to_any_language( string $post_lang, string $target_lang ) {
+		wp_set_current_user( self::$editor->ID );
+		$this->links = new PLL_Admin_Links( $this->pll_admin );
```
Chouby: Duplicated from `set_up()`.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+	protected static $contributor;
+
+	public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
+		parent::pllSetUpBeforeClass( $factory );
+
+		$factory->language->create_many( 2 );
+
+		self::$editor        = self::factory()->user->create_and_get( array( 'role' => 'editor' ) );
+		self::$translator_fr = self::factory()->user->create_and_get( array( 'role' => 'editor' ) );
+		self::$translator_fr->add_cap( 'translate_fr' );
```
Chouby: This capability must be tested in Polylang Pro.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+
+		self::require_api();
+	}
+
+	public function tear_down() {
+		// Reset user creator after each tests to avoid state bleeding.
+		Capabilities::set_user_creator( new Creator() );
+
+		parent::tear_down();
+	}
```
Chouby: Is this useful? It's not changed in tests.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		$options           = $this->create_options(
+			array(
+				'default_lang'  => 'en',
+				'media_support' => true,
+			)
+		);
+		$model                  = new PLL_Model( $options );
+		$links_model            = $model->get_links_model();
+		$this->pll_admin        = new PLL_Admin( $links_model );
+		$this->pll_admin->links = new PLL_Admin_Links( $this->pll_admin );
```
Chouby: I suggest not to use a property, but to create a new independent object for each test.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		$privacy_page_fr = self::factory()->post->create_and_get(
+			array(
+				'post_type' => 'page',
+				'lang'      => 'fr',
+			)
+		);
+		$this->pll_admin->model->post->save_translations( $privacy_page_en->ID, array( 'en' => $privacy_page_en->ID, 'fr' => $privacy_page_fr->ID ) );
+
+		// Create a third language.
+		self::create_language( 'de_DE_formal' );
```
Chouby: I suggest to move this to `pllSetUpBeforeClass()`.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		// Reinitialize the links object.
+		$options           = $this->create_options(
+			array(
+				'default_lang'  => 'en',
+				'media_support' => true,
+			)
+		);
+		$model             = new PLL_Model( $options );
+		$links_model       = $model->get_links_model();
+		$this->pll_admin   = new PLL_Admin( $links_model );
```
Chouby: Why is this necessary?

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+
+		$link = $this->links->get_new_post_translation_link( $post, $language, 'raw' );
+
+		$this->assertStringMatchesFormat( 'http://example.org/wp-admin/post-new.php?post_type=post&from_post=%d&new_lang=fr&_wpnonce=%s', $link, 'The link should be correct.' );
+	}
+
+	/**
+	 * Tests the link for page post type.
+	 */
+	public function test_page_link() {
```
Chouby: Does it test something different from other tests?

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		$link = $this->links->get_new_post_translation_link( $post, $language );
+
+		$this->assertStringMatchesFormat( 'http://example.org/wp-admin/post-new.php?post_type=post&amp;from_post=%d&amp;new_lang=fr&amp;_wpnonce=%s&custom_param=1', $link, 'The link should be correct and the filter should be applied.' );
+	}
+
+	/**
+	 * Tests that a translator cannot create a term translation in an unauthorized language.
+	 */
+	public function test_translator_cannot_translate_term_to_unauthorized_language() {
+		wp_set_current_user( self::$translator_fr->ID );
```
Chouby: Useless

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		$this->assertStringMatchesFormat( 'http://example.org/wp-admin/edit-tags.php?taxonomy=category&post_type=post&from_tag=%d&new_lang=fr&custom_term_param=1', $link, 'The link should be correct and the filter should be applied.' );
+	}
+
+	/**
+	 * Replace `PLL_Admin_Links::$user` with a user mock that can translate.
+	 *
+	 * @param array $methods An array of methods to mock and their return values.
+	 * @return void
+	 */
+	private function mock_links_user_as_translator( array $methods = array() ) {
```
Chouby: I suggest not to make reference to 'translator':
```suggestion
	private function mock_user_for_links( array $methods = array() ) {
```

Another possibility could be to restrict the usage of this method to `'can_translate' => false` as finally this should be the only useful usage. and in this case I suggest:
```suggestion
	private function disallow_translation_for_user() {
```

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		$this->links = new PLL_Admin_Links( $this->pll_admin );
+
+		$term = self::factory()->term->create_and_get( array( 'taxonomy' => 'category' ) );
+		$this->pll_admin->model->term->set_language( $term->term_id, 'en' );
+		$language = $this->pll_admin->model->get_language( 'fr' );
+
+		$link = $this->links->get_new_term_translation_link( $term, 'page', $language );
+
+		$this->assertStringMatchesFormat( 'http://example.org/wp-admin/edit-tags.php?taxonomy=category&post_type=page&from_tag=%d&new_lang=fr', $link, 'The link should be correct.' );
+	}
```
Chouby: Although permitted, this looks like a wrong usage of the method as the 'category' taxonomy is not registered for pages.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+		$this->links = new PLL_Admin_Links( $this->pll_admin );
+
+		$term = self::factory()->term->create_and_get( array( 'taxonomy' => 'post_tag' ) );
+		$this->pll_admin->model->term->set_language( $term->term_id, 'en' );
+		$language = $this->pll_admin->model->get_language( 'fr' );
+
+		$link = $this->links->get_new_term_translation_link( $term, 'post', $language );
+
+		$this->assertStringMatchesFormat( 'http://example.org/wp-admin/edit-tags.php?taxonomy=post_tag&post_type=post&from_tag=%d&new_lang=fr', $link, 'The link should be correct.' );
+	}
```
Chouby: This test doesn't add much value in my opinion. I you want to keep it I suggest, to merge it with another test and use `@testWith`.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+
+		$link = $this->links->get_new_term_translation_link( $term, 'post', $language );
+
+		$this->assertSame( '', $link, 'A French translator should not be able to create a term translation in English.' );
+	}
+
+	/**
+	 * Tests that a translator can create a term translation in their allowed language.
+	 */
+	public function test_translator_can_translate_term_to_authorized_language() {
```
Chouby: This test refers to Polylang Pro feature.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+	 */
+	public function test_translator_can_translate_to_authorized_language() {
+		wp_set_current_user( self::$translator_fr->ID );
+
+		$this->mock_links_user_as_translator(
+			array(
+				'can_translate' => true,
+				'has_cap'       => true,
+			)
+		);
```
Chouby: Setting the user is useless if you mock the user (and translators are related to Polylang Pro features). In fact, this test is redundant with the next one. You don't need to mock the user here as this is the default behavior for any user in Polylang.

tests/phpunit/tests/test-admin-links.php
```diff
@@ -0,0 +1,550 @@
...
+
+		$link = $this->links->get_new_term_translation_link( $term, 'post', $language );
+
+		$this->assertStringMatchesFormat( 'http://example.org/wp-admin/edit-tags.php?taxonomy=category&post_type=post&from_tag=%d&new_lang=fr', $link, 'A French translator should be able to create a term translation in French.' );
+	}
+
+	/**
+	 * Tests that an editor can translate a term to any language.
+	 */
+	public function test_editor_can_translate_term_to_any_language() {
```
Chouby: The name doesn't reflect the test. For posts you added a `@testWith` to test different languages, not here.

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+	public function test_returns_empty_fields_on_edit_media_panel(): void {
+		$GLOBALS['pagenow'] = 'post.php';
+
+		$attachment = self::factory()->attachment->create();
+		$this->pll_model->post->set_language( $attachment, 'en' );
+
+		$fields = $this->pll_env->filters_media->attachment_fields_to_edit( array(), get_post( $attachment ) );
+
+		$this->assertArrayNotHasKey( 'language', $fields );
+	}
```
Chouby, line 99: This looks duplicate from https://github.com/polylang/polylang/blob/3.8-beta1/tests/phpunit/tests/test-media.php#L69-L88
The original test is better because it calls `get_attachment_fields_to_edit()` instead of our method.
Having 2 separate tests as you did is better as the original test.

Adding a new test file is probably the cause for duplicating tests.

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+	/**
+	 * Tests that non-translator users see all languages and dropdown is enabled.
+	 *
+	 * @testWith ["en"]
+	 *           ["fr"]
+	 *           ["de"]
+	 *
+	 * @param string $lang_slug The language of the attachment.
+	 */
+	public function test_non_translator_sees_all_languages( string $lang_slug ): void {
```
Chouby: I would prefer that we don't make reference to a Polylang Pro feature in method name.

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+		$GLOBALS['pagenow'] = 'upload.php';
+
+		$attachment = self::factory()->attachment->create();
+		$this->pll_model->post->set_language( $attachment, $lang_slug );
+
+		$fields = $this->pll_env->filters_media->attachment_fields_to_edit( array(), get_post( $attachment ) );
```
Chouby: Calling `get_attachment_fields_to_edit()` is better in my opinion as this is what WP does.

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+	public function set_up() {
+		parent::set_up();
+
+		setUp();
+
+		$options                        = $this->create_options( $this->options );
+		$this->pll_model                = new PLL_Model( $options );
+		$links_model                    = $this->pll_model->get_links_model();
+		$this->capabilities             = new Capabilities();
+		Capabilities::set_user_creator( new Creator() );
```
Chouby: Is it useful?

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+		$this->pll_env                  = new PLL_Admin( $links_model );
+		$this->pll_env->posts           = new PLL_CRUD_Posts( $this->pll_env );
+		$this->pll_env->terms           = new PLL_CRUD_Terms( $this->pll_env );
```
Chouby: terms is useless for sure.
posts should be too. It used to useful but it shouldn't be anymore. I realize that I forgot to remove the property in `PLL_Admin_Filters_Media`. We could do this in a separate PR.

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+		$this->pll_env->links           = new PLL_Admin_Links( $this->pll_env );
+		$this->pll_env->posts           = new PLL_CRUD_Posts( $this->pll_env );
```
Chouby: Duplicate.

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+			->willReturn( true );
+		$translator_mock
+			->method( 'can_translate' )
+			->willReturnCallback( fn( $language ) => 'fr' === $language->slug );
+		$translator_mock
+			->method( 'get_preferred_language_slug' )
+			->willReturn( 'fr' );
+
+		// Brain\Monkey API doesn't support static methods mocking, so we need to use Patchwork.
+		redefine( Capabilities::class . '::get_user', fn() => $translator_mock );
```
Chouby: Couldn't we reuse the method `mock_translator()` that you proposed in the test case above?

tests/phpunit/tests/Capabilities/UI/test-Media.php
```diff
@@ -0,0 +1,312 @@
...
+
+		$fields = $this->pll_env->filters_media->attachment_fields_to_edit( array(), get_post( $attachment ) );
+		$html   = $fields['language']['html'];
+
+		$this->assert_languages_in_dropdown( array( 'en-US', 'en-US', 'fr-FR', 'de-DE' ), $html ); // Two times "en-US" because the empty option is prepend to a fuul list (no Pro feature active).
```
Chouby: About the comment, isn't this a bug that we should fix instead?

tests/phpunit/tests/Capabilities/Create/TestCase.php
```diff
@@ -0,0 +1,99 @@
...
+
+		setUp();
+
+		$options         = $this->create_options( array( 'default_lang' => 'en' ) );
+		$this->pll_model = new PLL_Model( $options );
+	}
+
+	public function tear_down() {
+		// Reset user creator after each tests to avoid state bleeding.
+		Capabilities::set_user_creator( new Creator() );
```
Chouby: Is it useful?

tests/phpunit/tests/Capabilities/Create/TestCase.php
```diff
@@ -0,0 +1,99 @@
...
+		parent::pllSetUpBeforeClass( $factory );
+
+		$factory->language->create_many( 3 );
+
+		self::$translator_fr = $factory->user->create_and_get( array( 'role' => 'editor' ) );
+		self::$translator_fr->add_cap( 'translate_fr' );
+
+		self::$translator_en = $factory->user->create_and_get( array( 'role' => 'editor' ) );
+		self::$translator_en->add_cap( 'translate_en' );
+
```
Chouby: Translators shouldn't be used in these tests.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+	 * @testWith ["en"]
+	 *           ["fr"]
+	 *           ["de"]
+	 *
+	 * @param string $lang The language code.
+	 */
+	public function test_returns_new_lang_from_get_param( string $lang ) {
+		$_GET['new_lang'] = $lang;
+
+		wp_set_current_user( self::$editor->ID );
```
Chouby: This seems useless.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * @group capabilities
+ */
+class Test_Post extends TestCase {
+
+	/**
+	 * @testWith ["en"]
+	 *           ["fr"]
+	 *           ["de"]
```
Chouby: Does this add value to more than 1 language?

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+		$post   = $this->create_post_capa_object();
+		$result = $post->get_language( 0 );
+
+		$this->assertSame( $lang, $result->slug );
+	}
+
+	public function test_returns_default_when_new_lang_is_invalid() {
+		$_GET['new_lang'] = 'invalid';
+
+		wp_set_current_user( self::$editor->ID );
```
Chouby: Seems useless.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+		$result = $post->get_language( 0 );
+
+		$this->assertSame( 'en', $result->slug );
+	}
+
+	/**
+	 * @testWith ["en"]
+	 *           ["fr"]
+	 *           ["de"]
+	 *
```
Chouby, line 46: Testing only a non default language should be sufficient. Am I wrong?

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+	 * @param string $lang The language code.
+	 */
+	public function test_returns_lang_from_request_param_on_frontend( string $lang ) {
+		$_REQUEST['lang'] = $lang;
+
+		// pref_lang is null to simulate frontend context.
```
Chouby, line 52: I would move this comment above `create_post_capa_object()`.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+		wp_set_current_user( self::$editor->ID );
```
Chouby: Seems useless.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+		$result = $post->get_language( 0 );
+
+		$this->assertSame( $lang, $result->slug );
+	}
+
+	public function test_ignores_request_lang_when_pref_lang_is_set() {
+		$_REQUEST['lang'] = 'de';
+
+		// pref_lang is set to simulate admin context.
+		wp_set_current_user( self::$editor->ID );
```
Chouby: Same as above.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+
+		$this->assertSame( 'fr', $result->slug );
+	}
+
+	public function test_returns_language_from_rest_request() {
+		$request = $this->createMock( Request::class );
+		$request->method( 'get_language' )
+			->willReturn( $this->pll_model->languages->get( 'de' ) );
+
+		wp_set_current_user( self::$editor->ID );
```
Chouby: Same here and everywhere else.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	public function test_returns_pref_lang_when_user_can_translate() {
+		wp_set_current_user( self::$editor->ID );
+
+		$post   = $this->create_post_capa_object( null, $this->pll_model->languages->get( 'fr' ), null );
+		$result = $post->get_language( 0 );
+
+		$this->assertSame( 'fr', $result->slug );
+	}
```
Chouby: I would group tests related to `pref_lang` together.

tests/phpunit/tests/Capabilities/Create/TestCase.php
```diff
@@ -0,0 +1,99 @@
...
+	/**
+	 * @var WP_User
+	 */
+	protected static $editor;
+
+	public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
+		parent::pllSetUpBeforeClass( $factory );
+
+		$factory->language->create_many( 3 );
```
Chouby: In lots of places you are getting language objects with `$this->pll_model->languages->get()`. Maybe it could be useful to create static language objects used everywhere you nee them. What do you think?

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+		wp_set_current_user( self::$editor->ID );
+
+		$post   = $this->create_post_capa_object( null, null, $this->pll_model->languages->get( 'de' ) );
+		$result = $post->get_language( 0 );
+
+		$this->assertSame( 'de', $result->slug );
+	}
+
+	public function test_returns_default_language_for_translator_allowed_to_translate_default() {
+		wp_set_current_user( self::$translator_en->ID );
```
Chouby: This is useless as you use the mock. The same everywhere else.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	public function test_returns_default_language_for_non_translator() {
+		wp_set_current_user( self::$editor->ID );
+
+		$post   = $this->create_post_capa_object();
+		$result = $post->get_language( 0 );
+
+		$this->assertSame( 'en', $result->slug );
+	}
```
Chouby, line 122: I would move this test above the 2 tests just above.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+
+		$child_id = self::factory()->post->create( array( 'post_type' => 'page', 'post_parent' => $parent_id ) );
+
+		wp_set_current_user( self::$editor->ID );
+
+		$post   = $this->create_post_capa_object( $request, null, null );
+		$result = $post->get_language( $child_id );
+
+		$this->assertSame( 'fr', $result->slug );
+	}
```
Chouby: I'm not sure that this is a desired feature.

tests/phpunit/tests/Capabilities/Create/test-Term.php
```diff
@@ -0,0 +1,299 @@
...
+namespace WP_Syntex\Polylang\Tests\Integration\modules\Capabilities\Create;
+
+use PLL_Language;
+use WP_Syntex\Polylang\REST\Request;
+use WP_Syntex\Polylang\Capabilities\Create\Term;
+
+/**
+ * @group capabilities
+ */
+class Test_Term extends TestCase {
```
Chouby, line 12: A lot of my comments for posts are valid for this class too.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,235 @@
...
+	 *
+	 * @param string $lang The language code.
+	 */
+	public function test_returns_new_lang_from_get_param( string $lang ) {
+		$_GET['new_lang'] = $lang;
+
+		wp_set_current_user( self::$editor->ID );
+
+		$post   = $this->create_post_capa_object();
+		$result = $post->get_language( 0 );
```
Chouby: Given that 0 is the default value:
```suggestion
		$result = $post->get_language();
```

tests/phpunit/tests/Capabilities/Create/test-Term.php
```diff
@@ -0,0 +1,299 @@
...
+	 *
+	 * @param string $lang The language code.
+	 */
+	public function test_returns_new_lang_from_get_param( string $lang ) {
+		$_GET['new_lang'] = $lang;
+
+		wp_set_current_user( self::$editor->ID );
+
+		$term   = $this->create_term_capa_object();
+		$result = $term->get_language( 0, '' );
```
Chouby: Given that these are the default values:
```suggestion
		$result = $term->get_language();
```

tests/phpunit/tests/Capabilities/Create/test-Term.php
```diff
@@ -0,0 +1,299 @@
...
+		$_GET['new_lang']          = 'de';
+		$_POST['term_lang_choice'] = 'fr';
+
+		wp_set_current_user( self::$editor->ID );
+
+		$term   = $this->create_term_capa_object();
+		$result = $term->get_language( 0, '' );
+
+		$this->assertSame( 'de', $result->slug );
+	}
```
Chouby: Not sure these tests make sense as test cases are mutually exclusive.

tests/phpunit/tests/test-admin-links.php
Hug0-Drelon: The second option cannot be used because we need to mock `has_cap` sometimes.

tests/phpunit/tests/Capabilities/Create/test-Post.php
Hug0-Drelon: Do you mean that you want to change `Post::get-language` implementation or remove the test?

tests/phpunit/tests/Capabilities/UI/test-Media.php
Hug0-Drelon: I'm not sure to understand, do you want me to remove `test_returns_empty_fields_on_edit_media_panel`?

tests/phpunit/tests/Capabilities/UI/test-Media.php
Hug0-Drelon: Returned values aren't the same + we don't use  `ReflectionProperty` here but monkey patch `Capabilities::get_user`.

tests/phpunit/tests/Capabilities/UI/test-Media.php
Hug0-Drelon: It doesn't occur with Polylang Pro.
This is because `$this->model->languages->filter( 'translator' )->get_list()` returns **all** languages in Polylang Free even if the use has `translate_*` capability.
But UI wise it could be an edge case to handle.

tests/phpunit/tests/Capabilities/Create/test-Post.php
Hug0-Drelon: I think testing at least both default and non-default langue is a good idea. So I propose to remove `de` only.

tests/phpunit/tests/Capabilities/Create/test-Post.php
Hug0-Drelon: Same as above, I suggest to keep only two languages (def and non-def).

tests/phpunit/tests/Capabilities/Create/test-Term.php
Hug0-Drelon: We don't tests the same query arguments if I'm correct.

tests/phpunit/tests/Capabilities/Create/test-Post.php
```diff
@@ -0,0 +1,205 @@
...
+		$parent_id = self::factory()->post->create( array( 'post_type' => 'page' ) );
+		$this->pll_model->post->set_language( $parent_id, 'de' );
+
+		$child_id = self::factory()->post->create( array( 'post_type' => 'page', 'post_parent' => $parent_id ) );
+
+		$post   = $this->create_post_capa_object( $request, null, null );
+		$result = $post->get_language( $child_id );
+
+		$this->assertSame( 'fr', $result->slug );
+	}
```
Chouby: I suggest to remove these 2 tests as they cover the current behavior but not necessarily something we want.

tests/phpunit/tests/Capabilities/Create/test-Term.php
```diff
@@ -0,0 +1,257 @@
...
+		$parent_id = self::factory()->term->create( array( 'taxonomy' => 'category' ) );
+		$this->pll_model->term->set_language( $parent_id, 'de' );
+
+		$child_id = self::factory()->term->create( array( 'taxonomy' => 'category', 'parent' => $parent_id ) );
+
+		$term   = $this->create_term_capa_object( $request, null, null );
+		$result = $term->get_language( $child_id, 'category' );
+
+		$this->assertSame( 'fr', $result->slug );
+	}
```
Chouby: Same as above. I suggest to remove these 2 tests as they cover the current behavior but not necessarily something we want.

tests/phpunit/tests/Capabilities/Create/test-Term.php
```diff
@@ -0,0 +1,257 @@
...
+
+	public function test_new_lang_takes_priority_over_term_lang_choice() {
+		$_GET['new_lang']          = 'de';
+		$_POST['term_lang_choice'] = 'fr';
+
+		$term   = $this->create_term_capa_object();
+		$result = $term->get_language();
+
+		$this->assertSame( 'de', $result->slug );
+	}
```
Chouby: I suggest to remove these 2 tests too.

tests/phpunit/tests/Capabilities/UI/test-Media.php
Chouby: As they are now I prefer to delete the test in https://github.com/polylang/polylang/blob/3.8-beta1/tests/phpunit/tests/test-media.php#L69-L88

tests/phpunit/tests/Capabilities/UI/test-Media.php
Chouby: Then I suggest to move this to a separate method to clarify the test.

tests/phpunit/tests/Capabilities/UI/test-Media.php
Chouby: Then let's fix this in a separate PR.

Chouby (changes requested):

tests/phpunit/tests/Capabilities/UI/test-Media.php
Hug0-Drelon: Here it is https://github.com/polylang/polylang/pull/1817

PR #1819, opened 2026-02-17
Fix slug of child term not following the same rules as WP

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/2193

### DIFF
--- a/src/term-slug.php
+++ b/src/term-slug.php
@@ -87,6 +87,18 @@ class PLL_Term_Slug {
 		}
 		$this->lang = $lang;
 
+		if ( ! $this->slug ) {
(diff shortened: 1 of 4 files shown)


Hug0-Drelon (approved):
I fear this create a regression in Polylang Pro, see:

```bash
1) Share_Slugs_Test::test_translated_terms_with_parents_sharing_same_name
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'test-test'
+'test-test-2'
```

Otherwise the code LGTM 👍

Chouby:
In fact, the proposal is totally wrong. I am calling `term_exists` with `$this->parent` not set yet.

PR #1820, opened 2026-02-18
Restore language filtering after middleware migration.

Marianne380:
## What
- Make `pllFilteredRoutes` globally accessible for external middleware.
- Remove duplicate REST middleware code.

## Why
Migration to https://github.com/polylang/polylang-react-library moved the REST middleware (`editorsRequestsFilter`) into an external package.

Before:
  - Middleware was in local in `js/src/packages/middlewares/`.
  - Same build/scope as let `pllFilteredRoutes`.
  - Middleware accessed variable directly: `typeof pllFilteredRoutes`.

After migration:
  - Middleware in external package, see https://github.com/polylang/polylang-react-library.
  - Different scope, checks `typeof window.pllFilteredRoutes`.
  - Variable defined as `let` (local scope) and so always `undefined`.
  - Middleware bypasses ALL requests so no `lang=` parameter added.

Impact: REST requests missing `lang=` parameter since migration, causing incorrect language filtering:
- Reusable blocks (`/wp/v2/blocks`): Returns ALL languages instead of current language.
- Template parts (`/wp/v2/template-parts`): Returns DEFAULT language instead of current language.
- etc...

Additionally, PLL (`js/src/block-editor.js`) was still using duplicate middleware logic instead of importing from the shared library.

## Fix
1. Changes `let pllFilteredRoutes` to `window.pllFilteredRoutes` in `PLL_Filter_REST_Routes::add_inline_script()`
2. Refactors `js/src/block-editor.js` to use `editorsRequestsFilter` from `@wpsyntex/polylang-react-library`
3. Removes unused `js/src/lib/filter-path-middleware.js`

This ensures both PLL and PLL Pro use the same middleware implementation.

Additionally, moves `block-editor.js` compilation from vanilla to reactified webpack config to properly resolve React dependencies from `@wpsyntex/polylang-react-library`.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -35,6 +35,7 @@ function configureWebpack( options ) {
 		'js/src/lib/**',
 		'js/src/packages/**',
 		'js/src/blocks/**',
+		'js/src/block-editor.js',
 		'**/*.config.js',
 		'**/*.min.js',
 	];
@@ -67,7 +68,10 @@ function configureWebpack( options ) {
 	];
 
 	const reactifiedConfig = getReactifiedConfig( {
-		entryPoints: { blocks: './js/src/blocks/index.js' },
+		entryPoints: {
+			blocks: './js/src/blocks/index.js',
+			'block-editor': './js/src/block-editor.js',
+		},
 		outputPath: workingDirectory,
 		libraryName: 'polylang',
 		isProduction,


Marianne380:
> LGTML 👍 I've tested in the editor, content (Ie. media or terms) is well filtered by language. As seen together, it'll need an adaptation in `webpack.config.js` in Polylang Pro.

Right, see https://github.com/polylang/polylang-pro/pull/2931.

PR #1822, opened 2026-02-20
Fix non expiring cache of `count_posts()`

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/1967

In `PLL_Model::count_post()`, the cache doesn't expire. This is an issue for sites using object cache.
This PR proposes to salt the cache key with `wp_cache_get_last_changed()`.
The 2 private methods are introduced to handle the backward compatibility with WP < 6.9. They could be removed in the future when we will stop supporting WP 6.8 and older.

### DIFF
--- a/src/model.php
+++ b/src/model.php
@@ -620,7 +620,7 @@ class PLL_Model {
 	 * @since 3.9
 	 *
 	 * @param string $cache_key The cache key.
-	 * @return mixed The counts grouped per language, false if the cache is empty.
+	 * @return array|false The counts grouped per language, false if the cache is empty.
 	 */
 	private function get_counts_cache( string $cache_key ) {
 		$last_changed = wp_cache_get_last_changed( 'posts' );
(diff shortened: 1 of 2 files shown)


src/model.php
```diff
@@ -613,4 +613,47 @@ public function set_language_in_mass( $lang = null, array $types = array() ): vo
...
+	 * @param string $cache_key The cache key.
+	 * @return array|false The counts grouped per language, false if the cache is empty.
+	 */
+	private function get_counts_cache( string $cache_key ) {
+		$last_changed = wp_cache_get_last_changed( 'posts' );
+
+		if ( ! function_exists( 'wp_cache_get_salted' ) ) {
+			// Backward compatibility with WordPress < 6.9.
+			$cache_key = "{$cache_key}:{$last_changed}";
+			$counts    = wp_cache_get( $cache_key, 'posts' );
```
Screenfeed, line 631: Any reason to change the group from `'counts'` to `'posts'`? It seems to me that `'counts'` is more appropriate.
Same goes for `wp_cache_set()` (obviously), `wp_cache_get_salted()` and `wp_cache_set_salted()`.

Chouby: Yes. Consistency. I used `wp_cache_get_last_changed()` with the group `'posts'`. So it's consistent to use the same group for `wp_cache_get()`. Note that WP doesn't set a last changed on the `'counts'` group.

Screenfeed: I'm not sure what is best honestly. Put it into `count` because it's a count, or into `posts` because it uses `wp_cache_get_last_changed()`.

PR #1825, opened 2026-02-28
Avoid repeating ourselves in metas tests

Chouby:
Follow-up of #1761 
Preparation for https://github.com/polylang/polylang-pro/issues/2767

This is a proposal to avoid duplicating tests between posts metas and terms, thus simplifying the maintenance.
Note: This removes specific tests for protected post metas as I don't see any difference with public metas.
If they are necessary, we could re-introduce them in the new `Post` class.

### DIFF
--- a/tests/phpunit/tests/Meta/TestCase.php
+++ b/tests/phpunit/tests/Meta/TestCase.php
@@ -26,7 +26,7 @@ abstract class TestCase extends PLL_UnitTestCase {
 		$this->pll_env->sync = new PLL_Sync( $this->pll_env );
 	}
 
-	public function data_provider() {
+	public function meta_values_provider() {
 		return array(
 			'string'  => array( 'the_value' ),
 			'list'    => array( array( 'the_value', 'the_value_2' ) ),
--- a/tests/phpunit/tests/Meta/trait-Add.php
+++ b/tests/phpunit/tests/Meta/trait-Add.php
@@ -4,7 +4,7 @@ namespace WP_Syntex\Polylang_Pro\Tests\Integration\modules\Meta;
 
 trait Add {
 	/**
-	 * @dataProvider data_provider
+	 * @dataProvider meta_values_provider
 	 *
 	 * @param mixed $value The value to add to the meta.
 	 * @return void
@@ -16,7 +16,7 @@ trait Add {
 	}
 
 	/**
-	 * @dataProvider data_provider
+	 * @dataProvider meta_values_provider
 	 *
 	 * @param mixed $value The value to add to the meta.
 	 * @return void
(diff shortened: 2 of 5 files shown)


tests/phpunit/tests/Meta/test-Post.php
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * @var string
+	 */
+	protected static $type = 'post';
+
+	/**
+	 * @var int[]
+	 */
+	protected static $objects;
```
Hug0-Drelon, line 24: I think you need to delete `self::$objects` in `self::wpTearDownAfterClass()`.

tests/phpunit/tests/Meta/test-Term.php
```diff
@@ -0,0 +1,50 @@
...
+
+	/**
+	 * @var string
+	 */
+	protected static $type = 'term';
+
+	/**
+	 * @var int[]
+	 */
+	protected static $objects;
```
Hug0-Drelon, line 24: I think you need to delete `self::$objects` in `self::wpTearDownAfterClass()`.

tests/phpunit/tests/Meta/TestCase.php
```diff
@@ -26,7 +26,7 @@ public function set_up() {
 		$this->pll_env->sync = new PLL_Sync( $this->pll_env );
 	}
 
-	public function meta_values_provider() {
+	public function data_provider() {
```
Hug0-Drelon: We usually name providers more meaningfully. Don't you like the former name?

tests/phpunit/tests/Meta/test-Post.php
Chouby: It's useless as it's already don by WP. See https://github.com/WordPress/wordpress-develop/blob/6.9.4/tests/phpunit/includes/abstract-testcase.php#L91

tests/phpunit/tests/Meta/TestCase.php
Chouby: I don't mind (I probably renamed it because Meta is already part of the namespace).
See 1e796b6259c9bf643f2c90dbad209bd81afcea8c

tests/phpunit/tests/Meta/test-Post.php
Hug0-Drelon: Well, I think we could simplify a whole lot of other tests then...
Good to know, thanks!

Hug0-Drelon:
> > I suggest to store `the_key` in a property, in case we need to have several ones in the future.
> 
> I'm not sure to understand the purpose.

It's been a long time. And reading that now makes me think it was not a good idea.

PR #1826, opened 2026-02-28
Fix synchronization of metas when updating a meta with multiple values

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/2767

Also add minor performance improvement by avoiding useless hook in `update_metas()`.

### DIFF
--- a/tests/phpunit/tests/Meta/trait-Update.php
+++ b/tests/phpunit/tests/Meta/trait-Update.php
@@ -33,7 +33,7 @@ trait Update {
 
 		$this->assertTrue( update_metadata( self::$type, self::$objects['en'], 'the_key', $value, 'another_value' ) );
 
-		$this->assertEqualSets(
+		$this->assertEqualSetsWithIndex(
 			array( $value, 'yet_another_value' ),
 			get_metadata( self::$type, self::$objects['fr'], 'the_key', false )
 		);


tests/phpunit/tests/Meta/trait-Update.php
```diff
@@ -21,6 +21,24 @@ public function test_update_meta( $value ) {
...
+	 * @param mixed $value The value to add to the meta.
+	 * @return void
+	 */
+	public function test_update_meta_with_several_values( $value ) {
+		add_metadata( self::$type, self::$objects['en'], 'the_key', 'another_value' );
+		add_metadata( self::$type, self::$objects['en'], 'the_key', 'yet_another_value' );
+
+		$this->assertTrue( update_metadata( self::$type, self::$objects['en'], 'the_key', $value, 'another_value' ) );
+
+		$this->assertEqualSets(
```
Hug0-Drelon:
```suggestion
		$this->assertSameSetsWithIndex(
```

src/modules/sync/sync-metas.php
```diff
@@ -276,10 +254,9 @@ public function update_meta( $mid, $id, $meta_key, $meta_value ) {
 						}
 					}
 				}
-				$this->restore_add_meta_action();
+				$this->add_all_meta_actions();
 			}
 
-			unset( $this->prev_value[ $hash ] );
```
Hug0-Drelon, line 282: If I understood correctly, this line is the source of the bug. But was it preventing another one?

Chouby: Not according to my tests. Probably it was added by "reflex" to clean up things, without enough tests on the consequences.

tests/phpunit/tests/Meta/trait-Update.php
Chouby: Added and reverted.
This doesn't work for object (expected).
Also the number and boolean are converted to string.

Hug0-Drelon: But we do care about the list order, am I right?
If so, should we consider `assertEqualSetsWithIndex`?

Chouby: Done. I didn't know that WP orders metas. But it does. Since v3.8. For the record: https://core.trac.wordpress.org/ticket/25511

PR #1828, opened 2026-03-09
Remove `$title` in `PLL_Language::get_flag_html()`

### LINKED ISSUE #1547: Filter pll_flag_title not work
## Prerequisites

- [ x] This is not a usage question (Those should be directed to the [community supported forum](https://wordpress.org/support/plugin/polylang), unless this is a question about Polylang Pro in which case you should [use the helpdesk](https://polylang.pro/support/)).
- [x ] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x ] The issue still exists against the latest `master` branch of Polylang and the latest WordPress version.
- [x ] This bug happens with only Polylang plugin active
- [x ] I can reproduce this bug consistently

## My environment

1. WordPress 6.6.2
2. Polylang 3.6.4

I use Filter `pll_flag_title` to custom Title of the language switcher but it doesn't work.
I tried with your sample code, but the Title still doesn't change as I want.
Please check
Thanks
🐻

Chouby:
Closes #1547 

The parameter `$title` is not used anymore. So I propose to remove it. As it's the second out of 3 and several 3rd party plugins use the method, I added a backward compatibility in case the 3rd argument is used. However if a developer uses only the second argument for an expected title, it will be used as alt text (since #579, nothing's displayed in this case).

The PR also removes the useless filter `pll_flag_title` as its output is not used since #579.

### DIFF
--- a/changelog/1828
+++ b/changelog/1828
@@ -1,1 +1,1 @@
-* Remove the filter `pll_flag_title`, inoperant since v3.0 #1828
+* Remove the filter `pll_flag_title`, inoperative since v3.0 #1828


src/language.php
```diff
@@ -435,11 +435,18 @@ public static function get_flag_information( $code ) {
...
+	public static function get_flag_html( $flag, $alt = '' ) {
 		if ( empty( $flag['src'] ) ) {
 			return '';
 		}
 
+		// Backward compatibility.
+		if ( func_num_args() > 2 ) {
+			_deprecated_argument( __METHOD__ . '()', '3.9', 'The parameter `$title` has been removed and replaced by `$alt` in second position' );
+			/** @var string */
+			$alt = func_get_arg( 2 );
```
Screenfeed, line 447: I think we should make sure to not overwrite any given alt:

```suggestion
			if ( empty( $alt ) ) {
				/** @var string */
				$alt = func_get_arg( 2 );
			}
```

Chouby: If we have 3 arguments (the old way), we are supposed to receive the title in `$alt` (the second argument) and the alt text should be in the 3rd argument, so I think we should intentionally overwrite `$alt`.

Screenfeed: You're right.

Issue #1836, opened 2026-03-17
Fatal error: Uncaught TypeError: array_merge()


Division70:
Our monitoring systems detected the following PHP error on  site:
2026/03/17 11:53:47 [error] 113518#113518: *2465348 FastCGI sent in stderr: "PHP message: PHP Fatal error: Uncaught TypeError: array_merge(): Argument #2 must be of type array, bool given in /public/wp-content/plugins/polylang/src/translatable-object.php:267

PR #1837, opened 2026-03-17
Prevent fatal when reading from cache in `PLL_Translatable_Object`.

Hug0-Drelon:
Fixes https://github.com/polylang/polylang/issues/1836

## What?
A fatal has been reported when updating to 3.8, see https://wordpress.org/support/topic/fatal-error-after-updating-to-verison-3-8-from-3-7-8/#post-18854300

## How?
Somehow, users have badly populated cache. I suspect that having `WP_DEBUG` set to true would show the problem because of consecutive call to `_get_non_cached_ids` and `_validate_cache_id` in `PLL_Translatable_Object::prime_object_term_cache()`.

So the proposed fixed only ensure we get correct values from the cache.

### DIFF
--- a/src/translatable-object.php
+++ b/src/translatable-object.php
@@ -349,7 +349,7 @@ abstract class PLL_Translatable_Object {
 			}
 
 			$id = reset( $term_ids );
-			if ( ! is_numeric( $id ) ) {
+			if ( ! is_numeric( $id ) || empty( $id ) ) {
 				continue;
 			}
 


src/translatable-object.php
```diff
@@ -261,22 +261,35 @@ protected function get_object_terms( array $object_ids, string $taxonomy ) {
...
+		if ( ! is_array( $cached_values ) ) {
+			return array();
+		}
+
 		// Flatten the array to prime the terms cache.
 		$all_term_ids = array();
 		foreach ( $cached_values as $term_ids ) {
+			if ( ! is_array( $term_ids ) ) {
+				continue;
+			}
```
Screenfeed: While at it, we could also return early if the list is empty. Also, if we remove the entry, no need to check again later.

```suggestion
		foreach ( $cached_values as $object_id => $term_ids ) {
			if ( ! is_array( $term_ids ) || empty( $term_ids ) ) {
				unset( $cached_values[ $object_id ] );
				continue;
			}
```

src/translatable-object.php
```diff
@@ -261,22 +261,35 @@ protected function get_object_terms( array $object_ids, string $taxonomy ) {
...
 		_prime_term_caches( $all_term_ids, false );
 
 		$terms = array();
 		foreach ( $cached_values as $object_id => $term_ids ) {
-			if ( ! empty( $term_ids ) ) {
-				$term_id = reset( $term_ids ); // There is only one term for language or translation groups.
+			if ( ! is_array( $term_ids ) || empty( $term_ids ) ) {
+				continue;
+			}
+
```
Screenfeed: With my previous suggestion, this would be superfluous.


```suggestion
```

src/translatable-object.php
```diff
@@ -340,10 +353,11 @@ protected function prime_object_term_cache( array $object_ids ) {
 	 * @param int[]  $object_ids Array of object IDs to retrieve terms for.
 	 * @param string $taxonomy   Taxonomy name.
 	 *
-	 * @return int[][]
+	 * @return mixed Ideally an array of int[], but can be anything else.
```
Screenfeed: Hahaha 🤣
When reality reminds us that we cannot trust the cache.

Hug0-Drelon: *Wordpress cache...
There are type-safe cache systems...

src/translatable-object.php
Hug0-Drelon: I'm not against but what's bother me is that we loop twice through the same array...
I don't know if it's the right place to change the logic here but...
Is it ok if we check with @Chouby?

src/translatable-object.php
Hug0-Drelon: And in reality, `_validate_cache_id` should do its job but if `WP_DEBUG` is `false` then it goes silent.

Screenfeed (approved):

src/translatable-object.php
```diff
@@ -261,22 +261,35 @@ protected function get_object_terms( array $object_ids, string $taxonomy ) {
 
 		$cached_values = $this->get_from_object_term_cache( $object_ids, $taxonomy );
 
+		if ( ! is_array( $cached_values ) ) {
+			return array();
+		}
```
Chouby: The test is useless as we always get an array. See https://github.com/WordPress/WordPress/blob/6.9.4/wp-includes/cache-compat.php#L89-L97

src/translatable-object.php
Chouby: Reading at your discussion, I would modify `get_object_term_cache()` instead.
```php
	protected function get_from_object_term_cache( array $object_ids, string $taxonomy ) {
		$this->prime_object_term_cache( $object_ids );

		$values = wp_cache_get_multiple( $object_ids, "{$taxonomy}_relationships" );
		
		foreach ( $values as $object_id => $term_ids ) {
			if ( empty( $term_ids ) ) { // Checking empty should be sufficient, but we can ckeck is_array() if you prefer.
				unset( $cached_values[ $object_id ] );
			}
		}
		
		return $values;
	}
```
That way we can trust that we have an `int[][]` as returned value (instead of `array<int[]|false>` and avoid all tests in this method. 

We can even envisage to do the `reset()` in this method to return `int[]`.

src/translatable-object.php
```diff
@@ -261,22 +261,35 @@ protected function get_object_terms( array $object_ids, string $taxonomy ) {
...
-				/** @var WP_Term $term */
-				$term                = get_term( $term_id );
-				$terms[ $object_id ] = $term;
+			if ( ! is_numeric( $term_id ) ) {
+				continue;
 			}
+
+			/** @var WP_Term $term */
+			$term                = get_term( (int) $term_id );
+			$terms[ $object_id ] = $term;
```
Chouby: I'm not convinced that this extra changes are useful.

src/translatable-object.php
Chouby: If we keep the method like this we return `array<int[]|false>` so we can be more precise than `mixed`.

tests/phpunit/tests/test-translated-post.php
```diff
@@ -363,4 +365,66 @@ public function test_should_not_update_translations_group_when_removing_all_tran
...
+			)
+		);
+
+		$terms = wp_get_object_terms( $posts['en'], 'language' );
+
+		$this->assertCount( 1, $terms );
+
+		$this->setExpectedIncorrectUsage( '_get_non_cached_ids' );
+
+		wp_cache_set( $posts['en'], array( 2 => false ), 'language_relationships' );
```
Chouby: These are the 2 lines meaningful in this test. So in my opinion, this doesn't test Polylang but WordPress (`_get_non_cached_ids`).

tests/phpunit/tests/test-translated-post.php
```diff
@@ -363,4 +365,66 @@ public function test_should_not_update_translations_group_when_removing_all_tran
...
+
+		$post_id = self::factory()->post->create(
+			array(
+				'lang' => 'en',
+			),
+		);
+
+		$terms = wp_get_object_terms( $post_id, 'language' );
+
+		$this->assertCount( 1, $terms );
```
Chouby: This does not seem necessary. The last 2 lines should be sufficient.

src/translatable-object.php
Screenfeed: In your suggestion, the `foreach` loop can be replaced by `array_filter()`.

tests/phpunit/tests/test-translated-post.php
```diff
@@ -363,4 +365,66 @@ public function test_should_not_update_translations_group_when_removing_all_tran
...
+	}
+
+	public function test_get_object_term_for_translation_should_not_fail_when_cache_corrupted() {
+		$posts = self::factory()->post->create_translated(
+			array(
+				'lang' => 'en',
+			),
+			array(
+				'lang' => 'fr',
+			)
```
Chouby: I guess the second post is not necessary for the test.

tests/phpunit/tests/test-translated-post.php
```diff
@@ -363,4 +365,66 @@ public function test_should_not_update_translations_group_when_removing_all_tran
...
+		);
+
+		$terms = wp_get_object_terms( $posts['en'], 'post_translations' );
+
+		$this->assertCount( 1, $terms );
```
Chouby: This doesn't seem necessary.

src/translatable-object.php
```diff
@@ -262,21 +262,14 @@ protected function get_object_terms( array $object_ids, string $taxonomy ) {
...
-				$term                = get_term( $term_id );
-				$terms[ $object_id ] = $term;
-			}
+		foreach ( $cached_values as $object_id => $term_id ) {
+			/** @var WP_Term $term */
+			$term                = get_term( $term_id );
+			$terms[ $object_id ] = $term;
 		}
 
 		return $terms;
```
Screenfeed, line 274: This would be simpler but then I guess PHPStan won't be happy:

```php
_prime_term_caches( array_values( $cached_values ), false );

return array_map( 'get_term', $cached_values );
```

By the way, the return type is incorrect in the doc now, it should be `array<int, WP_Term>`.

Hug0-Drelon: With your proposal we would lose the `$object_id` keys IMHO.

I'm looking at the doc.

Hug0-Drelon: Updated in da02fdee.

Screenfeed: We don't lose it but OK.

PR #1839, opened 2026-03-18
Fix translatable objects when cache is suspended

Chouby:
Follow-up of #1837

#1837 fixed the fatal error but kept the underlying bug, i.e. if nothing was found in the cache (previously primed), then nothing was returned.

In this PR, we first hit the cache. It will returns something like:
```
array (
  3 => false, // cache missed.
  7 => 
  array (
    0 => 13,  // found in cache.
  ),
)
```
Then call the new `update_object_term_cache()` which updates the cache for non cached ids and returns the corresponding object terms (which are added to the cache for future use). Here something like:
```
array (
  'language' => 
  array (
    3 => 
    array (
      0 => 13, 
    ),
  ),
  'post_translations' => 
  array (
    3 => 
    array (
    ),
  ),
)
```
We then replace the cache missed values by the freshly queried object terms for the given taxonomy.

The test with suspended cache is fixed.
A new test is added to handle the case of objects without language.

### DIFF
--- a/src/translatable-object.php
+++ b/src/translatable-object.php
@@ -290,7 +290,7 @@ abstract class PLL_Translatable_Object {
 		}
 
 		if ( empty( $non_cached_ids ) ) {
-			return array_fill_keys( $this->tax_to_cache, array() );
+			return array();
 		}
 
 		$terms = wp_get_object_terms(
@@ -303,7 +303,7 @@ abstract class PLL_Translatable_Object {
 		);
 
 		if ( ! is_array( $terms ) ) {
-			return array_fill_keys( $this->tax_to_cache, array() );
+			return array();
 		}
 
 		$object_terms = array();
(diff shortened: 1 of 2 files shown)


src/translatable-object.php
```diff
@@ -335,12 +337,8 @@ protected function prime_object_term_cache( array $object_ids ) {
 	 * @return int[] Array of term IDs with object ID as key.
 	 */
 	protected function get_from_object_term_cache( array $object_ids, string $taxonomy ): array {
-		$this->prime_object_term_cache( $object_ids );
 		$values = wp_cache_get_multiple( $object_ids, "{$taxonomy}_relationships" );
-
-		if ( ! is_array( $values ) ) {
-			return array();
-		}
+		$values = array_replace( $values, $this->update_object_term_cache( $object_ids )[ $taxonomy ] );
```
Hug0-Drelon: I think I prefer the former proposal ([eb49e11](https://github.com/polylang/polylang/pull/1839/commits/eb49e110b8d0e2435b0ad6df8aa270a59cd26c96)). As you pointed out, it prevent to have an extra argument useful only for one class.

That said, I think this line deserve a comment explaining "why".
Also, we shouldn't we check `$taxonomy` key existence before using the value?

PR #1843, opened 2026-03-26
Compute default language when the option is not set.

Hug0-Drelon:
## What?
Fixes https://github.com/polylang/polylang-pro/issues/2962

## How?
Compute default language in `Model\Languages::get_default`  when `default_lang` option is empty.
This ensures that `PLL_Choose_Lang::get_preferred_language` returns a `PLL_Language` object (preventing the fatal error in the issue).

### DIFF
--- a/src/Model/Languages.php
+++ b/src/Model/Languages.php
@@ -734,10 +734,6 @@ class Languages {
 		}
 
 		$default_term = reset( $terms );
-		if ( empty( $default_term ) ) {
-			return false;
-		}
-
 		$this->update_default( $default_term->slug );
 
 		return $this->get( $this->options['default_lang'] );


tests/phpunit/tests/Model/Languages/test-CRUD.php
```diff
@@ -12,13 +12,14 @@
...
+	protected $options;
 
 	public function set_up() {
 		parent::set_up();
 
-		$options         = self::create_options();
+		$this->options         = self::create_options();
 		$this->languages = new Languages(
-			$options,
+			$this->options,
```
Screenfeed:
```suggestion
		$this->options   = self::create_options();
		$this->languages = new Languages(
			$this->options,
```

Hug0-Drelon: GitHub won't apply the commit :/
See ea7382d3

src/Model/Languages.php
```diff
@@ -723,10 +723,22 @@ public function set_ready(): void {
...
+
+		$terms = $this->get_terms();
+		if ( empty( $terms ) ) {
 			return false;
 		}
 
+		$default_term = reset( $terms );
+		if ( empty( $default_term ) ) {
+			return false;
+		}
```
Chouby: This test (or the previous one) is useless. if `$terms` is empty, `reset()` returns false and if `reset()` returns false, then the array is empty.

src/Model/Languages.php
```diff
@@ -723,10 +723,22 @@ public function set_ready(): void {
 	 * @return PLL_Language|false Default language object, `false` if no language found.
 	 */
 	public function get_default() {
-		if ( empty( $this->options['default_lang'] ) ) {
+		if ( ! empty( $this->options['default_lang'] ) ) {
+			return $this->get( $this->options['default_lang'] );
+		}
+
+		$terms = $this->get_terms();
```
Chouby, line 731:
```suggestion
		// It happens that there the default language is lost. We require one, so let's select one arbitrarily.
		$terms = $this->get_terms();
```

src/Model/Languages.php
```diff
@@ -723,10 +723,22 @@ public function set_ready(): void {
...
 		}
 
+		$default_term = reset( $terms );
+		if ( empty( $default_term ) ) {
+			return false;
+		}
+
+		$this->update_default( $default_term->slug );
+
 		return $this->get( $this->options['default_lang'] );
```
Chouby, line 739:
```suggestion
		return $this->get( $default_term->slug );
```
I still wonder how the default language could disappear. One way I use is:
```php
add_filter( 'option_polylang', function( $options ) {
	unset( $options['default_lang'] );
	return $options;
} );
```
which would still break the things if we persist to get the value from the option.

src/Model/Languages.php
Hug0-Drelon: See b97597f8

src/Model/Languages.php
Hug0-Drelon: `Languages::update_default()` updates the option object, not the value in database directly (done on shutdown on normal installation).
So here we read from `Options` which must be up to date.

src/Model/Languages.php
```diff
@@ -723,10 +723,19 @@ public function set_ready(): void {
 	 * @return PLL_Language|false Default language object, `false` if no language found.
 	 */
 	public function get_default() {
-		if ( empty( $this->options['default_lang'] ) ) {
+		if ( ! empty( $this->options['default_lang'] ) ) {
+			return $this->get( $this->options['default_lang'] );
+		}
+
+		// It happens that there the default language is lost. We require one, so let's select one arbitrarily.
```
szepeviktor, line 730: The comment can be improved:

> If no default language is configured but languages still exist, restore one from the available language terms.

PR #1844, opened 2026-03-29
Fix conflict with Yoast Duplicate Post

### LINKED ISSUE #1536: Pro: Yoast SEO Rewrite & Republish on Synched Translations deletes the post
## Prerequisites

This issue is unrelated to #813.

1. Setup a clean WordPress with Polylang Pro and Duplicate Posts (by Yoast) installed.
2. Make sure at least two langauges are available.
3. Create a new post with a synched translation.
4. Use the "Rewrite & Republish" feature of Duplicate Posts on your new post.
5. "Republish" your new Post Draft.

Due to the synchronisation of the posts, all of the Duplicate Posts related meta data is copied from the new draft back to the original post. This leads to a fatal behaviour when executing "Republish" – Duplicate Posts replaces the original post with the Draft (which is expected behaviour) but then continues to delete the old draft _and_ the original post as well, since the meta data marks it as a duplicate of itself.

This can be observed also by deleting the draft instead of republishing it. Upon page reload you can see that the original post is labeled as "Rewrite & Republish" of _itself_ in the list.

I don't know if excluding all the Duplicate Post metadata from the translation process would be enough to fix this problem.

In `polylang-pro/vendor/wpsyntex/polylang/integrations/duplicate-post/duplicate-post.php`:

```php
add_filter( 'pll_copy_post_metas', 'exclude_duplicate_post_meta' );

function exclude_duplicate_post_meta( array $keys ) {
    $keys_to_remove = [
         "_dp_original",
         "_dp_is_rewrite_republish_copy",
         "_dp_has_rewrite_republish_copy",
         "_dp_creation_date_gmt"
    ];

    return array_diff( $keys, $keys_to_remove );
}
```

Alternatively, the function "Rewrite & Republish" should be disabled or come with a warning in conjunction with synched translations in Polylang Pro. Unexpected post deletion is a big problem a client of us had in their production environment due to this bug.

Chouby:
Fix #1536 

Fix synchronized posts deleted when using "Rewrite & Republish" by excluding post metas from the copy/synchronizations as suggested.

Also fixes the broken translations group when a post is republished.
To reproduce :
1. Create 2 translated posts (not necessarily synchronized), Let say English and French.
2. In the list table click on "Rewrite & Republish", let say for the English post. 
3. Once the editor is open, click on Republish.
4. Visit the posts list table and see how the French post has lost its link to the English post. 
image

To fix the issue, I had to partially revert a28f803d0a79ee6e0c535060cd7978e8b741a776. Indeed Duplicate Post breaks the translations group, but it's not possible to repair it (seen from the English post, the translations look ok, so if we attempted to repair it `should_update_translation_group()` would return false). So need to always update the translation group if we want to be able to repair it (I excluded only one case for terms when the translation group only includes one term). See [the test](https://github.com/polylang/polylang/blob/3.8.2/tests/phpunit/tests/test-translated-term.php#L223) which pushed me to exclude this specific case.

### DIFF
--- a/src/integrations/duplicate-post/duplicate-post.php
+++ b/src/integrations/duplicate-post/duplicate-post.php
@@ -48,7 +48,7 @@ class PLL_Duplicate_Post {
 	 * @param string[] $keys List of meta keys.
 	 * @return string[]
 	 */
-	public function exclude_post_metas( $keys ) {
+	public function exclude_post_metas( $keys ): array {
 		$to_remove = array(
 			'_dp_original',
 			'_dp_is_rewrite_republish_copy',
@@ -68,7 +68,7 @@ class PLL_Duplicate_Post {
 	 * @param int $post_id The original post's ID.
 	 * @return void
 	 */
-	public function after_rewriting( $copy_id, $post_id ) {
+	public function after_rewriting( $copy_id, $post_id ): void {
 		$language     = pll_get_post_language( $post_id );
 		$translations = pll_get_post_translations( $post_id );
 
--- a/src/integrations/duplicate-post/duplicate-post.php
+++ b/src/integrations/duplicate-post/duplicate-post.php
@@ -16,6 +16,7 @@ class PLL_Duplicate_Post {
 	 */
 	public function init() {
 		add_filter( 'option_duplicate_post_taxonomies_blacklist', array( $this, 'taxonomies_blacklist' ) );
+		add_filter( 'pll_copy_post_metas', array( $this, 'exclude_post_metas' ) );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


Screenfeed (changes requested):
Sadly this doesn't solve the issue entirely.

If 2 posts are synced and we use "Rewrite & Republish", it breaks the synchronization group. This can be seen in 2 ways:
1. when republishing: we can see that the sync group is broken when we visit the edit page of the translation (the field is empty).
2. when drafting the "republished" post: the synced post is also drafted and it becomes a copy of the republished post.

Chouby:
See the updated description.

src/integrations/duplicate-post/duplicate-post.php
```diff
@@ -34,4 +37,42 @@ public function taxonomies_blacklist( $taxonomies ) {
...
+	 * Exclude Duplicate Post metas from the copy/synchronization.
+	 *
+	 * This avoids synchronized posts to be deleted when using "Rewrite & Republish".
+	 *
+	 * @since 3.9
+	 *
+	 * @param string[] $keys List of meta keys.
+	 * @return string[]
+	 */
+	public function exclude_post_metas( $keys ) {
```
Screenfeed:
```suggestion
	public function exclude_post_metas( $keys ): array {
```

src/integrations/duplicate-post/duplicate-post.php
```diff
@@ -34,4 +37,42 @@ public function taxonomies_blacklist( $taxonomies ) {
...
+	/**
+	 * Fixes the translations group just after a post is republished.
+	 *
+	 * @since 3.9
+	 *
+	 * @param int $copy_id The copy's ID.
+	 * @param int $post_id The original post's ID.
+	 * @return void
+	 */
+	public function after_rewriting( $copy_id, $post_id ) {
```
Screenfeed:
```suggestion
	public function after_rewriting( $copy_id, $post_id ): void {
```

Screenfeed (changes requested):

Screenfeed (approved):

intelligence:
Is this slated for the next release?

Screenfeed:
It is scheduled for the version 3.9.

PR #1853, opened 2026-04-08
Fix error when the WXR import files has synchronized posts

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/2980

This PR fixes 2 bugs:
```
PHP Fatal error:  Uncaught TypeError: Cannot access offset of type array in isset or empty in /src/integrations/wp-importer/wp-import.php:197
```
This one is caused by the presence of additional data in the term description for the translation group. The importer expects only the array of translations. But it my contain synchronization information as well as 3rd party data. This is fixed by whitelisting the keys that we manipulate when remapping translation groups. A pair of synchronized posts are added to the update test import file to test the issue.  
 
```
PHP Warning:  Array to string conversion in /src/integrations/wp-importer/wp-import.php on line 163
```
This is a regression introduced by #1577. The `array_diff()` used to compare an array of strings when we have now an array of arrays. Using `array_udiff()` to convert arrays to strings before the comparison fixes the issue.

### DIFF
--- a/src/integrations/wp-importer/wp-import.php
+++ b/src/integrations/wp-importer/wp-import.php
@@ -132,50 +132,50 @@ class PLL_WP_Import extends WP_Import {
 	protected function remap_terms_relations( &$terms ) {
 		global $wpdb;
 
-		$trs = array();
+		$term_relationships = array();
 
 		foreach ( $terms as $term ) {
 			$translations = maybe_unserialize( $term['term_description'] );
 			foreach ( $translations as $slug => $old_id ) {
 				if ( $old_id && ! empty( $this->processed_terms[ $old_id ] ) && $lang = PLL()->model->get_language( $slug ) ) {
 					// Language relationship.
-					$trs[] = array( $this->processed_terms[ $old_id ], $lang->get_tax_prop( 'term_language', 'term_taxonomy_id' ) );
+					$term_relationships[] = array( $this->processed_terms[ $old_id ], $lang->get_tax_prop( 'term_language', 'term_taxonomy_id' ) );
 
 					// Translation relationship.
-					$trs[] = array( $this->processed_terms[ $old_id ], get_term( $this->processed_terms[ $term['term_id'] ], 'term_translations' )->term_taxonomy_id );
+					$term_relationships[] = array( $this->processed_terms[ $old_id ], get_term( $this->processed_terms[ $term['term_id'] ], 'term_translations' )->term_taxonomy_id );
 				}
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


sebastienserre:
This PR fixes the issue.

src/integrations/wp-importer/wp-import.php
```diff
@@ -160,7 +160,13 @@ protected function remap_terms_relations( &$terms ) {
 				$existing_trs[ $key ] = array( $tr->object_id, $tr->term_taxonomy_id );
 			}
 
-			$trs = array_diff( $trs, $existing_trs );
+			$trs = array_udiff(
+				$trs,
+				$existing_trs,
+				function ( $a, $b ) {
+					return strcmp( implode( ',', $a ), implode( ',', $b ) ); // An easy way to compare arrays (ok if values are int or numeric strings).
```
Hug0-Drelon, line 167:
```suggestion
					return $a === $b ? 0 : -1;
```
Isn't it easier?

src/integrations/wp-importer/wp-import.php
```diff
@@ -177,31 +183,33 @@ protected function remap_terms_relations( &$terms ) {
...
-	 * @param array $processed_objects array of posts or terms processed by WordPress Importer
+	 * @param array $terms             Array of terms in 'post_translations' or 'term_translations' taxonomies?
+	 * @param array $processed_objects Array of posts or terms processed by WordPress Importer.
 	 */
 	protected function remap_translations( &$terms, &$processed_objects ) {
 		global $wpdb;
 
-		$u = array();
+		$languages = pll_languages_list();
+		$u         = array();
```
Hug0-Drelon: Can we take the opportunity to rename this variable?

Hug0-Drelon (changes requested):

src/integrations/wp-importer/wp-import.php
Chouby: It wouldn't work if one array has numeric strings and the other has integers. That's the reason of the trick (modify my comment if it's not clear enough).

src/integrations/wp-importer/wp-import.php
Chouby: I started to code you wasn't born :)

src/integrations/wp-importer/wp-import.php
Hug0-Drelon: Thank you, it's clearer once you put it that way.

Hug0-Drelon (approved):

PR #1857, opened 2026-04-09
Harmonize flags in admin

Screenfeed:
Fixes partially https://github.com/polylang/polylang-pro/issues/2950 (similar work has been done in [PLL Pro](https://github.com/polylang/polylang-pro/pull/2992)).

## How

This introduces the method `PLL_Language::get_admin_flag()`, which is the admin version of `PLL_Language::get_display_flag()`. It is meant to be used in admin in two contexts:

1. `PLL_Language::get_admin_flag()` when the flag is not in a link: the flag must convey the language information. In this case, the attributes `lang` and `dir` are added because the `alt` attribute contains the name of the language.
2. `PLL_Language::get_admin_flag( 'minimal' )` when the flag is displayed in a link (the link already contain the information to convey) or next to text that already conveys the information: in this case the flag is only decorative. Note: in this case the attribute `tabindex="-1"` is added but I'm not entirely sure it is useful since these tags should not be focusable to start with.

Also important, if the flag is empty then a `<abbr>` tag is displayed instead.

## Example markups

In the following examples the `style`, `width`, and `height` attributes have been removed to make the markup more readable but are still printed.

### Posts list: table heading

The flag is not in a link, it must convey the information (via the `alt` attribute):

```html
<th scope="col" id="language_fr" class="manage-column column-language_fr">
	<img lang="fr-FR" dir="ltr" alt="Français" src="...">
</th>
```

If the flag is not available, the information is conveyed via the `title` attribute:

```html
<th scope="col" id="language_fr" class="manage-column column-language_fr">
	<abbr lang="fr-FR" dir="ltr" title="Français">FR</abbr>
</th>
```

### Posts list: cell content

The flag is in a link. The link has a `title` attribute and the link contains a "screen reader text". The flag is hidden to screen readers:

```html
<a href="..." class="pll_column_flag" title="Un article">
	<span class="screen-reader-text">Modifier cet élément en Français</span>
	<img aria-hidden="true" tabindex="-1" src="..." alt="Français">
</a>
```

With no flag:

```html
<a href="..." class="pll_column_flag" title="Un article">
	<span class="screen-reader-text">Modifier cet élément en Français</span>
	<abbr aria-hidden="true" tabindex="-1">FR</abbr>
</a>
```

### Media editor: classic metabox

The flag is not in a link but it preceeds language information. In this case I think the flag should not convey information:

```html
<td class="pll-media-language-column">
	<span class="pll-translation-flag">
		<img aria-hidden="true" tabindex="-1" src="..." alt="English">
	</span>
	English
</td>
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -933,7 +933,7 @@ parameters:
 		-
 			message: '#^@readonly property cannot have a default value\.$#'
 			identifier: property.readOnlyByPhpDocDefaultValue
-			count: 8
+			count: 7
 			path: src/language.php
 
 		-
@@ -952,10 +952,6 @@ parameters:
 			message: '#^Property PLL_Language\:\:\$fallbacks \(list\<non\-empty\-string\>\) does not accept array\<non\-empty\-string\>\.$#'
 			identifier: assign.propertyType
 			count: 1
-			path: src/language.php
-
-		-
-			message: '#^@readonly property PLL_Language\:\:\$admin_flag is assigned outside of the constructor\.$#'
 			path: src/language.php
 
 		-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -933,7 +933,7 @@ parameters:
 		-
 			message: '#^@readonly property cannot have a default value\.$#'
 			identifier: property.readOnlyByPhpDocDefaultValue
-			count: 7
+			count: 8
 			path: src/language.php
 
 		-
@@ -952,6 +952,10 @@ parameters:
 			message: '#^Property PLL_Language\:\:\$fallbacks \(list\<non\-empty\-string\>\) does not accept array\<non\-empty\-string\>\.$#'
 			identifier: assign.propertyType
 			count: 1
+			path: src/language.php
+
+		-
+			message: '#^@readonly property PLL_Language\:\:\$cache is assigned outside of the constructor\.$#'
 			path: src/language.php
 
 		-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -955,7 +955,7 @@ parameters:
 			path: src/language.php
 
 		-
-			message: '#^@readonly property PLL_Language\:\:\$cache is assigned outside of the constructor\.$#'
+			message: '#^@readonly property PLL_Language\:\:\$admin_flag is assigned outside of the constructor\.$#'
 			path: src/language.php
 
 		-
--- a/src/admin/admin-classic-editor.php
+++ b/src/admin/admin-classic-editor.php
@@ -265,7 +265,7 @@ class PLL_Admin_Classic_Editor {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => empty( $lang->flag ) ? esc_html( $lang->slug ) : $lang->flag ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag() ) );
 
 		// Sample permalink
 		$x->Add( array( 'what' => 'permalink', 'data' => get_sample_permalink_html( $post->ID ) ) );
--- a/src/admin/admin-classic-editor.php
+++ b/src/admin/admin-classic-editor.php
@@ -265,7 +265,7 @@ class PLL_Admin_Classic_Editor {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag() ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag( 'no-screen-reader' ) ) );
 
 		// Sample permalink
 		$x->Add( array( 'what' => 'permalink', 'data' => get_sample_permalink_html( $post->ID ) ) );
--- a/src/admin/admin-classic-editor.php
+++ b/src/admin/admin-classic-editor.php
@@ -265,7 +265,7 @@ class PLL_Admin_Classic_Editor {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag( 'no-screen-reader' ) ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag( 'aria-hidden' ) ) );
 
 		// Sample permalink
 		$x->Add( array( 'what' => 'permalink', 'data' => get_sample_permalink_html( $post->ID ) ) );
--- a/src/admin/admin-filters-term.php
+++ b/src/admin/admin-filters-term.php
@@ -509,7 +509,7 @@ class PLL_Admin_Filters_Term {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => empty( $lang->flag ) ? esc_html( $lang->slug ) : $lang->flag ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag() ) );
 
 		$x->send();
 	}
--- a/src/admin/admin-filters-term.php
+++ b/src/admin/admin-filters-term.php
@@ -509,7 +509,7 @@ class PLL_Admin_Filters_Term {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag() ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag( 'no-screen-reader' ) ) );
 
 		$x->send();
 	}
--- a/src/admin/admin-filters-term.php
+++ b/src/admin/admin-filters-term.php
@@ -509,7 +509,7 @@ class PLL_Admin_Filters_Term {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag( 'no-screen-reader' ) ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag( 'aria-hidden' ) ) );
 
 		$x->send();
 	}
--- a/src/language.php
+++ b/src/language.php
@@ -475,6 +475,51 @@ class PLL_Language extends PLL_Language_Deprecated {
 	}
 
 	/**
+	 * Returns the language flag or the language slug if there is no flag.
+	 *
+	 * @since 3.9
+	 *
+	 * @param string $context Optional. Allows to modify the markup depending on how the flag is used. Possible values are:
+	 *                        - `full`: the flag has everything,
+	 *                        - `minimal`: the flag is preceded or followed by a text that would make it redundant.
+	 *                        Default is `full`.
+	 * @return string
+	 *
+	 * @phpstan-param 'full'|'minimal' $context
+	 */
+	public function get_admin_flag( string $context = 'full' ): string {
+		if ( ! empty( $this->flag ) ) {
+			if ( 'minimal' === $context ) {
+				// Hidden from screen readers.
+				return str_replace( '<img ', '<img aria-hidden="true" tabindex="-1" ', $this->flag );
+			}
+
+			return str_replace(
+				'<img ',
+				sprintf(
+					'<img lang="%1$s" dir="%2$s" ',
+					esc_attr( $this->get_locale( 'display' ) ),
+					$this->is_rtl ? 'rtl' : 'ltr'
+				),
+				$this->flag
+			);
+		}
+
+		if ( 'minimal' === $context ) {
+			// Hidden from screen readers.
+			return sprintf( '<abbr aria-hidden="true" tabindex="-1">%s</abbr>', esc_html( strtoupper( $this->slug ) ) );
+		}
+
+		return sprintf(
+			'<abbr title="%1$s" lang="%2$s" dir="%3$s">%4$s</abbr>',
+			esc_attr( $this->name ),
+			esc_attr( $this->get_locale( 'display' ) ),
+			$this->is_rtl ? 'rtl' : 'ltr',
+			esc_html( strtoupper( $this->slug ) )
+		);
+	}
+
+	/**
 	 * Returns the html of the custom flag if any, or the default flag otherwise.
 	 *
 	 * @since 2.8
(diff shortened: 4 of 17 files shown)


src/language.php
```diff
@@ -474,6 +474,51 @@ function ( &$value, $key ) {
...
+	 * @param string $context Optional. Allows to modify the markup depending on how the flag is used. Possible values are:
+	 *                        - `full`: the flag has everything,
+	 *                        - `minimal`: the flag is preceded or followed by a text that would make it redundant.
+	 *                        Default is `full`.
+	 * @return string
+	 *
+	 * @phpstan-param 'full'|'minimal' $context
+	 */
+	public function get_admin_flag( string $context = 'full' ): string {
+		if ( ! empty( $this->flag ) ) {
```
Hug0-Drelon: `PLL_Language::$flag` is documented as `@phpstan-var non-empty-string`. Either we should change the doc, or the check is useless.

src/language.php
```diff
@@ -474,6 +474,51 @@ function ( &$value, $key ) {
 		);
 	}
 
+	/**
+	 * Returns the language flag or the language slug if there is no flag.
+	 *
+	 * @since 3.9
+	 *
+	 * @param string $context Optional. Allows to modify the markup depending on how the flag is used. Possible values are:
```
Hug0-Drelon: The `$context` argument doesn't make sense to me.

What you want to achieve is to get either a screen-readable text or not. If you want a single method, I suggest to changes this argument name and its values.

Otherwise, it would make sense to have to separate methods I think.

src/language.php
```diff
@@ -474,6 +474,51 @@ function ( &$value, $key ) {
...
+	 *                        - `full`: the flag has everything,
+	 *                        - `minimal`: the flag is preceded or followed by a text that would make it redundant.
+	 *                        Default is `full`.
+	 * @return string
+	 *
+	 * @phpstan-param 'full'|'minimal' $context
+	 */
+	public function get_admin_flag( string $context = 'full' ): string {
```
Hug0-Drelon: I suggest to cache the output. It doesn't change much but it's called a shitload of times in the admin.

What do you think?

Hug0-Drelon (changes requested):
I suggest to add the following to `phpcs.xml.dist` to avoid all `phpcsignore`:
```xml
<rule ref="WordPress.Security.EscapeOutput">
	<properties>
		<property name="customAutoEscapedFunctions" type="array">
			<element value="get_admin_flag"/>
		</property>
	</properties>
</rule>
```

Screenfeed:
> I suggest to add the following to `phpcs.xml.dist` to avoid all `phpcsignore`

God idea but it doesn't work: it works only with functions, not class methods.

src/language.php
Screenfeed: Fixed in [4cb9eb2](https://github.com/polylang/polylang/pull/1857/commits/4cb9eb217a45a329a8ee05c7f534f39c33482911).
A language can have an empty flag actually (in the UI we can select "no flag").

src/language.php
Screenfeed: Agree. Done in [6b4835c](https://github.com/polylang/polylang/pull/1857/commits/6b4835c0c8ff9a64965676242b4eade173d99f8e).

src/language.php
Screenfeed: Why not. Done in [b2a684e](https://github.com/polylang/polylang/pull/1857/commits/b2a684e6d11f4fcc717cca45d795d546379647d9).

src/language.php
```diff
@@ -293,6 +292,9 @@ class PLL_Language extends PLL_Language_Deprecated {
 	 * @phpstan-param LanguageData $language_data
 	 */
 	public function __construct( array $language_data ) {
+		/** @phpstan-var LanguageData */
+		$language_data = array_diff_key( $language_data, array( 'cache' => array() ) );
```
Hug0-Drelon: I don't get why you need that. Is it really necessary?

src/language.php
```diff
@@ -560,6 +613,7 @@ public function get_locale( $filter = 'raw' ) {
 	 */
 	public function to_array( $context = 'display' ) {
 		$language = get_object_vars( $this );
+		$language = array_diff_key( $language, array( 'cache' => array() ) );
```
Hug0-Drelon: Making `PLL_Language::$cache` static would avoid this hack.

src/language.php
```diff
@@ -474,6 +476,57 @@ function ( &$value, $key ) {
...
+	 * @phpstan-param 'screen-reader'|'no-screen-reader' $mode
+	 */
+	public function get_admin_flag( string $mode = 'screen-reader' ): string {
+		if ( isset( $this->cache['admin_flag'][ $mode ] ) ) {
+			return $this->cache['admin_flag'][ $mode ];
+		}
+
+		if ( ! empty( $this->flag ) ) {
+			if ( 'no-screen-reader' === $mode ) {
+				// Hidden from screen readers.
```
Hug0-Drelon:
```suggestion
```
With such self documented mode name, the comment becomes unnecessary.

src/language.php
```diff
@@ -474,6 +476,57 @@ function ( &$value, $key ) {
...
+						$this->is_rtl ? 'rtl' : 'ltr'
+					),
+					$this->flag
+				);
+			}
+			return $this->cache['admin_flag'][ $mode ];
+		}
+
+		if ( 'no-screen-reader' === $mode ) {
+			// Hidden from screen readers.
```
Hug0-Drelon:
```suggestion
```
With such self documented mode name, the comment becomes unnecessary.

Hug0-Drelon (changes requested):

src/language.php
Screenfeed: 1. REALLY necessary, no, but I'd like to prevent such thing.
2. I just realized that it's much simpler to do `unset( $language_data['cache'] );` 🤣
3. As I alternative we could put it in `PLL_Language_Factory::sanitize_data()` but I'd prefer to keep it here.

What do you think?

src/language.php
Screenfeed: Maybe but then the values would be shared across objects, which makes no sense IMHO.
(And like I said in the previous comment, I can change it to `unset( $language['cache'] );`, which is much simpler.)

src/language.php
Screenfeed: Removed in [c811e3d](https://github.com/polylang/polylang/pull/1857/commits/c811e3da03518ebd52a6e732f7e7b26cda2daf8a) because PHPStan is annoying when it's not in the mood.

src/language.php
Screenfeed: Use `unset()` in [c811e3d](https://github.com/polylang/polylang/pull/1857/commits/c811e3da03518ebd52a6e732f7e7b26cda2daf8a).

src/language.php
Screenfeed: Removed in [d49d08d](https://github.com/polylang/polylang/pull/1857/commits/d49d08dfc7a26f520cab02762b8dad54ba7a516f).

src/language.php
Screenfeed: Removed in [d49d08d](https://github.com/polylang/polylang/pull/1857/commits/d49d08dfc7a26f520cab02762b8dad54ba7a516f).

src/language.php
```diff
@@ -474,6 +473,55 @@ function ( &$value, $key ) {
...
+	 * @phpstan-param 'screen-reader'|'no-screen-reader' $mode
+	 */
+	public function get_admin_flag( string $mode = 'screen-reader' ): string {
+		if ( isset( $this->cache['admin_flag'][ $mode ] ) ) {
+			return $this->cache['admin_flag'][ $mode ];
+		}
+
+		if ( ! empty( $this->flag ) ) {
+			if ( 'no-screen-reader' === $mode ) {
+				$this->cache['admin_flag'][ $mode ] = str_replace( '<img ', '<img aria-hidden="true" tabindex="-1" ', $this->flag );
```
Hug0-Drelon: Having an extra space in the needle could have side effect, am I right?
For instance, if the haystack has `` or something.

Hug0-Drelon (changes requested):
One last question, otherwise LGTM.

src/language.php
Screenfeed: OK, it can happen. [7976b62](https://github.com/polylang/polylang/pull/1857/commits/7976b6227b1d4bfaf091fa093aa020806deb554c)

Hug0-Drelon (approved):

src/admin/admin-classic-editor.php
```diff
@@ -265,7 +265,7 @@ public function post_lang_choice() {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => empty( $lang->flag ) ? esc_html( $lang->slug ) : $lang->flag ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag() ) );
```
Chouby: For me, this flag is purely decorative (in front of the select which includes all necessary info). I would hide it to screen readers.

src/admin/admin-filters-term.php
```diff
@@ -509,7 +509,7 @@ public function term_lang_choice() {
 		}
 
 		// Flag
-		$x->Add( array( 'what' => 'flag', 'data' => empty( $lang->flag ) ? esc_html( $lang->slug ) : $lang->flag ) );
+		$x->Add( array( 'what' => 'flag', 'data' => $lang->get_admin_flag() ) );
```
Chouby: Same as above. It's in front of a select.

src/settings/table-languages.php
```diff
@@ -73,8 +73,11 @@ public function column_default( $item, $column_name ) {
 			case 'count':
 				return $item->get_tax_prop( 'language', $column_name );
 
+			case 'flag':
+				return $item->get_admin_flag();
+
 			default:
-				return $item->$column_name; // Flag.
+				return $item->$column_name;
```
Chouby: I suggest to remove these 2 lines are they aren't useful anymore.

src/walker-dropdown.php
```diff
@@ -87,7 +87,7 @@ public function walk( $elements, $max_depth, ...$args ) { // // phpcs:ignore Wor
 			$lang = reset( $current );
 			$output = sprintf(
 				'<span class="pll-select-flag">%s</span>',
-				empty( $lang->flag ) ? esc_html( $lang->slug ) : $lang->flag
+				$lang->get_admin_flag()
```
Chouby: Same as above. The info is included in the select. I would hide it to screen readers.

src/language.php
```diff
@@ -474,6 +473,55 @@ function ( &$value, $key ) {
...
+	 *
+	 * @param string $mode Optional. Allows to modify the markup depending on how the flag is used. Possible values are:
+	 *                     - `screen-reader`: the flag can be seen by screen readers,
+	 *                     - `no-screen-reader`: the flag is hidden from screen readers: it is preceded or followed by a
+	 *                     text (language name for example) that would make it redundant. Default is `screen-reader`.
+	 * @return string
+	 *
+	 * @phpstan-param 'screen-reader'|'no-screen-reader' $mode
+	 */
+	public function get_admin_flag( string $mode = 'screen-reader' ): string {
```
Chouby: Maybe it's as simple to have 2 different methods, let say: `get_admin_flag()` and `get_aria_hidden_flag()`. What do you think?

If you prefer to keep the param, I suggest `'' | 'aria-hidden'`.

src/language.php
```diff
@@ -474,6 +473,55 @@ function ( &$value, $key ) {
...
+		if ( isset( $this->cache['admin_flag'][ $mode ] ) ) {
+			return $this->cache['admin_flag'][ $mode ];
+		}
```
Chouby: I suggest to simplify with `$this->admin_flag[ $mode ]`
Or have 2 different properties `$this->admin_flag` and `$this->aria_hidden_flag` in case the method is split.

phpstan-baseline.neon
```diff
@@ -954,6 +954,10 @@ parameters:
 			count: 1
 			path: src/language.php
 
+		-
+			message: '#^@readonly property PLL_Language\:\:\$cache is assigned outside of the constructor\.$#'
+			path: src/language.php
+
```
Chouby: This new error pushes to build the properties in the language factory to keep a truly immutable object. The method(s) would just then be getter(s) for the private property.

Chouby (changes requested):

src/admin/admin-classic-editor.php
Screenfeed: Indeed, the selected value in the following selector contains the information. Done in [ae6fcf3](https://github.com/polylang/polylang/pull/1857/commits/ae6fcf3f7e1c9ed7d42cd0bf17b70381b04b8dce).

src/language.php
Screenfeed: Renamed the values in [13bfe09](https://github.com/polylang/polylang/pull/1857/commits/13bfe096df089160bf0408b1159e9b45ad4f3e9c).

src/language.php
Screenfeed: Done in [44241f8](https://github.com/polylang/polylang/pull/1857/commits/44241f8a11bee3043f63c0872bb155d43b198c83).

src/settings/table-languages.php
Screenfeed: Done in [27578fc](https://github.com/polylang/polylang/pull/1857/commits/27578fcb6e0ca1b06c97683eee4cc9f971bdf3ac).

Screenfeed: It can't be removed entirely because PHPStan wants us to return something.

phpstan-baseline.neon
Screenfeed: Will be handled in #1880.

Chouby: Not this one. #1880 aims to handle properties defined with default values. But all are set in the constructor. 

For this new property, PHPStan protests because it is set outside the constructor. And thus the object is no more immutable. That's why I proposed to build the property in the language factory as done currently for all other properties.

Screenfeed: Sorry I misunderstood, I forgot there were 2 issues reported by PHPStan. Done in [6f9f3aa](https://github.com/polylang/polylang/pull/1857/commits/6f9f3aafa22f128a02eb0848fd2797259fc53650).

PR #1858, opened 2026-04-09
Use `PLL_Language::get_predefined_flag()`

Screenfeed:
Follows #1864.

This PR calls `PLL_Language::get_predefined_flag()` instead of `PLL_Language::get_flag_html( PLL_Language::get_flag_information() )`, which is exactly what it does.

I also took the initiative to move the method up in the class, next to the methods it uses.

### DIFF
--- a/src/settings/view-tab-lang.php
+++ b/src/settings/view-tab-lang.php
@@ -130,7 +130,7 @@ defined( 'ABSPATH' ) || exit;
 								printf(
 									'<option value="%s" data-flag-html="%s"%s>%s</option>' . "\n",
 									esc_attr( $code ),
-									esc_html( PLL_Language::get_flag_html( PLL_Language::get_flag_information( $code ) ) ),
+									esc_html( PLL_Language::get_predefined_flag( $code ) ),
 									selected( isset( $edit_lang->flag_code ) && $edit_lang->flag_code === $code, true, false ),
 									esc_html( $label )
 								);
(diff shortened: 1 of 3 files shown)


src/settings/view-tab-lang.php
```diff
@@ -127,15 +127,16 @@
 							<?php
 							$flags = include __DIR__ . '/flags.php';
 							foreach ( $flags as $code => $label ) {
-								$flag = PLL_Language::get_flag_information( $code );
-								if ( empty( $flag['url'] ) ) {
+								$flag_html = PLL_Language::get_predefined_flag( $code );
+
+								if ( empty( $flag_html ) ) {
```
Marianne380, line 132: Is the behavior change in the `continue` condition intentional?

If a third-party sets `src` without `url` via `pll_flag`, the old code would skip the option while the new one would display it.

Marianne380 (changes requested):

src/settings/view-tab-lang.php
Screenfeed: It's intentional yes.

`$flag['url']` is used only [here](https://github.com/polylang/polylang/blob/40fbc54c55d04fbd35f4f746fb0afec06963bcdc/src/language.php#L422-L426): `get_flag_html()` doesn't use it. So, to me, checking for `$flag['src']` makes more sense than checking `$flag['url']`.

Maybe that's something that @Chouby should check out when doing his review.

Marianne380 (approved):

PR #1859, opened 2026-04-10
Add E2E test for language switcher block crashing in widget editor.

Marianne380:
Following https://github.com/polylang/polylang/pull/1856.

## What

Add a E2E regression test that inserts the language switcher block in the widget editor and verifies no error appears.


tests/e2e/specs/widgets/language-switcher-widget.spec.js
```diff
@@ -0,0 +1,102 @@
...
+	{ tag: [ '@widget-editor' ] },
+	() => {
+		/**
+		 * Before all tests:
+		 *     - Activate a theme with widget areas.
+		 *     - Create en_US and fr_FR languages.
+		 */
+		test.beforeAll( async ( { requestUtils } ) => {
+			// Activate a theme with widget areas.
+			await requestUtils.activateTheme( 'twentytwentyone' );
```
Hug0-Drelon, line 31: I think you need to activate the base theme in `afterAll`.

Hug0-Drelon (changes requested):

tests/e2e/specs/widgets/language-switcher-widget.spec.js
Marianne380: See https://github.com/polylang/polylang/pull/1859/commits/2d667cae30b07464b3f2f149e38ae5906c939514.

tests/e2e/specs/widgets/language-switcher-widget.spec.js
```diff
@@ -13,12 +13,20 @@ test.describe(
...
 		/**
 		 * Before all tests:
 		 *     - Activate a theme with widget areas.
 		 *     - Create en_US and fr_FR languages.
 		 */
 		test.beforeAll( async ( { requestUtils } ) => {
+			const [ activeTheme ] = await requestUtils.rest( {
+				path: '/wp/v2/themes',
+				params: { status: 'active' },
+			} );
```
Hug0-Drelon, line 27: Suggesting:
```js
await requestUtils
			.rest( {
				path: '/wp/v2/themes',
				method: 'GET',
				params: { status: 'active' },
			} )
			.then( ( response ) => response[ 0 ] );
```
(sorry, GH won't let me propose code directly for some reason...)

Hug0-Drelon (changes requested):

tests/e2e/specs/widgets/language-switcher-widget.spec.js
Marianne380: I'd keep the current version, mixing `.then()` with `async/await` doesn't really add anything here. What would we gain?

Hug0-Drelon: We would have a single statement. In my proposal, `await` is superflous.

But let's keep it like you've done.

Hug0-Drelon (approved):

PR #1862, opened 2026-04-15
Fix PHP warnings and improve nav menus option display in Site Health.

### LINKED ISSUE #1860: Site Health Debug warnings: Undefined array key "value"/"label"
**Describe the bug**
  Opening Site Health debug info triggers PHP warnings related to missing array keys (`value` / `label`) while WordPress formats debug data.
  The warnings appear on:

  `/wp-admin/site-health.php?tab=debug`

  and seem tied to Polylang Pro sections in Site Health debug output.

  **Reproduced with**
  - WordPress version: `6.8.5`
  - Polylang version: `3.8.2` (Polylang Pro)
  - PHP version: `8.4.18`
  - Multisite: `Yes`

  **To Reproduce**
  1. Go to `/wp-admin/site-health.php?tab=debug`
  2. Load the page
  3. Check PHP logs
  4. See warnings:
     - `Undefined array key "value"` in `wp-admin/includes/class-wp-debug-data.php` (line 1799)
     - `Undefined array key "value"` in `wp-admin/site-health-info.php` (lines 116, 125)
     - `Undefined array key "label"` in `wp-admin/site-health-info.php` (line 131)

  **Expected behavior**
  Site Health debug page should render without PHP warnings/notices.

  **Environment (please complete the following information):**
  - Device: `MacBook Pro`
  - OS: `MacOS Tahoe 26.3.1`
  - Browser: `Google Chrome`
  - Version: `147.0.7727.56`

  **Additional context, screenshots**
  - Related issue: https://github.com/polylang/polylang/issues/1753
  - This report appears similar in area (Site Health debug formatting), but reproduced on current setup (`WP 6.8.5` + `Polylang Pro 3.8.2` +
  `PHP 8.4.18`).

Hug0-Drelon:
## What?
Fixes #1860 

To reproduce :
- Polylang 3.8.2
- Two languages
- Activate classic theme (needs old menus)
- Create one menu per language
- Visit `http://localhost:8888/wp-admin/site-health.php?tab=debug#health-check-section-pll_options`
- See the warnings

## How?
- Ensure to always have `label` and  `value` keys set in `Nav_Menus::get_site_health_info` by calling `Abstract_Option::format_single_value_for_site_health_info` before returning the piece of information.
- Improve array formatting in `Abstract_Option::format_array_for_site_health_info()` (use brackets to format menu array containing language slugs as keys and post IDs as values).

### DIFF
--- a/src/Options/Abstract_Option.php
+++ b/src/Options/Abstract_Option.php
@@ -322,19 +322,14 @@ abstract class Abstract_Option {
 	 *
 	 * @since 3.8
 	 *
-	 * @param array $array An array of formatted data.
+	 * @param array<scalar> $array An array of formatted data.
 	 * @return string
 	 */
 	protected function format_array_for_site_health_info( array $array ): string {
 		array_walk(
 			$array,
 			function ( &$value, $key ) {
-				if ( is_array( $value ) ) {
-					$ids = implode( ' , ', $value );
-					$value = "$key => [$ids]";
-				} else {
-					$value = "$key => [$value]";
-				}
+				$value = "$key => [$value]";
 			}
 		);
 		return implode( ', ', $array );
--- a/src/Options/Abstract_Option.php
+++ b/src/Options/Abstract_Option.php
@@ -329,9 +329,10 @@ abstract class Abstract_Option {
 		array_walk(
 			$array,
 			function ( &$value, $key ) {
-				$value = "$key => [$value]";
+				$value = "$key => $value";
 			}
 		);
-		return implode( ', ', $array );
+
+		return '[' . implode( ', ', $array ) . ']';
 	}
 }
--- a/src/Options/Abstract_Option.php
+++ b/src/Options/Abstract_Option.php
@@ -331,12 +331,12 @@ abstract class Abstract_Option {
 			function ( &$value, $key ) {
 				if ( is_array( $value ) ) {
 					$ids = implode( ' , ', $value );
-					$value = "$key => $ids";
+					$value = "$key => [$ids]";
 				} else {
-					$value = "$key => $value";
+					$value = "$key => [$value]";
 				}
 			}
 		);
-		return implode( ' | ', $array );
+		return implode( ', ', $array );
 	}
 }
--- a/src/Options/Business/Nav_Menus.php
+++ b/src/Options/Business/Nav_Menus.php
@@ -60,7 +60,6 @@ class Nav_Menus extends Abstract_Option {
 		$parts = array();
 		foreach ( $nav_menus[ $current_theme ] as $location => $lang ) {
 			if ( empty( $lang ) ) {
-				/* translators: 1: menu location slug, 2: status when no menu is assigned */
 				$parts[] = sprintf( '%1$s: %2$s', $location, __( 'Not used', 'polylang' ) );
 			} else {
 				$parts[] = sprintf(
--- a/src/Options/Business/Nav_Menus.php
+++ b/src/Options/Business/Nav_Menus.php
@@ -57,18 +57,21 @@ class Nav_Menus extends Abstract_Option {
 			return array();
 		}
 
-		$fields = array();
+		$parts = array();
 		foreach ( $nav_menus[ $current_theme ] as $location => $lang ) {
-			$fields[ $location ]['label'] = sprintf( 'menu: %s', $location );
-
 			if ( empty( $lang ) ) {
-				/* translators: default value when a menu location is not used. */
-				$fields[ $location ]['value'] = __( 'Not used', 'polylang' );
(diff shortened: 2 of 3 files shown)


src/Options/Business/Nav_Menus.php
```diff
@@ -57,18 +57,21 @@ public function get_site_health_info( Options $options ): array { // phpcs:ignor
...
-		$fields = array();
+		$parts = array();
 		foreach ( $nav_menus[ $current_theme ] as $location => $lang ) {
-			$fields[ $location ]['label'] = sprintf( 'menu: %s', $location );
-
 			if ( empty( $lang ) ) {
-				/* translators: default value when a menu location is not used. */
-				$fields[ $location ]['value'] = __( 'Not used', 'polylang' );
+				/* translators: 1: menu location slug, 2: status when no menu is assigned */
+				$parts[] = sprintf( '%1$s: %2$s', $location, __( 'Not used', 'polylang' ) );
```
Screenfeed:
```suggestion
				$parts[] = sprintf( '%s: %s', $location, _x( 'Not used', 'Menu location (language slug)', 'polylang' ) );
```

The `translators` text is for `'Not used'`, not for `'%1$s: %2$s'`.
About the content of the `translators` text: IMHO it is more important to say what `Not used` refers to. If I was a translator, I think `status when no menu is assigned` doesn't help me. However I would like to know what it refers to (example: because in French: `Non utilisé` vs `Non utilisée`). And since the same text (`Not used`) can be translated into 2 things (`Non utilisé` vs `Non utilisée`), I think `_x()` is more appropriate than a comment.

src/Options/Abstract_Option.php
```diff
@@ -331,12 +331,12 @@ protected function format_array_for_site_health_info( array $array ): string {
 			function ( &$value, $key ) {
 				if ( is_array( $value ) ) {
 					$ids = implode( ' , ', $value );
-					$value = "$key => $ids";
+					$value = "$key => [$ids]";
 				} else {
-					$value = "$key => $value";
+					$value = "$key => [$value]";
```
Screenfeed: So `$value` is already a string representation of an array here? Otherwise I don't think we should add `[ ]`.


```suggestion
					$value = "$key => $value";
```

Note: I have the feeling that this method should be recursive (or use `print_r()`?).

Screenfeed (changes requested):

src/Options/Business/Nav_Menus.php
Hug0-Drelon: Firstly, this strings is already shipped in 3.8.x, so it's simpler to leave it untouched.
Secondly, "Not used' is for a single menu location; e.g. "footer: Not used".

src/Options/Abstract_Option.php
Hug0-Drelon: This is for display purpose only, without bracket, we can lose track of arrays when several are chained.

> Note: I have the feeling that this method should be recursive (or use print_r()?).

I don't see why, can you explain?

src/Options/Business/Nav_Menus.php
```diff
@@ -57,18 +57,21 @@ public function get_site_health_info( Options $options ): array { // phpcs:ignor
...
 
-		$fields = array();
+		$parts = array();
 		foreach ( $nav_menus[ $current_theme ] as $location => $lang ) {
-			$fields[ $location ]['label'] = sprintf( 'menu: %s', $location );
-
 			if ( empty( $lang ) ) {
-				/* translators: default value when a menu location is not used. */
-				$fields[ $location ]['value'] = __( 'Not used', 'polylang' );
+				/* translators: 1: menu location slug, 2: status when no menu is assigned */
```
Hug0-Drelon: Go to hell.

src/Options/Abstract_Option.php
Hug0-Drelon: Should be `array<array<scalar>>`.

src/Options/Business/Nav_Menus.php
Hug0-Drelon: Removed in 4578c01b.

src/Options/Abstract_Option.php
Hug0-Drelon: In fact it should be `array<scalar>`, see `NavMenusValue` phpstan type.

Fixed and simplified in 4578c01b.

Screenfeed: Indeed, I missed that. Those array shapes can be useful after all 😏

Screenfeed (approved):

Hug0-Drelon:
Menus should be display like so: `primary: [en => 13, fr => 13] | Social: Not used`.

Hug0-Drelon:
> Menus should be display like so: `primary: [en => 13, fr => 13] | Social: Not used`.

Done in 3a6aa2cd.

PR #1864, opened 2026-04-21
Use SVG flags by default

Chouby:
This PR replaces good old famfamfam PNG flags by [our new set of SVG flags](https://github.com/polylang/flags/).


src/language.php
```diff
@@ -674,4 +679,25 @@ public function get_prop( $property ) {
...
+	/**
+	 * Prepares a SVG image for use in data uri.
+	 *
+	 * @see https://codepen.io/tigt/post/optimizing-svgs-in-data-uris.
+	 * @since 3.8
+	 *
+	 * @string $svg A string representing an SVG image.
+	 * @return string Encode SVG.
+	 */
+	protected static function encode_svg( $svg ) {
```
Screenfeed:
```suggestion
	 * @param string $svg A string representing an SVG image.
	 * @return string Encode SVG.
	 */
	protected static function encode_svg( string $svg ): string {
```

src/settings/view-tab-lang.php
```diff
@@ -127,10 +127,15 @@
 							<?php
 							$flags = include __DIR__ . '/flags.php';
 							foreach ( $flags as $code => $label ) {
+								$flag = PLL_Language::get_flag_information( $code );
+								if ( empty( $flag['url'] ) ) {
+									continue;
+								}
```
Screenfeed, line 133: Ah! This makes [my PR](https://github.com/polylang/polylang/pull/1858) useless now 😅

src/language.php
```diff
@@ -378,18 +378,23 @@ public static function get_flag_information( $code ) {
...
 		// Polylang builtin flags.
-		if ( ! empty( $code ) && is_readable( POLYLANG_DIR . ( $file = '/flags/' . $code . '.png' ) ) ) {
-			$default_flag['url'] = plugins_url( $file, POLYLANG_FILE );
+		$file = "/vendor/wpsyntex/flags/{$code}.svg";
+		if ( is_readable( POLYLANG_DIR . $file ) ) {
+			$default_flag = array(
+				'url'    => plugins_url( $file, POLYLANG_FILE ),
+				'width'  => 18,
+				'height' => 12,
+			);
```
Screenfeed, line 393: Since `src` is a mandatory key, we should have it here (and it will make PHPStan more happy).
```suggestion
			$default_flag = array(
				'url'    => plugins_url( $file, POLYLANG_FILE ),
				'src'    => '',
				'width'  => 18,
				'height' => 12,
			);
```

src/language.php
```diff
@@ -378,18 +378,23 @@ public static function get_flag_information( $code ) {
...
 			// If base64 encoded flags are preferred.
 			if ( pll_get_constant( 'PLL_ENCODED_FLAGS', true ) ) {
-				$imagesize = getimagesize( POLYLANG_DIR . $file );
-				if ( is_array( $imagesize ) ) {
-					list( $default_flag['width'], $default_flag['height'] ) = $imagesize;
-				}
-				$file_contents       = file_get_contents( POLYLANG_DIR . $file ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
-				$default_flag['src'] = 'data:image/png;base64,' . base64_encode( $file_contents ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+				$content             = file_get_contents( POLYLANG_DIR . $file ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+				$default_flag['src'] = 'data:image/svg+xml,' . self::encode_svg( $content );
```
Screenfeed: `file_get_contents()` can return `false`. I prefer to prevent an easily avoidable fatal error over a "it should not".


```suggestion
				$content = file_get_contents( POLYLANG_DIR . $file ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents

				if ( ! empty( $content ) ) {
					$default_flag['src'] = 'data:image/svg+xml,' . self::encode_svg( $content );
				}
```

Screenfeed (changes requested):

Screenfeed:
There are also some tests to fix because of the change in the URL.

src/language.php
```diff
@@ -378,18 +378,23 @@ public static function get_flag_information( $code ) {
 			'src' => '',
 		);
 
+		if ( empty( $code ) ) {
+			return $default_flag;
+		}
```
Screenfeed, line 383: Note: we're changing the behavior, an empty flag won't go through the filter `'pll_flag'`.

Hug0-Drelon (commented):
I've investigated why e2e tests are failing. And I have been able to reproduce locally using a dockerized installation with `wp-env`.

This is because of `wp_remote_get()` called in https://github.com/polylang/polylang/blob/3.8.2/src/Model/Languages.php#L1067 that returns an error.

`wp_remote_get()` runs inside the `wp-env` container. The URL uses `localhost:8888`, but `8888` is only mapped on the host; inside the container the server listens on `80`, so nothing accepts `:8888`; leading to cURL error 7 (`Failed to connect to localhost port 8888 after 0 ms: Could not connect to server`). Browsers hit the host; PHP hits the container loopback, different port.

For the moment, I didn't find a solution that avoids to write ugly PHP code to bypass this behavior.

src/language.php
Chouby: We decided to keep like that.

src/settings/view-tab-lang.php
Screenfeed: @Chouby Suggestion, following what I've done in my PR (previous comment):

```php
foreach ( $flags as $code => $label ) {
	$flag_html = PLL_Language::get_predefined_flag( $code ); // <==
	if ( empty( $flag_html ) ) { // <==
		continue;
	}

	printf(
		'<option value="%s" data-flag-html="%s"%s>%s</option>' . "\n",
		esc_attr( $code ),
		esc_html( $flag_html ), // <==
		selected( isset( $edit_lang->flag_code ) && $edit_lang->flag_code === $code, true, false ),
		esc_html( $label )
	);
}
```

Hug0-Drelon:
- Playwright Docker tools added in ed9e94fe5cfe0c2e5dc28fe6dbf6e655c52978cf
- Snapshots re-generated in 1fad53ca2292a2931641b562805c0a2160f17d6a and b5f7937b98526da892369258844909b0ac5e845a (MacOS and Linux)

PR #1865, opened 2026-04-21
Update the languages and flags lists

Chouby:
Depends on #1864

- Rename some flag codes to match iso 3166 code (ex `catalonia` => `es-ct`).
- Rename some flag codes to match iso 639 code when the flag is related to a language (ex `esperanto` => `eo.lang`).
- Add new flags (mainly for regions).
- Update flag codes for languages, sometimes replacing a national flag by a regional flag.
- Add few languages
- Remove Veneto language (not a WP language).


src/settings/flags.php
```diff
@@ -15,255 +15,270 @@
...
+	'ae'       => __( 'United Arab Emirates', 'polylang' ),
+	'af'       => __( 'Afghanistan', 'polylang' ),
+	'ag'       => __( 'Antigua and Barbuda', 'polylang' ),
+	'ai'       => __( 'Anguilla', 'polylang' ),
+	'al'       => __( 'Albania', 'polylang' ),
+	'am'       => __( 'Armenia', 'polylang' ),
+	'an'       => __( 'Netherlands Antilles', 'polylang' ),
+	'ao'       => __( 'Angola', 'polylang' ),
+	'ar'       => __( 'Argentina', 'polylang' ),
+	'ar.lang'  => __( 'Arabic language', 'polylang' ),
```
Screenfeed, line 28: What happens to the languages whose flag codes have disappeared? I would expect their flags to trigger a 404 error.

Screenfeed:
Two tests are in error.

src/settings/flags.php
Chouby: I added a process to migrate codes like 'arab' => 'ar.lang'.
Otherwise, if the flag file is missing, everything happens as when we create a language without flag (the language code is displayed in backend, nothing in frontend).

Chouby:
Two tests are erroring because they create a language with a non-existing flag. In this PR, there's no flag for 'ar.lang'. Once #1864 is merged, everything should be ok for these 2 tests. I will merge this PR after #1864.

src/install/upgrade.php
```diff
@@ -250,6 +250,56 @@ protected function upgrade_3_8() {
...
+
+		if ( ! is_array( $terms ) ) {
+			return;
+		}
+
+		foreach ( $terms as $term ) {
+			$description = maybe_unserialize( $term->description );
+			$description = is_array( $description ) ? $description : array();
+
+			if ( isset( $description['flag_code'] ) && isset( $to_migrate[ $description['flag_code'] ] ) ) {
```
Screenfeed: 2 ways to fix the issue reported by PHPStan, I let you choose:


```suggestion
			$description = maybe_unserialize( $term->description );
			$description = is_array( $description ) ? $description : array();

			if ( isset( $description['flag_code'] ) && is_string( $description['flag_code'] ) && isset( $to_migrate[ $description['flag_code'] ] ) ) {
```
Or:

```suggestion
			/** @phpstan-var array{flag_code: string}|mixed */
			$description = maybe_unserialize( $term->description );
			$description = is_array( $description ) ? $description : array();

			if ( isset( $description['flag_code'], $to_migrate[ $description['flag_code'] ] ) ) {
```

Screenfeed (changes requested):

Screenfeed (approved):

Chouby:
Just updated the code for Arabic language in tests. As a consequence, there are now 6 tests failing or erroring, the same as #1864, for the opposite reason (one PR has the right flags files, the other has the right flags list).

PR #1866, opened 2026-04-21
Fix upgrade

Chouby:
When upgrading from a very old version (< 1.8), we expect to prevent the plugin activation and display the message
>Polylang has been deactivated because you upgraded from a too old version.
>Before upgrading to 3.9, please upgrade to <strong>2.9</strong>

However (I don't know since which WP version), WP doesn't display anymore our message but the meaningless:
>Plugin could not be activated because it triggered a fatal error.

To reproduce:
- Deactivate the plugin.
- Change the version stored in the option polylang (for ex to 0.9).
- Activate the plugin.

I propose not to prevent activation anymore, to avoid the meaningless WP message but to stop the execution, just displaying the notice.

Additionally, the PR:
- Removes `PLL_Upgrade::can_activate()` now useless.
- Removes all calls to `load_plugin_textdomain()`, mainly found in activation notices (this function is useless for a while now).
- Uses `wp_admin_notice()` to display admin notices (better to rely on WP API).

### DIFF
--- a/src/install/upgrade.php
+++ b/src/install/upgrade.php
@@ -82,7 +82,7 @@ class PLL_Upgrade {
 
 		$args = array(
 			'type'           => 'error',
-			'paragraph_wrap' => 'false',
+			'paragraph_wrap' => false,
 		);
 
 		wp_admin_notice( $message, $args );
--- a/src/install/upgrade.php
+++ b/src/install/upgrade.php
@@ -80,7 +80,12 @@ class PLL_Upgrade {
 			)
 		);
 
-		wp_admin_notice( $message, array( 'type' => 'error' ) );
+		$args = array(
+			'type'           => 'error',
+			'paragraph_wrap' => 'false',
+		);
+
+		wp_admin_notice( $message, $args );
 	}
 
 	/**
--- a/src/install/upgrade.php
+++ b/src/install/upgrade.php
@@ -62,14 +62,13 @@ class PLL_Upgrade {
 	}
 
 	/**
-	 * Displays a notice when upgrading from a too old version
+	 * Displays a notice when upgrading from a too old version.
 	 *
 	 * @since 1.0
 	 *
 	 * @return void
 	 */
 	public function admin_notices() {
-		load_plugin_textdomain( 'polylang' );
 		printf(
 			'<div class="error"><p>%s</p><p>%s</p></div>',
 			esc_html__( 'Polylang has been deactivated because you upgraded from a too old version.', 'polylang' ),
@@ -77,7 +76,7 @@ class PLL_Upgrade {
 				/* translators: %1$s and %2$s are Polylang version numbers */
 				esc_html__( 'Before upgrading to %2$s, please upgrade to %1$s.', 'polylang' ),
 				'<strong>2.9</strong>',
-				POLYLANG_VERSION // phpcs:ignore WordPress.Security.EscapeOutput
+				esc_html( POLYLANG_VERSION )
 			)
 		);
 	}
--- a/src/install/upgrade.php
+++ b/src/install/upgrade.php
@@ -69,8 +69,8 @@ class PLL_Upgrade {
 	 * @return void
 	 */
 	public function admin_notices() {
-		printf(
-			'<div class="error"><p>%s</p><p>%s</p></div>',
+		$message = sprintf(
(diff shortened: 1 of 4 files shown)


src/install/upgrade.php
```diff
@@ -77,24 +62,25 @@ public function can_upgrade() {
...
 			sprintf(
 				/* translators: %1$s and %2$s are Polylang version numbers */
 				esc_html__( 'Before upgrading to %2$s, please upgrade to %1$s.', 'polylang' ),
 				'<strong>2.9</strong>',
-				POLYLANG_VERSION // phpcs:ignore WordPress.Security.EscapeOutput
+				esc_html( POLYLANG_VERSION )
 			)
 		);
+
+		wp_admin_notice( $message, array( 'type' => 'error' ) );
```
Screenfeed: Since the message already contains paragraphs we must add `'paragraph_wrap' => false`. I also suggest to add `'dismissible' => true` since we do the same with the "you can't activate PLL and PLL Pro at the same time".


```suggestion
		wp_admin_notice(
			$message,
			array(
				'type'           => 'error',
				'dismissible'    => true,
				'paragraph_wrap' => false,
			)
		);
```

src/install/usable.php
```diff
@@ -60,6 +58,8 @@ public static function php_version_notice() {
 				esc_html( static::get_min_php_version() )
 			)
 		);
+
+		wp_admin_notice( $message, array( 'type' => 'error' ) );
```
Screenfeed, line 59: No need to wrap in a paragraph, `wp_admin_notice()` can do it for us.

```php
		$message = sprintf(
			/* translators: 1: Plugin name 2: Current PHP version 3: Required PHP version */
			esc_html__( '%1$s has deactivated itself because you are using an old version of PHP. You are using using PHP %2$s. %1$s requires PHP %3$s.', 'polylang' ),
			esc_html( static::get_plugin_name() ),
			esc_html( pll_get_constant( 'PHP_VERSION', '' ) ),
			esc_html( static::get_min_php_version() )
		);

		wp_admin_notice(
			$message,
			array(
				'type'        => 'error',
				'dismissible' => true,
			)
		);
```

src/install/usable.php
```diff
@@ -86,6 +84,8 @@ public static function wp_version_notice() {
 				esc_html( static::get_min_wp_version() )
 			)
 		);
+
+		wp_admin_notice( $message, array( 'type' => 'error' ) );
```
Screenfeed, line 82: Same here.


```php
		$message = sprintf(
			/* translators: 1: Plugin name 2: Current WordPress version 3: Required WordPress version */
			esc_html__( '%1$s has deactivated itself because you are using an old version of WordPress. You are using using WordPress %2$s. %1$s requires at least WordPress %3$s.', 'polylang' ),
			esc_html( static::get_plugin_name() ),
			esc_html( $wp_version ),
			esc_html( static::get_min_wp_version() )
		);

		wp_admin_notice(
			$message,
			array(
				'type'        => 'error',
				'dismissible' => true,
			)
		);
```

src/integrations/wp-importer/wordpress-importer.php
```diff
@@ -38,9 +38,6 @@ public function maybe_wordpress_importer() {
 	 * @since 1.2
 	 */
 	public function wordpress_importer_init() {
-		$class = new ReflectionClass( 'WP_Import' );
-		load_plugin_textdomain( 'wordpress-importer', false, basename( dirname( $class->getFileName() ) ) . '/languages' );
-
```
Screenfeed, line 43: Why this?

Screenfeed (changes requested):

src/integrations/wp-importer/wordpress-importer.php
Chouby: Since I saw `load_plugin_textdomain()` calls in the code I modified, I searched for all occurrences to remove them.

src/install/upgrade.php
Chouby: It wasn't dismissible previously. And I kept it like this intentionally. I will open the discussion during our next meeting.

Chouby: We decided to keep these notices non dismissible.

Screenfeed (approved):

sebastienserre:
Finally make it working

PR #1868, opened 2026-04-22
Prevent using unsupported block fields in wpml-config

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/2952.

This prevents to use unsupported "Gutenberg block" fields in `wpml-config.xml`.

## How

Are excluded:
- Fields with a `link`/`post-ids`/`taxonomy-ids` type.
- `key` fields with a `regex` value for `search-method`.

Related [doc](https://wpml.org/documentation/support/language-configuration-files/make-custom-gutenberg-blocks-translatable/).

### DIFF
--- a/src/modules/wpml/wpml-config.php
+++ b/src/modules/wpml/wpml-config.php
@@ -509,6 +509,10 @@ class PLL_WPML_Config {
 				}
 
 				foreach ( $block->children() as $child ) {
+					if ( ! $this->is_supported_field( $child ) ) {
+						continue;
+					}
+
 					$rule      = '';
 					$child_tag = $child->getName();
 
@@ -670,6 +674,10 @@ class PLL_WPML_Config {
 		$sub_attributes = array();
 
 		foreach ( $children as $child ) {
+			if ( ! $this->is_supported_field( $child ) ) {
+				continue;
+			}
+
 			$sub = $this->get_field_attributes( $child );
 
 			if ( empty( $sub ) ) {
@@ -680,6 +688,30 @@ class PLL_WPML_Config {
 		}
 
 		return $sub_attributes;
+	}
+
+	/**
+	 * Tells if the given field is supported.
+	 *
+	 * @since 3.9
+	 *
+	 * @param  SimpleXMLElement $field A XML node.
+	 * @return bool
+	 */
+	protected function is_supported_field( SimpleXMLElement $field ): bool {
+		$child_type = $this->get_field_attribute( $field, 'type' );
+
+		if ( in_array( $child_type, array( 'link', 'post-ids', 'taxonomy-ids' ), true ) ) {
+			return false;
+		}
+
+		if ( $field->getName() !== 'key' ) {
+			return true;
+		}
+
+		$search_method = $this->get_field_attribute( $field, 'search-method' );
+
+		return 'regex' !== $search_method;
 	}
 
 	/**
--- a/src/modules/wpml/wpml-config.php
+++ b/src/modules/wpml/wpml-config.php
@@ -709,9 +709,7 @@ class PLL_WPML_Config {
 			return true;
 		}
 
-		$search_method = $this->get_field_attribute( $field, 'search-method' );
-
-		return 'regex' !== $search_method;
+		return 'regex' !== $this->get_field_attribute( $field, 'search-method' );
 	}
 
 	/**
--- a/src/modules/wpml/wpml-config.php
+++ b/src/modules/wpml/wpml-config.php
@@ -695,10 +695,10 @@ class PLL_WPML_Config {
 	 *
 	 * @since 3.9
 	 *
-	 * @param  SimpleXMLElement $field A XML node.
+	 * @param SimpleXMLElement $field A XML node.
 	 * @return bool
 	 */
-	protected function is_supported_field( SimpleXMLElement $field ): bool {
+	private function is_supported_field( SimpleXMLElement $field ): bool {
 		$child_type = $this->get_field_attribute( $field, 'type' );
 
 		if ( in_array( $child_type, array( 'link', 'post-ids', 'taxonomy-ids' ), true ) ) {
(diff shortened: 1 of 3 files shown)


src/modules/wpml/wpml-config.php
```diff
@@ -682,6 +690,28 @@ private function get_field_attributes( SimpleXMLElement $field ): array {
 		return $sub_attributes;
 	}
 
+	/**
+	 * Tells if the given field is supported.
+	 *
+	 * @since 3.9
+	 *
+	 * @param  SimpleXMLElement $field A XML node.
```
Marianne380:
```suggestion
	 * @param SimpleXMLElement $field A XML node.
```

src/modules/wpml/wpml-config.php
```diff
@@ -682,6 +690,28 @@ private function get_field_attributes( SimpleXMLElement $field ): array {
...
+	 * @return bool
+	 */
+	protected function is_supported_field( SimpleXMLElement $field ): bool {
```
Marianne380: Why not `private` ?

Marianne380 (changes requested):

src/modules/wpml/wpml-config.php
Screenfeed: Because... reasons.

Image

src/modules/wpml/wpml-config.php
Screenfeed: Done in [2cc62ae](https://github.com/polylang/polylang/pull/1868/commits/2cc62aec95f03efe85e6307181a3813d5f4bba48).

src/modules/wpml/wpml-config.php
Screenfeed: Done in [2cc62ae](https://github.com/polylang/polylang/pull/1868/commits/2cc62aec95f03efe85e6307181a3813d5f4bba48).

Marianne380 (approved):

src/modules/wpml/wpml-config.php
Screenfeed: (Probably because I copy/pasted a method that was `protected`)

src/modules/wpml/wpml-config.php
```diff
@@ -682,6 +690,28 @@ private function get_field_attributes( SimpleXMLElement $field ): array {
...
+	 *
+	 * @param SimpleXMLElement $field A XML node.
+	 * @return bool
+	 */
+	private function is_supported_field( SimpleXMLElement $field ): bool {
+		$child_type = $this->get_field_attribute( $field, 'type' );
+
+		if ( in_array( $child_type, array( 'link', 'post-ids', 'taxonomy-ids' ), true ) ) {
+			return false;
+		}
```
Chouby: Shouldn't we act the opposite way and return `false` as soon as we have a type since we support none? In case new types appear in the future.

src/modules/wpml/wpml-config.php
```diff
@@ -682,6 +690,28 @@ private function get_field_attributes( SimpleXMLElement $field ): array {
 		return $sub_attributes;
 	}
 
+	/**
+	 * Tells if the given field is supported.
+	 *
+	 * @since 3.9
```
Chouby: I propose to merge this for the next minor version.
```suggestion
	 * @since 3.8.4
```

tests/phpunit/data/wpml-config.xml
```diff
@@ -80,8 +80,21 @@
...
+				<key name="url" type="link" />
+				<key name="PostIds">
+					<key name="*" type="post-ids" sub-type="post" />
+				</key>
+				<key name="TermIds">
+					<key name="*" type="taxonomy-ids" sub-type="category" />
+				</key>
+				<key name="regexData">
+					<key name="/^[^_]\S+$/" search-method="regex" />
+				</key>
```
Chouby, line 97: Could we add a search method wildcard which should pass.

src/modules/wpml/wpml-config.php
```diff
@@ -682,6 +690,28 @@ private function get_field_attributes( SimpleXMLElement $field ): array {
...
+
+		if ( in_array( $child_type, array( 'link', 'post-ids', 'taxonomy-ids' ), true ) ) {
+			return false;
+		}
+
+		if ( $field->getName() !== 'key' ) {
+			return true;
+		}
+
+		return 'regex' !== $this->get_field_attribute( $field, 'search-method' );
```
Chouby: Same. I would act the opposite way and accept only what we are able to handle (i.e the wildcard).

Chouby (changes requested):

src/modules/wpml/wpml-config.php
Screenfeed: Done in [74a7ce9](https://github.com/polylang/polylang/pull/1868/commits/74a7ce9658ae1f5f7f8ba333ed74b706c8d89380).

src/modules/wpml/wpml-config.php
Screenfeed: Done in [74a7ce9](https://github.com/polylang/polylang/pull/1868/commits/74a7ce9658ae1f5f7f8ba333ed74b706c8d89380).

src/modules/wpml/wpml-config.php
Screenfeed: Done in [6792b9b](https://github.com/polylang/polylang/pull/1868/commits/6792b9b2f3593b7f5fa9161bcad6cdecfd4b2fe6),

tests/phpunit/data/wpml-config.xml
Screenfeed: Done in [c699ab1](https://github.com/polylang/polylang/pull/1868/commits/c699ab174c34151f66d5d34161da64568811ecee).

Marianne380 (approved):

PR #1869, opened 2026-04-24
Rework multilingual biography

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/1124.

## How

Instead of printing `hidden` inputs, add an inline JS `const` that contains the related data.
Also:
- Use vanilla JS without jQuery.
- Improve accessibility by adding `lang` and `dir` HTML attributes, and `<label>` tags.
- Add language flags.

### Old markup

```html
<div>Français</div>
<textarea name="description_fr" id="description_fr" rows="5" cols="30">Biography.</textarea>
<br>
```

### New markup

```html
<p lang="fr-FR">
	<label for="description_fr" dir="ltr">
		<img alt="" src="..." width="16" height="11">
		Français
	</label>
	<br>
	<textarea name="description_fr" id="description_fr" dir="ltr" rows="5" cols="30">Biography.</textarea>
</p>
```

image


src/admin/admin-filters.php
```diff
@@ -56,25 +56,41 @@ public function personal_options_update( $user_id ) {
...
-				'<input type="hidden" class="biography" name="%s___%s" value="%s" />',
-				esc_attr( $lang->slug ),
-				esc_attr( $lang->name ),
-				sanitize_user_field( 'description', $description, $profileuser->ID, 'edit' )
+			$data[] = array(
+				'slug'        => $lang->slug,
+				'name'        => $lang->name,
+				'lang'        => $lang->get_locale( 'display' ),
+				'direction'   => $lang->is_rtl ? 'rtl' : 'ltr',
+				'flag'        => $lang::get_flag_information( $lang->flag_code ),
```
Marianne380: Why not `PLL_Language::get_flag_information( $lang->flag_code )` ?

changelog/1124
```diff
@@ -0,0 +1 @@
+ * Accessibility improvement of the multilingual biography fields on admin user profile page. #1828
```
Marianne380: Why #1828 ?

Marianne380 (changes requested):

js/src/user.js
```diff
@@ -1,33 +1,72 @@
...
+		} else {
+			document.addEventListener(
+				'DOMContentLoaded',
+				pllDescription.ready
+			);
+		}
+	},
+
+	/**
+	 * Called when the DOM is ready. Attaches the events to the wrapper.
```
Marianne380: It says **_"Attaches the events to the wrapper"_**, where are events attached here?

Marianne380 (changes requested):

src/admin/admin-filters.php
Screenfeed: It's the same. Do you prefer `PLL_Language` for clarity?

Marianne380: Yep, if that's okay with you.

js/src/user.js
Screenfeed: In another PR!
🦄

src/admin/admin-filters.php
Screenfeed: Done in [8b26a61](https://github.com/polylang/polylang/pull/1869/commits/8b26a612d6ed7abce5c701fceba5633007fa72e9).

changelog/1124
Screenfeed: I can't tell you, it's a secret.

Marianne380 (approved):

Chouby (changes requested):
Adding the text direction is a good idea. this however results in a strange layout:
image
image

Screenfeed:
> Adding the text direction is a good idea. this however results in a strange layout

Fixed:
image
image

Marianne380 (approved):

js/src/user.js
```diff
@@ -1,33 +1,73 @@
...
+
+		if ( ! originTextarea ) {
+			return;
+		}
+
+		const rows = [];
+
+		pllDescriptionData.forEach( ( data ) => {
+			const wrapper = document.createElement( 'p' );
+			wrapper.setAttribute( 'lang', data.lang );
```
Chouby: In the other rows of the table, paragraphs are use only for the description.
Thus I suggest to remove the 'p' element and move the attribute 'lang' to the label and textarea as you did for the attribute 'dir'.
What do you think?

src/admin/admin-filters.php
```diff
@@ -56,25 +56,50 @@ public function personal_options_update( $user_id ) {
 	}
 
 	/**
-	 * Outputs hidden information to modify the biography form with js.
+	 * Adds data to modify the biography form with JS.
+	 *
+	 * This is done here instead of `PLL_Admin_Base::add_inline_scripts()` because we have direct access to the user ID
+	 * And the user capability check has been done by WP.
```
Chouby: Since the `enqueue_script()` was moved from `PLL_Admin_Base` to here, this comment could be removed (or at least adapted).

src/admin/admin-filters.php
```diff
@@ -56,25 +56,50 @@ public function personal_options_update( $user_id ) {
...
 	 *
 	 * @since 0.4
+	 * @since 3.9 Doesn't print hidden inputs anymore.
```
Chouby: Not sure this changelog is useful. If you want to keep it, let's change it to 3.8.4.

Chouby (changes requested):

js/src/user.js
Screenfeed: The `<p>` is useful because it prevents a layout issue where the following `<label>` moves at the right of the `<textarea>`.
So if you don't want the `<p>`, I see 2 solutions:

1. Remove it and add a `<br>` after the `<textarea>`. Caveat: instead of inserting a single `<p>` into the DOM, we have to insert a `<label>`, a `<br>`, a `<textarea>`, and another `<br>`.
2. Replace it by a `<div>`. Caveat: a bit less margin between each biography but can be neglected, it's only a few pixels.

src/admin/admin-filters.php
Screenfeed: Removed in [7a111d7](https://github.com/polylang/polylang/pull/1869/commits/7a111d75a411c15d2980ea1e2d6bff78067f6f56).

src/admin/admin-filters.php
Screenfeed: Indeed I forgot. Removed in [7a111d7](https://github.com/polylang/polylang/pull/1869/commits/7a111d75a411c15d2980ea1e2d6bff78067f6f56). I also improved the method's description in [69f7a81](https://github.com/polylang/polylang/pull/1869/commits/69f7a814e9b3aad40900d8c3a3c098b4d02a0345) and renamed the method in [d9fda8d](https://github.com/polylang/polylang/pull/1869/commits/d9fda8dbbfd46a23636d561c9215f3e230bff591).

js/src/user.js
Screenfeed: Use a div in [5776295](https://github.com/polylang/polylang/pull/1869/commits/5776295fb2e6338ed295b52b0bc451e0dae75bdd).

PR #1872, opened 2026-04-27
Switcher visitor proposal

Hug0-Drelon:
## What?
Add a Visitor-based switcher architecture proposal in `switcher.md`, including UML diagram and flow charts for:

- `Switcher::print()` (HTML rendering)
- `Switcher::to_array()` (raw data output)
- typed `Element` variants + typed `Visitor` dispatch

> [!NOTE]
This PR is documentation-only. See `switcher.md` for full details.

> [!NOTE]
The heavy usage of interfaces is only here to show purposes and types. It can be replaced with abstract classes during full implementation.

### DIFF
--- a/src/modules/Switcher/Element/Current_Language_Element.php
+++ b/src/modules/Switcher/Element/Current_Language_Element.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @package Polylang
+ */
+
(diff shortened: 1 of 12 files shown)


Hug0-Drelon:
Won't use.

PR #1874, opened 2026-04-27
Sort the list of flags by alphabetical order

### LINKED ISSUE #533: Flag missorted in list
## Prerequisites

1. Go to Languages / Languages
2. Search the United Kingdom flag
3. United Kingdom seems to be alphabetically between Galicia and Grenada :-)

This flag should be listed in the "U" or renamed "Great Britain"
gb

Chouby:
Closes #533

This PR uses [`Collator`](https://www.php.net/manual/en/class.collator.php) to sort the flags by alphabetical order (depending on the user locale). I did not implement a locale conversion as the WordPress locale should be correctly interpreted in most cases.
If we have complaints that the order is crazy in some locales which could not be correctly interpreted, we may introduce a conversion in a second time.

If the `intl` extension is not present, then I fallback to standard sort (which will work correctly for English).

### DIFF
--- a/src/settings/settings.php
+++ b/src/settings/settings.php
@@ -356,7 +356,7 @@ class PLL_Settings extends PLL_Admin_Base {
 	}
 
 	/**
-	 * Get the list of predefined languages
+	 * Get the list of predefined languages.
 	 *
 	 * @since 2.3
 	 *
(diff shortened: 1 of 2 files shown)


src/settings/settings.php
```diff
@@ -408,4 +410,26 @@ public static function get_predefined_languages() {
...
+		if ( class_exists( 'Collator' ) ) {
+			$co = Collator::create( get_user_locale() );
+			if ( $co ) {
+				$co->asort( $flags );
+				return $flags;
+			}
+		}
+
+		asort( $flags );
+		return $flags;
```
Screenfeed: I would reverse the logic, so we don't pile up 2 `if`:


```suggestion
		if ( ! class_exists( 'Collator' ) ) {
			asort( $flags );
			return $flags;
		}

		$co = Collator::create( get_user_locale() );
		if ( $co ) {
			$co->asort( $flags );
		}

		return $flags;
```

What do you think?

Screenfeed (changes requested):

Screenfeed (approved):

PR #1875, opened 2026-04-28
Introduce POLYLANG_ROOT_DIR constant

Chouby:
Follow-up of #1864

With the flags introduced as a dependency, we now need to access to the vendor directory from Polylang, whether Polylang is standalone plugin or a dependency of Polylang Pro. It's convenient to introduce a `POLYLANG_ROOT_DIR` constant similar to the existing `POLYLANG_ROOT_FILE`.

- Use this new constant to access the flags in the vendor directory.
- Fix the flags URLs for Polylang Pro, by replacing `POLYLANG_ROOT_FILE` instead of `POLYLANG_FILE` as second param of `plugins_url()`.

### DIFF
--- a/polylang.php
+++ b/polylang.php
@@ -68,11 +68,12 @@ define( 'PLL_MIN_PHP_VERSION', '7.4' );
 define( 'POLYLANG_FILE', __FILE__ );
 define( 'POLYLANG_DIR', __DIR__ );
 
-// Whether we are using Polylang or Polylang Pro, get the filename of the plugin in use.
+// The filename of the plugin in use, whether it is Polylang or Polylang Pro.
 if ( ! defined( 'POLYLANG_ROOT_FILE' ) ) {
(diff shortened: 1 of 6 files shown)


polylang.php
```diff
@@ -73,6 +73,10 @@ static function () {
 	define( 'POLYLANG_ROOT_FILE', __FILE__ );
 }
 
+if ( ! defined( 'POLYLANG_ROOT_DIR' ) ) {
+	define( 'POLYLANG_ROOT_DIR', __DIR__ );
+}
```
Hug0-Drelon, line 79: I think these constants requires a bit more documentation. Especially about constant definition from Polylang Pro.

Hug0-Drelon (changes requested):
Can't we use `dirname( POLYLANG_ROOT_FILE );` instead of creating one more constant?

PR #1878, opened 2026-05-04
Remove wpcom_vip_get_page_by_path()

Chouby:
This function has been deprecated since the release of WP 6.1.


Marianne380 (approved):
Does this need a changelog entry?

Chouby:
I don't believe it's necessary as we don't support WP versions where this function is not deprecated.

PR #1879, opened 2026-05-04
Introduce functions to register and enqueue scripts and styles

Chouby:
The goal is to avoid to repeat ourselves.

Some decisions:
- We manipulate handles (prefixed) instead of filenames.
- The version is replaced by the time when `WP_DEBUG` is `true`. We don't use `SCRIPT_DEBUG` because this variable is intentionally set to `false` by the test team.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -172,3 +172,81 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
 
 	return 'GET' === $request->get_method() && 'edit' === $request->get_param( 'context' );
 }
+
+/**
+ * Enqueues a JS script.
+ *
+ * @since 3.9
+ *
+ * @param string   $name         The filename of the script (without the extension).
+ * @param string[] $dependencies Optional. An array of registered script handles this script depends on.
+ * @param array $args     {
+ *     Optional. An array of additional script loading strategies. Default empty array.
+ *
+ *     @type string    $strategy      Optional. If provided, may be either 'defer' or 'async'.
+ *     @type bool      $in_footer     Optional. Whether to print the script in the footer. Default 'false'.
+ *     @type string    $fetchpriority Optional. The fetch priority for the script. Default 'auto'.
+ * }
+ * @return void
+ */
+function pll_enqueue_script( string $name, array $dependencies = array(), array $args = array() ): void {
+	pll_register_script( $name, $dependencies, $args );
+	wp_scripts()->enqueue( 'pll-' . $name );
+}
+
+/**
+ * Registers a JS script.
+ *
+ * @since 3.9
+ *
+ * @param string   $name         The filename of the script (without the extension).
+ * @param string[] $dependencies Optional. An array of registered script handles this script depends on.
+ * @param array $args     {
+ *     Optional. An array of additional script loading strategies. Default empty array.
+ *
+ *     @type string    $strategy      Optional. If provided, may be either 'defer' or 'async'.
+ *     @type bool      $in_footer     Optional. Whether to print the script in the footer. Default 'false'.
+ *     @type string    $fetchpriority Optional. The fetch priority for the script. Default 'auto'.
+ * }
+ * @return void
+ */
+function pll_register_script( string $name, array $dependencies = array(), array $args = array() ): void {
+	$handle  = 'pll-' . $name;
+	$suffix  = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+	$file    = '/js/build/' . $name . $suffix . '.js';
+	$src     = plugins_url( $file, POLYLANG_ROOT_FILE );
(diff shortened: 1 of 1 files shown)


src/functions.php
```diff
@@ -172,3 +172,87 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
...
+ *     @type string    $fetchpriority Optional. The fetch priority for the script. Default 'auto'.
+ * }
+ * @return void
+ */
+function pll_register_script( string $handle, array $dependencies = array(), array $args = array() ): void {
+	$suffix  = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+	$file    = '/js/build/' . substr( $handle, 4 ) . $suffix . '.js';
+	$src     = plugins_url( $file, POLYLANG_ROOT_FILE );
+	$version = filemtime( POLYLANG_ROOT_DIR . $file );
+	wp_register_script( $handle, $src, $dependencies, $version, $args );
```
Hug0-Drelon:
```suggestion
	wp_register_script( $handle, $src, $dependencies, (string) $version, $args );
```
To please PHPStan.

src/functions.php
```diff
@@ -172,3 +172,87 @@ function pll_is_edit_rest_request( WP_REST_Request $request ): bool {
...
+ * @param string   $handle       The handle of the stylesheet.
+ * @param string[] $dependencies Optional. An array of registered stylesheet handles this stylesheet depends on.
+ * @return void
+ */
+function pll_register_style( string $handle, array $dependencies = array() ): void {
+	$suffix  = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+	$file    = '/js/build/' . substr( $handle, 4 ) . $suffix . '.css';
+	$src     = plugins_url( $file, POLYLANG_ROOT_FILE );
+	$version = filemtime( POLYLANG_ROOT_DIR . $file );
+	wp_register_style( $handle, $src, $dependencies, $version );
```
Hug0-Drelon:
```suggestion
	wp_register_style( $handle, $src, $dependencies, (string) $version );
```
Same suggestion as above.

Hug0-Drelon (changes requested):

src/functions.php
```diff
@@ -186,3 +186,87 @@ function pll_sanitize_ids( $ids ): array {
...
+ *
+ * @since 3.9
+ *
+ * @param string   $handle       The handle of the stylesheet.
+ * @param string[] $dependencies Optional. An array of registered stylesheet handles this stylesheet depends on.
+ * @return void
+ */
+function pll_register_style( string $handle, array $dependencies = array() ): void {
+	$suffix  = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+	$file    = '/js/build/' . substr( $handle, 4 ) . $suffix . '.css';
```
Hug0-Drelon:
```suggestion
	$file    = '/css/build/' . substr( $handle, 4 ) . $suffix . '.css';
```
:suspect:

Hug0-Drelon (changes requested):
I find using `substr( $handle, 4 )` a bit brittle. Could we use a constant instead and throw with something like:
```php
$string = str_starts_with( $string, PLL_HANDLE_PREFIX)
    ? substr( $string, strlen( PLL_HANDLE_PREFIX ) ) // Or hardcoded since the length should not vary.
    : _doing_it_wrong( /* ... */ );
```

Hug0-Drelon:
I wonder if the script/style handle should be dynamic so it can be used in our add-ons (e.g. `plle-` or `pllwc-`).

src/functions.php
Chouby: OMG

Chouby:
> I find using `substr( $handle, 4 )` a bit brittle. Could we use a constant instead and throw with something like:
> 
> ```
> $string = str_starts_with( $string, PLL_HANDLE_PREFIX)
>     ? substr( $string, strlen( PLL_HANDLE_PREFIX ) ) // Or hardcoded since the length should not vary.
>     : _doing_it_wrong( /* ... */ );
> ```

I don't like useless operations. So I prefer to hardcode the length.
However, I initially used the file name instead of the handle as param. See f2c202c381837ccd6dd07a513f38e9483df3d07e. Do you prefer the previous solution?

> I wonder if the script/style handle should be dynamic so it can be used in our add-ons (e.g. `plle-` or `pllwc-`).

The idea was to keep the number of params as low as possible (this the purpose of this PR). So I would rather create separate functions in the addons.
However, if we go to use the name instead of the handle, one possibility could be to use a 4th param `$prefix = 'pll-'` then define, for ex:
 ```php
function pllwc_register_script( $handle, $dependencies, $args ) {
	pll_register_script( $handle, $dependencies, $args, 'pllwc-' );
}
```

Hug0-Drelon (approved):

PR #1880, opened 2026-05-04
Readonly property cannot have a default value

Chouby:
Fix the PHPStan error.

I selected to initialize the default values in the constructor of `PLL_Language`. This could also have been done in the factory. To be discussed.

### DIFF
--- a/src/language.php
+++ b/src/language.php
@@ -111,7 +111,7 @@ class PLL_Language extends PLL_Language_Deprecated {
 	 *
 	 * @var string
 	 */
-	public $facebook = '';
+	public $facebook;
 
 	/**
 	 * Home URL in this language.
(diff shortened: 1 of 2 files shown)


src/language.php
```diff
@@ -293,7 +293,18 @@ class PLL_Language extends PLL_Language_Deprecated {
...
+		// Default values for optional params.
+		$defaults = array(
+			'facebook'        => '',
+			'page_on_front'   => 0,
+			'page_for_posts'  => 0,
+			'custom_flag_url' => '',
+			'custom_flag'     => '',
+			'active'          => true,
+			'fallbacks'       => array(),
+		);
```
Screenfeed, line 305: We have mismatch information between `LanguageData` and the DocBlock.

1. Only `facebook`, `custom_flag_url`, `custom_flag`, and `fallbacks` are marked as optional in `LanguageData`.
2. However, `page_on_front` and `page_for_posts` are marked optional in the DocBlock.
3. `active` is described with "Default `true`." in the DocBlock but doesn't say "Optional." and is not optional in `LanguageData`.
4. `fallbacks` is not marked as optional in the DocBlock, but it is in `LanguageData`.

Screenfeed (changes requested):

Screenfeed (approved):

PR #1881, opened 2026-05-07
Preserve empty query string and do not override existing query param in PLL_Filter_REST_Routes::add_query_parameters.

Hug0-Drelon:
## What?
Required by https://github.com/polylang/polylang-pro/pull/2994
There is two problem at once here:
- `add_query_arg()` is used to add parameters to preload paths, but it rewrite `key=` to `key` in the query string.
It is usually not a problem, but preload paths are cache in the browser and `apiFetch` cache middleware use `normalizePath` to canonicalize paths and then compare them. While `apiFetch` adds `key=` instead of `key` for empty string parameters. Leading to cache miss.
- `PLL_Filter_REST_Routes::add_query_parameters` override existing query parameters, leading to missed cache hit because if a path is added with `lang=` it'll be replace by `lang=current_lang_slug`.

## How?
- Use `build_query` directly instead of `add_query_arg()` to preserve `key=` queries.
- Check a parameter doesn't already exist in the preloaded path before adding it.

### DIFF
--- a/src/filter-rest-routes.php
+++ b/src/filter-rest-routes.php
@@ -74,8 +74,14 @@ class PLL_Filter_REST_Routes {
 			// Sort query params to put it in the same order as the preloading middleware does
 			ksort( $query_params );
 
-			// Replace the key by the correct path with query params reordered
-			$sorted_path = add_query_arg( urlencode_deep( $query_params ), $path_parts['path'] );
+			// Replace the key by the correct path with query params reordered.
+			// Uses `build_query()` instead of `add_query_arg()` to preserve empty values as `key=`.
+			$sorted_query = build_query( urlencode_deep( $query_params ) );
+			$sorted_path  = $path_parts['path'];
+
+			if ( '' !== $sorted_query ) {
+				$sorted_path .= '?' . $sorted_query;
+			}
 
 			if ( is_array( $path ) ) {
 				$preload_paths[ $k ][0] = $sorted_path;
--- a/src/filter-rest-routes.php
+++ b/src/filter-rest-routes.php
@@ -79,11 +79,11 @@ class PLL_Filter_REST_Routes {
 			// Sort query params to put it in the same order as the preloading middleware does
 			ksort( $query_params );
 
-			// Replace the key by the correct path with query params reordered.
-			// Uses `build_query()` instead of `add_query_arg()` to preserve empty values as `key=`.
 			$sorted_query = build_query( urlencode_deep( $query_params ) );
 			$sorted_path  = $path_parts['path'];
 
+			// Replace the key by the correct path with query params reordered.
+			// Uses `build_query()` instead of `add_query_arg()` to preserve empty values as `key=`.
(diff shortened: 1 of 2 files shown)


src/filter-rest-routes.php
```diff
@@ -68,14 +68,25 @@ public function add_query_parameters( array $preload_paths, array $args ): array
...
-			// Replace the key by the correct path with query params reordered
-			$sorted_path = add_query_arg( urlencode_deep( $query_params ), $path_parts['path'] );
+			// Replace the key by the correct path with query params reordered.
+			// Uses `build_query()` instead of `add_query_arg()` to preserve empty values as `key=`.
+			$sorted_query = build_query( urlencode_deep( $query_params ) );
+			$sorted_path  = $path_parts['path'];
+
+			if ( '' !== $sorted_query ) {
+				$sorted_path .= '?' . $sorted_query;
+			}
```
Marianne380, line 89: Should we add a comment  to explain why we have to do this now that we're not using `add_query_arg` anymore ?

Marianne380 (changes requested):

src/filter-rest-routes.php
Hug0-Drelon: Isn't it sufficient to you; https://github.com/polylang/polylang/pull/1881/changes#diff-bdfe7a4648c187f4c5aa1786d553a4dd78fc319f24ba4a787f1f07c23ec96ce0R82-R83 ?

Marianne380: Yes, that's sufficient! Could you move it just above the `if` though?

Marianne380 (changes requested):

src/filter-rest-routes.php
Hug0-Drelon: Sure, done in 974e0a57.

src/filter-rest-routes.php
```diff
@@ -68,14 +68,25 @@ public function add_query_parameters( array $preload_paths, array $args ): array
...
 			ksort( $query_params );
 
-			// Replace the key by the correct path with query params reordered
-			$sorted_path = add_query_arg( urlencode_deep( $query_params ), $path_parts['path'] );
+			$sorted_query = build_query( urlencode_deep( $query_params ) );
+			$sorted_path  = $path_parts['path'];
+
+			// Replace the key by the correct path with query params reordered.
+			// Uses `build_query()` instead of `add_query_arg()` to preserve empty values as `key=`.
+			if ( '' !== $sorted_query ) {
```
Marianne380, line 87:
```suggestion
            // Uses `build_query()` instead of `add_query_arg()` to preserve empty values as `key=`.
			$sorted_query = build_query( urlencode_deep( $query_params ) );
			$sorted_path  = $path_parts['path'];

			// Replace the key by the correct path with query params reordered.
			if ( '' !== $sorted_query ) {
```

Marianne380 (changes requested):

Marianne380 (approved):

PR #1884, opened 2026-05-11
Add ESLint on Gutenberg blocks JS source code.

Marianne380:
Fix https://github.com/polylang/polylang-pro/issues/2999.

## What

Add ESLint on the Gutenberg blocks JS source code.

## Why

When the blocks were in Polylang Pro, ESLint was running on their source code. Since the migration to Polylang (#1811 ), this was no longer the case.

## How

- Add `.eslintrc` with rules matching those used in Polylang Pro, scoped to the blocks code
- Update the `static-analysis.yml` workflow to lint `js/src/blocks/` in addition to `tests/e2e/`
- Fix prettier issues raised by ESLint in `webpack.config.js`, `switcher-container.js`, and `navigation-language-switcher/components/navigation-switcher-container.js`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,19 +1,5 @@
 {
 	"extends": [ "plugin:@wordpress/eslint-plugin/recommended" ],
-	"ignorePatterns": [
-		"js/src/admin.js",
-		"js/src/block-editor.js",
-		"js/src/classic-editor.js",
-		"js/src/lib/**",
-		"js/src/media.js",
-		"js/src/nav-menu.js",
-		"js/src/post.js",
-		"js/src/settings.js",
-		"js/src/term.js",
-		"js/src/user.js",
-		"js/src/widgets.js",
-		"src/modules/wizard/js/languages-step.js"
-	],
 	"settings": {
 		"import/core-modules": [
 			"@wordpress/block-editor",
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -1,6 +1,53 @@
...
+		"js/src/lib/**",
+		"js/src/media.js",
+		"js/src/nav-menu.js",
+		"js/src/post.js",
+		"js/src/settings.js",
+		"js/src/term.js",
+		"js/src/user.js",
+		"js/src/widgets.js",
+		"src/modules/wizard/js/languages-step.js"
+	],
```
Hug0-Drelon: I wouldn't add this rule to the main config. If for some reason we want to run the linter on this file locally, it'll force us to remove these lines.

You already ensured the CI won't run on these files, which is enough IMHO.

Hug0-Drelon (changes requested):

.eslintrc
Marianne380: You're right, we'll manage locally, see https://github.com/polylang/polylang/pull/1884/changes/5f625d0a960d3906593bdc4196bbfda4d66be068.

Hug0-Drelon (approved):

PR #1887, opened 2026-05-16
Fix renamed media files deleted when deleting the first media translation

### LINKED ISSUE #1883: Bug: When an image is edited within WordPress, deleting the image will delete all its files (even when it has translations)
**Describe the bug**

In WordPress, when the user deletes a media item, all the files (.jpg, .png, etc) are also deleted, for all of them: The pic, the  "original_image" (if any), and the thumbnails.                                                                                                 
                                                                                    
With Polylang, when the image has translations, deleting the image in any language will not delete the files. Polylang prevents that in `PLL_CRUD_Posts`, method `wp_delete_file`.
                                                                                                                                                 
However, that logic fails when you edit the image in WordPress (eg: you scale it down), as it'll create another pic. Eg: if the uploaded pic is "golden-triangle.jpg",  it'll create "golden-triangle-scaled.jpg" ("golden-triangle.jpg" remains as the "original_image")

Then, deleting an image that was edited in WP will delete its files, even when there are translations.

**Reproduced with**
- WordPress version: 6.9.4
- Polylang version: 3.8.3

**To Reproduce**
1. Go to Media
2. Translate any media item
3. Edit the media item in any way, using the Media editor (eg: scale it down)

Image

Image

4. Delete the media item
5. All its thumbnails will have been deleted

**Expected behavior**
Images should not be deleted.

Chouby:
Fixes #1883 

We prevent the file deletion when a media translation is deleted.
In the process, we use to by re-build the filename from the meta `_wp_attached_file` to match it against the deleted file.
This is however not reliable in cases where WP renames files (when they are edited or scaled at upload).
In this PR we match directly the deleted file against a filename stored in the meta `_wp_attachment_data` (serialized array).


szepeviktor (changes requested):
That means intermediate-size/backups matching is filename-only, not path-qualified. If two unrelated attachments in different upload subdirectories happen to share the same basename, deleting one file could incorrectly appear to belong to another translated attachment and block deletion.

Example risk:

    2025/05/photo-scaled.jpg
    2026/06/photo-scaled.jpg

Chouby:
Thank you @szepeviktor. I overlooked this case. I added a test for it and modified the SQL query to fix this + another bug I found out when I reviewed my code.

szepeviktor:
Glad to contribute to Polylang.

src/crud-posts.php
```diff
@@ -269,21 +269,25 @@ public function delete_post( $post_id ) {
 	public function wp_delete_file( $file ) {
 		global $wpdb;
 
-		$uploadpath = wp_upload_dir();
+		$upload   = wp_upload_dir();
+		$basefile = basename( $file );
+		$subdir   = str_replace( $upload['basedir'], '', $file );
```
Hug0-Drelon: Should we check for `$upload['error']` before going any further?

src/crud-posts.php
```diff
@@ -269,21 +269,25 @@ public function delete_post( $post_id ) {
...
+		$subdir   = trim( str_replace( $basefile, '', $subdir ), '/' );
```
Hug0-Drelon: If the file is directly under `uploads/` (no date folder), we get `LIKE '/%'` for `pm1.meta_value`.

Could it be a problem?

tests/phpunit/tests/test-media.php
```diff
@@ -85,4 +93,79 @@ public function test_create_media_translation_with_slashes() {
...
+		}
+
+		wp_delete_attachment( $fr );
+		foreach ( $filenames as $filename ) {
+			$this->assertFileDoesNotExist( $filename, 'Deleting all translations must delete the files.' );
+		}
+	}
+
+	protected static function upload_dir_filter( $time ) {
+		return static function ( $upload ) use ( $time ) {
```
Hug0-Drelon, line 211: ~~Returning a static function inside a static function seems overkill.
Could we make `upload_dir_filter` public instead (with a use keyword)?~~

Not relevant.

tests/phpunit/tests/test-media.php
```diff
@@ -85,4 +93,79 @@ public function test_create_media_translation_with_slashes() {
...
+			$this->assertFileExists( $filename, 'Deleting a translation must not delete the files' );
+		}
+
+		wp_delete_attachment( $fr );
+		foreach ( $filenames as $filename ) {
+			$this->assertFileDoesNotExist( $filename, 'Deleting all translations must delete the files.' );
+		}
+	}
+
+	protected static function upload_dir_filter( $time ) {
```
Hug0-Drelon:
```suggestion
	protected static function set_upload_dir_time( $time ) {
```
Proposing a more meaningful name.

tests/phpunit/tests/test-media.php
```diff
@@ -85,4 +93,79 @@ public function test_create_media_translation_with_slashes() {
...
+		add_filter( 'upload_dir', $filter );
+		self::factory()->attachment->create_upload_object( $filename );
+		remove_filter( 'upload_dir', $filter );
+
+		// And a new one with the same name.
+		$en = self::factory()->attachment->create_upload_object( $filename );
+		$fr = $this->pll_admin->model->post->create_media_translation( $en, 'fr' );
+
+		$uploads_dir = wp_upload_dir();
+		$filenames   = glob( "{$uploads_dir['basedir']}{$uploads_dir['subdir']}/big-image*" );
```
Hug0-Drelon:
```suggestion
		$filenames   = glob( "{$uploads_dir['basedir']}**/big-image*" );
```
Shouldn't it be a wildcard for subdirectories?

Hug0-Drelon (changes requested):

src/crud-posts.php
Chouby: I believe it's useless because it should be always false. Errors may occur only when attempting to create a directory. Here we are deleting files, so the directory already exists.

src/crud-posts.php
Chouby: Good catch! I believe it will.

tests/phpunit/tests/test-media.php
Chouby: No. We want to check files in only one subdirectory.

src/crud-posts.php
```diff
@@ -269,21 +269,25 @@ public function delete_post( $post_id ) {
...
-		$attached_file = preg_replace( '#-\d+x\d+\.([a-z]+)$#', '.$1', $attached_file );
-
-		$ids = $wpdb->get_col(
+		// Search in the serialized array of the attachment metadata.
+		$count = $wpdb->get_var(
 			$wpdb->prepare(
-				"SELECT post_id FROM $wpdb->postmeta
-				WHERE meta_key = '_wp_attached_file' AND meta_value = %s",
-				$attached_file
+				"SELECT COUNT(*) FROM $wpdb->postmeta AS pm1
```
szepeviktor: RoboViktor opens his laser eyes.

> src/crud-posts.php:280 regresses deletion safety for non-image attachments. The new guard only counts sibling attachments that also have _wp_attachment_metadata or _wp_attachment_backup_sizes. That works for images, but attachments like ZIPs, text files, and many other media types commonly have only _wp_attached_file. In those cases the JOIN finds nothing, $count becomes 0, and deleting one translation will **delete the shared physical file** even if another translation still points to it. The previous exact-match lookup on _wp_attached_file did not have this hole.

szepeviktor: (my secret prompt for Codex CLI: `please review this branch on top of master`)

tests/phpunit/tests/test-media.php
Chouby: But it can be simplified. See ce6c34d48e49ea72bdb07a75b42d66fa9c356ce5

src/crud-posts.php
Chouby: See a34fdbb8ae9808a64bad1e5a3813f09f6ebb58b6

src/crud-posts.php
Chouby: My head will explode! I will look at that tomorrow.

src/crud-posts.php
```diff
@@ -269,21 +269,37 @@ public function delete_post( $post_id ) {
...
+					JOIN $wpdb->postmeta AS pm2 ON pm1.post_id = pm2.post_id
+					WHERE pm1.meta_key = '_wp_attached_file' AND pm1.meta_value LIKE %s
+					AND pm2.meta_key IN ( '_wp_attachment_metadata', '_wp_attachment_backup_sizes' )
+					AND ( pm2.meta_value LIKE %s OR pm2.meta_value LIKE %s )",
+					$wpdb->esc_like( $subdir ) . '/%', // The subdir is always present in '_wp_attached_file'.
+					'%"' . $wpdb->esc_like( $basefile ) . '"%', // Intermediate sizes in '_wp_attachment_metadata' + '_wp_attachment_backup_sizes'.
+					'%"' . $wpdb->esc_like( "$subdir/$basefile" ) . '"%' // Main file in '_wp_attachment_metadata'.
+				)
+			);
+		}
```
Hug0-Drelon:
```suggestion
		$upload        = wp_upload_dir();
		$basefile      = basename( $file );
		$subdir        = str_replace( $upload['basedir'], '', $file );
		$subdir        = trim( str_replace( $basefile, '', $subdir ), '/' );
		$basefile_like = '%"' . $wpdb->esc_like( $basefile ) . '"%';

		// Search in the serialized array of the attachment metadata.
		$count = $wpdb->get_var(
			$wpdb->prepare(
				"SELECT COUNT(*) FROM $wpdb->postmeta AS pm1
				LEFT JOIN $wpdb->postmeta AS pm2
					ON pm1.post_id = pm2.post_id
					AND %s != ''
					AND pm2.meta_key IN ( '_wp_attachment_metadata', '_wp_attachment_backup_sizes' )
					AND ( pm2.meta_value LIKE %s OR pm2.meta_value LIKE %s )
				WHERE (
					%s = ''
					AND pm1.meta_key IN ( '_wp_attachment_metadata', '_wp_attachment_backup_sizes' )
					AND pm1.meta_value LIKE %s
				) OR (
					%s != ''
					AND pm1.meta_key = '_wp_attached_file'
					AND pm1.meta_value LIKE %s
					AND pm2.post_id IS NOT NULL
				)",
				$subdir,
				$basefile_like,
				'%"' . $wpdb->esc_like( "$subdir/$basefile" ) . '"%',
				$subdir,
				$basefile_like,
				$subdir,
				$wpdb->esc_like( $subdir ) . '/%'
			)
		);
```
Proposal; using a single SQL query is easier to maintain. Bonus, it lets SQL manage the empty `$subdir` case.

tests/phpunit/tests/test-media.php
Hug0-Drelon, line 1: I'd suggest to do `$this->assertNotEmpty( $filenames );` before `foreach`. It is good hygiene to check an array is not empty before doing assertions on each items.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

Issue #1057, opened 2022-06-07
Feat: Add admin table filters to Post Type and Taxonomies


jpSimkins:
I am aware of the admin toolbar feature to filter between: All Languages or each language

The issue with that toolbar, is that while it does filter the content, it's not really useful from a management perspective. The admin toolbar does not alter the totals for the WP statuses.

I propose having a secondary filter. This should allow you to filter any language or all languages.

This is what WPML has:
![Screenshot from 2022-06-07 11-56-53](https://user-images.githubusercontent.com/4512118/172426712-6660e279-55a1-4243-b045-7a4ad43c0fe5.png)

While I understand this is not WPML, I believe that filter is more usable than the admin toolbar filter.
- It has proper counts for each language
  - Can easily check the total against business expectations.  The current system requires loading the page for each language to get the count from the table. As more languages are added, this can get time consuming.
- It's clear which language you are in 
  - I cannot count how many times I had to point this out to others that it's in the admin topbar

As someone who is migrating from WPML to PolyLang, this is perhaps one of the largest issues with the switch. I am working on a small plugin to add this to my sites but it would be nice if this were part of PolyLang itself.

Thank you

Chouby:
This looks like a duplicate of #133. Am I wrong?

jpSimkins:
@Chouby Yes and no. You are correct about it being similar due to that the totals are not reflected but what I am asking for is a similar filter like this on the tables. This is because they have the filter in the topbar but it's not very obvious (most my team didn't even notice it). 

So yes, the issue is similar but this is asking for the filter to be added to the table (which would also update the totals properly then). Hope that makes sense.

PR #1890, opened 2026-05-20
Improve count_posts() to use only one query for all post statuses

Chouby:
Preparatory work for #1057

### DIFF
--- a/src/model.php
+++ b/src/model.php
@@ -370,12 +370,11 @@ class PLL_Model {
 
 		if ( ! is_array( $counts ) ) {
 			$counts  = array();
-			$select  = "SELECT pll_tr.term_taxonomy_id, COUNT( * ) AS num_posts FROM {$wpdb->posts}";
+			$select  = "SELECT pll_tr.term_taxonomy_id, post_status, COUNT( * ) AS num_posts FROM {$wpdb->posts}";
 			$join    = $this->post->join_clause();
-			$where   = sprintf( " WHERE post_status = '%s'", esc_sql( $q['post_status'] ) );
-			$where  .= sprintf( " AND {$wpdb->posts}.post_type IN ( '%s' )", implode( "', '", esc_sql( $q['post_type'] ) ) );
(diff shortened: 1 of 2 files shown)


src/model.php
```diff
@@ -424,15 +422,21 @@ public function count_posts( $lang, $q = array() ): int {
...
 		}
 
 		$term_taxonomy_id = $lang->get_tax_prop( 'language', 'term_taxonomy_id' );
-		return empty( $counts[ $term_taxonomy_id ] ) ? 0 : $counts[ $term_taxonomy_id ];
+
+		if ( 'any' === $q['post_status'] ) {
+			return (int) array_sum( $counts[ $term_taxonomy_id ] );
+		}
+
+		return $counts[ $term_taxonomy_id ][ $q['post_status'] ] ?? 0;
```
Hug0-Drelon, line 439: Could we add a `WHERE` clause on the `post_status` (with some adaptations for `any` status)?

I'd prevent to compute in PHP (better perf).

Hug0-Drelon, line 439: I wonder if it'd be better to use SQL `COUNT` function directly.
Performance will increase and no filters seem to be applied here.

Hug0-Drelon (changes requested):

src/model.php
Chouby: Sorry I don't see what you have in mind. I removed [`WHERE post_status`](https://github.com/polylang/polylang/pull/1890/files#diff-97a585a968d4ef643af0f92f137cc1d2c65b9c040017a04ddbacf4e7e1639aafL375) to query them all at once to avoid multiple sql queries in case we need the count of posts in multiple statuses.

Chouby: Here too, I don't understand what you mean. `COUNT` is already used to get the count of posts for each post status. 'any' is not a post status so we need to sum all the counts we have. I guess we could do the sum in sql with extra complexity, but I don't see the added value.

Hug0-Drelon (approved):

PR #1891, opened 2026-05-21
Translate taxonomy filter in post list table when switching the language filter

Chouby:
Fix https://github.com/polylang/polylang-pro/issues/2115

### DIFF
--- a/src/admin/admin-base.php
+++ b/src/admin/admin-base.php
@@ -583,6 +583,8 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	 *
 	 * @param object $language The language or an object for all languages.
 	 * @return string
+	 *
+	 * @phpstan-param object{'slug': string} $language
 	 */
 	protected function get_admin_bar_menu_url( $language ): string {
 		global $pagenow;
@@ -593,20 +595,35 @@ abstract class PLL_Admin_Base extends PLL_Base {
 			return $url;
 		}
 
-		foreach ( get_object_taxonomies( get_post_type(), 'objects' ) as $tax ) {
+		// Attempt to translate the category (taxonomy) filter if present.
+		$post_type = get_post_type();
+
+		if ( ! $post_type ) {
+			return $url;
+		}
+
+		foreach ( get_object_taxonomies( $post_type, 'objects' ) as $tax ) {
 			if ( ! $this->model->is_translated_taxonomy( $tax->name ) ) {
 				continue;
 			}
 
 			if ( 'category' === $tax->name ) {
-				$query_var = 'cat';
 				$field     = 'term_id';
+				$query_var = 'cat';
+				$qv        = get_query_var( $query_var );
+				if ( ! is_int( $qv ) ) {
+					continue;
+				}
 			} else {
-				$query_var = $tax->query_var;
 				$field     = 'slug';
+				$query_var = (string) $tax->query_var;
+				$qv        = get_query_var( $query_var );
+				if ( ! is_string( $qv ) ) {
+					continue;
+				}
 			}
 
-			$qv = $this->model->term->get_by( $field, get_query_var( $query_var ), $language, $tax->name );
+			$qv = $this->model->term->get_by( $field, $qv, $language, $tax->name );
 			if ( ! empty( $qv ) ) {
 				$url = add_query_arg( $query_var, $qv, remove_query_arg( $query_var, $url ) );
 			}
--- a/src/admin/admin-base.php
+++ b/src/admin/admin-base.php
@@ -551,7 +551,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 			array(
 				'id'    => 'languages',
 				'title' => $selected->flag . $title,
-				'href'  => esc_url( add_query_arg( 'lang', $selected->slug, remove_query_arg( 'paged' ) ) ),
+				'href'  => esc_url( $this->get_admin_bar_menu_url( $selected ) ),
 				'meta'  => array(
 					'title' => __( 'Filters content by language', 'polylang' ),
 					'class' => 'all' === $selected->slug ? '' : 'pll-filtered-languages',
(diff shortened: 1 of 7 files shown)


src/translated-term.php
```diff
@@ -449,6 +449,40 @@ public function update( int $term_id, array $args = array() ) {
 		return $term;
 	}
 
+	/**
+	 * Get a translated term by term field
```
Hug0-Drelon:
```suggestion
	 * Gets a translated term by term field
```

src/translated-term.php
```diff
@@ -449,6 +449,40 @@ public function update( int $term_id, array $args = array() ) {
...
+	 *
+	 * @since 3.9
+	 *
+	 * @param string                  $field    The term field. Accepted values are 'term_id', 'slug', 'name', 'term_taxonomy_id'.
+	 * @param int|string              $value    Search for this term value.
+	 * @param PLL_Language|string|int $lang     Language (object, slug, or term ID).
```
Hug0-Drelon: We tend to use only `PLL_Language` for new methods. Is it applicable here?

src/admin/admin-base.php
```diff
@@ -569,13 +569,69 @@ public function admin_bar_menu( $wp_admin_bar ) {
...
+			return $url;
+		}
+
+		foreach ( get_object_taxonomies( $post_type, 'objects' ) as $tax ) {
+			if ( ! $this->model->is_translated_taxonomy( $tax->name ) ) {
+				continue;
+			}
+
+			if ( 'category' === $tax->name ) {
+				$field     = 'term_id';
```
Hug0-Drelon: Should we set the variable only if the next condition is not met?

src/admin/admin-base.php
```diff
@@ -569,13 +569,69 @@ public function admin_bar_menu( $wp_admin_bar ) {
...
+				$query_var = 'cat';
+				$qv        = get_query_var( $query_var );
+				if ( ! is_int( $qv ) ) {
+					continue;
+				}
+			} else {
+				$field     = 'slug';
```
Hug0-Drelon: Same question as above.

Hug0-Drelon (changes requested):

src/translated-term.php
Chouby: We could. And in fact all the calls use `PLL_Language` but the truth is that it works with other cases too. The reason is that it's used in `[$this->get()](https://github.com/polylang/polylang/pull/1891/files#diff-0b0baa5e764f3d952223632601993f15eff6e7e491fc12866299de2dd0d42228R480) which accepts them all.

If you prefer, we can change that and add strict typing (that way the doc would still tell the truth).

src/admin/admin-base.php
Chouby: I believe that it didn't matter but I changed that in cae23e73c69d50b162b7d93dc003cdcb8452b530 as I discovered a bug.

Indeed, for categories, we can use the filter above the list table which will add `cat` in the url. Or we can click on a category name which will add `category_name` in the url. In both case, both values are present in the `WP_Query`. The previous code resulted in a conflict I we clicked on the category name (`category_name` was present with original category and `cat` was present with the translated category, resulting in an empty list). cae23e73c69d50b162b7d93dc003cdcb8452b530 should fix that.

src/translated-term.php
Hug0-Drelon: I think using stricter type is a good idea for this new method.

tests/phpunit/tests/test-admin.php
```diff
@@ -1,10 +1,11 @@
...
+	 * @param PLL_UnitTest_Factory $factory
+	 * @return void
 	 */
-	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
-		parent::wpSetUpBeforeClass( $factory );
+	public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
+		parent::pllSetUpBeforeClass( $factory );
 
 		self::create_language( 'en_US' );
 		self::create_language( 'fr_FR' );
```
Hug0-Drelon:
```suggestion
$factory->language->create_many( 2 );
```

src/admin/admin-base.php
Chouby: Done in 0a37f36eebcb8286ca630d33ab99875503abbbb5

tests/phpunit/tests/test-admin.php
Chouby: Applied in 4ec0e6012337ba4f80ea1a5f04d25ba913ac0b71

src/translated-term.php
```diff
@@ -449,6 +449,41 @@ public function update( int $term_id, array $args = array() ) {
...
+
+		$t = reset( $terms );
+		if ( ! $t instanceof WP_Term ) {
+			return $empty;
+		}
+
+		$tr_id   = $this->get( $t->term_id, $lang );
+		$tr_term = get_term( $tr_id );
+
+		return $tr_term instanceof WP_Term ? $tr_term->$field : $empty;
```
Hug0-Drelon, line 484: Should we validate `WP_Term::$field` properties beforehand?

src/frontend/frontend-auto-translate.php
```diff
@@ -329,14 +300,18 @@ protected function translate_terms_list( $query_var, $taxonomy ) {
 			$slugs = explode( $sep, $query_var );
 		}
 
-		foreach ( $slugs as &$slug ) {
-			if ( ! is_string( $slug ) ) {
-				// We got an unexpected query var, let return it unchanged.
-				return $query_var;
+		if ( ! empty( $this->curlang ) ) {
+			foreach ( $slugs as &$slug ) {
```
Hug0-Drelon: Just a suggestion, but I think references in loops are hard to read and make the code less expressive 😅

Hug0-Drelon (changes requested):

src/translated-term.php
Chouby: In theory you are right. But if we want to increase the safety of the method, we would then also have to check that the `$value` type is in agreement with the `$field`. Too much bloat for low added value (devs must read the doc).
NB: Together with Greg, we tried to document this properly with PHPStan, with no luck.

src/frontend/frontend-auto-translate.php
Chouby: I'm thinking the opposite. I find this elegant. But I'm not attached to my way.
See dd66fd6f8c18b6b12b05b456c4a8c7075a707478

Hug0-Drelon (approved):

PR #1892, opened 2026-05-22
Allow to filter untranslated posts in the posts list table

Chouby:
This PR adds a new filter on top of the posts list table allowing to filter the list for untranslated posts in a a given language.
In order to achieve that, a new query var `untranslated_in` is added to `WP_Query`. As we need a custom sql clause, this requires to remove the `suppress_filters` query var if present in the query (generally when using `get_posts()`).

### DIFF
--- a/src/crud-posts.php
+++ b/src/crud-posts.php
@@ -73,6 +73,11 @@ class PLL_CRUD_Posts {
 			add_action( 'delete_attachment', array( $this, 'delete_post' ) );
 			add_filter( 'wp_delete_file', array( $this, 'wp_delete_file' ) );
 		}
+
+		// Filter query.
+		add_filter( 'query_vars', array( $this, 'add_query_vars' ) );
+		add_action( 'parse_query', array( $this, 'parse_query' ) );
+		add_filter( 'posts_clauses', array( $this, 'posts_clauses' ), 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/crud-posts.php
```diff
@@ -442,4 +447,75 @@ private function translate_term( WP_Term $term, string $taxonomy, PLL_Language $
...
+			return $clauses;
+		}
+
+		$clauses['where'] .= $wpdb->prepare(
+			" AND {$wpdb->posts}.ID NOT IN (
+				SELECT {$wpdb->posts}.ID FROM {$wpdb->posts}
+					LEFT JOIN {$wpdb->term_relationships} AS pllutr ON {$wpdb->posts}.ID = pllutr.object_id
+					INNER JOIN {$wpdb->term_taxonomy} AS pllutt ON pllutt.term_taxonomy_id = pllutr.term_taxonomy_id
+					WHERE ( pllutr.term_taxonomy_id IN (%d) )
+					OR ( pllutt.taxonomy = 'post_translations' AND pllutt.description LIKE %s )
```
Screenfeed: Now imagine my language's slug is `syn`, or `fr` (while another language is `fra`) 🥸
Example data:
```
a:4:{s:4:"sync";a:2:{s:2:"es";s:3:"fra";s:3:"fra";s:3:"fra";}s:3:"fra";i:52;s:2:"es";i:53;s:2:"en";i:55;}
```
Just for clarity, this is the serialized value of:
```
array(
  'sync' => array(
    'es'  => 'fra',
    'fra' => 'fra',
  ),
  'fra' => 52,
  'es'  => 53,
  'en'  => 55,
)
```

Moreover, the description can contain anything added by a 3rd party plugin, or by ourselves in a future release.
Right now I don't have a solution for you. This type of query with `description LIKE %s` seems doomed.

src/crud-posts.php
```diff
@@ -442,4 +447,75 @@ private function translate_term( WP_Term $term, string $taxonomy, PLL_Language $
...
+
+		$untranslated_in = PLL()->model->languages->get( $query->query['untranslated_in'] );
+
+		if ( empty( $untranslated_in ) ) {
+			return $clauses;
+		}
+
+		$clauses['where'] .= $wpdb->prepare(
+			" AND {$wpdb->posts}.ID NOT IN (
+				SELECT {$wpdb->posts}.ID FROM {$wpdb->posts}
```
Screenfeed: I don't know about the performance of this if the sub-query returns a lot of post IDs 🤔

Screenfeed (changes requested):

src/crud-posts.php
Chouby: Well I just copied [some existing code](https://github.com/polylang/polylang/blob/2f44ec86615466aa28aa3448d24f28e09570b31a/src/translated-post.php#L434-L438) (which I plan to rewrite once this query filter is merged), without additional thinking. And I see now that I also made other mistakes (term_id instead of term_taxonomy_id).
What do you think of:
```php
		$tt_id = $untranslated_in->get_tax_prop( 'language', 'term_taxonomy_id' );

		$clauses['where'] .= $wpdb->prepare(
			" AND {$wpdb->posts}.ID NOT IN (
				SELECT pllutr.object_id
				FROM {$wpdb->term_relationships} AS pllutr
				WHERE pllutr.term_taxonomy_id = %d
			)",
			$tt_id
		);

		$clauses['where'] .= $wpdb->prepare(
			" AND {$wpdb->posts}.ID NOT IN (
				SELECT pllutr1.object_id
				FROM {$wpdb->term_relationships} AS pllutr1
				JOIN {$wpdb->term_taxonomy} AS pllutt ON pllutt.term_taxonomy_id = pllutr1.term_taxonomy_id
				JOIN {$wpdb->term_relationships} AS pllutr2 ON pllutr2.term_taxonomy_id = pllutt.term_taxonomy_id
				JOIN {$wpdb->term_relationships} AS pllutr3 ON pllutr3.object_id = pllutr2.object_id
				WHERE pllutt.taxonomy = 'post_translations'
				AND pllutr3.term_taxonomy_id = %d
			)",
			$tt_id
		);
```
Here I wrote two subqueries instead of the OR in the WHERE clause. Perhaps, this improves readability but I don't know about the performance. I will check that later

Screenfeed: Far better: it doesn't try to "guess" from the descriptions, and it seems to work 👍

Chouby:
I have made several performance tests on a set of 12000 posts (5000 translated, 1000 untranslated in each language).
The current version takes 0.021 s - 0.027s to return 100 posts (to compare to 0.004s for an unfiltered query).

I have tested an alternative with `UNION ALL` between the two subqueries. Results are about 0.06s.

For the record, a unique subquery with `WHERE pllutr1.term_taxonomy_id = %d OR (pllutt.taxonomy = 'post_translations' AND pllutr3.term_taxonomy_id = %d)` takes approximately 18s!

So the current version seems the most acceptable.

Screenfeed (approved):

PR #1894, opened 2026-05-22
[Language switcher rework] Classic widget

Screenfeed:
New possibilities for the classic widget.

- Move our two widgets to a dedicated folder.
- Replace `PLL_Widget_Languages` by `Widgets\Languages`.
- Namespace the calendar widget.
- `widgets.js`: remove jQuery dependency by using vanilla JS in the whole file.
- `widgets.js`: fix the flag that should appear in the widget header when the widget is displayed only for 1 language. Note: this still doesn't work in the customizer because WP inserts the form on-the-fly.
- Enqueue styles like it is done in `WP_Widget_Recent_Comments::__construct()`.
- `Fields_Interface` is replaced by `Abstract_Fields`.

## Widgets page with Classic Widgets

image
image

## Widgets page without Classic Widgets

These are, of course, what WP calls "legacy widgets".

image

## Customizer

image


src/Switcher/Fields/Widget.php
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @phpstan-import-type FieldsData from Fields_Interface
+ */
+class Widget implements Fields_Interface {
+	/**
+	 * Returns setting field data available for the language switcher.
+	 *
+	 * @since 3.9
+	 *
+	 * @return array[]
```
Marianne380, line 23: Could it be `@return array<string, array>`or `FieldsData`below is enough ?

js/src/widgets.js
```diff
@@ -1,151 +1,230 @@
 /**
- * Adds a flag to the widgets filtered by a language.
- *
- * @package Polylang
+ * Handles the options in the language switcher widget.
```
Marianne380, line 2: Do we still need jQuery loaded on the widgets and customizer screens (in `admin-base.php`) ?

css/src/admin.css
```diff
@@ -372,6 +372,24 @@ td.column-default_lang .icon-default-lang:before,
 	margin-right: 10px;
 }
 
+/* Widgets, nav menus */
+.polylang-language-switcher-widget-content {
+	width: 100%;
+	border-collapse: collapse;
+	margin-bottom: 20px;
+}
+.polylang-language-switcher-widget-content th {
```
Marianne380, line 381: Should we add `rtl`?

src/Switcher/Assets.php
```diff
@@ -15,6 +17,51 @@
...
+	 */
+	public static function maybe_enqueue_admin_styles(): void {
+		$screen = get_current_screen();
+
+		if ( empty( $screen ) ) {
+			return;
+		}
+
+		if ( 'widgets' === $screen->base || 'customize' === $screen->base ) {
+			wp_enqueue_style( 'polylang_admin' );
```
Marianne380: Isn't `polylang_admin` already enqueued on all admin pages via `polylang_dialog's` dependency in `admin-base.php`?

If so, `Assets::maybe_enqueue_admin_styles()` might be redundant, or is it meant as a safety net for contexts where `admin-base.php `doesn't run?

js/src/widgets.js
```diff
@@ -1,151 +1,230 @@
...
-			widgets_container = $( '.edit-widgets-main-block-list' );
-			widgets_selector  = '.widget';
+			const key = event.target.getAttribute( 'data-key' );
+
+			// Show/Hide rows.
+			wrapper
+				.querySelectorAll(
+					`:scope [class*="pll-hidden-if-${ key }-"]:not(.pll-hidden-if-${ key }-${ value })` // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword, Generic.ControlStructures.InlineControlStructure.NotAllowed, PHPCS detects `-if-` like a `if(`.
+				)
+				.forEach( ( input ) => {
```
Marianne380:
```suggestion
				.forEach( ( row ) => {
```
`row` or `element` seems more appropriate, same below.

src/Widgets/Languages.php
```diff
@@ -0,0 +1,287 @@
+<?php
```
Marianne380, line 1: `widget-calendar.php`now lives in `src/Widgets`without a namespace, next to `Languages.php`which is namespaced, should we align it ?

js/src/widgets.js
```diff
@@ -1,151 +1,230 @@
...
+					}
+				} );
+
+				if ( 'customize-theme-controls' !== wrapper.id ) {
+					// No flags for the customizer.
+					if (
+						wrapper.classList.contains( 'blocks-widgets-container' )
+					) {
+						// Without Classic Widgets: wait a second that everything is inserted into the page by WP.
+						setTimeout( () => {
```
Marianne380, line 79: The 1 second `setTimeout`  feels fragile, is there a more reliable way to know when it's ready here ?

src/Switcher/Assets.php
```diff
@@ -39,4 +86,50 @@ public static function enqueue_frontend_scripts(): void {
...
+		$active_widgets_by_sidebars = array_intersect_key( wp_get_sidebars_widgets(), $wp_registered_sidebars );
+
+		if ( empty( $active_widgets_by_sidebars ) ) {
+			// No widgets in the registered sidebars.
+			return false;
+		}
+
+		// phpcs:ignore Squiz.PHP.CommentedOutCode.Found
+		// Array( 'search-2', 'polylang-3', 'polylang-2', 'polylang-5' ).
+		$active_widgets = array_merge( ...array_values( $active_widgets_by_sidebars ) );
```
Marianne380: Should we filter for arrays before the spread?

src/Widgets/Languages.php
```diff
@@ -0,0 +1,287 @@
...
+	 * Displays the widget form.
+	 *
+	 * @since 3.9
+	 *
+	 * @param array $instance Current settings.
+	 * @return void
+	 *
+	 * @phpstan-param NewInstance|OldInstance $instance
+	 */
+	public function form( $instance ): void {
```
Marianne380, line 163: What do you think of : 
```php
// Inside form()
$html  = '<table role="presentation" class="polylang-language-switcher-widget-content"><tbody>';

$html .= $this->get_select( 'layout', $labels_and_data, $validated );
$html .= $this->get_select( 'alignment', $labels_and_data, $validated );
$html .= $this->get_checkbox( 'show_flags', $labels_and_data, $validated );
// ...

$html .= '</tbody></table>';

echo $html;

// The field method
private function get_select( string $key, array $labels_and_data, array $values ): string {
	$out  = $this->get_wrapper_start( $key, $labels_and_data, $values );
	// ... build the select tags ...
	$out .= '</select></td>';
	$out .= $this->get_wrapper_end();

	return $out;
}
```

Marianne380 (changes requested):
Are there PHPUnit or E2E tests planned for the classic widget form?

src/Switcher/Fields/Widget.php
Screenfeed: `array<string, array>` is not a valid DocBlock notation (only `array` and `array[]` are) so we have to rely on `FieldsData` for a more precise type.

src/Widgets/Languages.php
Screenfeed: That's something I thought about: namespace it and alias it to the old class name, but I don't think we already did this somewhere.

js/src/widgets.js
Screenfeed: I agree but I don't know any way to detect when the block editor has finish to load all its things, do you?
(The script is now loaded in the footer so it should help.)

src/Widgets/Languages.php
Screenfeed: Printing things directly avoids using:
```php
// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

What is the benefit to use getters instead of "printers"?

js/src/widgets.js
Screenfeed: No! (I forgot)
Dependency removed in [becc5ae](https://github.com/polylang/polylang/pull/1894/commits/becc5ae9f25e97167ad09383d72b034c5879d77f). The script is also loaded in the footer now, instead of the header.

css/src/admin.css
Screenfeed: Good catch. Done in [37b69eb](https://github.com/polylang/polylang/pull/1894/commits/37b69ebf9615674f6304447576f1c0d68dbfb185).

src/Switcher/Assets.php
Screenfeed: It is enqueued by `polylang_dialog` indeed. I did that as a safety net: some day we will change this method in `admin-base.php` and I didn't feel comfortable relying on `polylang_dialog` (someday it may not be enqueued on the pages where we need `polylang_admin`). But I think you're right, even if we change things, we'll still enqueue `polylang_admin` everywhere.
Removed in 927e6e5.

js/src/widgets.js
Screenfeed: You're right. Renamed in [557cb7d](https://github.com/polylang/polylang/pull/1894/commits/557cb7dce851cc2c2b23eca996c73abec49305d9).

src/Switcher/Assets.php
Screenfeed: Why not. Done in [618936a](https://github.com/polylang/polylang/pull/1894/commits/618936aaa421be239359c7f0c8680648707882b9).

js/src/widgets.js
Marianne380: @Hug0-Drelon, do you have any ideas ?

src/Switcher/Assets.php
```diff
@@ -39,4 +65,51 @@ public static function enqueue_frontend_scripts(): void {
...
+
+		foreach ( $active_widgets as $widget ) {
+			if ( is_string( $widget ) && preg_match( "/^{$pll_widget_id}-\d+$/", $widget ) ) {
+				return true;
+			}
+		}
```
Chouby: More readable (in my opinion) and without global var:
```suggestion
		foreach ( wp_get_sidebars_widgets() as $sidebar => $widgets ) {
			if ( 'wp_inactive_widgets' === $sidebar || empty( $widgets ) ) {
				continue;
			}

			foreach ( $widgets as $widget ) {
			if ( is_string( $widget ) && preg_match( "/^polylang-\d+$/", $widget ) ) {
				return true;
			}
		}
```

src/Switcher/Assets.php
```diff
@@ -39,4 +65,51 @@ public static function enqueue_frontend_scripts(): void {
...
+	 */
+	private static function has_classic_widget(): bool {
+		global $wp_widget_factory, $wp_registered_sidebars;
+
+		if ( ! $wp_widget_factory instanceof WP_Widget_Factory || ! is_array( $wp_registered_sidebars ) ) {
+			// If this happens, the site owner has bigger problems.
+			return false;
+		}
+
+		$pll_widget_id = 'polylang';
```
Chouby: Is it necessary? The intermediate variable reduces the readability in my opinion.
```suggestion
```

src/Widgets/Languages.php
Marianne380: At first, it seemed weird to me to have a `get_wrapper_start` but no `get_wrapper_end`.

I do not see any particular benefits.

Screenfeed:
```php
private function print_wrapper_end(): void {
	echo '</tr>;
}
```

😏

Screenfeed: Would you prefer a method that returns the wrapper's CSS classes?

Screenfeed: Like [fb5a135](https://github.com/polylang/polylang/pull/1894/commits/fb5a1350170c833a750345b00a1fc836d05b37e3).

Screenfeed:
> Are there PHPUnit or E2E tests planned for the classic widget form?

Not right now because we're in a hurry but we should add some indeed.

On another subject, I applied what I suggested about the deprecation notices (legacy settings) and filters (settings). I added a 2nd parameter to `Settings`' constructor. Maybe there is a better way to do this.

Chouby (changes requested):

src/Switcher/Assets.php
Screenfeed: ~~The global vars must be used, otherwise we don't know if the widgets are used.~~
~~Example:~~
~~1. I enable Twenty Twenty-One.~~
~~2. I setup (classic) widgets.~~
~~3. `wp_get_sidebars_widgets()` will return the list of the said widgets.~~
~~4. I switch to Twenty Twenty-Five.~~
~~5. This block theme doesn't support widgets.~~
~~6. `wp_get_sidebars_widgets()` will still return the same list of widgets while they are not used.~~

~~However you have a point with `wp_inactive_widgets`, I forgot to exclude it.~~

So, I got all wrong about `$wp_registered_sidebars`: this global var still contains things when the theme doesn't support widgets, so checking it is useless.

I'll use your code and `get_theme_support( 'widgets' )`.

src/Widgets/Languages.php
Screenfeed: After discussing with Fred, namespaced in [4fea6a9](https://github.com/polylang/polylang/pull/1894/commits/4fea6a9c8b521e4293b207523ee8009ab3a44f4f) + [9c1e9c4](https://github.com/polylang/polylang/pull/1894/commits/9c1e9c476757258d20dd05667ea8af26be1c6a9b) + [19f37ef](https://github.com/polylang/polylang/pull/1894/commits/19f37ef9d7633a93dffbdcf78caf74641052ff30).

src/Switcher/Assets.php
Screenfeed: Removed in [23a1323](https://github.com/polylang/polylang/pull/1894/commits/23a1323d88e03f42d4caa74b0156f3e409ccafaf).

src/Switcher/Assets.php
Screenfeed: Done in [ddeb82b](https://github.com/polylang/polylang/pull/1894/commits/ddeb82b5301a85d412031a84ba4f46ea55d198b9).

src/Widgets/Languages.php
Marianne380: It doesn't really matter to me, it was really the fact that there was a `start` method but no `end` method that struck me the first time I read it.

src/Switcher/Settings/Abstract_Settings_Legacy.php
```diff
@@ -124,23 +132,25 @@ protected function is_legacy( array $settings ): bool {
...
+		}
 
 		if ( ! isset( $settings['show_wrapper'] ) ) {
 			// `PLL_Walker_Dropdown` displays the wrapper (`<select>`) while `PLL_Walker_List` didn't.
 			$settings['show_wrapper'] = ! empty( $settings['dropdown'] );
 		}
 
 		if ( isset( $settings['layout'], $settings['dropdown'] ) ) {
-			// Set a new value to `layout` only if the value of `layout` and `dropdown` don't match.
+			// Set a new value to `layout` only if the value of `layout` and `dropdown` don't match, so we don't loose `layout`'s real value.
```
Marianne380:
```suggestion
			// Set a new value to `layout` only if the value of `layout` and `dropdown` don't match, so we don't lose `layout`'s real value.
```

Marianne380 (approved):

src/Switcher/Settings/Abstract_Settings_Legacy.php
Screenfeed: I often make this one. Thanks!

js/src/widgets.js
```diff
@@ -1,151 +1,230 @@
...
 			}
-		);
 
-		function pll_toggle( a, test ) {
-			test ? a.show() : a.hide();
-		}
+			const langSlug = event.target.value;
+			const icon =
+				langSlug &&
+				pllWidget.displayFlags.flags.hasOwnProperty( langSlug )
```
Hug0-Drelon, line 197: `pllWidget.displayFlags` object type is not checked here.

js/src/widgets.js
```diff
@@ -1,151 +1,230 @@
...
+		}
+	},
+
+	/**
+	 * Called when the DOM is ready. Attaches the events to the wrapper.
+	 */
+	ready: () => {
+		if ( window.pll_widgets?.flags ) {
+			pllWidget.displayFlags.flags = window.pll_widgets.flags;
+		}
```
Hug0-Drelon: I suggest to return early here if `window.pll_widgets?.flags` is falsy.

src/Switcher/Fields/Abstract_Fields.php
```diff
@@ -32,15 +33,23 @@ interface Fields_Interface {
...
-	public static function validate( array $settings ): array;
+	public static function filter( Settings $settings ): array {
+		$validated = array();
+
+		foreach ( static::get() as $name => $data ) {
+			$validated[ $name ] = $settings->$name;
+		}
+
+		return $validated;
+	}
```
Hug0-Drelon, line 80: Now you have extra arguments in `Settings`, I wonder if this method is still useful.
Correct me if I'm wrong but you wrote this to bypass depreciation notices and hooks in the first place?

src/Switcher/Settings/Settings.php
```diff
@@ -147,18 +147,27 @@ class Settings extends Abstract_Settings_Legacy {
...
-		 * Filter the language switcher settings.
-		 *
-		 * @since 3.9
-		 *
-		 * @param array $settings Settings.
-		 */
-		$settings = apply_filters( 'pll_language_switcher_settings', $settings );
+	public function __construct( array $settings, array $args = array() ) {
+		$this->args = array_merge( $this->args, $args );
+		$settings   = $this->maybe_filter_legacy( $settings );
```
Hug0-Drelon: Why don't you honor `filter_settings` argument here?

src/Switcher/Assets.php
```diff
@@ -15,6 +17,30 @@
...
+
+	/**
+	 * Maybe enqueues CSS in frontend.
+	 *
+	 * @since 3.9
+	 *
+	 * @return void
+	 */
+	public static function maybe_enqueue_frontend_styles(): void {
+		if ( self::has_classic_widget() ) {
```
Hug0-Drelon: Since you used public static method, I think this condition should be found in a widget specific class. Here it prevents to enqueue the script for other kinds of switchers.

src/Switcher/Assets.php
```diff
@@ -39,4 +65,40 @@ public static function enqueue_frontend_scripts(): void {
...
+			if ( 'wp_inactive_widgets' === $sidebar || empty( $widgets ) ) {
+				continue;
+			}
+
+			foreach ( $widgets as $widget ) {
+				if ( is_string( $widget ) && preg_match( '/^polylang-\d+$/', $widget ) ) {
+					return true;
+				}
+			}
+		}
```
Hug0-Drelon: Having two `foreach` like that is O(n^2) complex, moreover you added a condition at line ~91 that runs for every items (although it'd match a single key that could be extracted before traversing the array).

What about serializing `wp_get_sidebars_widgets()` returned value and passing it to a Regex?

src/Widgets/Languages.php
```diff
@@ -0,0 +1,273 @@
...
+	 *     before_sidebar: string,
+	 *     after_sidebar: string,
+	 *     show_in_rest: boolean,
+	 *     widget_id: string,
+	 *     widget_name: string
+	 * } $args
+	 * @phpstan-param NewInstance|OldInstance $instance
+	 */
+	public function widget( $args, $instance ): void {
+		if ( empty( PLL()->links ) ) {
```
Hug0-Drelon: What about injecting the instance in the constructor?

src/Widgets/Languages.php
```diff
@@ -0,0 +1,273 @@
...
+			return;
+		}
+
+		$instance['unique_id']    = "pll-switcher-widget-{$this->number}";
+		$instance['show_wrapper'] = true;
+
+		if ( isset( $instance['layout'] ) ) {
+			// For backward compatibility, some legacy options are saved along the new ones (see the end of `Languages::update()`).
+			unset( $instance['dropdown'], $instance['show_names'] );
+		}
```
Hug0-Drelon: Should we use `Fields` class instead?

Hug0-Drelon (changes requested):

js/src/widgets.js
Screenfeed: Not here, because it will stop everything, even things that are not related to flags. It is done [here](https://github.com/polylang/polylang/pull/1894/changes#diff-9f39fe0cdd5fd597743a381ee4ba832ac93085369954ef415089d6209af2dcf0R44).

src/Switcher/Fields/Abstract_Fields.php
Screenfeed: All the validation is done in `Settings` indeed. The aim is to limit the returned array to a subset of all the settings. This will be used before storing the settings in the DB for example, but we don't need to store everything, we only need the settings that appear in `static::get()`.
This could also be done another way:

```php
return array_intersect_key( get_object_vars( $settings ), static::get() );
```

Do you prefer this? I think it makes the intent more clear.

src/Switcher/Settings/Settings.php
Screenfeed: Because `maybe_filter_legacy()` doesn't only filter, it also converts legacy settings to the new ones. It is honored inside `maybe_filter_legacy()`.

src/Switcher/Assets.php
Screenfeed: It doesn't prevent anything: it enqueues the styles for one case (classic widgets), but `enqueue_frontend_styles()` is public and can be used any time.
Unless I misunderstood your suggestion? 🤔

Hug0-Drelon: The main issue for me is related to the structure. I would expect to find widget related logic in widget classe(s).

src/Widgets/Languages.php
Screenfeed: I like the idea. Instead of:
```php
register_widget( Widgets\Languages::class );
```

we would do:
```php
register_widget( new Widgets\Languages( PLL()->links ) );
```

But what if `PLL()->links` is empty at that point? :(
I don't think we're going to solve anything.

Side note: we register the widget in `PLL_Base` but the the property `$links` is declared in `PLL_Frontend` and `PLL_Admin_Base`, so we can't use `$this->links`.

js/src/widgets.js
Screenfeed: I don't think it's necessary, we know it's either an empty object or `window.pll_widgets.flags`, which is an object: https://github.com/polylang/polylang/pull/1894/changes#diff-9f39fe0cdd5fd597743a381ee4ba832ac93085369954ef415089d6209af2dcf0R21 + https://github.com/polylang/polylang/pull/1894/changes#diff-9f39fe0cdd5fd597743a381ee4ba832ac93085369954ef415089d6209af2dcf0R40.

However, I improved things a bit [55ae3f8](https://github.com/polylang/polylang/pull/1894/commits/55ae3f8a93d3521227209d145b6343d4be06d67d).

js/src/widgets.js
Screenfeed: Improved a bit in [55ae3f8](https://github.com/polylang/polylang/pull/1894/commits/55ae3f8a93d3521227209d145b6343d4be06d67d)

src/Switcher/Assets.php
Screenfeed: Why not. I used JSON but the idea is the same: [e0327df](https://github.com/polylang/polylang/pull/1894/commits/e0327df8b7eac775577191c11c1e6a36ffff194f).

src/Widgets/Languages.php
Screenfeed: Done in [59a9baa](https://github.com/polylang/polylang/pull/1894/commits/59a9baa18391367a9557dd539f7636ca54c7376b).

src/Widgets/Languages.php
Screenfeed: Done in [f1a6413](https://github.com/polylang/polylang/pull/1894/commits/f1a6413a65714f873c899e53410eb362a57d2b4f).

src/Switcher/Assets.php
Screenfeed: Done in [92336e9](https://github.com/polylang/polylang/pull/1894/commits/92336e9e6b160189aa3ffe26d0b3098b7ee73efc).

src/Switcher/Assets.php
```diff
@@ -5,6 +5,8 @@
 
 namespace WP_Syntex\Polylang\Switcher;
 
+use WP_Widget_Factory;
+
```
Hug0-Drelon:
```suggestion
```

src/Switcher/Fields/Widget.php
```diff
@@ -0,0 +1,110 @@
...
+	 *
+	 * @param Settings $settings Switcher settings.
+	 * @return array
+	 */
+	public static function to_db( Settings $settings ): array {
+		$validated = static::filter( $settings );
+
+		// Keep some legacy keys in database for backward compatibility.
+		$validated['dropdown']   = 'select' === $validated['layout'] ? 1 : 0;
+		$validated['show_names'] = ! empty( $validated['show_labels'] ) ? 1 : 0;
```
Hug0-Drelon: Should it be kept only for old installations?

Hug0-Drelon (changes requested):

src/Switcher/Fields/Widget.php
Screenfeed: @Hug0-Drelon The answer is not as simple as I first thought.

We store these legacy keys in case the user has an annoying bug (not necessarily with the widgets) and rolls back the plugin. So we could say that this can happen with old and new installations.
On the other hand, keeping these is not the perfect solution, because if the user chose the new `dropdown` layout, it will be stored as `$validated['dropdown'] = 0;`, which translates to a `vertical` layout in the legacy format, because we had only 2 possible layouts instead of 4 now.

Then if we look at the menu, the old format has 2 layouts, but the new one too. This means the menu should stay roughly the same after the rollback.

So, IMHO:
1. Those legacy keys can be useful for old and new installations.
2. But it won't really work if the user has selected a layout `horizontal` or `dropdown` for the widget because those didn't exist in the old format.
3. `show_names` is OK, unless the user selected `Language codes`, which will translate to `Language names` in the old format.

Alternative: store those legacy keys only if the new layout matches the old one.

src/Switcher/Assets.php
Screenfeed: I found a better way with [9d696fd](https://github.com/polylang/polylang/pull/1894/commits/9d696fd2b5164570113728acb5d91c6893960568).

src/Switcher/Fields/Widget.php
Hug0-Drelon: I'm not sure this is a requirement for new installation. But let's see what @Chouby prefers.

Hug0-Drelon (approved):

src/Switcher/Settings/Abstract_Settings_Legacy.php
```diff
@@ -49,6 +49,14 @@ abstract class Abstract_Settings_Legacy {
 		'admin_current_lang'     => null, // Use the global current language.
 	);
 
+	/**
+	 * @var bool[]
+	 */
+	protected array $args = array(
+		'filter_settings'    => true,
+		'deprecate_settings' => false,
+	);
```
Chouby: - I don't understand the need for the `filter_settings` arg. Why would we want to bypass the new `pll_language_switcher_settings` filter?
- I would like to avoid the `deprecate_settings` arg too. My proposal is to move the call to `_deprecated_argument()` to `pll_the_languages()`. I foresee that this would require to call `Settings::is_legacy()` from this function.

Let's discuss that in our next meeting.

src/Widgets/Languages.php
```diff
@@ -0,0 +1,277 @@
...
+	private ?PLL_Links $links;
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 3.9
+	 *
+	 * @param PLL_Base $polylang The Polylang object.
+	 */
+	public function __construct( PLL_Base &$polylang ) {
```
Chouby, line 59: If necessary, I would pass only `PLL_Links`.

src/Switcher/Fields/Abstract_Fields.php
```diff
@@ -0,0 +1,82 @@
...
+	 * @param array $raw_settings Raw settings.
+	 * @return array
+	 */
+	public static function from_db( array $raw_settings ): array {
+		if ( isset( $raw_settings['layout'] ) ) {
+			// For backward compatibility, some legacy options are saved along the new ones.
+			unset( $raw_settings['dropdown'], $raw_settings['show_names'] );
+		}
+		return $raw_settings;
+	}
```
Chouby: I'm not convinced by this method (the same for `to_db()`. 
- This splits handling of legacy settings in too many places. 
- I'm not sure we should handle this edge case.
 
Let's discuss that if our next meeting.

Chouby (changes requested):

src/Switcher/Fields/Abstract_Fields.php
Screenfeed: This is not an edge case, this will happen to all our users once they have saved the settings of the widget and menu. We save `dropdown` and `show_names` along the new settings in case someone needs to roll back our plugin, so the widget and menu will still work. Then we unset them here after retrieving the settings from the DB, to prevent the switcher to detect them as legacy settings, then convert them to new settings, overwriting the new settings that were previously saved into the DB.

src/Switcher/Settings/Abstract_Settings_Legacy.php
Screenfeed: > * I don't understand the need for the `filter_settings` arg. Why would we want to bypass the new `pll_language_switcher_settings` filter?

To prevent them from being filtered in the admin (widget and menu form).

> * I would like to avoid the `deprecate_settings` arg too. My proposal is to move the call to `_deprecated_argument()` to `pll_the_languages()`. I foresee that this would require to call `Settings::is_legacy()` from this function.

~In that case, the conversion from legacy settings to the new ones must be done directly in `pll_the_languages()`. This implies the class `Switcher` to not accept legacy settings at all.~
We fell into that trap again: this is not possible, classes extending `Settings` must have access to this mechanism.

src/Widgets/Languages.php
Screenfeed: This class is instantiated in `PLL_Base`: the property `$links` doesn't exist in `PLL_Base`.

src/Switcher/Settings/Abstract_Settings_Legacy.php
Screenfeed: Other solutions implying that we separate the legacy class and the settings class:

1. Create an upgrade script that migrates legacy settings to the new ones for the classic widget and classic menu. Then we don't have legacy settings in them anymore. This implies the blocks can manage this by themselves.
2. Let the deprecation notices trigger in the widget and menu.
3. Create 2 classes that extend the legacy class, but they don't trigger the deprecation notices: one for the widget, one for the menu.

src/Switcher/Fields/Abstract_Fields.php
Screenfeed: Comments improved in [5254b20](https://github.com/polylang/polylang/pull/1894/commits/5254b20a725579595d4dc347ffea433fe5ff9660).

src/Switcher/Settings/Abstract_Settings_Legacy.php
Screenfeed: Done in [5254b20](https://github.com/polylang/polylang/pull/1894/commits/5254b20a725579595d4dc347ffea433fe5ff9660).

In `pll_the_languages()` we'll add:

```php
	if ( Settings::is_legacy( $args ) ) {
		_deprecated_argument(
			'pll_the_languages()',
			'3.9',
			sprintf(
				/* translators: %s is a function name. */
				esc_html__( "See %s's documentation.", 'polylang' ),
				'pll_the_languages()'
			)
		);
	}
```

Alternative, so everything is kept in the class:

```php
	Settings::maybe_deprecate_arguments( $args );
```

src/Switcher/Fields/Abstract_Fields.php
```diff
@@ -0,0 +1,82 @@
...
+	 * @return array
+	 */
+	public static function filter( Settings $settings ): array {
+		$validated = array();
+
+		foreach ( static::get() as $name => $data ) {
+			$validated[ $name ] = $settings->$name;
+		}
+
+		return $validated;
```
Chouby: Would this work?
```suggestion
		return array_intersect_key( get_object_vars( $settings ), static::get() );
```

src/Switcher/Fields/Abstract_Fields.php
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * Removes the legacy keys that were stored in the database for backward compatibility.
+	 *
+	 * @since 3.9
+	 *
+	 * @param array $raw_settings Raw settings.
+	 * @return array
+	 */
+	public static function from_db( array $raw_settings ): array {
```
Chouby: I suggest to rename the method to clarify its objective:
```suggestion
	 * Removes the legacy keys that are stored in the database.
	 *
	 * This method is part of the plan to support rollbacks to versions < 3.9.
	 * Backward compatibility with Polylang < 3.9.
	 *
	 * @since 3.9
	 *
	 * @param array $raw_settings Raw settings.
	 * @return array
	 */
	public static function remove_legacy_settings( array $raw_settings ): array {
```

src/Switcher/Fields/Widget.php
```diff
@@ -0,0 +1,110 @@
...
+	 * backward compatibility.
+	 *
+	 * @since 3.9
+	 *
+	 * @param Settings $settings Switcher settings.
+	 * @return array
+	 */
+	public static function to_db( Settings $settings ): array {
+		$validated = static::filter( $settings );
+
```
Chouby: I propose to rename this method and alter its behavior to handle only the legacy settings:
```suggestion
	 * Add legacy settings to the data saved in DB.
	 *
	 * This would be useful in case a user rollbacks to a version < 3.9.
	 * Backward compatibility with Polylang < 3.9.
	 *
	 * @since 3.9
	 *
	 * @param Settings $settings Switcher settings.
	 * @return array
	 */
	public static function add_legacy_settings( Settings $settings ): array {
```
Then in the calling method:
```php
$validated = Fields::filter( new Settings( $new_instance, array( 'filter_settings' => false ) ) );
$validated = Fields::add_legacy_settings( $settings );
```
That way, we will simply have to remove the legacy methods when we'll remove the support to rollback to versions < 3.9.

Chouby (changes requested):

src/Switcher/Fields/Abstract_Fields.php
Screenfeed: Good idea. Done in [4adfadf](https://github.com/polylang/polylang/pull/1894/commits/4adfadf8fde6aa26ca08ef5da1348dcda4bd99e2).

src/Switcher/Fields/Abstract_Fields.php
Screenfeed: Done in [6161a76](https://github.com/polylang/polylang/pull/1894/commits/6161a762e11637eaac18820a4221960a2dcc9b25).

src/Switcher/Fields/Widget.php
Screenfeed: Done in [9fbbaaa](https://github.com/polylang/polylang/pull/1894/commits/9fbbaaaec3a4b86048164fbc3566fe760c97fa38) and [57c9ab5](https://github.com/polylang/polylang/pull/1894/commits/57c9ab59a34b3b8bf59556b1b0d68fc81d170547).

PR #1895, opened 2026-05-26
Fix PHPstan errors in REST controllers

Chouby:
Fix errors appeared with WordPress stubs 7.0
See also https://github.com/polylang/polylang-pro/pull/3014

### DIFF
--- a/src/modules/REST/V1/Languages.php
+++ b/src/modules/REST/V1/Languages.php
@@ -36,6 +36,22 @@ class Languages extends Abstract_Controller {
 	private $translatable_objects;
 
 	/**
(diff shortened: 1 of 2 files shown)


src/modules/REST/V1/Languages.php
```diff
@@ -58,6 +72,7 @@ public function __construct( PLL_Model $model ) {
 	 */
 	public function register_routes(): void {
 		register_rest_route(
+			/** @phpstan-var non-falsy-string */
```
Screenfeed: Is this necessary? 🤔

Screenfeed (changes requested):

Screenfeed (approved):

PR #1896, opened 2026-05-27
Fix fatal error when `$_REQUEST['lang']` is a locale in AJAX context.

Marianne380:
See https://secure.helpscout.net/conversation/3321243736/40542#thread-10167215994.

## What

Fix a fatal error triggered when an AJAX front request includes a locale-format `lang` parameter (e.g. `en_GB` instead of slug `en`).

## Why

`sanitize_key()` lowercases the value, turning `en_GB` into `en_gb`. `get_language('en_gb')` returns `false`, causing `set_language(false)` to fall through to `get_current_language()`, which calls `get_query_var()` on a `$wp_query` not yet initialized at `plugins_loaded` in AJAX context.

Observed with the Goldish theme, whose live search JS sends `lang=get_locale()` in its AJAX requests.

## Fix

* `Languages`: add an `is_locale()` static method and use it in `validate_lang()` to replace the inline `preg_match` call.
* `PLL_Choose_Lang::init()`: use `Languages::is_locale()` to detect locale-format values and safely preserve their case (handles `en_GB`), otherwise fallback to `sanitize_key()` as before. Also adds an `isset()` and `is_string()` guard to satisfy PHPStan.
* `PLL_Choose_Lang_Content::get_current_language()`: add a guard on `$wp_query` following the same pattern as WP core conditional tags (`is_404()` etc.), as a safety net for any code path calling this method before the query runs.
* `test-copied-functions.php`: remove `test_sanitize_locale_name()`. `sanitize_locale_name` is native since WP 6.2.1 and we require WP 6.5, so there is no need to monitor it anymore.

### DIFF
--- a/src/Model/Languages.php
+++ b/src/Model/Languages.php
@@ -936,8 +936,8 @@ class Languages {
 	 * @param string $value The value to check.
 	 * @return bool
 	 */
-	public static function is_locale( string $value ): bool {
-		return (bool) preg_match( '#' . self::LOCALE_PATTERN . '#', $value );
+	public static function is_locale( $value ): bool {
+		return is_string( $value ) && (bool) preg_match( '#' . self::LOCALE_PATTERN . '#', $value );
 	}
 
 	/**
--- a/src/Model/Languages.php
+++ b/src/Model/Languages.php
@@ -929,6 +929,18 @@ class Languages {
 	}
 
 	/**
+	 * Checks if a given value matches the locale format.
+	 *
+	 * @since 3.8.5
+	 *
+	 * @param string $value The value to check.
+	 * @return bool
+	 */
+	public static function is_locale( string $value ): bool {
+		return (bool) preg_match( '#' . self::LOCALE_PATTERN . '#', $value );
+	}
+
+	/**
 	 * Builds the language metas into an array and serializes it, to be stored in the term description.
 	 *
 	 * @since 3.4
@@ -1037,7 +1049,7 @@ class Languages {
 		$errors = new WP_Error();
 
 		// Validate locale with the same pattern as WP 4.3. See #28303.
-		if ( empty( $args['locale'] ) || ! preg_match( '#' . self::LOCALE_PATTERN . '#', $args['locale'], $matches ) ) {
+		if ( empty( $args['locale'] ) || ! self::is_locale( $args['locale'] ) ) {
 			$errors->add( 'pll_invalid_locale', __( 'Enter a valid WordPress locale', 'polylang' ) );
 		}
 
--- a/src/frontend/choose-lang-content.php
+++ b/src/frontend/choose-lang-content.php
@@ -51,7 +51,6 @@ class PLL_Choose_Lang_Content extends PLL_Choose_Lang {
 	protected function get_current_language() {
 		global $wp_query;
 		if ( ! isset( $wp_query ) ) {
-			// `$wp_query` is not yet initialized before the query runs (e.g. during `plugins_loaded` in AJAX context).
 			return false;
 		}
 
--- a/src/frontend/choose-lang-content.php
+++ b/src/frontend/choose-lang-content.php
@@ -49,6 +49,12 @@ class PLL_Choose_Lang_Content extends PLL_Choose_Lang {
 	 * @return PLL_Language|false
 	 */
 	protected function get_current_language() {
+		global $wp_query;
+		if ( ! isset( $wp_query ) ) {
+			// `$wp_query` is not yet initialized before the query runs (e.g. during `plugins_loaded` in AJAX context).
+			return false;
+		}
+
 		// No language set for 404.
 		if ( is_404() || ( is_attachment() && ! $this->options['media_support'] ) ) {
 			return false;
--- a/src/frontend/choose-lang.php
+++ b/src/frontend/choose-lang.php
@@ -3,6 +3,8 @@
  * @package Polylang
  */
 
+use WP_Syntex\Polylang\Model\Languages;
+
 /**
  * Base class to choose the language
  *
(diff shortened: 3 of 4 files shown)


src/frontend/choose-lang.php
```diff
@@ -61,7 +61,13 @@ public function __construct( PLL_Frontend &$polylang ) {
 	 */
 	public function init() {
 		if ( Polylang::is_ajax_on_front() || ! wp_using_themes() ) {
-			$this->set_language( empty( $_REQUEST['lang'] ) ? $this->get_preferred_language() : $this->model->get_language( sanitize_key( $_REQUEST['lang'] ) ) ); // phpcs:ignore WordPress.Security.NonceVerification
+			$lang = isset( $_REQUEST['lang'] ) && is_string( $_REQUEST['lang'] ) ? $_REQUEST['lang'] : ''; // phpcs:ignore WordPress.Security.NonceVerification, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			// Use `sanitize_locale_name` for locale-format values (e.g. `en_GB`) to preserve case, `sanitize_key` otherwise.
+			$this->set_language(
+				empty( $lang )
+					? $this->get_preferred_language()
+					: $this->model->get_language( str_contains( $lang, '_' ) ? sanitize_locale_name( $lang ) : sanitize_key( $lang ) )
```
Screenfeed: Instead of using `str_contains()` + `sanitize_locale_name()`, we could use `preg_match()` with `WP_Syntex\Polylang\Model\Languages::LOCALE_PATTERN` only (which is what `\WP_Syntex\Polylang\Model\Languages::validate_lang()` does).

Maybe we could create a method `Model\Languages::is_locale()` that does that, and use it here and in `Languages::validate_lang()`.

Marianne380: Good idea! See https://github.com/polylang/polylang/pull/1896/changes/242c954cb79ca5cfc99971652abcf02237432002.

Screenfeed (approved):

src/Model/Languages.php
```diff
@@ -928,6 +928,18 @@ public function filter( string $key ): Languages_Proxies {
...
+	/**
+	 * Checks if a given value matches the locale format.
+	 *
+	 * @since 3.8.5
+	 *
+	 * @param string $value The value to check.
+	 * @return bool
+	 */
+	public static function is_locale( string $value ): bool {
+		return (bool) preg_match( '#' . self::LOCALE_PATTERN . '#', $value );
```
Chouby: What do you think to make an exception and not to type the param.? This will allow not to repeat the `is_string()` check each time the method is called. 
```suggestion
	public static function is_locale( $value ): bool {
		return is_string( $value ) && (bool) preg_match( '#' . self::LOCALE_PATTERN . '#', $value );
```

src/frontend/choose-lang-content.php
```diff
@@ -49,6 +49,12 @@ protected function set_language( $curlang = false ): void {
 	 * @return PLL_Language|false
 	 */
 	protected function get_current_language() {
+		global $wp_query;
+		if ( ! isset( $wp_query ) ) {
+			// `$wp_query` is not yet initialized before the query runs (e.g. during `plugins_loaded` in AJAX context).
```
Chouby: The comment is not very clear but I don't think one is useful.
```suggestion
```

src/frontend/choose-lang.php
```diff
@@ -61,7 +63,13 @@ public function __construct( PLL_Frontend &$polylang ) {
...
 	public function init() {
 		if ( Polylang::is_ajax_on_front() || ! wp_using_themes() ) {
-			$this->set_language( empty( $_REQUEST['lang'] ) ? $this->get_preferred_language() : $this->model->get_language( sanitize_key( $_REQUEST['lang'] ) ) ); // phpcs:ignore WordPress.Security.NonceVerification
+			$lang = isset( $_REQUEST['lang'] ) && is_string( $_REQUEST['lang'] ) ? $_REQUEST['lang'] : ''; // phpcs:ignore WordPress.Security.NonceVerification, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+			// Use Languages locale pattern for locale-format values (e.g. `en_GB`), `sanitize_key` otherwise.
+			$this->set_language(
+				empty( $lang )
+					? $this->get_preferred_language()
+					: $this->model->get_language( Languages::is_locale( $lang ) ? $lang : sanitize_key( $lang ) )
+			);
```
Chouby: There is an issue (which was already present) if `get_language()` doesn't return a `PLL_Language`).
 
Moreover I propose to reorder the method to clarify the code:
```php
	public function init() {
		add_action( 'pre_comment_on_post', array( $this, 'pre_comment_on_post' ) ); // sets the language of comment.
		add_action( 'parse_query', array( $this, 'parse_main_query' ), 2 ); // sets the language in special cases.
		add_action( 'wp', array( $this, 'maybe_setcookie' ), 7 );

		if ( ! Polylang::is_ajax_on_front() && wp_using_themes() ) {
			return;
		}

		if ( isset( $_REQUEST['lang'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
			// Let's accept either WordPress locales or language codes.
			$lang = $this->model->languages->get( Languages::is_locale( $_REQUEST['lang'] ) ? $_REQUEST['lang'] : sanitize_key( $_REQUEST['lang'] ) ); // phpcs:ignore WordPress.Security.NonceVerification, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
		}

		$this->set_language( empty( $lang ) ? $this->get_preferred_language() : $lang );
	}
```

Alternatively, with no modification outside this method and using the sanitization function of WP:
```php
	public function init() {
		add_action( 'pre_comment_on_post', array( $this, 'pre_comment_on_post' ) ); // sets the language of comment.
		add_action( 'parse_query', array( $this, 'parse_main_query' ), 2 ); // sets the language in special cases.
		add_action( 'wp', array( $this, 'maybe_setcookie' ), 7 );

		if ( ! Polylang::is_ajax_on_front() && wp_using_themes() ) {
			return;
		}

		if ( isset( $_REQUEST['lang'] ) && is_string( $_REQUEST['lang'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
			// Let's accept language codes.
			$lang = $this->model->languages->get( sanitize_key( $_REQUEST['lang'] ) );

			// Or WordPress locales.
			if ( empty( $lang ) ) {
				$lang = $this->model->languages->get( sanitize_locale_name( $_REQUEST['lang'] ) ); // phpcs:ignore WordPress.Security.NonceVerification
			}
		}

		$this->set_language( empty( $lang ) ? $this->get_preferred_language() : $lang );
	}
```

Chouby (changes requested):

src/Model/Languages.php
Marianne380: Yes, see https://github.com/polylang/polylang/pull/1896/commits/3ed5fb4c5f87bad6b42f7d931635e1f99d265322.

src/frontend/choose-lang.php
Marianne380: Good idea, see https://github.com/polylang/polylang/pull/1896/commits/3ed5fb4c5f87bad6b42f7d931635e1f99d265322.

PR #1898, opened 2026-06-01
New language switcher block

Hug0-Drelon:
## What

Modernizes the **Language Switcher** block (`polylang/language-switcher`) with new layout and display options, a richer editor UI, and frontend rendering aligned with the new `Switcher` class.

- New block attributes: `layout` (vertical, horizontal, dropdown, select), `show_labels` (names, codes, none), and flag styling (`flag_aspect_ratio`, `flag_border_radius`, `flag_width`, `flag_margin_right`)
- Block supports for spacing, alignment, color, and typography
- New editor components: `RenderedSwitcher`, `SwitcherControls`, and `SwitcherLinkElement`
- Dedicated editor stylesheet (`language-switcher-editor-style.css`)
- PHP render callback refactored to use `Switcher` / `Settings`, with flag styling applied via `WP_HTML_Tag_Processor`

## Why

The block still relied on legacy attributes (`dropdown`, `show_names`) and a simplified editor preview that did not match the frontend. This aligns the block with the newer switcher API and gives users more control over layout and flag appearance directly in the editor.

## How

- **`block.json`**: Replaced boolean attributes with the new schema; registered frontend, editor, and view scripts/styles.
- **Editor (JS)**: Replaced the old UI components with toolbar/inspector controls and a disabled preview that mirrors frontend markup.
- **PHP (`Block.php`)**: Builds output through `Switcher`; post-processes flag HTML with CSS custom properties; registers the editor style on `init`.
- **CSS**: Extended `frontend-switcher.css` for new layouts and flag variables; added editor-only overrides.
- **Build**: Webpack bundles the new editor CSS; ESLint config updated for experimental Gutenberg APIs used in controls.

> [!NOTE]
> `polylang/navigation-language-switcher` is kept untouched and will be refactored later (using new controls from this PR).
> This is because the navigation block relies on different PHP rendering logic (dynamic core blocks wrapped) and to avoid complex review.

> [!WARNING]
> Tests a failing temporarily.
> Backward compatibility is not handled yet, I'd prefer to validate as it is and add `deprecated.js` once done.


css/src/blocks/language-switcher-editor-style.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+.pll-switcher.pll-layout-dropdown ul > li:first-of-type a {
+	display: inline-flex;
+	align-items: center;
+	flex-grow: 0;
+	gap: 0.25em;
+	padding: 0;
+	text-decoration: none;
+}
```
Screenfeed: Instead of relying on `:first-of-type` and `:not`, there is simpler ways (IMHO):


```suggestion
.pll-switcher.pll-layout-dropdown ul > li {
	display: inline-flex;
	align-items: center;
	flex-wrap: nowrap;
}

.pll-switcher.pll-layout-dropdown ul > li + li {
	display: none;
}

.pll-switcher.pll-layout-dropdown ul > li a {
	display: inline-flex;
	align-items: center;
	flex-grow: 0;
	gap: 0.25em;
	padding: 0;
	text-decoration: none;
}
```

If you still want to use `:first-of-type`, an alternative could be to replace:

```CSS
.pll-switcher.pll-layout-dropdown ul > li:not( :first-of-type ) {
```

By:

```CSS
.pll-switcher.pll-layout-dropdown ul > li:first-of-type ~ li {
```

All of these solutions, yours included, will give the same result. I think mine is a bit more efficient than using `:not( :first-of-type )` though, because it doesn't use pseudo-classes, only basic selectors, but I could be wrong.

css/src/frontend-switcher.css
```diff
@@ -118,10 +121,20 @@
 }
 
 /* Links */
-.pll-switcher a {
+.pll-switcher:not( .pll-layout-dropdown ) a {
```
Screenfeed: Why not for dropdowns? This is basically and horizontal layout with an extra top level link. If you're afraid to affect the top level link, the next block you added will overwrite this one.

```suggestion
.pll-switcher a {
```

js/src/blocks/language-switcher/components/rendered-switcher.js
```diff
@@ -0,0 +1,182 @@
...
+		flag_width,
+		flag_margin_right,
+	} = attributes;
+	const { languages } = useContext( LanguagesContext );
+	const currentLanguage = useCurrentLanguage( languages );
+	const curatedLanguages = useCuratedLanguages( languages, currentLanguage );
+
+	if ( layout === 'dropdown' ) {
+		return (
+			<nav className={ `pll-switcher pll-layout-${ layout }` }>
```
Screenfeed: This will behave like the php switcher and prevent to add the `data-aspect-ratio` attribute on each element:

```suggestion
			<nav className={ `pll-switcher pll-layout-${ layout } pll-aspect-ratio-{ flagAspectRatio.replace( ':', '' ) }` }>
```

This will also result in less CSS by not duplicating code. For example, this:

```css
.pll-aspect-ratio-32 .pll-switcher-flag {
	aspect-ratio: 3/2;
}

.pll-aspect-ratio-11 .pll-switcher-flag {
	aspect-ratio: 1/1;
}
```

Would replace:

```css
.pll-aspect-ratio-11 .pll-switcher-flag img {
	height: 18px;
}
.pll-aspect-ratio-32 .pll-switcher-flag img {
	height: 12px;
}
```

Note: we seem to apply the ratio differently. I was applying it to the image directly, while you seem to apply it to the span (when it's there). I think we should harmonize everything and probably use your method.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Screenfeed: 1. I think you mixed up `dropdown` and `select`.
2. For `show_labels` the "empty" value is an empty string (`block.json`, `switcher-controls.js`), so unless I'm missing something, we can't have `'no'`.
3. I think `$attributes['alignment']` can be removed.
4. The label and the wrapper are already present in the switcher.

```php
$attributes['unique_id']    = 'select' === $attributes['layout'] ? "lang_choice_{$dropdown_id}" : '';
$attributes['show_wrapper'] = true; // Just to make sure no-one can mess with this.

$settings        = new Settings( $attributes );
$switcher_output = ( new Switcher( $settings, $this->links ) )->get();

return sprintf( '<div %1$s>%2$s</div>', get_block_wrapper_attributes(), $switcher_output );
```

We'll probably have to do something like this at some point too, in other PRs I suppose:

```php
if ( ! empty( $attributes['className'] ) ) {
	$attributes['wrapper_classes'] = explode( ' ', trim( $attributes['className'] ) );
	unset( $attributes['className'] );
}
```

Also, maybe the `pll-alignment-***` will be an issue for you, so here's a first idea:

```php
$attributes['alignment'] = 'left'; // Force any value.
// ...
$switcher_output = ( new Switcher( $settings, $this->links ) )->get();
$switcher_output = str_replace( ' pll-alignment-left', '', $switcher_output ); // Then remove the class.
```

But since it's hacky, I got another idea: introduce a 5th choice in `Settings` for the alignment: `'none'`. In that case, we don't apply any alignment-related style (the CSS class `'pll-alignment-none'` could still be printed, so our users can use it in their styles). And then we can simply force `$attributes['alignment'] = 'none';` here.

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +121,93 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+		$processor = new \WP_HTML_Tag_Processor( $html );
+
+		while ( $processor->next_tag( array( 'tag_name' => 'SPAN', 'class_name' => 'pll-switcher-flag' ) ) ) {
+			$processor->set_attribute( 'data-aspect-ratio', $aspect_ratio );
+			$processor->set_attribute( 'style', $style );
+
+			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+				$processor->remove_attribute( 'width' );
+				$processor->remove_attribute( 'height' );
+				$processor->remove_attribute( 'style' );
```
Screenfeed: We shouldn't need this. [See](https://github.com/polylang/polylang/blob/698f8a900e6ae42c631eb1b5f460d60ac29ef4e5/src/Switcher/Element/Nav.php#L89).

```suggestion
```

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +121,93 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+	 * @param string $style         Inline CSS custom properties for the flag.
+	 * @return string
+	 */
+	private function apply_flag_styles_to_flag_images( string $html, string $aspect_ratio, string $style ): string {
+		$processor = new \WP_HTML_Tag_Processor( $html );
+
+		while ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+			$processor->remove_attribute( 'width' );
+			$processor->remove_attribute( 'height' );
+			$processor->remove_attribute( 'style' );
```
Screenfeed: Same here.

```suggestion
```

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +121,93 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+		$border_radius = max( 0, min( 100, (int) ( $attributes['flag_border_radius'] ?? 0 ) ) );
+		$flag_width    = max( 1, min( 1000, (int) ( $attributes['flag_width'] ?? 18 ) ) );
+		$style         = sprintf(
+			'--pll-flag-border-radius:%1$d;--pll-flag-width:%2$dpx',
+			$border_radius,
+			$flag_width
+		);
+
+		if ( '' === $attributes['show_labels'] ) {
+			return $this->apply_flag_styles_to_flag_images( $html, $aspect_ratio, $style );
```
Screenfeed: What would you think of always wrapping the image with the span, even when there is no label to display? It may simplify things. Then this method won't be required, and it may harmonize a few things maybe (CSS).
We'd only need to modify `Element\Nav::get_label()`, it will simplify it.

I started a bit in this way [in my PR](https://github.com/polylang/polylang/pull/1903/changes#diff-9055a22eba8f902141438892676bd414d14977c3ccf948b8f7156335de5ef6f3R80) but maybe we need to go all the way down this idea and always display the spans around the flag and the label: it may simplify things like here and bring consistency.

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -16,6 +18,41 @@
...
+	 */
+	public function register_editor_style(): void {
+		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+
+		Assets::register_styles();
+		Assets::register_scripts();
+
+		wp_register_style(
+			'pll-language-switcher-editor-style',
+			plugins_url( 'css/build/language-switcher-editor-style' . $suffix . '.css', POLYLANG_ROOT_FILE ),
```
Screenfeed: Just a suggestion, my personal limit is 1 concat' 😅

```suggestion
			plugins_url( "css/build/language-switcher-editor-style{$suffix}.css", POLYLANG_ROOT_FILE ),
```

src/Switcher/Assets.php
```diff
@@ -39,4 +39,28 @@ public static function enqueue_frontend_scripts(): void {
...
 	}
+
+	/**
+	 * Registers frontend CSS.
+	 *
+	 * @since 3.9
+	 *
+	 * @return void
+	 */
+	public static function register_styles(): void {
```
Screenfeed, line 50: Then we can simplify `enqueue_frontend_styles()` and `enqueue_frontend_scripts()` by doing this:

```php
public static function enqueue_frontend_styles(): void {
	self::register_styles();
	wp_enqueue_style( self::FRONTEND_ASSET_HANDLE );
}
```

I think we should rename those methods `register_frontend_styles()` and `register_frontend_scripts()`.

src/modules/Blocks/Language_Switcher/Standard/block.json
```diff
@@ -8,35 +8,73 @@
     "description": "Language switcher to insert in content or as a widget.",
     "example": {},
 	"attributes": {
-		"dropdown": {
-			"type": "boolean",
-			"default": 0
+		"layout": {
+			"type": "string",
+			"enum": ["vertical", "horizontal", "dropdown", "select"],
+			"default": "horizontal"
```
Screenfeed: The default value in `Settings` [is `vertical`](https://github.com/polylang/polylang/blob/f60927b541d04a6cfc31cae98885f2f8ff46fde8/src/Switcher/Settings/Settings.php#L23) (because it was displayed vertically before this new switcher). Should we harmonize?

src/modules/Blocks/Language_Switcher/Standard/block.json
```diff
@@ -8,35 +8,73 @@
...
 			"type": "boolean",
-			"default": 0
+			"default": false
 		},
 		"hide_if_no_translation": {
 			"type": "boolean",
-			"default": 0
+			"default": false
 		},
+    "flag_aspect_ratio": {
```
Screenfeed: This new block of properties and the following (`supports`, etc) are indented with space characters while the rest of the file is indented with tabs.

src/modules/Blocks/Language_Switcher/Standard/block.json
```diff
@@ -8,35 +8,73 @@
...
+      "default": 0
+    },
+    "flag_width": {
+      "type": "number",
+      "default": 18
+    },
+    "flag_margin_right": {
+      "type": "string",
+      "default": "0px"
+    },
```
Screenfeed: In current CSS file:

```css
.pll-switcher-flag + span {
	margin-inline-start: .3em;
	writing-mode: horizontal-tb;
}
```

We should harmonize here, instead of using a left margin on the label in one case, and a right margin on the flag in the other case.

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,416 @@
...
+
+const MARGIN_RIGHT_UNITS = [ 'px', 'em', 'rem', 'vh', 'vw' ];
+
+const MARGIN_RIGHT_UNIT_LABELS = {
+	px: __( 'px', 'polylang' ),
+	em: __( 'em', 'polylang' ),
+	rem: __( 'rem', 'polylang' ),
+	vh: __( 'vh', 'polylang' ),
+	vw: __( 'vw', 'polylang' ),
+};
```
Screenfeed: I believe all those units don't need to be translated 😅

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,416 @@
...
+						options={ [
+							{
+								label: __( 'Names', 'polylang' ),
+								value: 'names',
+							},
+							{
+								label: __( 'Codes', 'polylang' ),
+								value: 'codes',
+							},
+							{ label: __( 'None', 'polylang' ), value: '' },
```
Screenfeed: Do you know if this choice can be hidden or disabled when `show_flags` is `false` or `layout` is `'select'`?

Maybe this would work but it feels hacky and there's probably other things to think about:

```js
{ ( ! show_flags || 'select' === layout ) && (
	<SelectControl
		label={ __( 'Labels', 'polylang' ) }
		value={ show_labels }
		options={ [
			{
				label: __( 'Names', 'polylang' ),
				value: 'names',
			},
			{
				label: __( 'Codes', 'polylang' ),
				value: 'codes',
			},
		] }
		onChange={ ( value ) => { /* ... */ } }
		__next40pxDefaultSize
	/>
) }
{ ( show_flags && 'select' !== layout ) && (
	<SelectControl
		label={ __( 'Labels', 'polylang' ) }
		value={ show_labels }
		options={ [
			{
				label: __( 'Names', 'polylang' ),
				value: 'names',
			},
			{
				label: __( 'Codes', 'polylang' ),
				value: 'codes',
			},
			{ label: __( 'None', 'polylang' ), value: '' },
		] }
		onChange={ ( value ) => { /* ... */ } }
		__next40pxDefaultSize
	/>
) }
```

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,416 @@
...
+								// Reset margin right unit and value if labels are hidden.
+								updateMarginRightUnit( 'px' );
+								updateMarginRightValue( 0 );
+							}
+
+							setAttributes( { show_labels: value } );
+						} }
+						__next40pxDefaultSize
+					/>
+					<ToggleControl
```
Screenfeed: What do you think about this?

```js
( ( 'select' !== layout ) && {
	<ToggleControl
```

What I did in `Fields\Widget::get()`:
- Hide `show_flags` if `'select' === layout`,
- Hide `flag_aspect_ratio` if `'select' === layout` or `! show_flags`,
- Hide `show_labels` if `'select' === layout` (this means we can't select `'codes'` neither),
- Hide `hide_current` if `'select' === layout`.

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,416 @@
...
+									} )
+								}
+								__next40pxDefaultSize
+							/>
+							{ '' !== show_labels && (
+								<>
+									<ToggleGroupControl
+										__next40pxDefaultSize
+										label={ __(
+											'Margin right unit',
```
Screenfeed: The unit before the value? 🤔

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,416 @@
...
+							},
+							{
+								title: __( 'Codes', 'polylang' ),
+								onClick: () =>
+									setAttributes( { show_labels: 'codes' } ),
+							},
+							{
+								title: __( 'None', 'polylang' ),
+								onClick: () =>
+									setAttributes( { show_labels: '' } ),
```
Screenfeed: So I think there is nothing preventing to set `show_flags` to `false` && `show_labels` to `''` here, am I right?

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,416 @@
...
+ *
+ * @param {string} currentValue The current margin right value.
+ * @param {string} nextUnit     The next unit.
+ * @return {string} The margin right value.
+ */
+const getMarginRightValueFromUnit = ( currentValue, nextUnit ) => {
+	const regex = new RegExp(
+		`^(-?\\d+(?:\\.\\d+)?)(?:${ MARGIN_RIGHT_UNITS.map( ( unit ) =>
+			unit.replace( /([.*+?^${}()|\[\]\/\\])/g, '\\$1' )
+		).join( '|' ) })?$`
```
Screenfeed: Is it possible to have a space character between the value and the unit?
Also, why the regex for `MARGIN_RIGHT_UNITS`?


```suggestion
		`^(-?\\d+(?:\\.\\d+)?) *(?:${ MARGIN_RIGHT_UNITS.join( '|' ) })?$`
```

Same remarks for `getMarginRightValueFromValue`.

js/src/blocks/language-switcher/components/rendered-switcher.js
```diff
@@ -0,0 +1,182 @@
...
+										flag_border_radius,
+										flag_width,
+										flag_margin_right
+									) }
+								/>
+								{ isTopLevel && (
+									<span className="wp-block-navigation__submenu-icon">
+										<SubmenuIcon />
+									</span>
+								) }
```
Screenfeed: The top-level item should be displayed before the `<ul>`.

Idea:

```js
{ useCuratedLanguages( languages, currentLanguage, true ).map( ( language, index ) => {
	return (
		<li key={ language.slug }>
			<SwitcherLinkElement
				label={ getLabel( language, show_labels ) }
				flag={ getFlag(
					language,
					show_flags,
					flag_aspect_ratio,
					flag_border_radius,
					flag_width,
					flag_margin_right
				) }
			/>
			<span className="wp-block-navigation__submenu-icon">
				<SubmenuIcon />
			</span>
		</li>
	);
} ) }
<ul>
	{ useCuratedLanguages( languages, currentLanguage ).map( ( language, index ) => {
		return (
			<li key={ language.slug }>
				<SwitcherLinkElement
					label={ getLabel( language, show_labels ) }
					flag={ getFlag(
						language,
						show_flags,
						flag_aspect_ratio,
						flag_border_radius,
						flag_width,
						flag_margin_right
					) }
				/>
			</li>
		);
	} ) }
</ul>
```

I even wonder if we should render the `<ul>` and what's inside.

js/src/blocks/language-switcher/components/rendered-switcher.js
```diff
@@ -0,0 +1,182 @@
...
+							{ getLabel( language, show_labels ) }
+						</option>
+					);
+				} ) }
+			</select>
+		);
+	}
+
+	return (
+		<div className={ `pll-switcher pll-layout-${ layout }` }>
```
Screenfeed: Not a `<nav>` tag?

Screenfeed (changes requested):
I'll take a closer look at the CSS later, I think we have some things to see together before that.

Mainly, what I saw:
- A few small bugs, nothing dramatic IMHO.
- We must harmonize some things between what I've done and what you've done:
  - Instead of the `data-aspect-ratio` attribute on each item, we should use the CSS class on the wrapper (like in php): the previous CSS will work. However, things can be improved in the styles:
  - For the flag aspect ratio I mainly added style to the ``, while you mainly added style to the `<span>`. Let's find a way to harmonize all that.
  - And what I think will help, is to always wrap the label and the flag into spans (not only when they are both present).
  - I applied a margin-left on the label while you apply a margin right on the flag.
- Maybe we could create a `none` value for the `alignment` in the `Settings` class (and force it in the block) that wouldn't apply alignment-related styles to the switcher: this will help you apply the block's styles without being annoyed by the basic styles.

css/src/blocks/language-switcher-editor-style.css
Hug0-Drelon: See 2509bc1b.

js/src/blocks/language-switcher/components/rendered-switcher.js
Hug0-Drelon: See 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: See 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: Remove in 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: Removed in 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: Wrapped with span in 2509bc1b.

src/Switcher/Assets.php
Hug0-Drelon: Since these styles are enqueued in the editor as well, I'm not sure. What do you think?

src/modules/Blocks/Language_Switcher/Standard/block.json
Hug0-Drelon: Changed in 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/block.json
Hug0-Drelon: Fixed in 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/block.json
Hug0-Drelon: Harmonized in 2509bc1b.

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: i18n removed in 2509bc1b.

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: See a proposition in 2509bc1b.

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: Good idea, see 2509bc1b.

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: It's a matter of taste I'd say.

Hug0-Drelon: (we don't have access to a specific Gutenberg component yet https://developer.wordpress.org/block-editor/reference-guides/components/unit-control/)

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: There are notices, see 2509bc1b.

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: > Is it possible to have a space character between the value and the unit?
I don't think so, be see 2509bc1b for improvement.

js/src/blocks/language-switcher/components/rendered-switcher.js
Hug0-Drelon: See 2509bc1b.

js/src/blocks/language-switcher/components/rendered-switcher.js
Hug0-Drelon: Changed for `<nav>` in 2509bc1b.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Screenfeed: Maybe we should add a class to the 2nd span. I used `pll-switcher-label` in my PR. (It won't change anything for us, but it will help our users to target it if the flag is not displayed.)

src/modules/Blocks/Language_Switcher/Standard/block.json
Screenfeed: Nice 👍 
Btw, in case you were wondering, this `.3em` value comes from this:
https://github.com/polylang/polylang/blob/34e0ff929a874bd862b572f7bacce47fe1cdbed8/src/walker-list.php#L55

css/src/blocks/language-switcher-editor-style.css
```diff
@@ -0,0 +1,30 @@
+/*
+ * Block editor only. Base switcher styles are in frontend-switcher.css.
+ */
+
+/* Dropdown layout preview: hide the submenu list in the editor. */
+.wp-block-polylang-language-switcher .pll-switcher.pll-layout-dropdown ul,
+.pll-switcher.pll-layout-dropdown ul {
```
Screenfeed: ~I don't think the 2nd selector is useful.~

```suggestion
.wp-block-polylang-language-switcher .pll-switcher.pll-layout-dropdown ul {
```

EDIT: now that I can test it, it's the opposite, there's only one wrapper.

```suggestion
.pll-switcher.pll-layout-dropdown ul {
```

css/src/frontend-switcher.css
```diff
@@ -9,34 +9,45 @@
...
+	gap: 0.25em;
 	position: relative;
+	background-color: inherit;
 }
-.pll-layout-dropdown > .current-item {
-	display: block;
-	z-index: 1;
+
+.wp-block-polylang-language-switcher .pll-layout-dropdown,
+.pll-layout-dropdown.wp-block-polylang-language-switcher {
```
Screenfeed, line 23: I think one of those should be removed, but I'm not sure which one since I cannot test locally 🤔

Basically we have this:

```css
.wp-block-polylang-language-switcher .pll-layout-dropdown,
.wp-block-polylang-language-switcher.pll-layout-dropdown {
```

EDIT: now that I can test it, the 1st line can be removed, there's only 1 wrapper with both classes.

```css
.wp-block-polylang-language-switcher.pll-layout-dropdown {
```

css/src/frontend-switcher.css
```diff
@@ -137,20 +169,73 @@
...
+	/*
+	 * Map the block control value (0–100) to a CSS percentage radius.
+	 * 50% is the maximum useful value (fully rounded); the 0.5 factor
+	 * converts 100 → 50%, keeping the slider responsive across the range.
+	 */
+	border-radius: calc( var( --pll-flag-border-radius, 0 ) * 0.5 * 1% );
+}
+
+.pll-aspect-ratio-32 .pll-switcher-flag,
+.pll-switcher-flag[data-aspect-ratio="32"] {
```
Screenfeed: We don't need the `[data-aspect-ratio]` selectors anymore 👍

js/src/blocks/language-switcher/components/rendered-switcher.js
```diff
@@ -0,0 +1,210 @@
...
+										show_flags,
+										flag_border_radius,
+										flag_width
+									) }
+									labelSpacing={ labelSpacing }
+								/>
+							</li>
+						);
+					} ) }
+				</ul>
```
Screenfeed: Since we are in a `if ( layout === 'dropdown' ) {` statement, all of this can be removed. Then we won't even need to hide it we CSS.
I know we just discussed this, but seeing the code again, I don't understand why we should print this and hide it 🤔

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,437 @@
...
+			value: 'codes',
+		},
+	];
+
+	if ( show_flags ) {
+		labelOptions.push( {
+			label: __( 'None', 'polylang' ),
+			value: '',
+		} );
+	}
```
Screenfeed, line 71: Nice, much simpler than what I was thinking 👍

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,437 @@
...
+								<>
+									<ToggleGroupControl
+										__next40pxDefaultSize
+										label={ __(
+											'Label spacing unit',
+											'polylang'
+										) }
+										value={ labelSpacingUnit }
+										onChange={ updateLabelSpacingUnit }
+										isBlock
```
Screenfeed: Where does this `isBlock` comes from? 🤔

js/src/blocks/language-switcher/components/switcher-controls.js
```diff
@@ -0,0 +1,437 @@
...
+									<RangeControl
+										label={ __(
+											'Label spacing',
+											'polylang'
+										) }
+										initialPosition={ 0.3 }
+										value={ labelSpacingValue }
+										min={ 0 }
+										max={ 100 }
+										step={ 0.1 }
```
Screenfeed: Maybe the step should be `0.1` when the unit is `em` or `rem`, and `1` when `px`. I'm not sure for `vh` and `vw`. Do you think it's possible and relevant? Maybe we can take a look at how it's done elsewhere.

EDIT: Combined cursor and unit (buttons block):
image
image

And the `step` is indeed dynamic, it changes according to the chosen unit.

js/src/blocks/language-switcher/components/rendered-switcher.js
```diff
@@ -0,0 +1,210 @@
...
+						flag={ getFlag(
+							currentLanguageItem,
+							show_flags,
+							flag_border_radius,
+							flag_width
+						) }
+						labelSpacing={ labelSpacing }
+					/>
+				) }
+				<span className="wp-block-navigation__submenu-icon">
```
Screenfeed: Do you have any reason to use this CSS class?

If you don't, here is an idea:
1. Use the class `pll-submenu-toggle`, it's the one used by the real button.
2. In `frontend-switcher.css`, use `.pll-submenu-toggle` instead of `button`.
3. In `language-switcher-editor-style.css`, we'll probably be able to remove the related styles.

What do you think?

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +116,120 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+			$flag_width
+		);
+
+		$label_style = '';
+		if ( '' !== $attributes['show_labels'] ) {
+			$label_style = sprintf(
+				'--pll-flag-label-spacing:%1$s',
+				$attributes['flag_label_spacing'] ?? '0.3em'
+			);
+		}
```
Screenfeed, line 126: Since we provide a default value with `var( --pll-flag-label-spacing, 0.3em )`, I would suggest to add the spacing only if needed:

```suggestion
		$label_style   = '';
		$label_spacing = $attributes['flag_label_spacing'] ?? '0.3em';
		if ( '' !== $attributes['show_labels'] && '' !== $label_spacing && '0.3em' !== $label_spacing ) {
			$label_style = sprintf( '--pll-flag-label-spacing:%1$s', $label_spacing );
		}
```

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +116,120 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+			return $html;
+		}
+
+		$border_radius = max( 0, min( 100, (int) ( $attributes['flag_border_radius'] ?? 0 ) ) );
+		$flag_width    = max( 1, min( 1000, (int) ( $attributes['flag_width'] ?? 18 ) ) );
+		$flag_style    = sprintf(
+			'--pll-flag-border-radius:%1$d;--pll-flag-width:%2$dpx',
+			$border_radius,
+			$flag_width
+		);
```
Screenfeed, line 118: Here too, no need to add styles if they're set to their default value.

```suggestion
		$flag_style    = array();
		$border_radius = max( 0, min( 100, (int) ( $attributes['flag_border_radius'] ?? 0 ) ) );
		$flag_width    = max( 1, min( 1000, (int) ( $attributes['flag_width'] ?? 18 ) ) );
		if ( 0 !== $border_radius ) {
			$flag_style[] = "--pll-flag-border-radius:{$border_radius}";
		}
		if ( 18 !== $flag_width ) {
			$flag_style[] = "--pll-flag-width:{$flag_width}px";
		}
		$flag_style = implode( ';', $flag_style );
```

And then:

```php
if ( empty( $flag_style ) && '' === $label_style ) {
	return $html;
}

$processor = new \WP_HTML_Tag_Processor( $html );

while ( /* ... */ ) {
	if ( ! empty( $flag_style ) ) {
		$processor->set_attribute( 'style', implode( ';', $flag_style ) );
	}
```

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +116,120 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+
+		while ( $processor->next_tag( array( 'tag_name' => 'SPAN', 'class_name' => 'pll-switcher-flag' ) ) ) {
+			$processor->set_attribute( 'style', $flag_style );
+
+			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+				$processor->remove_attribute( 'width' );
+				$processor->remove_attribute( 'height' );
+			}
+
+			if ( '' !== $label_style && $processor->next_tag( array( 'tag_name' => 'SPAN' ) ) ) {
```
Screenfeed: If we add a class to the span like I suggested earlier, we can use it here:


```suggestion
			if ( '' !== $label_style && $processor->next_tag( array( 'tag_name' => 'SPAN', 'class_name' => 'pll-switcher-label' ) ) ) {
```

src/Switcher/Element/Nav.php
```diff
@@ -71,16 +71,18 @@ public function get_link(): string {
...
+
 		if ( empty( $this->label ) ) {
-			return (string) preg_replace( '/ style="[^"]*"/', '', $this->flag );
+			return sprintf( '<span class="pll-switcher-flag">%s</span>', $flag );
 		}
 
 		return sprintf(
 			'<span class="pll-switcher-flag">%s</span><span>%s</span>',
-			(string) preg_replace( '/ style="[^"]*"/', '', $this->flag ),
+			$flag,
```
Screenfeed: About my suggestion to add a class to the 2nd span, and since Github wrecked my suggestion:

```php
	public function get_label(): string {
		$label = '';

		if ( ! empty( $this->label ) ) {
			$label .= sprintf( '<span class="pll-switcher-label">%s</span>', esc_html( $this->label ) );
		}

		if ( ! empty( $this->flag ) ) {
			$flag   = (string) preg_replace( '/ style="[^"]*"/', '', $this->flag );
			$label .= sprintf( '<span class="pll-switcher-flag">%s</span>', $flag );
		}

		return $label;
	}
```

Or:

```php
	public function get_label(): string {
		if ( empty( $this->flag ) ) {
			return sprintf( '<span class="pll-switcher-label">%s</span>', esc_html( $this->label ) );
		}

		$flag = (string) preg_replace( '/ style="[^"]*"/', '', $this->flag );

		if ( empty( $this->label ) ) {
			return sprintf( '<span class="pll-switcher-flag">%s</span>', $flag );
		}

		return sprintf(
			'<span class="pll-switcher-flag">%s</span><span class="pll-switcher-label">%s</span>',
			$flag,
			esc_html( $this->label )
		);
	}
```

Screenfeed (changes requested):
A few things to tweak IMHO, but nice overall 👍

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -82,4 +116,120 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
...
+
+		$processor = new \WP_HTML_Tag_Processor( $html );
+
+		while ( $processor->next_tag( array( 'tag_name' => 'SPAN', 'class_name' => 'pll-switcher-flag' ) ) ) {
+			$processor->set_attribute( 'style', $flag_style );
+
+			if ( $processor->next_tag( array( 'tag_name' => 'IMG' ) ) ) {
+				$processor->remove_attribute( 'width' );
+				$processor->remove_attribute( 'height' );
+			}
```
Screenfeed: I don't think we need to remove them. We have styles to overwrite them and I believe they are useful to the browsers.

Screenfeed:
So there's an issue with rounded flags:
- Ratio 1/1.
- Border radius set to 100.

The top-level on a dropdown is round, but the ones in the list are not. The issue seems to come from `display: flex` on the parent `<a>`.

Screenfeed:
<details>
<summary>Preview of the current controls</summary>

image

</details>

css/src/blocks/language-switcher-editor-style.css
```diff
@@ -0,0 +1,30 @@
...
+	width: 0.6em;
+	height: 0.6em;
+	stroke: currentColor;
+}
+
+/* Override Gutenberg core styles on flag images. */
+.pll-switcher-flag img {
+	width: 100% !important;
+	height: 100% !important;
+}
```
Screenfeed: No required, there's already this in the other file:

```css
.pll-switcher-flag[style] img,
.pll-aspect-ratio-32 .pll-switcher-flag img,
.pll-aspect-ratio-11 .pll-switcher-flag img,
.pll-switcher-flag[data-aspect-ratio] img {
	width: 100%;
	height: 100%;
}
```


```suggestion
```

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: https://developer.wordpress.org/block-editor/reference-guides/components/toggle-group-control/#isblock-boolean

js/src/blocks/language-switcher/components/switcher-controls.js
Hug0-Drelon: As discussed, let's try to back port https://developer.wordpress.org/block-editor/reference-guides/components/unit-control/.
I'll ping once done.

js/src/blocks/language-switcher/components/rendered-switcher.js
Hug0-Drelon: Yes, this allows WordPress to handle most of the design.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: I don't think it worth it. It makes maintenance harder (more default values everywhere) and `''` string is not a proper value.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: As answered previously, I find it easier that way.

js/src/blocks/language-switcher/components/rendered-switcher.js
Hug0-Drelon: Removed in ab375c71.

css/src/frontend-switcher.css
Hug0-Drelon: Removed in ab375c71.

css/src/frontend-switcher.css
Hug0-Drelon: Applied in ab375c71.

css/src/blocks/language-switcher-editor-style.css
Hug0-Drelon: Applied in ab375c71.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: Applied in a5176c4d.

src/Switcher/Element/Nav.php
Hug0-Drelon: Second suggestion applied in a5176c4d.

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: Code removed in 403fe6af.

Hug0-Drelon:
> So there's an issue with rounded flags:
> 
> * Ratio 1/1.
> * Border radius set to 100.
> 
> The top-level on a dropdown is round, but the ones in the list are not. The issue seems to come from `display: flex` on the parent `<a>`.

Fixed in f1759640.

js/src/blocks/language-switcher/components/rendered-switcher.js
Hug0-Drelon: I applied your suggestion in 77dbe008, took the opportunity to remove `language-switcher-editor-style.css` file.

css/src/frontend-switcher.css
Screenfeed: Several are still present but we'll see that later, once you're done with the new controls. Maybe I could try to push a commit then if you're OK with that.

js/src/blocks/language-switcher/components/switcher-controls.js
Screenfeed: OK, understood, it's not a var coming from nowhere as I was supposing. I'm not used to this writing yet.

Hug0-Drelon: It's not a var, it's a prop.

css/src/frontend-switcher.css
Hug0-Drelon: Final sweep in c1a2d200.

js/src/blocks/language-switcher/components/switcher-controls.js
Screenfeed: Yes 👍

css/src/frontend-switcher.css
Hug0-Drelon: Applied in 45ef90a6.

Screenfeed (approved):

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -16,6 +18,32 @@
...
+	 * Registers the editor style for the language switcher block.
+	 *
+	 * @since 3.9
+	 *
+	 * @return void
+	 */
+	public function register_editor_style(): void {
+		Assets::register_styles();
+		Assets::register_scripts();
+	}
```
Chouby:
```suggestion
		add_action( 'init', array( Assets:class, 'register_styles' ) );
		add_action( 'init', array( Assets:class, 'register_scripts' ) );
	}
```

src/Switcher/Settings/Settings.php
```diff
@@ -208,7 +209,7 @@ public function get_switcher_class(): string {
 	protected function validate( array $settings ): array {
 		$choices   = array(
 			'layout'                 => array( 'horizontal', 'vertical', 'dropdown', 'select' ),
-			'alignment'              => array( 'left', 'center', 'right', 'stretched' ),
+			'alignment'              => array( 'left', 'center', 'right', 'stretched', 'none' ),
```
Chouby: What's the purpose of this alignment value 'none'? To have a dynamic value depending on the text direction of the current language? In this case, shouldn't we call it 'default' and rewrite the doc to say the default value is 'default' which results in an alignment varying between left and right?

src/modules/Blocks/Language_Switcher/Standard/Block.php
```diff
@@ -44,31 +72,23 @@ public function render( $attributes, $content, $block ) { //phpcs:ignore Variabl
 		static $dropdown_id = 0;
 		++$dropdown_id;
 
-		// Sets a unique id for dropdown in PLL_Switcher::the_language().
-		$attributes['dropdown'] = empty( $attributes['dropdown'] ) ? 0 : $dropdown_id;
-
-		$attributes = $this->set_attributes_for_block( $attributes );
+		$attributes['unique_id']    = 'select' === $attributes['layout'] ? 'lang_choice_' . $dropdown_id : '';
+		$attributes['show_wrapper'] = true;
+		$attributes['alignment']    = 'none';
```
Chouby: See my comment above. The doc says 'left' or 'right' as default value, not 'none'.

Chouby (changes requested):

src/Switcher/Settings/Settings.php
Hug0-Drelon: The `none` alignment disables the settings entirely.

This is because the block settings matching this behavior is handled by WordPress directly, not from one of ours attribute. ([see](https://github.com/polylang/polylang/blob/new-language-switcher-block/src/modules/Blocks/Language_Switcher/Standard/block.json#L65))

src/modules/Blocks/Language_Switcher/Standard/Block.php
Hug0-Drelon: The `$attribute` variable name may be confusing, but `alignement` is a core switcher setting, not a block attribute.

Screenfeed:
@Hug0-Drelon I just spotted a bug. In PLL Pro, the URL of the assets is `https://example.com/wp-content/plugins/polylang-pro/vendor/wpsyntex/polylang/css/build/frontend-switcher.css?ver=3.9-dev`, which returns a 404 error.

In `Assets::register_styles()` and `Assets::register_scripts()` we must use `POLYLANG_ROOT_FILE` instead of `POLYLANG_FILE`.

PR #1899, opened 2026-06-04
Add views for languages in posts list table

### LINKED ISSUE #1057: Feat: Add admin table filters to Post Type and Taxonomies

Chouby:
Closes #1057

This PR adds views for languages with post count above the posts list table. This counts all posts whatever the status.
The language views can be combined with the status views already available.
The WordPress 'All' view resets the languages views.

This feature somewhat duplicates the admin language filter available in the admin bar for the posts lists screens with the additional information of the posts count. This means that unlike WordPress status views, the language views are persistent (stored in DB).

This may help users on wp.com who don't have access to the admin bar.

Here is how it looks like:
image

### DIFF
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -297,6 +297,9 @@ class PLL_Admin_Filters_Post {
 				(int) $this->model->count_posts( $language, $q )
 			);
 		}
+
+		$views['all'] = str_replace( 'edit.php?', 'edit.php?lang=all&', $views['all'] );
+
 		return $views;
 	}
 }
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -46,6 +46,11 @@ class PLL_Admin_Filters_Post {
 
 		// Sets the language in Tiny MCE
 		add_filter( 'tiny_mce_before_init', array( $this, 'tiny_mce_before_init' ) );
+
+		// Languages views in post list tables.
+		foreach ( $this->model->get_translated_post_types() as $post_type ) {
+			add_filter( "views_edit-{$post_type}", array( $this, 'views_edit_post' ) );
+		}
 	}
 
 	/**
@@ -262,5 +267,36 @@ class PLL_Admin_Filters_Post {
 			$mce_init['directionality'] = $this->curlang->is_rtl ? 'rtl' : 'ltr';
 		}
 		return $mce_init;
+	}
+
+	/**
+	 * Add languages filter views to the list of views available for the post list table.
+	 *
+	 * @since 3.9
+	 *
+	 * @param string[] $views Array of available list table views.
+	 */
+	public function views_edit_post( $views ) {
+		global $typenow;
+
+		if ( empty( $typenow ) ) {
+			return $views;
+		}
+
+		$q = array(
+			'post_type'   => $typenow,
+			'post_status' => 'any',
+		);
+
+		foreach ( $this->model->languages->get_list() as $language ) {
+			$views[ $language->slug ] = sprintf(
+				'<a href="%s" %s>%s <span class="count">(%d)</span></a>',
+				esc_url( add_query_arg( 'lang', $language->slug, remove_query_arg( 'paged' ) ) ),
+				! empty( $this->curlang ) && $this->curlang->slug === $language->slug ? 'class="current" aria-current="page"' : '',
+				esc_html( $language->name ),
+				(int) $this->model->count_posts( $language, $q )
+			);
+		}
+		return $views;
 	}
 }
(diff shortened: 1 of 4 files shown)


src/admin/admin-filters-post.php
```diff
@@ -263,4 +268,39 @@ public function tiny_mce_before_init( $mce_init ) {
...
+
+	/**
+	 * Add languages filter views to the list of views available for the post list table.
+	 *
+	 * @since 3.9
+	 *
+	 * @param string[] $views Array of available list table views.
+	 * @return string[]
+	 */
+	public function views_edit_post( $views ) {
```
Hug0-Drelon:
```suggestion
	public function add_per_language_edit_post_views( $views ) {
```
Suggesting a more descriptive name.

src/admin/admin-filters-post.php
```diff
@@ -46,6 +46,11 @@ public function __construct( &$polylang ) {
 
 		// Sets the language in Tiny MCE
 		add_filter( 'tiny_mce_before_init', array( $this, 'tiny_mce_before_init' ) );
+
+		// Languages views in post list tables.
+		foreach ( $this->model->get_translated_post_types() as $post_type ) {
+			add_filter( "views_edit-{$post_type}", array( $this, 'views_edit_post' ) );
```
Hug0-Drelon:
```suggestion
			add_filter( "views_edit-{$post_type}", array( $this, 'add_per_language_edit_post_views' ) );
```
Follows the suggestion bellow.

src/admin/admin-filters-post.php
```diff
@@ -263,4 +268,39 @@ public function tiny_mce_before_init( $mce_init ) {
...
+			'post_status' => 'any',
+		);
+
+		foreach ( $this->model->languages->get_list() as $language ) {
+			$views[ $language->slug ] = sprintf(
+				'<a href="%s" %s>%s <span class="count">(%d)</span></a>',
+				esc_url( add_query_arg( 'lang', $language->slug, remove_query_arg( 'paged' ) ) ),
+				! empty( $this->curlang ) && $this->curlang->slug === $language->slug ? 'class="current" aria-current="page"' : '',
+				esc_html( $language->name ),
+				(int) $this->model->count_posts( $language, $q )
```
Hug0-Drelon: Should we use [number_format_i18n()](https://developer.wordpress.org/reference/functions/number_format_i18n/) here?

src/admin/admin-filters-post.php
```diff
@@ -263,4 +268,39 @@ public function tiny_mce_before_init( $mce_init ) {
...
+			);
+		}
+
+		$views['all'] = str_replace( 'edit.php?', 'edit.php?lang=all&', $views['all'] );
```
Hug0-Drelon:
```suggestion
		if ( ! isset( $views['all'] ) ) {
			return $views;
		}

		$views['all'] = str_replace( 'edit.php?', 'edit.php?lang=all&', $views['all'] );
```
Since this is a filter, a third party could remove the key for some reason.

Hug0-Drelon (changes requested):

src/admin/admin-filters-post.php
Chouby: Not. It's only for float. Here we have an int.

src/admin/admin-filters-post.php
Chouby: See my proposal in 9d04026a2570db1a114312c69574476028b1e0f4

src/admin/admin-filters-post.php
Chouby: Fixed by e7e857f82ee0f5ae890019db6d6cac0b223ae5d5

Chouby:
Rebase + force push to resolve a merge conflitct.

src/admin/admin-filters-post.php
Hug0-Drelon: What about `1,000` (`en_US`) vs `1 000` (`fr_FR`)?

Chouby: Yes! I forgot this :(

Hug0-Drelon (approved):

PR #1903, opened 2026-06-10
[Language switcher rework] Classic menu

Screenfeed:
This implements the new switcher classes for the classic menu.

Some details:

- Small tweaks to CSS: align flags to the baseline instead of middle. This allows to better align items with and without a flag.
- Small tweaks to the markup: when an is supposed to display a flag but the related language doesn't have one, still print the `<span>` around the label. It doesn't change anything to the display but it improves consistency with the other items that have this `<span>`, and it adds a possibility for our users to customize it with custom CSS.
- The admin JS file now passes ESLint.

image

### DIFF
--- a/js/src/nav-menu.js
+++ b/js/src/nav-menu.js
@@ -52,7 +52,7 @@ const pllNavMenu = {
 
 			const metabox = event.target
 				.closest( '.menu-item' )
-				.querySelector( '.menu-item-settings' );
+				?.querySelector( '.menu-item-settings' );
 
 			if ( ! metabox?.id ) {
 				// Should not happen.
@@ -292,6 +292,10 @@ const pllNavMenu = {
 			// Hide rows.
 			const key = event.target.getAttribute( 'data-key' );
 
+			if ( ! key ) {
+				return;
+			}
+
 			wrapper
 				.querySelectorAll(
 					`[class*="pll-hidden-if-${ key }-"]:not(.pll-hidden-if-${ key }-${ value })` // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword, Generic.ControlStructures.InlineControlStructure.NotAllowed
@@ -311,7 +315,7 @@ const pllNavMenu = {
 					'[data-key="show_flags"]'
 				);
 
-				if ( true !== otherInput.checked ) {
+				if ( otherInput && true !== otherInput.checked ) {
 					otherInput.checked = true;
 					otherInput.dispatchEvent(
 						new Event( 'change', { bubbles: true } )
@@ -322,7 +326,7 @@ const pllNavMenu = {
 					'[data-key="show_labels"]'
 				);
 
-				if ( '' === otherInput.value ) {
+				if ( otherInput && '' === otherInput.value ) {
 					otherInput.value = 'names';
 					otherInput.dispatchEvent(
 						new Event( 'change', { bubbles: true } )
--- a/js/src/nav-menu.js
+++ b/js/src/nav-menu.js
@@ -1,7 +1,5 @@
 /**
  * Handles the options in the language switcher nav menu metabox.
- *
- * @package Polylang
  */
 
 const pllNavMenu = {
@@ -34,8 +32,10 @@ const pllNavMenu = {
 		}
 
 		pllNavMenu.wrapper.addEventListener( 'click', pllNavMenu.printMetabox );
-		pllNavMenu.wrapper.addEventListener( 'change', pllNavMenu.ensureContent );
-		pllNavMenu.wrapper.addEventListener( 'change', pllNavMenu.showHideRows );
+		pllNavMenu.wrapper.addEventListener(
+			'change',
+			pllNavMenu.manageRowsAndValues
+		);
 	},
 
 	printMetabox: {
@@ -50,71 +50,118 @@ const pllNavMenu = {
 				return;
 			}
 
-			const metabox = event.target.closest( '.menu-item' ).querySelector( '.menu-item-settings' );
+			const metabox = event.target
+				.closest( '.menu-item' )
+				.querySelector( '.menu-item-settings' );
 
 			if ( ! metabox?.id ) {
 				// Should not happen.
 				return;
 			}
 
-			if ( ! metabox.querySelectorAll( 'input[value="#pll_switcher"][type=text]' ).length ) {
+			if (
+				! metabox.querySelectorAll(
+					'input[value="#pll_switcher"][type=text]'
+				).length
+			) {
 				// Not our metabox, or already replaced.
 				return;
 			}
 
 			// Remove default fields we don't need.
 			[ ...metabox.children ].forEach( ( el ) => {
-				if ( 'P' === el.nodeName && ! el.classList.contains( 'field-move' ) ) {
+				if (
+					'P' === el.nodeName &&
+					! el.classList.contains( 'field-move' )
+				) {
 					el.remove();
 				}
 			} );
 
-			const t      = pllNavMenu.printMetabox;
-			const itemId = Number( metabox.id.replace( 'menu-item-settings-', '' ) );
+			const t = pllNavMenu.printMetabox;
+			const itemId = Number(
+				metabox.id.replace( 'menu-item-settings-', '' )
+			);
 
-			metabox.append( t.createHiddenInput( 'title', itemId, pll_data.title ) ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
-			metabox.append( t.createHiddenInput( 'url', itemId, '#pll_switcher' ) ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+			// phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+			metabox.append(
+				t.createHiddenInput( 'title', itemId, pll_data.title )
+			);
+			// phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+			metabox.append(
+				t.createHiddenInput( 'url', itemId, '#pll_switcher' )
+			);
 			metabox.append( t.createHiddenInput( 'pll-detect', itemId, 1 ) ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
 
-			const ids          = Array( 'hide_if_no_translation', 'hide_current', 'force_home', 'show_flags', 'show_names', 'dropdown' ); // Reverse order.
-			const isValDefined = typeof( pll_data.val[ itemId ] ) !== 'undefined';
+			const menuValues =
+				typeof pll_data.val[ itemId ] !== 'undefined'
+					? pll_data.val[ itemId ]
+					: {};
 
-			ids.forEach( ( optionName ) => {
-				// Create the checkbox's wrapper.
-				const inputWrapper = t.createElement( 'p', { class: 'description' } );
+			Object.keys( pll_data.data )
+				.reverse()
+				.forEach( ( optionName ) => {
+					const optionData = pll_data.data[ optionName ];
+					const optionValue =
+						typeof menuValues[ optionName ] !== 'undefined'
+							? menuValues[ optionName ]
+							: optionData.default;
+					// Create the wrapper.
+					const wrapperAtts = { class: 'description' };
 
-				if ( 'hide_current' === optionName && isValDefined && 1 === pll_data.val[ itemId ].dropdown ) {
-					// Hide the `hide_current` checkbox if `dropdown` is checked.
-					inputWrapper.classList.add( 'hidden' );
-				}
+					if ( optionData.hide_if ) {
+						Object.keys( optionData.hide_if ).forEach(
+							( conditionName ) => {
+								const conditionValue =
+									optionData.hide_if[ conditionName ];
+								wrapperAtts.class += ` pll-hidden-if-${ conditionName }-${ conditionValue }`; // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword, Squiz.ControlStructures.ControlSignature.NewlineAfterOpenBrace
 
-				metabox.prepend( inputWrapper ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
+								if (
+									typeof menuValues[ conditionName ] !==
+										'undefined' &&
+									conditionValue ===
+										menuValues[ conditionName ]
+								) {
+									wrapperAtts.class += ` pll-hidden-by-${ conditionName }`;
+								}
+							}
+						);
+					}
 
-				// Create the checkbox's label.
-				const inputId = `edit-menu-item-${ optionName }-${ itemId }`;
-				const label   = t.createElement( 'label', { 'for': inputId } );
-				label.innerText = ` ${ pll_data.strings[ optionName ] }`;
+					const inputWrapper = t.createElement( 'p', wrapperAtts );
 
-				inputWrapper.append( label ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+					metabox.prepend( inputWrapper ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
 
-				// Create the checkbox.
-				const cb = t.createElement( 'input', {
-					type:  'checkbox',
-					id:    inputId,
-					name:  `menu-item-${ optionName }[${ itemId }]`,
-					value: 1,
+					// Create the label.
+					const label = t.createElement( 'label', {
+						for: `edit-menu-item-${ optionName }-${ itemId }`, // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword
+					} );
+					label.innerText = ` ${ optionData.label } `;
+
+					inputWrapper.append( label ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+
+					// Create the input.
+					if ( optionData.choices ) {
+						const input = t.createSelectInput(
+							optionName,
+							itemId,
+							optionValue,
+							optionData.choices
+						);
+						label.append( input ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+					} else {
+						const input = t.createCheckboxInput(
+							optionName,
+							itemId,
+							optionValue
+						);
+						label.prepend( input ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
+					}
 				} );
-
-				if ( ( isValDefined && 1 === pll_data.val[ itemId ][ optionName ] ) || ( ! isValDefined && 'show_names' === optionName ) ) { // `show_names` as default value.
-					cb.checked = true;
-				}
-
-				label.prepend( cb ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
-			} );
 		},
 
 		/**
-		 * Creates and returns a `<input type=hidden"/>` element.
+		 * Creates and returns a `<input type="hidden"/>` element.
 		 *
 		 * @param {string}        id     An identifier for this input. It will be part of the final `id` attribute.
 		 * @param {number}        itemId The ID of the menu element (post ID).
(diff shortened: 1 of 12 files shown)


src/Switcher/Element/Nav.php
```diff
@@ -70,18 +70,17 @@ public function get_link(): string {
...
+			$flag   = (string) preg_replace( '/ style="[^"]*"/', '', $this->flag );
+			$label .= sprintf( '<span class="pll-switcher-flag">%s</span>', $flag );
 		}
 
-		return sprintf(
-			'<span class="pll-switcher-flag">%s</span><span>%s</span>',
-			(string) preg_replace( '/ style="[^"]*"/', '', $this->flag ),
-			esc_html( $this->label )
-		);
+		return $label;
```
Chouby, line 83: This modifies the behavior if both `$this->label` and `$this->flag` are empty.

src/Switcher/Settings/Menu.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @return array
+	 */
+	protected function convert_to_legacy( array $settings ): array {
+		$args = parent::convert_to_legacy( $settings );
+
+		if ( isset( $settings['layout'] ) && 'dropdown' === $settings['layout'] ) {
+			$args['dropdown'] = 1;
+		} else {
+			$args['dropdown'] = 0;
+		}
```
Chouby, line 79: In my opinion this could be merged with https://github.com/polylang/polylang/blob/c0dc4811c2c51337d3c2e57d35fc28b48ea5996f/src/Switcher/Settings/Abstract_Settings_Legacy.php#L168-L170

src/Switcher/Settings/Menu.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @return array
+	 */
+	protected function validate( array $settings ): array {
+		if ( isset( $settings['layout'] ) ) {
+			if ( 'select' === $settings['layout'] ) {
+				$settings['layout'] = 'dropdown';
+			} elseif ( 'vertical' === $settings['layout'] ) {
+				$settings['layout'] = 'horizontal';
+			}
+		}
```
Chouby: I would simplify and have only one default value. Moreover, if I don't misread the code, if I we pass `trucmuche` as value, it will end to `vertical`.
```suggestion
		if ( ! isset( $settings['layout'] ) || 'dropdown' !=== $settings['layout'] ) {
			$settings['layout'] = 'horizontal';
		}
```

src/Switcher/Element/Nav.php
Screenfeed: `Settings` ensures that [it isn't possible](https://github.com/polylang/polylang/blob/4b2088a73671359cb677a540f9638921621ee07e/src/Switcher/Settings/Settings.php#L257-L260).

src/Switcher/Settings/Menu.php
Screenfeed: `select` in `Abstract_Settings_Legacy.php`, `dropdown` here. How would we do that?

src/Switcher/Settings/Menu.php
Screenfeed: What you suggest doesn't do the same thing that what the current code does.

Current code:
- `horizontal` => stays `horizontal` (which is the default value).
- `vertical` => converted to `horizontal`.
- `dropdown` => stays `dropdown` (which corresponds to `$args['dropdown'] = 1;`).
- `select` => converted to `dropdown` (which corresponds to `$args['dropdown'] = 1;`).
- `trucmuche` => converted to `vertical` by `Settings`, which is indeed a bug. I will fix this.

With your suggestion, `select` is converted to `horizontal` while it should be converted to `dropdown`.

Screenfeed: Fixed in [c3cded1](https://github.com/polylang/polylang/pull/1903/commits/c3cded14185d911ef6fbab36a9d4f3aaa0fc0091).

js/src/nav-menu.js
```diff
@@ -50,83 +50,200 @@ const pllNavMenu = {
...
+						label.append( input ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
+					} else {
+						const input = t.createCheckboxInput(
+							optionName,
+							itemId,
+							optionValue
+						);
+						label.prepend( input ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
+					}
+				} );
```
Hug0-Drelon, line 136:
```suggestion
			for ( const [ optionName, optionData ] of Object.entries(
				pll_data.data
			).reverse() ) {
				const optionValue =
					typeof menuValues[ optionName ] !== 'undefined'
						? menuValues[ optionName ]
						: optionData.default;
				// Create the wrapper.
				const wrapperAtts = { class: 'description' };
				if ( optionData.hide_if ) {
					Object.keys( optionData.hide_if ).forEach(
						( conditionName ) => {
							const conditionValue =
								optionData.hide_if[ conditionName ];
							wrapperAtts.class += ` pll-hidden-if-${ conditionName }-${ conditionValue }`; // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword, Squiz.ControlStructures.ControlSignature.NewlineAfterOpenBrace
							if (
								typeof menuValues[ conditionName ] !==
									'undefined' &&
								conditionValue === menuValues[ conditionName ]
							) {
								wrapperAtts.class += ` pll-hidden-by-${ conditionName }`;
							}
						}
					);
				}
				const inputWrapper = t.createElement( 'p', wrapperAtts );
				metabox.prepend( inputWrapper ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
				// Create the label.
				const label = t.createElement( 'label', {
					for: `edit-menu-item-${ optionName }-${ itemId }`, // phpcs:ignore Squiz.ControlStructures.ControlSignature.SpaceAfterKeyword
				} );
				label.innerText = ` ${ optionData.label } `;
				inputWrapper.append( label ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
				// Create the input.
				if ( optionData.choices ) {
					const input = t.createSelectInput(
						optionName,
						itemId,
						optionValue,
						optionData.choices
					);
					label.append( input ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.append
				} else {
					const input = t.createCheckboxInput(
						optionName,
						itemId,
						optionValue
					);
					label.prepend( input ); // phpcs:ignore WordPressVIPMinimum.JS.HTMLExecutingFunctions.prepend
				}
			}
```

Proposing an alternative; using `forEach` along `reverse()` then getting data from object with `pll_data.data[ optionName ]` is not optimized IMHO. Moreover, the intent is not really clear.
I'd suggest to use a simpler loop (`for` in that case) and to re-order appended items.

src/frontend/frontend-nav-menu.php
```diff
@@ -102,57 +90,67 @@ public function wp_get_nav_menu_items( $items ) {
...
+					// Since it is added to the parent `<li>`, no need to add it to the child elements too.
+					$item->classes[] = 'pll-aspect-ratio-' . str_replace( ':', '', $settings->flag_aspect_ratio );
+				}
+
 				$new_items[] = $item;
+				++$offset;
 			}
+
+			$i = 0; // For incrementation of menu order only in case of dropdown.
+			foreach ( $elements as $element ) {
```
Hug0-Drelon, line 131: If for some reason `$elements` is empty, the first dropdown item would be created anyway. Should we check `empty( $elements )` around line 106?

src/Switcher/Element/Nav.php
```diff
@@ -70,18 +70,17 @@ public function get_link(): string {
 	 * @return string
 	 */
 	public function get_label(): string {
-		if ( empty( $this->flag ) ) {
-			return esc_html( $this->label );
+		$label = '';
+
+		if ( ! empty( $this->flag ) ) {
+			$flag   = (string) preg_replace( '/ style="[^"]*"/', '', $this->flag );
```
Hug0-Drelon: I think `WP_HTML_Tag_Processor` would be a good fit for this use case.

src/Switcher/Fields/Menu.php
```diff
@@ -0,0 +1,88 @@
...
+	 * @since 3.9
+	 *
+	 * @param array $settings Switcher settings.
+	 * @return array
+	 */
+	public static function add_legacy_settings( array $settings ): array {
+		$settings['dropdown']   = isset( $settings['layout'] ) && 'dropdown' === $settings['layout'] ? 1 : 0;
+		$settings['show_names'] = ! empty( $settings['show_labels'] ) ? 1 : 0;
+		return $settings;
+	}
```
Hug0-Drelon, line 87: This is duplicated from `WP_Syntex\Polylang\Switcher\Fields\Widget`, isn't it?

src/Switcher/Fields/Menu.php
```diff
@@ -0,0 +1,88 @@
...
+			'force_home'             => array(
+				'label' => __( 'Force link to front page', 'polylang' ),
+			),
+			'hide_current'           => array(
+				'label' => __( 'Hide the current language', 'polylang' ),
+			),
+			'hide_if_no_translation' => array(
+				'label' => __( 'Hide languages with no translation', 'polylang' ),
+			),
+		);
```
Hug0-Drelon, line 68: I wonder if the maintenance is easier with (mostly) duplicated array between `Menu` and `Widget`, or if we should merge both array in the abstract class (and override in children).

src/Switcher/Settings/Menu.php
```diff
@@ -0,0 +1,83 @@
...
+	 *
+	 * @since 3.9
+	 *
+	 * @param array $settings The settings.
+	 * @return array
+	 */
+	protected function convert_from_legacy( array $settings ): array {
+		$settings = parent::convert_from_legacy( $settings );
+
+		$settings['layout'] = ! empty( $settings['dropdown'] ) ? 'dropdown' : 'horizontal';
```
Hug0-Drelon, line 58: This is the same logic for widgets, am I right?
If so, should it be refactored?

js/src/nav-menu.js
```diff
@@ -146,71 +263,71 @@ const pllNavMenu = {
...
 
-			if ( event.target.checked ) {
-				// Uncheck after hiding.
-				hideCb.checked = false;
-				hideCb.dispatchEvent( new Event( 'change' ) );
+			// Forbid disabling both flag and name/code.
+			if ( 'show_labels' === key && '' === value ) {
+				const otherInput = wrapper.querySelector(
+					'[data-key="show_flags"]'
+				);
```
Hug0-Drelon, line 314:
```suggestion
				const otherInput = wrapper.querySelector(
					'[data-key="show_flags"]'
				);
				
				if ( ! otherInput ) {
				    return;
			    }
```
Otherwise it could throw later.

js/src/nav-menu.js
```diff
@@ -50,83 +50,200 @@ const pllNavMenu = {
 				return;
 			}
 
-			const metabox = event.target.closest( '.menu-item' ).querySelector( '.menu-item-settings' );
+			const metabox = event.target
+				.closest( '.menu-item' )
+				.querySelector( '.menu-item-settings' );
```
Hug0-Drelon:
```suggestion
				?.querySelector( '.menu-item-settings' );
```
Safer

js/src/nav-menu.js
```diff
@@ -146,71 +263,71 @@ const pllNavMenu = {
...
 
-			if ( ! hideCb ) {
-				// Should not happen.
+			if ( ! wrapper ) {
 				return;
 			}
 
-			const description = hideCb.closest( '.description' );
+			// Hide rows.
+			const key = event.target.getAttribute( 'data-key' );
```
Hug0-Drelon, line 291:
```suggestion
			const key = event.target.getAttribute( 'data-key' );
			
			if ( ! key ) {
			    return;
			}
```
Just in case, prevent unnecessary computing.

Hug0-Drelon (changes requested):

js/src/nav-menu.js
Screenfeed: There are fields at the end of the metabox: if we don't insert things in reverse order, those fields appear at the top instead of the bottom.

image

src/Switcher/Element/Nav.php
Screenfeed: OK why not. Now I have a complementary idea: instead of removing the attribute here, let's remove it in `Abstract_Element`'s constructor. So, like `set_url()`, we may have `set_flag()`.

```php
private function set_flag(): void {
	if ( ! $this->settings->show_flags ) {
		return;
	}

	$this->flag = $this->language->get_display_flag( $this->settings->show_labels ? 'no-alt' : 'alt' );

	if ( empty( $this->flag ) ) {
		return;
	}

	$flag = new WP_HTML_Tag_Processor( $this->flag );

	if ( ! $flag->next_tag( array( 'tag_name' => 'img' ) ) ) {
		return;
	}

	if ( ! $flag->remove_attribute( 'style' ) ) {
		return;
	}

	$this->flag = $flag->get_updated_html();
}
```

src/Switcher/Fields/Menu.php
Screenfeed: No, in `Fields\Widget` the layout is compared to `select`, not `dropdown`.

src/Switcher/Fields/Menu.php
Screenfeed: That's what I did at the beginning, but Fred and you said it was too complicated: choices in `layout` must be changed, `alignment` must be removed, things in `hide_if` must be modified each time it implies `layout`.

src/Switcher/Settings/Menu.php
Screenfeed: No, like `add_legacy_settings()`, the difference is:

```php
// Abstract_Settings_Legacy
$settings['dropdown'] = ( $settings['layout'] === 'select' );

// Settings\Menu
$settings['dropdown'] = ( $settings['layout'] === 'dropdown' );
```

js/src/nav-menu.js
Screenfeed: Done in [8ee39ad](https://github.com/polylang/polylang/pull/1903/commits/8ee39ad44b9729723274e265842b3fb70774328e).

js/src/nav-menu.js
Screenfeed: Done in [8ee39ad](https://github.com/polylang/polylang/pull/1903/commits/8ee39ad44b9729723274e265842b3fb70774328e).

js/src/nav-menu.js
Screenfeed: Done in [8ee39ad](https://github.com/polylang/polylang/pull/1903/commits/8ee39ad44b9729723274e265842b3fb70774328e).

src/Switcher/Element/Nav.php
Screenfeed: Done in [328a922](https://github.com/polylang/polylang/pull/1903/commits/328a9229943b33e4bf26ebd4bdbbb888cbc7b9bc). Honestly I don't `WP_HTML_Tag_Processor::next_tag()` + `WP_HTML_Tag_Processor::remove_attribute()` + `WP_HTML_Tag_Processor::get_updated_html()` is better than a single `preg_replace()`.

src/frontend/frontend-nav-menu.php
Screenfeed: Done in [e6a6baf](https://github.com/polylang/polylang/pull/1903/commits/e6a6bafaa1ad214821b1b641b716f8975ad4f869).

src/Switcher/Settings/Menu.php
Hug0-Drelon: Do you think it worth to have an abstract method such as `get_dropdown_legacy_fallback` and override it in child classes?

src/Switcher/Fields/Menu.php
Hug0-Drelon: Does this suggestion apply https://github.com/polylang/polylang/pull/1903#discussion_r3436600094?

src/Switcher/Element/Nav.php
Hug0-Drelon: Safer, faster and less error prone. And it's a matter of taste, but easier to understand.

js/src/nav-menu.js
Hug0-Drelon: Sorry, I should have tested the code.

I updated the suggestion, keeping the `reverse()`. The logic inside the loop is mostly the same, but the loop itself is easier and use object desctructuring to get key/value.

See the result:

Image

Hug0-Drelon (changes requested):

src/Switcher/Element/Nav.php
Screenfeed: Safer, less error prone: true in the absolute, false here because we control the source markup.
Faster: I tested with a 100.000 loop. `WP_HTML_Tag_Processor`: 3.452430s, `preg_replace`: 0.045659s.

Hug0-Drelon: Indeed, not faster. My bad, but takes less memory (at least for long strings).

src/Switcher/Settings/Menu.php
Screenfeed: I thought about it. It may be possible, not sure it's worthy.

For example [in `Abstract_Settings_Legacy::convert_from_legacy()`](https://github.com/polylang/polylang/blob/2f7014bed15b262c00334ba941af533cc8509122/src/Switcher/Settings/Abstract_Settings_Legacy.php#L128-L137).
- We'll be able to replace `'select'` by `$this->get_dropdown_legacy_fallback()`.
- But we also use `'vertical'` here, which is the default value, so we'll need to call `array_diff_key( get_class_vars( self::class ), array( 'increment' => 0 ) )` to retrieve it (because it's `'horizontal'` for the menu). It works but it gets a bit more complicated. No big deal though.

Same thing for `convert_to_legacy()`.

Then we can remove `Menu::convert_from_legacy()` and `Menu::convert_to_legacy()`, and we created 2 (simpler) methods (1 in `Abstract_Settings_Legacy` and 1 in `Menu`).

Then we'll (re-)create `get_defaults()`, or `get_default( $key )`, as a shorthand for `array_diff_key( get_class_vars( self::class ), array( 'increment' => 0 ) )`, because we use it 3 times.

We can't remove `validate()` though, because we need to know [the list of the possible values](https://github.com/polylang/polylang/blob/2f7014bed15b262c00334ba941af533cc8509122/src/Switcher/Settings/Settings.php#L210).

Pros:
- We remove `Menu::convert_from_legacy()` and `Menu::convert_to_legacy()`, and replace them with 2 simpler methods.

Cons:
- `Abstract_Settings_Legacy::convert_from_legacy()` and `Abstract_Settings_Legacy::convert_to_legacy()` become a bit more complicated.
- We force `Abstract_Settings_Legacy`'s child classes to have something related to the layout.

What do you think?

src/Switcher/Fields/Menu.php
Screenfeed: They are related.

In `Fields\Widget`:
```php
$settings['dropdown'] = isset( $settings['layout'] ) && 'select' === $settings['layout'] ? 1 : 0;
```

In `Fields\Menu`:
```php
$settings['dropdown'] = isset( $settings['layout'] ) && 'dropdown' === $settings['layout'] ? 1 : 0;
```

Maybe we can have a `get_dropdown_legacy_fallback()` here too. What do you think?

js/src/nav-menu.js
Screenfeed: Yes, works well (and I prefer this form). Thanks 👍
Done in [c9e51cf](https://github.com/polylang/polylang/pull/1903/commits/c9e51cf75a25394317fe41f346972a8b94953cb3).

src/Switcher/Settings/Menu.php
Hug0-Drelon: Okay! Let's see what @Chouby thinks of it.

Hug0-Drelon (changes requested):
LGTM, there is one last question though.

src/Switcher/Settings/Menu.php
Chouby: Let's keep as it is.

src/Switcher/Settings/Menu.php
Chouby: My idea was simply:
```php
		if ( isset( $settings['layout'] ) && ( 'select' === $settings['layout'] || 'dropdown' === $settings['layout']) ) {
			$args['dropdown'] = 1;
		}
```
Forget. As I foresee a drawback and no real advantage now. My initial goal was to have the legacy code in one place but it's not possible for `convert_from_legacy()` anyway.

PR #1906, opened 2026-06-15
Fix namespace for metas tests

Chouby:
The namespace wrongly refers to Polylang Pro. I also removed the unnecessary 'modules`.


Hug0-Drelon (changes requested):
I think we can remove `Integration` from namespaces too. This is relevant only in Polylang Pro (`unit` vs  `integration` PHPUnit suite).

Hug0-Drelon (approved):

PR #1909, opened 2026-06-16
Fix conflict between language code and sitemap URL

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3017.

## What

Polylang makes the sitemap multilingual by duplicating URLs for each language. However, when a post type or term name ends with a language code (ex: `broken` ends with `en`), this language code is stripped off the post type name, leading to an incorrect URL.

## Why

Each URL is created by `WP_Sitemaps_Provider::get_sitemap_url()`, and it accepts only 2 parameters: `$name` (a post type name, a taxonomy name) and `$page` (an integer), while Polylang needs a language to be able to add it to the URL. The solution in use is to suffix `$name` with a language code, like `page-en`: this is done by `get_sitemap_data()` in `get_sitemap_type_data()`. Then [`get_sitemap_url()` can retrieve the language code from `$name`](https://github.com/polylang/polylang/commit/596fef929b6ef487e92f33110fafa41ff21f8df2#diff-9a35433bdf5e91acdddfbd216e76b49ea291fcc552da29c16a7ba51a491b6412R206). The issue is easy to guess: create a custom post type named `foobar-fr` and tell Polylang not to translate it (otherwize Polylang will add another language code to it): when receiving `foobar-fr` in `get_sitemap_url()`, Polylang will believe it's the post type `foobar` in `fr`. Worse, the regex patterns allows `foobarfr` to match.

## How

1. Instead of using `{post_type}-{lang_code}`, use a more unique separator: `---pll-sep---`.
2. Make the regex pattern more robust.

Notes:
1. Use 1 `preg_match()` to match everything instead of: 1 `preg_match()` to match the language code, and 1 `preg_replace()` to replace it.
2. The new pattern allows `SUBTYPE` to be empty (`(?<SUBTYPE>.*)`) for the same reason that the previous patterns allowed to match without the `-` character:
  - `post`, `page`, and any CPT are post types, they are some sort of "sub-types".
  - `category`, `post_tag`, etc are taxonomies, they are also "sub-types".
  - But what about users? They don't have sub-types, that's why the sub-type is allowed to be empty. That's also why `get_sitemap_type_data()` loops through the languages to fill the output when `$object_subtypes` is empty.

### DIFF
--- a/src/modules/sitemaps/multilingual-sitemaps-provider.php
+++ b/src/modules/sitemaps/multilingual-sitemaps-provider.php
@@ -18,7 +18,7 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
 	 * Pattern to match a name with language.
 	 *  `.*` in `(?<SUBTYPE>.*)` because users don't have sub-types. See `get_sitemap_data()`.
 	 */
-	private const PATTERN = '#^(?<SUBTYPE>.*)---pll-sep---(?<LANG>.+)$#';
+	private const PATTERN = '#^(?<SUBTYPE>.*)' . self::SEPARATOR . '(?<LANG>.+)$#';
 
 	/**
 	 * The decorated sitemaps provider.
--- a/src/modules/sitemaps/multilingual-sitemaps-provider.php
+++ b/src/modules/sitemaps/multilingual-sitemaps-provider.php
@@ -9,7 +9,7 @@
  * @since 2.8
  */
 class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
-	public const PLL_SEPARATOR = '---pll-sep---';
+	public const SEPARATOR = '---pll-sep---';
 
 	/**
 	 * The decorated sitemaps provider.
@@ -68,7 +68,7 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
 		$this->links_model = &$links_model;
 		$this->model       = &$links_model->model;
 
-		self::$pattern = sprintf( '#^(?<SUBTYPE>.*)%s(?<LANG>.+)$#', self::PLL_SEPARATOR ); // `.*` because users don't have sub-types. See `get_sitemap_data()`.
+		self::$pattern = sprintf( '#^(?<SUBTYPE>.*)%s(?<LANG>.+)$#', self::SEPARATOR ); // `.*` because users don't have sub-types. See `get_sitemap_data()`.
 	}
 
 	/**
@@ -135,7 +135,7 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
 		self::$filter_lang = $lang;
 
 		$return = array(
-			'name'  => sprintf( "{$object_subtype_name}%s{$lang}", self::PLL_SEPARATOR ),
+			'name'  => sprintf( "{$object_subtype_name}%s{$lang}", self::SEPARATOR ),
 			'pages' => $this->get_max_num_pages( $object_subtype_name ),
 		);
 
--- a/src/modules/sitemaps/multilingual-sitemaps-provider.php
+++ b/src/modules/sitemaps/multilingual-sitemaps-provider.php
@@ -48,6 +48,11 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
 	private static $filter_lang = '';
 
 	/**
+	 * @var string
+	 */
+	private static string $pattern;
+
+	/**
 	 * Constructor.
 	 *
 	 * @since 2.8
@@ -62,6 +67,8 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
 		$this->provider    = $provider;
 		$this->links_model = &$links_model;
 		$this->model       = &$links_model->model;
+
+		self::$pattern = sprintf( '#^(?<SUBTYPE>.*)%s(?<LANG>.+)$#', self::PLL_SEPARATOR ); // `.*` because users don't have sub-types. See `get_sitemap_data()`.
 	}
 
 	/**
@@ -196,9 +203,7 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
 	 */
 	public function get_sitemap_url( $name, $page ) {
 		// Check if a language was added in `$name`.
-		$pattern = sprintf( '#^(?<SUBTYPE>.*)%s(?<LANG>.+)$#', self::PLL_SEPARATOR ); // `.*` because users don't have sub-types. See `get_sitemap_data()`.
-
-		if ( preg_match( $pattern, $name, $matches ) ) {
+		if ( preg_match( self::$pattern, $name, $matches ) ) {
 			$lang = $this->model->get_language( $matches['LANG'] );
 
 			if ( ! empty( $lang ) ) {
--- a/src/modules/sitemaps/multilingual-sitemaps-provider.php
+++ b/src/modules/sitemaps/multilingual-sitemaps-provider.php
@@ -9,7 +9,16 @@
  * @since 2.8
  */
 class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
+	/**
+	 * Separator between name and language slug.
+	 */
 	public const SEPARATOR = '---pll-sep---';
+
+	/**
+	 * Pattern to match a name with language.
+	 *  `.*` in `(?<SUBTYPE>.*)` because users don't have sub-types. See `get_sitemap_data()`.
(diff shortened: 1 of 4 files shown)


src/modules/sitemaps/multilingual-sitemaps-provider.php
```diff
@@ -9,6 +9,8 @@
  * @since 2.8
  */
 class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
+	public const PLL_SEPARATOR = '---pll-sep---';
```
Hug0-Drelon:
```suggestion
	public const SEPARATOR = '---pll-sep---';
```
No need to prefix a class constant.

src/modules/sitemaps/multilingual-sitemaps-provider.php
```diff
@@ -183,22 +195,24 @@ public function get_sitemap_type_data() {
 	 * @return string The composed URL for a sitemap entry.
 	 */
 	public function get_sitemap_url( $name, $page ) {
-		// Check if a language was added in $name.
-		$pattern = '#(' . implode( '|', $this->model->get_languages_list( array( 'fields' => 'slug' ) ) ) . ')$#';
+		// Check if a language was added in `$name`.
+		$pattern = sprintf( '#^(?<SUBTYPE>.*)%s(?<LANG>.+)$#', self::PLL_SEPARATOR ); // `.*` because users don't have sub-types. See `get_sitemap_data()`.
```
Hug0-Drelon: Since you removed the dynamic part of the pattern (language slugs), I'd suggest to create a dedicated constant for the pattern. No need to compute it at runtime anymore.

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -23,23 +21,29 @@ protected function init( $sitemap_class = 'PLL_Sitemaps' ) {
...
-		register_post_type( 'cpt', array( 'public' => true ) ); // *Untranslated* custom post type.
-		register_taxonomy( 'tax', 'cpt' ); // *Untranslated* custom tax.
-
-		$links_model = self::$model->get_links_model();
-		if ( method_exists( $links_model, 'init' ) ) {
-			$links_model->init();
-		}
+		// Use a custom post type and a taxonomy whose slug ends with a language slug. See https://github.com/polylang/polylang-pro/issues/3017.
+		register_post_type( 'cpten', array( 'public' => true ) ); // *Untranslated* custom post type.
+		register_taxonomy( 'taxen', 'cpten' ); // *Untranslated* custom tax.
```
Hug0-Drelon:
```suggestion
		register_taxonomy( 'tax-en', 'cpten' ); // *Untranslated* custom tax.
```
Suggesting more data diversity.

Hug0-Drelon (changes requested):

src/modules/sitemaps/multilingual-sitemaps-provider.php
Screenfeed: I prefixed it because we extend a WP class. So, even if it will probably never happen, `WP_Sitemaps_Provider` could get a `SEPARATOR` constant some day, and then we'll be f*cked 😅

src/modules/sitemaps/multilingual-sitemaps-provider.php
Screenfeed: Done in [9fb0432](https://github.com/polylang/polylang/pull/1909/commits/9fb0432da1b7baef937c3865776a38d6b1cc9cb4).

src/modules/sitemaps/multilingual-sitemaps-provider.php
Screenfeed: OK, done in [0acd815](https://github.com/polylang/polylang/pull/1909/commits/0acd8158fc156a6e43530308dbfece3aef8521a3).

tests/phpunit/tests/test-sitemaps.php
Screenfeed: Good idea. Done in [4f70d77](https://github.com/polylang/polylang/pull/1909/commits/4f70d778fb0b21851ff9620a15c6035bf2b338ac).

src/modules/sitemaps/multilingual-sitemaps-provider.php
```diff
@@ -55,12 +61,14 @@ class PLL_Multilingual_Sitemaps_Provider extends WP_Sitemaps_Provider {
...
+		$this->name        = $provider->name;
 		$this->object_type = $provider->object_type;
 
-		$this->provider = $provider;
+		$this->provider    = $provider;
 		$this->links_model = &$links_model;
-		$this->model = &$links_model->model;
+		$this->model       = &$links_model->model;
+
+		self::$pattern = sprintf( '#^(?<SUBTYPE>.*)%s(?<LANG>.+)$#', self::SEPARATOR ); // `.*` because users don't have sub-types. See `get_sitemap_data()`.
```
Hug0-Drelon: Sorry, I wasn't clear enough. I meant that the pattern should be stored in a class constant (like `SEPARATOR`).
There is no dynamic data in the pattern anymore, and thus calling `sprintf` is useless IMHO.

Hug0-Drelon (changes requested):

src/modules/sitemaps/multilingual-sitemaps-provider.php
Screenfeed: I did that to avoid repeating the separator. Done in [0607071](https://github.com/polylang/polylang/pull/1909/commits/0607071c4461013c0859f39082aa7cfeb9b8c705).

Hug0-Drelon (approved):

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -23,23 +21,29 @@ protected function init( $sitemap_class = 'PLL_Sitemaps' ) {
...
+		$options = array_merge(
+			array(
+				'default_lang' => 'en',
+				'hide_default' => false,
+				'force_lang'   => 1,
+			),
+			$options
+		);
+		$this->pll_env = ( new PLL_Context_Frontend( array( 'options' => $options ) ) )->get();
+		$this->pll_env->links_model->init();
```
Chouby: Isn't already done  by `PLL_Context_Frontend`?

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -23,23 +21,29 @@ protected function init( $sitemap_class = 'PLL_Sitemaps' ) {
...
 		$this->pll_env->sitemaps = new $sitemap_class( $this->pll_env );
 		$this->pll_env->sitemaps->init();
 
 		wp_sitemaps_get_server(); // Allows to register sitemaps rewrite rules.
 
 		// Refresh languages.
-		self::$model->clean_languages_cache();
-		self::$model->get_languages_list();
+		$this->pll_env->model->clean_languages_cache();
+		$this->pll_env->model->get_languages_list();
```
Chouby: I know it was there but is it still necessary?

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -79,17 +83,19 @@ public function test_sitemaps_page() {
...
-		self::$model->post->set_language( $en, 'en' );
+		$post_en = self::factory()->post->create( array( 'post_author' => 1, 'lang' => 'en' ) );
+		$cat_en  = self::factory()->category->create( array( 'lang' => 'en' ) );
+		wp_set_post_terms( $post_en, array( $cat_en ), 'category' );
 
-		$fr = self::factory()->post->create( array( 'post_author' => 1 ) );
-		self::$model->post->set_language( $fr, 'fr' );
+		$post_fr = self::factory()->post->create( array( 'post_author' => 1, 'lang' => 'fr' ) );
+		$cat_fr  = self::factory()->category->create( array( 'lang' => 'fr' ) );
+		wp_set_post_terms( $post_fr, array( $cat_fr ), 'category' );
```
Chouby, line 93: Strange the category sitemap worked before.

Chouby (changes requested):

tests/phpunit/tests/test-sitemaps.php
Screenfeed: Removed in [41777b9](https://github.com/polylang/polylang/pull/1909/commits/41777b9c51baf95949430233c19c9504619ab15e).

tests/phpunit/tests/test-sitemaps.php
Screenfeed: Removed in [41777b9](https://github.com/polylang/polylang/pull/1909/commits/41777b9c51baf95949430233c19c9504619ab15e).

tests/phpunit/tests/test-sitemaps.php
Screenfeed: I supposed there was a remaining category somewhere from another test.

PR #1910, opened 2026-06-16
[Language switcher rework] Add tests for the classic widget

Screenfeed:
This adds some tests for the new classic widget.


Screenfeed:
I won't lose a single second on this.

PR #1914, opened 2026-06-17
[Language switcher rework] Use the new switcher in `pll_the_languages()`

Screenfeed:
All is in the title.

> [!IMPORTANT]
> `PLL_Walker_Dropdown` [adds extra data](https://github.com/polylang/polylang/blob/4b2088a73671359cb677a540f9638921621ee07e/src/walker-dropdown.php#L38-L52) to the switcher (the `value` attribute can contain something else than a URL, and there's a `data-lang` attribute. `pll_the_languages()` doesn't add them anymore for simplicity.


src/api.php
```diff
@@ -5,28 +5,47 @@
...
+ *     @type string   $show_labels            Display the labels. Possible values are an empty string (no labels),
+ *                                            `names` (language names), `codes` (languages codes). Default is `names`.
+ *     @type bool     $hide_if_empty          Hide languages that don't have any posts. Default is `true`.
+ *     @type bool     $hide_if_no_translation Hide languages that don't have translations. Default is `false`.
+ *     @type bool     $hide_current           Hide the current language. Default is `false`.
+ *     @type bool     $force_home             Force elements to link to the home pages instead of the translations.
+ *                                            Default is `false`.
+ *     @type int      $post_id                Build the links according to the translations of the given post ID.
+ *                                            Default is `0`.
+ *     @type bool     $preserve_spacing       Preserve or discard white space characters between tags.
```
Hug0-Drelon: I wouldn't document this setting. This is a legacy one and I don't think it should be used anymore.

src/api.php
```diff
@@ -35,8 +54,71 @@ function pll_the_languages( $args = array() ) {
 		return empty( $args['raw'] ) ? '' : array();
 	}
 
-	$switcher = new PLL_Switcher();
-	return $switcher->the_languages( PLL()->links, $args );
+	if ( PLL()->links instanceof PLL_Admin_Links ) {
+		// Backward compatibility. See https://github.com/polylang/polylang/blob/bb2f55cb322f657a73d006d5c064c74b98c65a55/src/switcher.php#L237-L240.
+		$args['hide_if_no_translation'] = false;
+	}
```
Hug0-Drelon: :suspect:

src/api.php
```diff
@@ -35,8 +54,71 @@ function pll_the_languages( $args = array() ) {
...
+
+		return $elements;
+	}
+
+	if ( $return_html ) {
+		return $switcher->get();
+	}
+
+	$switcher->print();
+	return '';
```
Hug0-Drelon: I'd update the returned type with `void` in such case.

Hug0-Drelon (changes requested):

src/api.php
Screenfeed: Previously it was printing the switcher AND was returning it too 😅
So I chose to not return it in that case, but I didn't want to change the type so I chose to return an empty string. We'll discuss this during the daily.

src/api.php
Screenfeed: I think I see what you mean by "it's legacy", we've added it only for backward compatibility. Let's bring this up during the daily too and decide all together.

src/api.php
Screenfeed: Done in [62b3a29](https://github.com/polylang/polylang/pull/1914/commits/62b3a29ae3ae4382cc6658934b232362f80784fe).

src/api.php
Screenfeed: Done in [62b3a29](https://github.com/polylang/polylang/pull/1914/commits/62b3a29ae3ae4382cc6658934b232362f80784fe).

tests/phpunit/tests/test-pll-the-languages.php
```diff
@@ -0,0 +1,389 @@
...
+			)
+		);
+		$this->pll_model = new PLL_Model( $options );
+		$links_model     = $this->pll_model->get_links_model();
+		$this->frontend  = new PLL_Frontend( $links_model );
+		$this->frontend->init();
+
+		// De-activate cache for links
+		$this->frontend->links->cache = $this->getMockBuilder( 'PLL_Cache' )->getMock();
+		$this->frontend->links->cache->method( 'get' )->willReturn( false );
```
Hug0-Drelon: Could we flush instead of deactivating it? Otherwise I fear that the integration test doesn't match production context.

Screenfeed: OK why not. Done in [4c155ef](https://github.com/polylang/polylang/pull/1914/commits/4c155ef7d5c16264c9285e992a7b9206e9cb88cf).

Hug0-Drelon (approved):

phpcs.xml.dist
```diff
@@ -77,6 +77,14 @@
...
 	</rule>
 
+	<rule ref="WordPress.NamingConventions.ValidVariableName">
+		<properties>
+			<property name="allowed_custom_properties" type="array">
+				<element value="nodeValue"/>
+			</property>
+		</properties>
+	</rule>
+
```
Chouby, line 87: Why do we suddenly need this? `nodeValue` is already used in existing tests: https://github.com/search?q=repo%3Apolylang%2Fpolylang%20nodeValue&type=code

src/Switcher/Element/Nav.php
```diff
@@ -70,18 +70,17 @@ public function get_link(): string {
...
+		if ( ! empty( $this->label ) ) {
+			$label .= sprintf( '<span class="pll-switcher-label">%s</span>', esc_html( $this->label ) );
 		}
 
-		return sprintf(
-			'<span class="pll-switcher-flag">%s</span><span>%s</span>',
-			(string) preg_replace( '/ style="[^"]*"/', '', $this->flag ),
-			esc_html( $this->label )
-		);
+		return $label;
```
Chouby, line 83: Same as https://github.com/polylang/polylang/pull/1903#pullrequestreview-4513542989
Moreover if you modify it in 2 PRs, this will conflict.

src/api.php
```diff
@@ -5,38 +5,118 @@
...
 		return empty( $args['raw'] ) ? '' : array();
 	}
 
-	$switcher = new PLL_Switcher();
-	return $switcher->the_languages( PLL()->links, $args );
+	if ( PLL()->links instanceof PLL_Admin_Links ) {
+		// Backward compatibility. See https://github.com/polylang/polylang/blob/bb2f55cb322f657a73d006d5c064c74b98c65a55/src/switcher.php#L237-L240.
+		$args['hide_if_no_translation'] = false;
+	}
+
```
Chouby: This shouldn't be useful anymore. This function was used in back when the blocks were rendered server-side. It's not the case for a while now. 
```suggestion
```

src/api.php
```diff
@@ -5,38 +5,118 @@
...
+	$return_raw  = ! empty( $args['raw'] );
+	$return_html = isset( $args['echo'] ) && empty( $args['echo'] );
+	unset( $args['raw'], $args['echo'] );
```
Chouby: In my opinion, it's useless to remove these params. They just wouldn't be interpreted.

src/api.php
```diff
@@ -5,38 +5,118 @@
...
+				esc_html__( "See %s's documentation.", 'polylang' ),
+				'pll_the_languages()'
+			)
+		);
+	}
+
+	$settings = new Settings( $args );
+
+	// Backward compatibility. See https://github.com/polylang/polylang/blob/bb2f55cb322f657a73d006d5c064c74b98c65a55/src/switcher.php#L141-L143.
+	$settings->hide_current = ! empty( $args['hide_current'] ) && ! ( ! empty( $args['dropdown'] ) && ! $return_raw );
```
Chouby: Isn't this the exact opposite of what we want now https://github.com/polylang/polylang-pro/issues/2958 ?

src/api.php
```diff
@@ -5,38 +5,118 @@
...
+
+	$switcher = new Switcher( $settings, PLL()->links );
+
+	if ( $return_raw ) {
+		$languages       = PLL()->links->model->languages->get_list();
+		$keyed_languages = array_combine( array_column( $languages, 'slug' ), $languages );
+		$elements        = array();
```
Chouby: This seems an over-complicated way to get the language (in fact the flag if I well understand). Since I guess you want to keep `$language` private in `Abstract_Element`, I propose to add a method in this class either to get the language or `get_display_flag_url()`.

src/api.php
```diff
@@ -5,38 +5,118 @@
...
+				'locale'         => $element->locale,
+				'is_rtl'         => 'rtl' === $element->direction,
+				'name'           => 'codes' === $settings->show_labels ? $element->slug : $language->name,
+				'url'            => $element->url,
+				'flag'           => ! empty( $settings->show_flags ) ? $element->flag : $language->get_display_flag_url(),
+				'current_lang'   => $element->is_current,
+				'no_translation' => ! $element->has_translations,
+				'classes'        => $element->item_classes,
+				'link_classes'   => $element->link_classes,
+			);
```
Chouby: I would simplify with `get_object_vars()` to return all properties then only add properties necessary for backward compat.

tests/phpunit/tests/test-pll-the-languages.php
```diff
@@ -0,0 +1,387 @@
...
+		$this->assertNotEmpty( $option, 'There should be an option tag.' );
+		$this->assertSame( 'selected', $option->getAttribute( 'selected' ) );
+		$this->assertNotEmpty( $xpath->query( '//select/option[.="Français"]' )->length );
+		$this->assertEmpty( $xpath->query( '//script' )->length ); // No `script` tag anymore.
+		$lang_attributes = $xpath->query( '//select/option/@lang' );
+		$this->assertSame( 'en-US', $lang_attributes->item( 0 )->value );
+		$this->assertSame( 'fr-FR', $lang_attributes->item( 1 )->value );
+	}
+
+	public function test_with_hide_if_no_translation_option_in_admin_context() {
```
Chouby: If I'm not wrong and we don't use the language switcher in admin context, then we should remove this test.

Chouby (changes requested):

phpcs.xml.dist
Screenfeed: Because I wrote:
```php
$span = $xpath->query( '//li/a[@lang="en-US"]/span[@class="pll-switcher-label"]' )->item( 0 );
$this->assertNotEmpty( $span );
$this->assertSame( 'English', $span->nodeValue ); // $span->nodeValue triggers PHPCS.
```

If I write like this, it will pass:
```php
$spans = $xpath->query( '//li/a[@lang="en-US"]/span[@class="pll-switcher-label"]' );
$this->assertNotEmpty( $spans->item( 0 ) );
$this->assertSame( 'English', $spans->item( 0 )->nodeValue ); // $spans->item( 0 )->nodeValue doesn't.
```

And we used the second form everywhere.
Do you want me to use the 2nd notation here?

src/Switcher/Element/Nav.php
Screenfeed: It was need at the time I wrote this, the change in the other PR happened after. Fixed after rebasing.

tests/phpunit/tests/test-pll-the-languages.php
Screenfeed: Indeed. Removed in [64be039](https://github.com/polylang/polylang/pull/1914/commits/64be039f4fa176e48d2374c419eeb10c3cd58ff5).

src/api.php
Screenfeed: Removed in [64be039](https://github.com/polylang/polylang/pull/1914/commits/64be039f4fa176e48d2374c419eeb10c3cd58ff5).

src/api.php
Screenfeed: Done in [28fd7a4](https://github.com/polylang/polylang/pull/1914/commits/28fd7a4fbd596d7d932bf98e3cfcfa45d989033f).

src/api.php
Screenfeed: Done in [a2c8dfd](https://github.com/polylang/polylang/pull/1914/commits/a2c8dfd01c6a6833e92a38802ec34c16868b0b63).

src/api.php
Screenfeed: Done in [0f85108](https://github.com/polylang/polylang/pull/1914/commits/0f851080c64c74ee54331ca1ced73c23df38115f).

src/api.php
Screenfeed: It seems so. But this issue should be handled in another PR IMHO, because there are other things to change: currently the "Hide current" field [is hidden](https://github.com/polylang/polylang/blob/9440ad96c6ec716e59d475056ecc21a3868ffb05/src/Switcher/Fields/Widget.php#L81) when the `select` layout is chosen, and [the value is overwiten](https://github.com/polylang/polylang/blob/9440ad96c6ec716e59d475056ecc21a3868ffb05/src/Switcher/Settings/Settings.php#L254). Now I even wonder if this line is necessary 🤔

Chouby: OK for a separate PR.

PR #1915, opened 2026-06-19
Handle `polylang/language-swticher` backward compatibility.

Hug0-Drelon:
## What?
Depends on https://github.com/polylang/polylang/pull/1898

Implement migration process in the editor.
Alternatively, adapt PHPUnit tests with legacy attributes that ensure backward compatibility.

## Why?
Base [PR](https://github.com/polylang/polylang/pull/1898) added new attributes that must be converted when the user saves a post containing an old `polylang/language-swticher`.

## How?
- Convert `"dropdown": true` to `"layout": "select"`
- Convert `"show_names": true` to `"show_labels": "names"`
- Toggle `"show_flags"` if needed
- Implement `isEligible` and return `true` when either `dropdown` or `show_flags` legacy attribute is present
- Adapt `tests/phpunit/tests/test-switcher-block.php` to use frontend render instead of REST render
- Change some assertions in `tests/phpunit/tests/test-switcher-block.php` to match new markup

> [!NOTE]
> GitHub Actions workflows run temporarily, revert c5870dc6 before merge.


tests/phpunit/tests/test-switcher-block.php
```diff
@@ -4,32 +4,47 @@
...
+	 */
 	protected $structure = '/%postname%/';
 
 	/**
-	 * @param WP_UnitTest_Factory $factory
+	 * @var PLL_Frontend
 	 */
-	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
-		parent::wpSetUpBeforeClass( $factory );
+	protected $pll_env;
```
Screenfeed, line 15: I don't think this is necessary, this property is already declared [in `PLL_UnitTestCase_Trait`](https://github.com/polylang/polylang/blob/726c5186b15dacf28e04cd9f3cf5259275775b0f/tests/phpunit/includes/testcase-trait.php#L34-L37).

tests/phpunit/tests/test-switcher-block.php
```diff
@@ -101,25 +116,24 @@ public function test_switcher_block( $block_name ) {
...
+			),
+			array(
+				'lang'         => 'fr',
+				'post_content' => '<!-- wp:' . $block_name . ' /-->',
+			)
+		);
+		$en = $posts['en'];
+		$fr = $posts['fr'];
 
 		self::$model->clean_languages_cache(); // To get an exact count of posts for the languages.
```
Screenfeed: Maybe not necessary anymore, since you don't use the static model.

tests/phpunit/tests/test-switcher-block.php
```diff
@@ -131,17 +145,17 @@ public function test_switcher_block_in_get_posts( $block_name ) {
...
-		$request = new WP_REST_Request( 'GET', '/wp/v2/posts/' . $en );
-		$response = $this->server->dispatch( $request );
-		$this->assertEquals( 200, $response->get_status() );
-		$en_post = $response->get_data();
-		$this->assertNotEmpty( $en_post['content'] );
-		$doc = new DomDocument();
-		$doc->loadHTML( $en_post['content']['rendered'] );
+		$rendered = $this->render_post_blocks( $en );
+		$doc      = new DomDocument();
+		$doc->loadHTML( $rendered );
```
Screenfeed, line 157: No need to silence errors here? Strange 🤔

Edit: Oh, right, no `add_theme_support()`.

Note: I made [this](https://github.com/polylang/polylang/blob/726c5186b15dacf28e04cd9f3cf5259275775b0f/tests/phpunit/tests/test-pll-the-languages.php#L388-L393) recently, maybe we could move it to a trait and reuse it everywhere (not necessarily in this PR).

tests/phpunit/tests/test-switcher-block.php
```diff
@@ -183,60 +193,78 @@ public function test_custom_class_list_switcher_block( $block_name, $class_query
...
+
+		$this->assertTrue( $processor->next_tag(), 'The rendered output should start with a tag.' );
+		$this->assertSame( 'NAV', $processor->get_tag(), 'The first tag should be a nav element.' );
+		$this->assertFalse( $processor->is_tag_closer(), 'The first tag should be an opening nav tag.' );
+
+		$id = $processor->get_attribute( 'id' );
+		$this->assertIsString( $id, 'The nav tag should have an id attribute.' );
+		$this->assertStringStartsWith( 'pll-switcher-', $id, 'The nav id should start with pll-switcher-.' );
+
+		$this->assertEqualSets(
```
Screenfeed:
```suggestion
		$this->assertSameSets(
```

tests/phpunit/tests/test-switcher-block.php
```diff
@@ -183,60 +193,78 @@ public function test_custom_class_list_switcher_block( $block_name, $class_query
...
-		$request->set_param( 'attributes', array( 'dropdown' => 1 ) );
-		$response = $this->server->dispatch( $request );
-		$switcher = $response->get_data();
+		$rendered = $this->render_switcher_block(
+			'polylang/language-switcher',
+			array( 'dropdown' => 1 )
+		);
 
-		$this->assertStringNotContainsString( '<nav>', $switcher['rendered'] );
+		$this->assertStringNotContainsString( '<nav>', $rendered );
```
Screenfeed: I think this will also be true if `'dropdown' => 0` because of the final `>`.


```suggestion
		$this->assertStringNotContainsString( '<nav ', $rendered );
```

Screenfeed (changes requested):

tests/phpunit/tests/test-switcher-block.php
Hug0-Drelon: TBH, I don't like `$frontend` property name. It is not explicit at all. We could have `$pll_frontend` instead. But in the meantime, I suggest to keep `$pll_env` set with `PLL_Frontend` I you will.

tests/phpunit/tests/test-switcher-block.php
Hug0-Drelon: Maybe later then.

tests/phpunit/tests/test-switcher-block.php
Hug0-Drelon: Nice catch! Applied in a79b802c.

tests/phpunit/tests/test-switcher-block.php
Hug0-Drelon: Applied in a79b802c.

tests/phpunit/tests/test-switcher-block.php
Hug0-Drelon: Indeed, removed in a79b802c.

tests/phpunit/tests/test-switcher-block.php
Screenfeed: Oh my bad, I didn't highlighted the right property. I meant [this one](https://github.com/polylang/polylang/blob/726c5186b15dacf28e04cd9f3cf5259275775b0f/tests/phpunit/includes/testcase-trait.php#L44-L47).

Hug0-Drelon: Okay, then I don't see the problem 🤔 

I re-define `$pll_env` with `PLL_Frontend` (sub class of `PLL_Base`) for stricter type and IDE discovering features.

Screenfeed (approved):

PR #1918, opened 2026-06-23
Bring back missing language names in Screen Options

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3032.

Regression introduced by https://github.com/polylang/polylang/pull/1857/changes#diff-d6f3bf462714cc89a7ed43ae3abadb594e878c5900a7c3f25ef920f327e0b410L88.

## Why

`WP_Screen::render_list_table_columns_preferences()` applies `wp_strip_all_tags()` in the screen options, which removes our flags and leave an empty label.

## How

Add the language name next to the flag, but as a hidden text (wrapped in `<span class="screen-reader-text">`).

1. `WP_Screen::render_list_table_columns_preferences()` will remove the flag, and reveal the language name (that looses its `<span>`).
2. `WP_List_Table::print_column_headers()` doesn't apply any escaping function in the column headers: the flag will be displayed, and the language name will still be present but hidden.

### DIFF
--- a/src/admin/admin-filters-columns.php
+++ b/src/admin/admin-filters-columns.php
@@ -90,9 +90,13 @@ class PLL_Admin_Filters_Columns {
 			 * - `WP_Screen::render_list_table_columns_preferences()` applies `wp_strip_all_tags()` in the screen options:
 			 *   this will remove the flag, and reveal the language name (that looses its `<span>`).
 			 * - `WP_List_Table::print_column_headers()` doesn't apply any escaping function in the column headers:
-			 *   the flag will be displayed, and the language name will still be present but hidden.
+			 *   the flag will be displayed, and the language name will still be present but visually hidden.
 			 */
-			$columns[ 'language_' . $language->slug ] = sprintf( '%s<span class="hidden">%s</span>', $language->get_admin_flag(), esc_html( $language->name ) );
+			$columns[ "language_{$language->slug}" ] = sprintf(
+				'%s<span class="screen-reader-text">%s</span>',
+				$language->get_admin_flag( 'aria-hidden' ),
+				esc_html( $language->name )
+			);
 		}
 
 		return isset( $end ) ? array_merge( $columns, $end ) : $columns;
--- a/src/admin/admin-filters-columns.php
+++ b/src/admin/admin-filters-columns.php
@@ -85,7 +85,14 @@ class PLL_Admin_Filters_Columns {
 		}
 
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[ 'language_' . $language->slug ] = $language->get_admin_flag();
+			/*
+			 * Hack:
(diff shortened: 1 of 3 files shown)


src/admin/admin-filters-columns.php
```diff
@@ -85,7 +85,14 @@ protected function add_column( array $columns, string $before ): array {
...
 		foreach ( $this->model->get_languages_list() as $language ) {
-			$columns[ 'language_' . $language->slug ] = $language->get_admin_flag();
+			/*
+			 * Hack:
+			 * - `WP_Screen::render_list_table_columns_preferences()` applies `wp_strip_all_tags()` in the screen options:
+			 *   this will remove the flag, and reveal the language name (that looses its `<span>`).
+			 * - `WP_List_Table::print_column_headers()` doesn't apply any escaping function in the column headers:
+			 *   the flag will be displayed, and the language name will still be present but hidden.
+			 */
+			$columns[ 'language_' . $language->slug ] = sprintf( '%s<span class="hidden">%s</span>', $language->get_admin_flag(), esc_html( $language->name ) );
```
Chouby: With the previous code in v3.8.x in mind:
```php
$columns[ 'language_' . $language->slug ] = $this->links->get_flag_html( $language ) . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
```
Would this work? And what do you think of it?
```suggestion
			$columns[ 'language_' . $language->slug ] = $language->get_admin_flag( 'aria-hidden' ) . '<span class="screen-reader-text">' . esc_html( $language->name ) . '</span>';
```

Chouby (changes requested):

src/admin/admin-filters-columns.php
Screenfeed: @Chouby Same result. Which one do you prefer?

Screenfeed: Done in [831d669](https://github.com/polylang/polylang/pull/1918/commits/831d669902f37c8fcb19d84c13152baecba26a23).

Marianne380 (approved):

tests/phpunit/includes/testcase.php
```diff
@@ -15,6 +15,17 @@ function ( $value ) {
 			}
 		);
 
+		Functions\when( 'get_column_headers' )->alias(
```
Marianne380, line 18: Why not only on `test-columns.php` ?

Marianne380 (changes requested):

tests/phpunit/includes/testcase.php
Chouby: Whatever the test calling this function, we may have issues with the static variable which could have values coming from a previous test. (I'm not sure it"s used only in test-columns.php and even less sure that we won't call it in future tests)

Marianne380 (approved):

PR #1920, opened 2026-06-25
Fix a typo in `pll_get_requested_url` doc

sebastienserre:
fix a typo in the doc description

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -26,7 +26,7 @@ function pll_is_cache_active() {
 }
 
 /**
- * Get the the current requested url
+ * Get the current requested url.
  *
  * @since 2.6
  *


src/functions.php
```diff
@@ -26,7 +26,7 @@ function pll_is_cache_active() {
 }
 
 /**
- * Get the the current requested url
+ * Get the current requested url.
```
Hug0-Drelon, line 29:
```suggestion
 * Gets the current requested url.
```
Doc standards force to conjugate.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

PR #1922, opened 2026-07-01
Refactors the Navigation Language Switcher block to match the standard Language Switcher block.

Hug0-Drelon:
## What

Refactors the **Navigation Language Switcher** block to align with the standard **Language Switcher** block: shared editor components, v2 attributes, new Switcher PHP API, split CSS, and consistent flag rendering in editor and frontend.

## Why

The navigation block duplicated logic from the standard switcher, still used legacy attributes (`dropdown`, `show_names`), no flag styling, and no v1→v2 migration path.

## How

**JS**
- Extract shared UI into `js/src/blocks/components/` (`SwitcherControls`, `SwitcherLanguageLink`, utils).
- Give each block its own `RenderedSwitcher`; merge nav container + UI into one preview component.
- Add `deprecated.js` for v1→v2 migration (`dropdown` → `layout`, `show_names` → `show_labels`).
- Add `useMemoizedFlag` hook to avoid reprocessing flag HTML on every render.
- Remove legacy files (`language-switcher-edit.js`, per-block duplicate components).

**PHP**
- Render navigation switcher via `Switcher` + `Settings\Menu` + `get_elements()`, delegating to core navigation blocks.
- Use `pll_label_markup` and shared `apply_flag_styles_to_markup()` from `Abstract_Block`.
- Align `block.json` attributes with the standard switcher (default layout: `horizontal`).

**CSS**
- Introduce shared styles in `css/src/lib/` (`switcher-flags.css`, `navigation-language-switcher-alignment.css`).
- Split frontend (`navigation-language-switcher.css`) and editor (`navigation-language-switcher-editor.css`) entry points.
- Fix flag-only vertical alignment (editor + frontend) and editor spacing via `--wp--style--block-gap`.

**Tests**
- Update PHPUnit fixtures for new markup (`pll-switcher-flag` / `pll-switcher-label`).
- Update E2E selectors for v2 inspector controls.

## Test plan

- [ ] Navigation editor: horizontal + dropdown layouts, flags only / names only / both
- [ ] Frontend: same layouts match editor (alignment, spacing, aspect ratio, border radius)
- [ ] v1 blocks migrate correctly via `deprecated.js`


tests/phpunit/tests/test-switcher-block-frontend.php
```diff
@@ -96,7 +96,9 @@ public function test_render_polylang_navigation_switcher_block( $options, $conte
 	public function switcher_options_provider() {
 		global $wp_version;
 
-		$suffix = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wp69' : '';
+		$suffix              = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wp69' : '';
+		$wp71_suffix         = version_compare( $wp_version, '7.1-alpha', '>=' ) ? '-wp71' : '';
+		$wp69_or_wp71_suffix = $wp71_suffix ?: $suffix;
```
Screenfeed: This seems complicated:
- For `$suffix` we add `-wpXX` if WP < 7.0,
- But for `$wp71_suffix` we add `-wpXX` if WP >= 7.1.
- Then what about in a year when we'll have to test if WP is >= 7.2?

Maybe we can make this a bit easier to understand.

Currently:
- `$suffix` can be `'-wp69'` (< 7.0) or `''` (>= 7.0).
- `$wp71_suffix` can be `''` (<7.1) or `'-wp71'` (>= 7.1).
- `$wp69_or_wp71_suffix` can be `'-wp69'` (< 7.0), `''` (> 7.0 and < 7.1), `'-wp71'` (>= 7.1).

So I suggest (`lt` means "lower than" here):
- `$suffix_70` can be `'-wp-lt-70'` (< 7.0) or `''` (>= 7.0).
- `$suffix_71` can be `'-wp-lt-71'` (<7.1) or `''` (>= 7.1).
- `$suffix` can be `'-wp-lt-70'` (< 7.0), `'-wp-lt-71'` (> 7.0 and < 7.1), `''` (>= 7.1).

So what about something like this:

```php
$suffix_70 = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wplt70' : '';
$suffix_71 = version_compare( $wp_version, '7.1-alpha', '<' ) ? '-wplt71' : '';
if ( empty( $suffix_71 ) ) {
	$suffix = '';
} elseif ( empty( $suffix_70 ) ) {
	$suffix = $suffix_71;
} else {
	$suffix = $suffix_70;
}
```

Then if we need to do the same thing for WP 7.2:

```php
$suffix_70 = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wplt70' : '';
$suffix_71 = version_compare( $wp_version, '7.1-alpha', '<' ) ? '-wplt71' : '';
$suffix_72 = version_compare( $wp_version, '7.2-alpha', '<' ) ? '-wplt72' : '';
if ( empty( $suffix_72 ) ) {
	$suffix = '';
} elseif ( empty( $suffix_71 ) ) {
	$suffix = $suffix_72;
} elseif ( empty( $suffix_70 ) ) {
	$suffix = $suffix_71;
} else {
	$suffix = $suffix_70;
}
```

What do you think?

src/modules/Blocks/Language_Switcher/Navigation/Block.php
```diff
@@ -6,8 +6,13 @@
...
 
 use WP_Block;
-use PLL_Switcher;
+use PLL_Language;
 use WP_HTML_Tag_Processor;
+use WP_Syntex\Polylang\Switcher\Assets;
+use WP_Syntex\Polylang\Switcher\Switcher;
+use WP_Syntex\Polylang\Switcher\Element\Abstract_Element;
+use WP_Syntex\Polylang\Switcher\Element\Nav as NavElement;
+use WP_Syntex\Polylang\Switcher\Settings\Menu as MenuSettings;
```
Screenfeed: Not in Pascal_Case, I'd suggest:

```php
use WP_Syntex\Polylang\Switcher\Element\Nav as Nav_Element;
use WP_Syntex\Polylang\Switcher\Settings\Menu as Menu_Settings;
```

Or, probably simpler:

```php
use WP_Syntex\Polylang\Switcher\Element\Nav as Element;
use WP_Syntex\Polylang\Switcher\Settings\Menu as Settings;
```

src/modules/Blocks/Language_Switcher/Navigation/Block.php
```diff
@@ -83,35 +96,37 @@ protected function get_block_name() {
...
+			}
 
-				if ( $switcher_element['current_lang'] && ! $attributes['hide_current'] ) {
-					$top_level_lang = $switcher_element;
-				}
+			$top_level_element = $this->get_top_level_element( $settings );
+
+			if ( empty( $top_level_element ) ) {
+				return '';
 			}
```
Screenfeed: Wouldn't it make more sense to deal with the top element first?

```php
$top_level_element = $this->get_top_level_element( $settings );

if ( empty( $top_level_element ) ) {
	return '';
}

$inner_nav_link_blocks = array();

foreach ( $elements as $element ) {
	$nav_link_block_args = array(
		'blockName' => 'core/navigation-link',
		'attrs'     => $this->get_core_block_attributes( $attributes, $element ),
	);

	$inner_nav_link_blocks[] = new WP_Block( $nav_link_block_args, $block->context );
}
```

js/src/blocks/navigation-language-switcher/rendered-switcher.js
```diff
@@ -0,0 +1,83 @@
...
+	const isDropdown = layout === 'dropdown';
+	const curatedLanguages = useCuratedLanguages(
+		languages,
+		currentLanguage,
+		false
+	);
+	const languagesToRender =
+		isDropdown && curatedLanguages?.length
+			? [ curatedLanguages[ 0 ] ]
+			: curatedLanguages;
```
Screenfeed: What would you think of this?

```suggestion
	const currentLanguage = useCurrentLanguage( languages );
	const isDropdown = layout === 'dropdown';
	const languagesToRender = useCuratedLanguages(
		languages,
		currentLanguage,
		isDropdown
	);
```

js/src/blocks/navigation-language-switcher/rendered-switcher.js
```diff
@@ -0,0 +1,83 @@
...
+		showLabels: show_labels,
+		showFlags: show_flags,
+		flagBorderRadius: flag_border_radius,
+		flagWidth: flag_width,
+		labelSpacing,
+	};
+
+	return (
+		<>
+			{ languagesToRender?.map( ( language ) => {
```
Screenfeed: Is the `?` really useful? Since `useCuratedLanguages()` returns `Object[]`, I think it can be removed.

js/src/blocks/navigation-language-switcher/rendered-switcher.js
```diff
@@ -0,0 +1,83 @@
...
+		false
+	);
+	const languagesToRender =
+		isDropdown && curatedLanguages?.length
+			? [ curatedLanguages[ 0 ] ]
+			: curatedLanguages;
+	const labelSpacing = getLabelSpacing( show_labels, flag_label_spacing );
+	const withSubmenuIcon = Boolean(
+		( showSubmenuIcon || openSubmenusOnClick ) && isDropdown
+	);
```
Screenfeed: Can we have `showSubmenuIcon` or `openSubmenusOnClick` at `true` but `isDropdown` at `false`? 🤔

js/src/blocks/components/switcher-controls.js
```diff
@@ -104,31 +156,18 @@ export const SwitcherControls = ( { attributes, setAttributes } ) => {
 		} );
 	}
 
+	const hideCurrentVisible =
+		'select' !== layout &&
+		( ! hideCurrentInDropdown || 'dropdown' !== layout );
```
Screenfeed: I don't understand why we change this. The "Hide current" should be hidden only if `'select' === layout` (which, btw, will be always shown in [my PR](https://github.com/polylang/polylang/pull/1923)).

Screenfeed (changes requested):

tests/phpunit/tests/test-switcher-block-frontend.php
Hug0-Drelon: Indeed, the code was ugly as fuck 😅 
See 1fdaabfd9a955ca7b4ebe53911fc6bbf44ce5e1d.

src/modules/Blocks/Language_Switcher/Navigation/Block.php
Hug0-Drelon: Oops, see 10c24bfd5bd87fe2c35eb916ea3f662fb20080ef.

src/modules/Blocks/Language_Switcher/Navigation/Block.php
Hug0-Drelon: Better indeed, see ae54f15b49a112e54a55518bf29cc49b6900e2bd.

js/src/blocks/navigation-language-switcher/rendered-switcher.js
Hug0-Drelon: Nice! Applied in 48960e64f568f2699527181e19678a34d302fe2a.

js/src/blocks/navigation-language-switcher/rendered-switcher.js
Hug0-Drelon: Not useful, removed in f0c49af2ba01b7cc15e5218efb1e9d99f3c65fa6.

js/src/blocks/navigation-language-switcher/rendered-switcher.js
Hug0-Drelon: No, we can't. Fixed in c4209867d004ceab902d122785460b944860c5ff.

js/src/blocks/components/switcher-controls.js
Hug0-Drelon: Now "Hide current" is always shown, see 44b90477.

tests/phpunit/tests/test-switcher-block-frontend.php
Screenfeed: Sorry it is still very confusing to me. An empty suffix should be used for the most recent versions of WP.

From what I see in the test:

| Description | Possibility 1 | Possibility 2 | Possibility 3 | Suffixes | OK | Should be |
|:------------|:-------------:|:-------------:|:-------------:|:---------|:--:|:----------|
| Display as list | <7.0 | >=7.0 | | `-wp69` / ` ` | ✅ ||
| Display as list with font and colors | <7.1 | >=7.1 | | ` ` / `-wp71` | ❌ | `-wp70` / ` ` |
| Display as list with custom font and custom colors | < 7.0 | < 7.1 | >= 7.1 | `-wp69` / ` ` / `-wp71` | ❌ | `-wp69` / `-wp70` / ` ` |
| Display as dropdown | <7.0 | >=7.0 | | `-wp69` / ` ` | ✅ ||
| Display as dropdown with flags | <7.0 | >=7.0 | | `-wp69` / ` ` | ✅ ||
| Display as dropdown with submenu icon before WP 7.0 | <7.0 | | | `-wp69` | ✅ ||
| Display as dropdown with submenu icon after WP 7.0 | >=7.0 | | | ` ` | ✅ ||
| Display as dropdown with open on click | <7.0 | >=7.0 | | `-wp69` / ` ` | ✅ ||
| Display as dropdown with font and colors | <7.1 | >=7.1 | | ` ` / `-wp71` | ❌ | `-wp70` / ` ` |
| Display as dropdown with custom font and custom colors | < 7.0 | < 7.1 | >= 7.1 | `-wp69` / ` ` / `-wp71` | ❌ | `-wp69` / `-wp70` / ` ` |
| Displayed as dropdown with unauthorized CSS | < 7.0 | < 7.1 | >= 7.1 | `-wp69` / ` ` / `-wp71` | ❌ | `-wp69` / `-wp70` / ` ` |

What I mean:

Currently:
- `$suffix` can be `'-wp69'` (< 7.0) or `''` (>= 7.0). ✅
- `$suffix_wp71` can be `''` (<7.1) or `'-wp71'` (>= 7.1). ❌
- `$suffix_or_wp71` can be `'-wp69'` (< 7.0), `''` (> 7.0 and < 7.1), `'-wp71'` (>= 7.1). ❌

So I come back to my previous comment, while keeping `-wp69`'s logic. The idea is "if WP < XX then suffix is `-wpXW`":
- `$suffix_70` should be `'-wp69'` (< 7.0) or `''` (>= 7.0).
- `$suffix_71` should be `'-wp70'` (<7.1) or `''` (>= 7.1).
- `$suffix` should be `'-wp69'` (< 7.0), `'-wp70'` (> 7.0 and < 7.1), `''` (>= 7.1).
- That's all, no need of other variables.

If `$suffix_70` can be `''` or `-wp69`, then `$suffix_71` should be `''` or `-wp70`, not the other way around. Another way of saying it is "if `-wp69` means <7.0, then `-wp71` should mean <7.2", but it doesn't. This means some files must be renamed.

js/src/blocks/navigation-language-switcher/rendered-switcher.js
Screenfeed: Wat? You removed `Boolean()` and reversed the checks 😅

Screenfeed (changes requested):

tests/phpunit/tests/test-switcher-block-frontend.php
Hug0-Drelon: Here is a proposal with your previous suggestion 25b1b99a.

js/src/blocks/navigation-language-switcher/rendered-switcher.js
Hug0-Drelon: At second sight, it appears that `Boolean()` is redundant. The expression already evaluates to a value React treats correctly in `{ withSubmenuIcon && ( … ) }`.

About the reversed check, it addresses the issue you found. Checking `isDropdown` first acts as a guard. Now the condition means; *show the submenu icon only when layout is dropdown and at least one of the navigation context flags is set*.

Keep in mind that `showSubmenuIcon` and `openSubmenusOnClick` are inherited context values, they're not computed regarding `polylang/nacigation-language-switcher` block attributes.

tests/phpunit/tests/test-switcher-block-frontend.php
Screenfeed: I just saw that we could use the same number for the var names (`$suffix_69` => `-wp69`), but it's OK, I already annoyed you enough with this x)
Does it make sense to you at least?

Screenfeed (approved):

tests/phpunit/tests/test-switcher-block-frontend.php
Hug0-Drelon: Yep, I got it I think. Var names weren't clear enough in the end.

TBH I hesitate to create 3 different datasets. It'd be easier to maintain, but would grow the provider greatly...

src/modules/Blocks/Language_Switcher/Abstract_Block.php
```diff
@@ -232,4 +233,50 @@ protected function set_attributes_for_block( $attributes ) {
...
+		if ( empty( $attributes['show_flags'] ) ) {
+			return $html;
+		}
+
+		$border_radius = max( 0, min( 100, (int) ( $attributes['flag_border_radius'] ?? 0 ) ) );
+		$flag_width    = $attributes['flag_width'] ?? '18px';
+		$flag_style    = sprintf(
+			'--pll-flag-border-radius:%1$d;--pll-flag-width:%2$s',
+			$border_radius,
+			$flag_width
```
Chouby: Doesn't this need to be escaped?

src/modules/Blocks/Language_Switcher/Abstract_Block.php
```diff
@@ -232,4 +233,50 @@ protected function set_attributes_for_block( $attributes ) {
...
+		);
+
+		$label_style = '';
+		if ( ! empty( $attributes['show_labels'] ) ) {
+			$label_style = sprintf(
+				'--pll-flag-label-spacing:%1$s',
+				$attributes['flag_label_spacing'] ?? '0.3em'
```
Chouby: same

tests/phpunit/tests/test-switcher-block-frontend.php
```diff
@@ -96,28 +96,36 @@ public function test_render_polylang_navigation_switcher_block( $options, $conte
 	public function switcher_options_provider() {
 		global $wp_version;
 
-		$suffix = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wp69' : '';
+		$suffix_70 = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wp69' : '';
+		$suffix_71 = version_compare( $wp_version, '7.1-alpha', '<' ) ? '-wp70' : '';
```
Chouby: It seems more logical to name the variable the same way as the string it contains. 
```suggestion
		$suffix_69 = version_compare( $wp_version, '7.0-alpha', '<' ) ? '-wp69' : '';
		$suffix_70 = version_compare( $wp_version, '7.1-alpha', '<' ) ? '-wp70' : '';
```
Also backward compatibility comment(s) would be welcome to be able to easily remove the suffix when we won't need them anymore.

tests/phpunit/tests/test-switcher-block-frontend.php
```diff
@@ -96,28 +96,36 @@ public function test_render_polylang_navigation_switcher_block( $options, $conte
...
+				'options'  => array( 'layout' => 'horizontal' ),
 				'context'  => array(
 					'textColor'              => 'primary',
 					'backgroundColor'        => 'secondary',
 					'overlayTextColor'       => 'pale-pink',
 					'overlayBackgroundColor' => 'vivid-red',
 					'fontSize'               => 'x-large',
 				),
-				'expected' => 'navigation-language-switcher-list-with-font-and-color.html',
+				'expected' => 'navigation-language-switcher-list-with-font-and-color' . $suffix_71 . '.html',
```
Chouby: This means that the expected result is the same for WP 6.9 and WP 7.1 but different fror WP 7.0. I am correct?

Chouby (changes requested):

src/modules/Blocks/Language_Switcher/Abstract_Block.php
Hug0-Drelon: Indeed, it does. I thought that attribute passed to `wp_kses_post` would be sufficient, but Gutenberg uses `safecss_filter_attr`.
Applied in db6ce73774c9520f9d046f95ddaf4ae97939fa89.

src/modules/Blocks/Language_Switcher/Abstract_Block.php
Hug0-Drelon: Applied in db6ce73774c9520f9d046f95ddaf4ae97939fa89.

tests/phpunit/tests/test-switcher-block-frontend.php
Hug0-Drelon: See 50b17b87

tests/phpunit/tests/test-switcher-block-frontend.php
Hug0-Drelon: No, it means that 6.9 and 7.0 share the same fixture; 7.1+ differs.

PR #1925, opened 2026-07-02
Rework select-menu's style to match WP 7.0

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/3027.

WP 7.0 changed a few things to form element's styles.
This PR adapts the styles of select-menu to match the real `<select>`'s styles.

> [!NOTE]
> 1. Removed support of IE.
> 2. Removed support of jQuery UI < 1.12. WP 6.5 uses 1.13.2.
> 3. Removed code that was overriding WC Bookings styles: WC Bookings doesn't enqueue its styles on our pages.
> 4. Use of WP's CSS variables with fallback (there are no WP CSS var on our Wizard page).
> 5. Target only our selectors with CSS classes starting with `pll-*`.

WP 7.0's tyle for a real `<select>` tag:
image

New select-menu styles:
image
image

The orange border is the "focus highlight" style with the admin style Sunrise.

In the wizard:
image

### DIFF
--- a/css/src/selectmenu.css
+++ b/css/src/selectmenu.css
@@ -1,239 +1,143 @@
-/* Greatly modified version of the jquery-ui.css */
-
-.ui-widget-overlay {
-	position: fixed;
-	top: 0;
-	left: 0;
-	width: 100%;
-	height: 100%;
-}
-
-.ui-menu {
-	list-style: none;
-	padding: 0;
-	margin: 0;
-	display: block;
-	outline: none;
-}
-
-.ui-menu .ui-menu {
-	position: absolute;
-}
-
-.ui-menu .ui-menu-item {
-	position: relative;
-	margin: 0;
-	padding: 3px 1em 3px .4em;
+/* Main wrapper */
+.pll-selectmenu-button,
+.pll-selectmenu-menu .ui-menu {
+	font-size: 14px;
+	font-family: inherit;
+	color: #1e1e1e;
+	background: #fff;
+	border-radius: 2px;
 	cursor: pointer;
-	min-height: 0; /* support: IE7 */
-	/* support: IE10, see #8844 */
-	list-style-image: url("data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7");
 }
-
-/* for jQuery UI 1.12 which introduces a wrapper */
-.ui-menu .ui-menu-item:not([role]) {
-	padding: 0;
-}
-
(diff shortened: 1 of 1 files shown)


css/src/selectmenu.css
```diff
@@ -1,239 +1,143 @@
...
-	padding: 0;
+.pll-selectmenu-button:focus {
+	border-color: var(--wp-admin-theme-color, #2271b1);
+	box-shadow: 0 0 0 var(--wp-admin-border-width-focus, 1.5px) var(--wp-admin-theme-color, #2271b1);
+	outline: 2px solid transparent;
 }
-
-.ui-menu-item-wrapper {
-	padding: 3px 1em 3px 2em;
+.pll-selectmenu-button-open {
```
Hug0-Drelon: jQuery UI adds `ui-selectmenu-button-open`, not `pll-selectmenu-button-open`.

css/src/selectmenu.css
```diff
@@ -1,239 +1,143 @@
...
-	height: 16px;
+.pll-selectmenu-menu .ui-menu-item-wrapper {
+	position: relative;
+	padding: 3px 1em 3px 2em;
 }
-
-.pll-selectmenu-button.ui-button:hover,
-.pll-wizard .ui-button:hover,
-.pll-wizard .ui-button:focus {
-	background: #fff; /* To override jQuery ui-dialog styles provided by WordPress */
```
Hug0-Drelon, line 233: I'm not sure this is back ported.

Should we add:
```css
.pll-selectmenu-button:hover,
.pll-selectmenu-button:focus {
	background: #fff;
}
```

Hug0-Drelon (changes requested):
~~I think we should have two separate stylesheets until we support WP7.0 at least.
Otherwise the new style is applied only to the select for WP < 7.0.~~ The issue doesn't require backward compatibility.

css/src/selectmenu.css
Screenfeed: Probably a search/replace that went wrong. I removed it in [53e80f4](https://github.com/polylang/polylang/pull/1925/commits/53e80f4f17b54397dc30a1fd8a095b6e3f9836bb) anyway because it does nothing.

css/src/selectmenu.css
Screenfeed: It was missing indeed. Added in [64bdb55](https://github.com/polylang/polylang/pull/1925/commits/64bdb55330267aa3688626593f9d19ebdd5976ac).

Hug0-Drelon (approved):

PR #1926, opened 2026-07-03
Test for PLL_Sitemaps::set_language_from_query()

accolette:
Fixes https://github.com/polylang/polylang-pro/issues/2881

## Why
The method `set_language_from_query()` didn't have the correct types, which was fixed in a previous PR. To prevent regression, we need to add tests for this method.

## What
A test has been created, covering 4 possible cases for `set_language_from_query()`, which takes two arguments.
It uses the same structure as other sitemap tests (with `init()`, not `set_up()`).


tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,51 @@ public function test_users() {
...
+		$query = new WP_Query( $query_args );
+
+		if ( $expected_slug_or_default === 'default_lang' ) {
+			$expected = $this->pll_env->model->get_default_language();
+		} else {
+			$expected = $this->pll_env->model->get_language( $expected_slug_or_default );
+		}
+		$actual = $this->pll_env->sitemaps->set_language_from_query( $lang, $query );
+
+		// Assert
```
Hug0-Drelon:
```suggestion
```
I think we can remove this comment.

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,51 @@ public function test_users() {
...
+		$this->assertSame( $expected->slug, $actual->slug );
+	}
+
+	public function lang_query_values_provider() {
```
Hug0-Drelon, line 313: Can you add PHPDoc to this provider method?

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,51 @@ public function test_users() {
...
+			array(
+				'fr',
+				array( 'sitemap' => 'posts', 'lang' => '' ),
+				'default_lang',
+			),
+			array(
+				false,
+				array( 'sitemap' => 'posts', 'lang' => '' ),
+				'default_lang',
+			),
```
Hug0-Drelon: Can you add keys to first level values?
This way, if a test fails, we'll get a pretty error message pointing to the correct data set.

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,51 @@ public function test_users() {
...
+	 * @param string|false $lang_slug
+	 * @param array        $query_args
+	 * @param string       $expected_slug_or_default
+	 */
+	public function test_it_should_return_expected_lang_slug( $lang_slug, $query_args, $expected_slug_or_default ) {
+		$this->init();
+		$lang = $this->pll_env->model->get_language( $lang_slug );
+		$query = new WP_Query( $query_args );
+
+		if ( $expected_slug_or_default === 'default_lang' ) {
```
Hug0-Drelon:
```suggestion
		if ( 'default_lang' === $expected_slug_or_default ) {
```
Use Yoda Condition checks, you must.

Hug0-Drelon (commented):
I suggest to add a few cases in your provider as well. For instance:
- `array( 'sitemap' => 'foo', 'lang' => 'fr' )` for unknown sitemap.
- `array( 'sitemap' => 'posts', 'lang' => 'foo' )` for unknown language.
- `array( 'foo' => 'bar' )` for WTF happens?

Hug0-Drelon (approved):

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,81 @@ public function test_users() {
...
+	public function test_it_should_return_expected_lang_slug( $lang_slug, $query_args, $expected_slug_or_default ) {
+		$this->init();
+		$lang = $this->pll_env->model->get_language( $lang_slug );
+		$query = new WP_Query( $query_args );
+
+		if ( 'default_lang' === $expected_slug_or_default ) {
+			$expected = $this->pll_env->model->get_default_language();
+		} else {
+			$expected = $this->pll_env->model->get_language( $expected_slug_or_default );
+		}
```
Chouby: This kind of pattern (add a test to dynamically change the expected result) generally means that this could be simplified by rewriting the provider.

Here the default language will always be 'en'. So instead of 'default_lang' in provider, you could directly use 'en', possibly adding a comment to precise it's the default language if you like.

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,81 @@ public function test_users() {
 		);
 		$this->assertSameSets( $expected, wp_list_pluck( $providers['users']->get_sitemap_entries(), 'loc' ) );
 	}
+
+	/**
+	 * @dataProvider lang_query_values_provider
+	 * @param string|false $lang_slug
```
Chouby:
```suggestion
	 * @dataProvider lang_query_values_provider
	 *
	 * @param string|false $lang_slug
```

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,81 @@ public function test_users() {
...
+		$actual = $this->pll_env->sitemaps->set_language_from_query( $lang, $query );
+
+		$this->assertSame( $expected->slug, $actual->slug );
+	}
+
+	/**
+	 * Provides data about queries, lang and expected values.
+	 *
+	 * @return array $data {
+	 *     @type string|false $lang_slug                Current language code.
```
Chouby: I suggest to pass an empty string instead of false to avoid a type change. `get_language()` will then convert it to false as you expect to pass to `set_language_from_query()`.
```suggestion
	 *     @type string $lang_slug                Current language code.
```

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,81 @@ public function test_users() {
...
+
+	/**
+	 * @dataProvider lang_query_values_provider
+	 * @param string|false $lang_slug
+	 * @param array        $query_args
+	 * @param string       $expected_slug_or_default
+	 */
+	public function test_it_should_return_expected_lang_slug( $lang_slug, $query_args, $expected_slug_or_default ) {
+		$this->init();
+		$lang = $this->pll_env->model->get_language( $lang_slug );
```
Chouby: The method `get_language()` is not explicitly defined anymore. [It still work for backward compatibility
](https://github.com/polylang/polylang/blob/95750b63087e76ce1650a73f8a8ef198103a655a/src/model.php#L149) but we may deprecate it in the future so let use the new way directly.
```suggestion
		$lang = $this->pll_env->model->languages->get( $lang_slug );
```

Chouby (changes requested):

tests/phpunit/tests/test-sitemaps.php
```diff
@@ -284,4 +284,78 @@ public function test_users() {
...
+	 */
+	public function test_it_should_return_expected_lang_slug( $lang_slug, $query_args, $expected_slug ) {
+		$this->init();
+		$lang = $this->pll_env->model->languages->get( $lang_slug );
+		$query = new WP_Query( $query_args );
+
+		$expected = $this->pll_env->model->languages->get( $expected_slug );
+		$actual = $this->pll_env->sitemaps->set_language_from_query( $lang, $query );
+
+		$this->assertSame( $expected->slug, $actual->slug );
```
Chouby: You can further simplify as `$expected->slug` and `$expected_slug` are always the same:
```suggestion
		$lang   = $this->pll_env->model->languages->get( $lang_slug );
		$query  = new WP_Query( $query_args );
		$actual = $this->pll_env->sitemaps->set_language_from_query( $lang, $query );
		
		$this->assertSame( $expected_slug, $actual->slug );
```

Chouby (changes requested):

PR #1930, opened 2026-07-09
Add Tag Pre-release

sebastienserre:


Hug0-Drelon (approved):
I validate as it is, but let's see with @Chouby if the idea of tagging then moving these tests in a separate CI fit the needs.

PR #1934, opened 2026-07-16
fix(language-switcher): honor block textAlign

Hug0-Drelon:
## What?
Make the standard language switcher block respect typography `textAlign` (left / center / right) in the editor and on the frontend for vertical, horizontal, dropdown, and select layouts.

## Why?
`textAlign` had little or no effect: flex links ignored `text-align`, the dropdown stayed `fit-content` without packing, and the editor select markup did not match the frontend flex wrapper.

## How?
- CSS: pack layouts with `justify-content`, stretch the block for center/right, position the dropdown submenu, and add editor styles for nested previews.
- JS: pass `style.typography.textAlign` into switcher classes; wrap the select preview like the frontend.
- PHP: map `textAlign` to switcher `alignment` and register `pll-language-switcher-editor`.
- JSON: declare the block `editorStyle` handle.

### ~Proposal: segregate classic vs block CSS~
~Commit [`297d60ee`](https://github.com/polylang/polylang/commit/297d60eea55bf726dfb217eca822abed3eeec24c) proposes splitting block-only rules into `css/src/language-switcher.css` (which `@import`s the shared classic `frontend-switcher.css`) and registering block styles from `Block::register_styles()` instead of `Assets::register_styles`. Classic widgets keep the shared stylesheet; the block loads the combined entry.~

## Test plan
- [ ] Vertical / horizontal / dropdown / select: set Text align left, center, right in the editor — preview matches.
- [ ] Same layouts on the frontend after save — alignment matches the editor.
- [ ] Dropdown submenu panel anchors left / center / right with the control.
- [ ] Select uses a wrapper + inner `select.pll-switcher-select` in the editor (same as frontend).
- [ ] Hard-refresh so rebuilt CSS/JS are loaded.
- ~[ ] Classic Languages widget still styles correctly (shared `frontend-switcher.css` only).~
- ~[ ] Block frontend/editor still get shared + block rules via `language-switcher.css`.~
- [ ] Change the background color of the dropdown's sub-menu by selecting a background color for the switcher itself.
- [ ] Change the background color of the dropdown's sub-menu by using the CSS variables described in [this comment (5th point)](https://github.com/polylang/polylang/pull/1934#issuecomment-5058902099).
- [ ] Widgets: Vertical / horizontal / dropdown / select: set alignment to left, center, right, stretched. This can be tested with the classic widget and a block widget containing a block switcher (for example, change the value of `textAlign` in: `<!-- wp:polylang/language-switcher {"layout":"horizontal","style":{"typography":{"textAlign":"stretched"}}} /-->`).


Screenfeed:
What I did:
- Regroup dropdown-related styles.
- Remove some useless declarations.
- Fix some alignment-related issues with the dropdown layout. This required to add another `<div>` in the switcher for this layout: this new element follows the "text alignment" the same way the other layouts do, while its child elements remain "packed".
- Add a bit of padding to the top-level link and button (dropdown).
- Introduce the CSS vars `--pll-dropdown-background-color` and `--pll-dropdown-text-color` to tweak the colors of the dropdown sub-menu. By default, the sub-menu's background will be transparent (unless the theme gives a background to the element wrapping the block, which won't happen often IMHO). Using the UI to give a background color to the switcher works, the sub-menu inherits it, but with these CSS variables the user can give a background to the sub-menu without changing the switcher's background. Ex:
  1. Select the block.
  2. Open the "Advanced" section at the bottom.
  3. Fill the "Additional CSS" box with the following:
    ```css
    --pll-dropdown-background-color: rebeccapurple;
    --pll-dropdown-text-color: #fff;
    ```
  Note: in the widgets this background is white, but still customizable with CSS vars.
- Improve the `stretched` alignment (used only in widgets).
- Remove the CSS file dedicated to blocks: its content was less than 10 lines in the end, not enough to justify a separated file.
- Remove the use of the `inherit` value in the block: this was introduced when we wanted to use another way to align the switcher's content. Now that we use the `alignment` key:
  - `inherit` is useless.
  - `inherit` is not one of the possible choices (which are `left`, `center`, `right`): if the user selects one of the 3 values, they can't set it back to `inherit`, so it introduces a weird behavior.

Screenfeed (approved):

PR #1935, opened 2026-07-16
test(e2e) test e2e-action-cache

Hug0-Drelon:
:suspect:


Hug0-Drelon:
https://github.com/polylang/actions/pull/20 merged. Went from ~4-5 min run to less than 3 min 🥳

PR #1936, opened 2026-07-16
fix(phpdoc): qualify class names in Languages model docs

Hug0-Drelon:
## What?
Qualify class names in PHPDoc `@param` / `@return` tags in `WP_Syntex\Polylang\Model\Languages` (e.g. `PLL_Language` → `\PLL_Language`, `WP_Error` → `\WP_Error`, `WP_Term` → `\WP_Term`). Also correct a wrong `@param Language` type to `Languages` on a deprecated filter.

## Why?
This class lives in a namespaced context. Unqualified PHPDoc types do not resolve correctly for third parties using Polylang auto-generated stubs when there is no native type hint on the argument or return.

## How?
Prefix global / root classes with `\` in PHPDoc where stubs rely on documentation rather than PHP type hints.

## Test plan
- [ ] Confirm stub generators (or IDE inspection of stubs) resolve `\PLL_Language`, `\WP_Error`, and `\WP_Term` correctly for methods without native type hints

### DIFF
--- a/src/Model/Languages.php
+++ b/src/Model/Languages.php
@@ -115,7 +115,7 @@ class Languages {
 	 *                     this is to prevent confusion between `term_id` and `term_taxonomy_id`.
 	 * @return \PLL_Language|false Language object, false if no language found.
 	 *
-	 * @phpstan-param PLL_Language|WP_Term|int|string $value
+	 * @phpstan-param \PLL_Language|\WP_Term|int|string $value
 	 */
 	public function get( $value ) {
 		if ( $value instanceof PLL_Language ) {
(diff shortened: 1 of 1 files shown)


Screenfeed:
I split into several commits in case there are things you don't want.

Hug0-Drelon (commented):
There is two problems IMHO:
- Adding full namespaces or `\` prefix when the argument/return/property is type-hinted is not useful and adds noise to the PHPDoc ([php-stubs/generator](https://github.com/php-stubs/generator) already handles it).
- Type-hinting public methods that weren't is really risky, moreover on this public language model that is heavily used. It'd require a lot of tests to ensure we don't break anything (PLLPro, PLLWC, PLLE and even third-parties)

What do you think?

Screenfeed:
@Hug0-Drelon I reverted [9c54813](https://github.com/polylang/polylang/pull/1936/commits/9c54813533ad89ad53316a1a7c28fc6061320ec1) and [8d0f25a](https://github.com/polylang/polylang/pull/1936/commits/8d0f25ae66ef8433c9bbf8c27aa248116384cbe3).

Screenfeed (approved):

PR #1939, opened 2026-07-21
Use a WP_Query for `get_untranslated()`

Chouby:
This PR is a follow-up of #1892 

In #1187, we switched from a `WP_Query` to a custom DB query for `PLL_Translated_Post;;get_untranslated()`.
in #1892, we introduced a `WP_Query` filter to get untranslated posts. Thus it's now better to re-use the `WP_Query` and use this filter (via the query var `untranslated_in`) to avoid to repeat ourselves. Moreover the query used in #1892 is more robust than the previous custom query in `get_untranslated()` (See https://github.com/polylang/polylang/pull/1892#discussion_r3311844063).

NB: This method is already covered by tests.

### DIFF
--- a/src/translated-post.php
+++ b/src/translated-post.php
@@ -429,7 +429,7 @@ class PLL_Translated_Post extends PLL_Translated_Object implements PLL_Translata
 		$args = apply_filters( 'pll_ajax_posts_not_translated_args', $args );
 
 		/** @var WP_Post[] */
-		$posts = new WP_Query()->query( $args );
+		$posts = ( new WP_Query() )->query( $args );
 
 		foreach ( $posts as $i => $post ) {
 			if ( ! $this->current_user_can_read( $post->ID, 'edit' ) ) {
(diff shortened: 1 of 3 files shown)


src/translated-post.php
```diff
@@ -402,59 +402,40 @@ public function create_media_translation( $post_id, $lang ) {
...
 	public function get_untranslated( $type, PLL_Language $untranslated_in, PLL_Language $lang, $search = '' ) {
-		global $wpdb;
+		$args = array(
+			's'                => $search,
+			'suppress_filters' => false, // To make the post_fields filter work.
+			'lang'             => '', // Avoid admin language filter.
+			'untranslated_in'  => $untranslated_in,
+			'numberposts'      => 20, // Limit to 20 posts by default.
+			'post_status'      => 'any',
+			'post_type'        => $type,
```
Hug0-Drelon, line 412:
```suggestion
			'post_type'        => $type,
			'no_found_rows'    => true,
```

Hug0-Drelon (changes requested):

Issue #26, opened 2016-06-20
Menu shown in default language on the backend.


grappler:
We are using Shortcode on the backend to visually show the shortcode. The shortcode includes code `wp_nav_menu()` to display the menu. 

The front end shows the menu in the correct language but on the backend the menu is shown in the default language.

Chouby:
I am not surprised. wp_nav_menu is not filtered on backend (as other functions that I believed to be frontend specific). Could you explain more on the context? Are you using a specific plugin to visualize your shortcode on the backend?

grappler:
I was also not surprised. I am using [Shortcake](https://github.com/wp-shortcake/shortcake) to visualize the shortcode on the backend.

https://wordpress.org/plugins/shortcode-ui/
https://wptavern.com/shortcake-is-now-a-wordpress-feature-plugin

Chouby:
I downloaded the plugin. But I did not find any call to `wp_nav_menu()`. I guess you are using it in combination with a second plugin.

grappler:
Yes, we are running it with some custom code. This is the code you should be able to use. https://gist.github.com/grappler/067845cc9716e581be5ae930caec06d3

Chouby:
On backend, I create fictive nav menu locations, one per language. This locations are filled with the correct menu, thanks to the filter `'theme_mod_nav_menu_locations'`.

On frontend, I don't hack the nav menu locations, but fill the location with the menu in the current language, again thanks to the filter `'theme_mod_nav_menu_locations'`.

So, the same filter is used in two different ways on backend and frontend. That's why when calling `wp_nav_menu()` on backend with the usual nav menu location, you always get the menu in the default language.

To get access to the correct location on backend, you could use:
`PLL()->nav_menu->combine_location( $location, $language );`
Here the expected language is an object, not just the language code. It could be obtained with:
`PLL()->model->get_language( pll_get_post_language( $post_id ) );`

grappler:
Do you mean editing the shortcode to something like this:?

```php
$language = PLL()->model->get_language( pll_get_post_language( $post_id ) );
$location = PLL()->nav_menu->combine_location( $a['menuposition'], $language );
<?php wp_nav_menu( array(
	'menu_class'     => 'no-list inline-list category-tags',
	'container'      => false,
	'theme_location' => $location,
) ); ?>
```

Chouby:
Yes. It's what I meant. If you can get `$post_id` in your function.

grappler:
[...]

PR #1940, opened 2026-07-22
Add PLL_Updater_Registry to elect a single updater leader.

Marianne380:
Fixes partially https://github.com/polylang/polylang-pro/issues/3040.

## What
Adds `PLL_Updater_Registry` and `PLL_Updater_Interface`, a small coordination layer in core.

## Why
Several add-ons can run at once, each with its own isolated copy of the updater.

Only one of them should set up the shared licenses UI, so they need a single common point to pick which updater does it.

As discussed in daily, Polylang core seems like the best place for that shared point.

The interface belongs here for the same reason: each add-on ships the updater under its own namespace, so this is the only type they all share, and the only one the registry can type hint against.

## How
- Add-ons announce themselves via `register( PLL_Updater_Interface $updater )` from their `Updater` constructor.
- On `pll_init`, the highest updater version wins (first registered on a tie) and its setup runs once.
- `get_elected()` exposes the elected updater, so we can reach it later if we ever need to.
- Pure coordination layer: no HTTP or license logic, that stays in the updater package. An interface rather than an abstract class, so no updater code lands in core.


src/updater-registry.php
```diff
@@ -0,0 +1,68 @@
...
+	 * @return void
+	 */
+	public static function elect(): void {
+		if ( empty( self::$candidates ) ) {
+			return;
+		}
+
+		$leader = null;
+		foreach ( self::$candidates as $candidate ) {
+			if ( null === $leader || version_compare( $candidate['version'], $leader['version'], '>' ) ) {
```
Screenfeed: Now that I think about it, we were supposed to use `Updater::get_version()`, which is what you do indirectly in #26.
So why not store `Updater` instances in this registry instead? That way we can get rid of the array. It would require this:

```php
class Updater {
	public static function get_version(): string {
		// ...
	}

	public function get_id(): string {
		return $this->license->id;
	}

	public function boot_leader(): void {
		// ...
	}
}
```

So we'd probably need an interface or abstract class.
What do you think?

Note:
If we store instances then it is possible to have a `get_elected()` method, to be able to access the hook callbacks (we may need to tweak things some day). But maybe it's not the time to add this, we can decide later.

Screenfeed (changes requested):

src/updater-registry.php
Marianne380: Good idea, see https://github.com/polylang/polylang/pull/1940/changes/a4e40338aa9a70a8a232891e4f0c571235a64193 (and https://github.com/polylang/updater/pull/26/changes/e2ba30c3d4964bf2838f467c251f5fc05d1fafef).

src/updater-interface.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @return string
+	 */
+	public static function get_version(): string;
+
+	/**
+	 * Returns the product id, also the storage key in the `polylang_licenses` option.
+	 *
+	 * @since 3.9
+	 *
+	 * @return string
```
Screenfeed: Maybe we should enforce non-empty strings?

```suggestion
	 * @return string
	 * 
	 * @phpstan-return non-empty-string
	 */
	public static function get_version(): string;

	/**
	 * Returns the product id, also the storage key in the `polylang_licenses` option.
	 *
	 * @since 3.9
	 *
	 * @return string
	 * 
	 * @phpstan-return non-empty-string
```

src/updater-interface.php
```diff
@@ -0,0 +1,45 @@
...
+	 * Returns the updater version, used to elect the leader.
+	 *
+	 * @since 3.9
+	 *
+	 * @return string
+	 */
+	public static function get_version(): string;
+
+	/**
+	 * Returns the product id, also the storage key in the `polylang_licenses` option.
```
Screenfeed:
```suggestion
	 * Returns the product ID, also the storage key in the `polylang_licenses` option.
```

src/updater-registry.php
```diff
@@ -0,0 +1,85 @@
...
+			if ( null === $elected || version_compare( $candidate::get_version(), $elected::get_version(), '>' ) ) {
+				$elected = $candidate;
+			}
+		}
+
+		if ( null === $elected ) {
+			return;
+		}
+
+		self::$elected = $elected;
```
Screenfeed: I don't think the `$elected` var is necessary.

```suggestion
		foreach ( self::$candidates as $candidate ) {
			if ( null === $elected || version_compare( $candidate::get_version(), $elected::get_version(), '>' ) ) {
				self::$elected = $candidate;
			}
		}

		if ( empty( self::$elected ) ) {
			return;
		}
```

src/updater-registry.php
```diff
@@ -0,0 +1,85 @@
...
+
+		$elected->boot_leader( $polylang );
+	}
+
```
Screenfeed: Do you think it would be useful?

```suggestion

	/**
	 * Tells if an updater has been elected.
	 *
	 * @since 3.9
	 *
	 * @return bool
	 * 
	 * @phpstan-assert-if-true PLL_Updater_Interface $this->get_elected()
	 */
	public static function has_elected(): bool {
		return ! empty( self::$elected );
	}

```

Screenfeed (changes requested):

src/updater-interface.php
Screenfeed: Meh, forget it, not important.

src/updater-registry.php
Marianne380: `@phpstan-assert-if-true` needs a `$this->…()` or a variable, which we don't have here since `has_elected()` is static.

So I dropped it, callers can null-check `get_elected()` directly and PHPStan narrows it on its own.

src/updater-registry.php
Marianne380: Kept the local `$elected` for readability: it keeps the loop's working value separate from the exposed `self::$elected`, compute locally, publish once.

Dropped the early return too, see https://github.com/polylang/polylang/pull/1940/commits/9461096b9ed8b5c4d3adea4b13b4115e5e614d94.

src/updater-registry.php
```diff
@@ -0,0 +1,83 @@
...
+	 *
+	 * @since 3.9
+	 *
+	 * @param PLL_Updater_Interface $updater The add-on's updater.
+	 * @return void
+	 */
+	public static function register( PLL_Updater_Interface $updater ): void {
+		if ( empty( self::$candidates ) ) {
+			add_action( 'pll_init', array( self::class, 'elect' ), 0 );
+		}
```
Chouby: As it's a static method, adding it again doesn't duplicate it in called actions:
```suggestion
		add_action( 'pll_init', array( self::class, 'elect' ), 0 );
```

src/updater-registry.php
```diff
@@ -0,0 +1,83 @@
...
+
+		self::$candidates[ $updater->get_id() ] = $updater;
```
Chouby: I don't see that the key is used anywhere so I would simply remove the method `get_id()`.
```suggestion
		self::$candidates[] = $updater;
```
The only interest I foresee is to avoid duplicates, but I hope we are able avoid that ;) and it wouldn't harm the feature.

src/updater-registry.php
```diff
@@ -0,0 +1,83 @@
...
+	/**
+	 * Returns the elected updater, or null before the election has run.
+	 *
+	 * @since 3.9
+	 *
+	 * @return PLL_Updater_Interface|null
+	 */
+	public static function get_elected(): ?PLL_Updater_Interface {
+		return self::$elected;
+	}
```
Chouby: I haven't seen this method in use. What is it useful for? Removing it would also allow to remove the static property.

src/install/updater-registry.php
```diff
@@ -0,0 +1,83 @@
...
+			if ( null === $elected || version_compare( $candidate::get_version(), $elected::get_version(), '>' ) ) {
+				$elected = $candidate;
+			}
+		}
+
+		self::$elected = $elected;
+
+		if ( null !== $elected ) {
+			$elected->boot_leader( $polylang );
+		}
```
Chouby, line 59: What do you think of this? Also proposing to rename `boot_leader()`.
```suggestion
	public static function elect( PLL_Base $polylang ): void {
		$elected = array_reduce( 
			self::$candidates,
			function ( $carry, $item ) {
				return version_compare( $item::get_version(), $carry::get_version(), '>' ) ? $item : $carry;
			}
		);
		
		if ( ! empty( $elected ) ) {
			$elected->init( $polylang );
		}
```

src/install/updater-interface.php
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Contract an add-on's updater must fulfil to take part in the leader election.
+ *
+ * Lives in core because each add-on ships the updater under its own namespace: this interface is the only type they
+ * all share, and therefore the only one the registry can type hint against.
+ *
+ * @since 3.9
+ */
+interface PLL_Updater_Interface {
```
Chouby, line 16: I suggest to move the 2 files to the install directory.

Chouby (changes requested):

src/updater-registry.php
Marianne380: See https://github.com/polylang/polylang/pull/1940/changes/e7ed43734f43e846884098fd1c2798ae0a5a8db3.

src/updater-registry.php
Marianne380: Dropped it.

The idea was to maybe unhook some methods later (e.g. on `EDD_SL_Plugin_Updater::init()`), but nothing needs it yet. Easy to add back if it ever does. Removes the `$elected` property too.

See https://github.com/polylang/polylang/pull/1940/changes/63e1c7a63fe850a86d31dd7b60c505d2cb338b94.

src/install/updater-registry.php
Marianne380: Won't this fatal on the first item? With no initial value, `$carry` is `null`, so `$carry::get_version()` is called on `null`.

I also find the `foreach` easier to follow, the `null === $elected` check makes the first-item case explicit.

Fine to switch if you prefer, though.

See https://github.com/polylang/polylang/pull/1940/commits/b88c2b4f11d41e76120ab1898c853e87e3c0bbc7 for the method name.

src/install/updater-interface.php
Marianne380: See https://github.com/polylang/polylang/pull/1940/commits/e799ac8b50554689b7b52df999bb27a1650653c3.

PR #1944, opened 2026-07-23
[Language switcher rework] Enqueue styles when only the classic menu is used

Screenfeed:
## What

Flags are not displayed correctly in the classic menu.

## Why

If the switcher is used only in the classic menu (no widgets, no blocks), no styles are enqueued for the flags.

## How

If the switcher's styles are not already enqueued, detect when a classic menu uses a switcher with flags, and enqueue the styles accordingly.

Note: at first I wanted to enqueue only `css/src/lib/switcher-flags.css`, but it seems we tend to prefer enqueueing les files.

### DIFF
--- a/src/base.php
+++ b/src/base.php
@@ -6,7 +6,6 @@
 use WP_Syntex\Polylang\Widgets;
 use WP_Syntex\Polylang\REST\Request;
 use WP_Syntex\Polylang\Capabilities\Capabilities;
-use WP_Syntex\Polylang\Switcher\Assets as Switcher_Assets;
 
 /**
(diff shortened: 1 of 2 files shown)


Hug0-Drelon (approved):
Logic in `PLL_Frontend_Nav_Menu::maybe_enqueue_styles` seems fine.

But could we simplify and always enqueue styles when rendering the menu?

Calling `Assets::enqueue_frontend_styles()` twice if a block is rendered as well shouldn't be a problem.

What do you think @Chouby?

Chouby:
Closed in favor of #1970

PR #1947, opened 2026-07-28
Fix `get_pages()` post_type on hierarchical CPT screens

accolette:
Fixes [#3052](https://github.com/polylang/polylang-pro/issues/3052)

## Why
In `admin_enqueue_scripts()`, hierarchical post types were using `get_pages()` with default values (post type = page).
So with a CPT, the wrong pages were sent to the JS.
As a consequence, `update_post_caches()` was also priming the cache with the wrong content because of the mismatched arguments.
On top of that, `wp_dropdown_pages()` ended up being called twice: even though the code intended to avoid this ("Same arguments as the parent pages dropdown to avoid an extra query"), the arguments actually differed, so two separate queries were run.
On top of that, `is_post_type_hierarchical()` alone doesn't guarantee the "Parent" dropdown is actually displayed on screen: WordPress only renders it when the post type also supports `page-attributes` (see `WP_Posts_List_Table::inline_edit()`).

## What
Added the current post_type to the `get_pages()` arguments to fix this.
Same for the dropdown: the arguments are now identical, so only one query is run.

Replaced `update_post_caches()` by `update_object_term_cache()`: only the term cache is actually needed here (post ID and language), so priming the full post cache was unnecessary. `get_pages()` can return `false`, so a guard was added before iterating over the result.

Added a `post_type_supports( $screen->post_type, 'page-attributes' )` check alongside `is_post_type_hierarchical()`, to avoid building `pll_page_languages` when the dropdown will never be shown.

## Tests
Added to `tests/phpunit/tests/test-admin-filters-post.php`:
- `test_inline_script_for_pages`: base case on the `page` post type, checks `pll_page_languages` for two languages.
- `test_inline_script_for_hierarchical_cpt`: a translated hierarchical CPT screen only returns that CPT's items, not `page` posts (the actual regression this PR fixes).
- `test_inline_script_conditions_for_dropdown`: parameterized test (`@testWith`) covering the hierarchical / page-attributes / translated combinations, replacing `test_inline_script_for_non_hierarchical_cpt` and `test_inline_script_for_hierarchical_untranslated_cpt`, and adding coverage for the new `page-attributes` condition above.
- `test_get_pages_query_is_cached_for_dropdown`: checks that Polylang and the dropdown share the same cached query, resulting in a single SQL query.

Also merged the two post-type registration helpers into a single one, reused across these tests.

### DIFF
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -99,7 +99,7 @@ class PLL_Admin_Filters_Post {
 
 		// Hierarchical post types
 		if ( 'edit' == $screen->base && is_post_type_hierarchical( $screen->post_type ) ) {
-			$pages = get_pages( array( 'sort_column' => 'menu_order, post_title' ) ); // Same arguments as the parent pages dropdown to avoid an extra query.
+			$pages = get_pages( array( 'post_type' => $screen->post_type, array( 'sort_column' => 'menu_order, post_title' ) ) ); // Same arguments as the parent pages dropdown to avoid an extra query.
 
 			update_post_caches( $pages, $screen->post_type, true, false );
 
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -99,7 +99,12 @@ class PLL_Admin_Filters_Post {
 
 		// Hierarchical post types
 		if ( 'edit' == $screen->base && is_post_type_hierarchical( $screen->post_type ) ) {
-			$pages = get_pages( array( 'post_type' => $screen->post_type, array( 'sort_column' => 'menu_order, post_title' ) ) ); // Same arguments as the parent pages dropdown to avoid an extra query.
+			$pages = get_pages(
+				array(
+					'sort_column' => 'menu_order, post_title',
+					'post_type' => $screen->post_type,
+				)
+			); // Same arguments as the parent pages dropdown to avoid an extra query.
 
 			update_post_caches( $pages, $screen->post_type, true, false );
 
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -102,7 +102,7 @@ class PLL_Admin_Filters_Post {
 			$pages = get_pages(
 				array(
 					'sort_column' => 'menu_order, post_title',
-					'post_type' => $screen->post_type,
+					'post_type'   => $screen->post_type,
 				)
 			); // Same arguments as the parent pages dropdown to avoid an extra query.
 
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -98,7 +98,7 @@ class PLL_Admin_Filters_Post {
 		}
 
 		// Hierarchical post types
-		if ( 'edit' == $screen->base && is_post_type_hierarchical( $screen->post_type ) ) {
+		if ( 'edit' == $screen->base && is_post_type_hierarchical( $screen->post_type ) && post_type_supports( $screen->post_type, 'page-attributes' ) ) {
 			$pages = get_pages(
 				array(
 					'sort_column' => 'menu_order, post_title',
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -106,6 +106,10 @@ class PLL_Admin_Filters_Post {
 				)
 			); // Same arguments as the parent pages dropdown to avoid an extra query.
 
+			if ( ! is_array( $pages ) ) {
+				return;
+			}
+
 			update_object_term_cache( wp_list_pluck( $pages, 'ID' ), $screen->post_type );
 
 			$page_languages = array();
--- a/src/admin/admin-filters-post.php
+++ b/src/admin/admin-filters-post.php
@@ -106,7 +106,7 @@ class PLL_Admin_Filters_Post {
 				)
 			); // Same arguments as the parent pages dropdown to avoid an extra query.
 
-			update_post_caches( $pages, $screen->post_type, true, false );
+			update_object_term_cache( wp_list_pluck( $pages, 'ID' ), $screen->post_type );
 
 			$page_languages = array();
 
--- a/tests/phpunit/tests/test-admin-filters-post.php
+++ b/tests/phpunit/tests/test-admin-filters-post.php
@@ -503,7 +503,7 @@ class Admin_Filters_Post_Test extends PLL_UnitTestCase {
 	}
 
 	private function register_post_type_for_dropdown_test( array $args ): void {
-		if ( in_array( 'translated', $args ) ) {
+		if ( in_array( 'translated', $args, true ) ) {
 			add_filter(
 				'pll_get_post_types',
 				function ( $post_types ) {
@@ -516,8 +516,8 @@ class Admin_Filters_Post_Test extends PLL_UnitTestCase {
 			'doc',
 			array(
 				'public'       => true,
-				'hierarchical' => in_array( 'hierarchical', $args ),
-				'supports'     => in_array( 'page-attributes', $args ) ? array( 'page-attributes' ) : array(),
+				'hierarchical' => in_array( 'hierarchical', $args, true ),
+				'supports'     => in_array( 'page-attributes', $args, true ) ? array( 'page-attributes' ) : array(),
 			)
 		);
 	}
--- a/tests/phpunit/tests/test-admin-filters-post.php
+++ b/tests/phpunit/tests/test-admin-filters-post.php
@@ -502,19 +502,28 @@ class Admin_Filters_Post_Test extends PLL_UnitTestCase {
 		);
 	}
 
-	private function register_translated_post_type( string $post_type, bool $hierarchical ): void {
-		add_filter(
-			'pll_get_post_types',
-			function ( $post_types ) use ( $post_type ) {
-				$post_types[] = $post_type;
-				return $post_types;
-			}
+	private function register_post_type_for_dropdown_test( string $post_type, bool $hierarchical, bool $page_attributes = true, bool $translated = true ): void {
+		if ( $translated ) {
+			add_filter(
+				'pll_get_post_types',
+				function ( $post_types ) use ( $post_type ) {
+					$post_types[] = $post_type;
+					return $post_types;
+				}
+			);
+		}
+		register_post_type(
+			$post_type,
+			array(
+				'public'       => true,
+				'hierarchical' => $hierarchical,
+				'supports'     => $page_attributes ? array( 'page-attributes' ) : array(),
+			)
 		);
-		register_post_type( $post_type, array( 'public' => true, 'hierarchical' => $hierarchical, 'supports' => array( 'page-attributes' ) ) );
 	}
 
 	public function test_inline_script_for_hierarchical_cpt() {
-		$this->register_translated_post_type( 'doc', true );
+		$this->register_post_type_for_dropdown_test( 'doc', true );
 		$en = self::factory()->post->create( array( 'post_type' => 'doc' ) );
 		self::$model->post->set_language( $en, 'en' );
 		$page = self::factory()->post->create( array( 'post_type' => 'page' ) );
@@ -528,30 +537,38 @@ class Admin_Filters_Post_Test extends PLL_UnitTestCase {
 		$this->assertStringContainsString( wp_json_encode( $result ), $footer );
 	}
 
-	public function test_inline_script_for_non_hierarchical_cpt() {
-		$this->register_translated_post_type( 'doc', false );
+	/**
+	 * @testWith [true, true, true, true]
+	 *           [false, true, true, false]
+	 *           [true, false, true, false]
+	 *           [true, true, false, false]
+	 *
+	 * @param bool $hierarchical    Whether the post type is hierarchical.
+	 * @param bool $page_attributes Whether the post type supports 'page-attributes'.
+	 * @param bool $translated      Whether the post type is registered as translated by Polylang.
+	 * @param bool $expects_output  Whether pll_page_languages is expected in the footer.
+	 *
+	 * @return void
+	 */
+	public function test_inline_script_conditions_for_dropdown( bool $hierarchical, bool $page_attributes, bool $translated, bool $expects_output ): void {
+		$this->register_post_type_for_dropdown_test( 'doc', $hierarchical, $page_attributes, $translated );
 		$en = self::factory()->post->create( array( 'post_type' => 'doc' ) );
 		self::$model->post->set_language( $en, 'en' );
 
 		$this->set_current_edit_screen( 'doc' );
 		$footer = $this->get_admin_footer_scripts();
 
-		$this->assertStringNotContainsString( 'pll_page_languages', $footer );
-	}
-
-	public function test_inline_script_for_hierarchical_untranslated_cpt() {
-		register_post_type( 'doc', array( 'public' => true, 'hierarchical' => true, 'supports' => array( 'page-attributes' ) ) );
-
-		$this->set_current_edit_screen( 'doc' );
-		$footer = $this->get_admin_footer_scripts();
-
-		$this->assertStringNotContainsString( 'pll_page_languages', $footer );
+		if ( $expects_output ) {
+			$this->assertStringContainsString( 'pll_page_languages', $footer );
+		} else {
+			$this->assertStringNotContainsString( 'pll_page_languages', $footer );
+		}
 	}
 
 	public function test_get_pages_query_is_cached_for_dropdown() {
 		global $post;
 
-		$this->register_translated_post_type( 'doc', true );
+		$this->register_post_type_for_dropdown_test( 'doc', true );
 
 		// touch_time() (called internally by inline_edit() further) reads the global $post
 		// via get_post() with no argument — not set outside The Loop, so we set it up manually here.
--- a/tests/phpunit/tests/test-admin-filters-post.php
+++ b/tests/phpunit/tests/test-admin-filters-post.php
@@ -502,28 +502,28 @@ class Admin_Filters_Post_Test extends PLL_UnitTestCase {
 		);
 	}
 
-	private function register_post_type_for_dropdown_test( string $post_type, bool $hierarchical, bool $page_attributes = true, bool $translated = true ): void {
-		if ( $translated ) {
+	private function register_post_type_for_dropdown_test( array $args ): void {
+		if ( in_array( 'translated', $args ) ) {
 			add_filter(
 				'pll_get_post_types',
-				function ( $post_types ) use ( $post_type ) {
-					$post_types[] = $post_type;
+				function ( $post_types ) {
+					$post_types[] = 'doc';
 					return $post_types;
 				}
 			);
 		}
 		register_post_type(
-			$post_type,
+			'doc',
 			array(
 				'public'       => true,
-				'hierarchical' => $hierarchical,
-				'supports'     => $page_attributes ? array( 'page-attributes' ) : array(),
+				'hierarchical' => in_array( 'hierarchical', $args ),
+				'supports'     => in_array( 'page-attributes', $args ) ? array( 'page-attributes' ) : array(),
 			)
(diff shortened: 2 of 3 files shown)


tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -486,4 +497,107 @@ public function test_parent_pages_script_data_in_footer() {
...
+				'exclude_tree' => $post->ID,
+				'selected'     => $post->post_parent,
+				'sort_column'  => 'menu_order, post_title',
+				'echo'         => 0,
+			)
+		);
+		$after_dropdown = $wpdb->num_queries;
+
+		$this->assertSame( $after_pll, $after_dropdown );
+		$this->assertSame( 1, $after_pll - $before );
```
Marianne380: I'd drop `assertSame( 1, $after_pll - $before )`.

The meaningful check is `assertSame( $after_pll, $after_dropdown )`,  the dropdown adds zero queries, so the cache is shared.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -486,4 +497,107 @@ public function test_parent_pages_script_data_in_footer() {
 
 		$this->assertNotFalse( strpos( $footer, 'var pll_page_languages = ' . wp_json_encode( $pages ) ) );
 	}
+
+	public function filter_translated_post_type_in_settings( $post_types ) {
```
Marianne380:
```suggestion
	public function add_doc_to_translated_post_types( $post_types ) {
```

Marianne380 (changes requested):
The screen/scripts setup + footer capture is repeated across the 4 tests.
Two small private helpers would cover all of them and read better:
```php
    private function set_current_edit_screen( string $post_type ): void {
        $GLOBALS['hook_suffix'] = 'edit.php';
        $_REQUEST['post_type']  = $post_type;
        set_current_screen();
        $GLOBALS['wp_scripts'] = new WP_Scripts();
        wp_default_scripts( $GLOBALS['wp_scripts'] );
    }

    private function get_admin_footer(): string {
        do_action( 'admin_enqueue_scripts' );
        ob_start();
        do_action( 'admin_print_footer_scripts' );
        return (string) ob_get_clean();
    }
```

The 3 footer tests use both. 
The query-count test reuses only `set_current_edit_screen()` and keeps its own measurement around `admin_enqueue_scripts`, so a single combined helper wouldn't work for it.

Marianne380 (approved):

src/admin/admin-filters-post.php
```diff
@@ -99,7 +99,12 @@ public function admin_enqueue_scripts() {
 
 		// Hierarchical post types
 		if ( 'edit' == $screen->base && is_post_type_hierarchical( $screen->post_type ) ) {
-			$pages = get_pages( array( 'sort_column' => 'menu_order, post_title' ) ); // Same arguments as the parent pages dropdown to avoid an extra query.
+			$pages = get_pages(
+				array(
+					'sort_column' => 'menu_order, post_title',
+					'post_type' => $screen->post_type,
```
Chouby: Formatting. [The rule is disabled](https://github.com/polylang/polylang/blob/3.8.6/phpcs.xml.dist#L33) in Polylang due to lot of legacy code.
```suggestion
					'sort_column' => 'menu_order, post_title',
					'post_type'   => $screen->post_type,
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -39,6 +39,17 @@ public function set_up() {
...
+		parent::tear_down();
+
+		remove_filter( 'pll_get_post_types', array( $this, 'add_doc_to_translated_post_types' ) );
+		_unregister_post_type( 'doc' );
+		_unregister_post_type( 'article' );
+		_unregister_post_type( 'doc_not_translated' );
+		unset( $GLOBALS['current_screen'] );
+		unset( $GLOBALS['hook_suffix'], $_REQUEST['post_type'] );
+	}
+
```
Chouby, line 46:
```suggestion
```
This is useless:
- WP restores hooks before each test: https://github.com/WordPress/wordpress-develop/blob/7.0.2/tests/phpunit/includes/abstract-testcase.php#L120
- And we clean up the global variables in the test case: https://github.com/polylang/polylang/blob/3.8.6/tests/phpunit/includes/testcase-trait.php#L143-L146

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +475,103 @@ public function test_categories_script_data_in_footer() {
...
+		$fr = self::factory()->post->create( array( 'post_type' => 'page' ) );
+		self::$model->post->set_language( $fr, 'fr' );
+
+		$this->set_current_edit_screen( 'page' );
+		$footer = $this->get_admin_footer();
 
 		$pages = array( 'en' => array( $en ), 'fr' => array( $fr ) );
 
 		$this->assertNotFalse( strpos( $footer, 'var pll_page_languages = ' . wp_json_encode( $pages ) ) );
```
Chouby: Use `assertContains()` instead

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +475,103 @@ public function test_categories_script_data_in_footer() {
...
 	}
+
+	public function add_doc_to_translated_post_types( $post_types ) {
+		$post_types[] = 'doc';
+		return $post_types;
+	}
+
+	public function test_hierarchical_cpt_script_data_in_footer() {
+		add_filter( 'pll_get_post_types', array( $this, 'add_doc_to_translated_post_types' ) );
```
Chouby: For a small method like that, you have the possibility to use closure here if you prefer (here it doesn't harm the readability in my opinion): 
```suggestion
		add_filter( 
			'pll_get_post_types',
			function ( $post_types ) {
				$post_types[] = 'doc';
				return $post_types;
			}
		);
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +475,103 @@ public function test_categories_script_data_in_footer() {
...
+	}
+
+	public function test_non_hierarchical_cpt_script_data_in_footer() {
+		register_post_type( 'article', array( 'public' => true, 'hierarchical' => false ) );
+
+		$this->set_current_edit_screen( 'article' );
+		$footer = $this->get_admin_footer();
+
+		$this->assertFalse( strpos( $footer, 'pll_page_languages' ) );
+	}
```
Chouby: You don't test what the method name let think. Here you are testing that the script isn't there if there are no posts in the list. 
Also I suggest to use `assertNotContains()`.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +475,103 @@ public function test_categories_script_data_in_footer() {
...
+
+	public function test_hierarchical_untranslated_cpt_script_data_in_footer() {
+		register_post_type( 'doc_not_translated', array( 'public' => true, 'hierarchical' => true ) );
+		$en = self::factory()->post->create( array( 'post_type' => 'doc_not_translated' ) );
+		self::$model->post->set_language( $en, 'en' );
```
Chouby, line 554: You can't assign a language since the post type isn't translatable.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +475,103 @@ public function test_categories_script_data_in_footer() {
...
+		// Same arguments as the real 'parent page' dropdown built by wp_dropdown_pages() in WordPress core.
+		wp_dropdown_pages(
+			array(
+				'post_type'    => $post->post_type,
+				'exclude_tree' => $post->ID,
+				'selected'     => $post->post_parent,
+				'sort_column'  => 'menu_order, post_title',
+				'echo'         => 0,
+			)
+		);
```
Chouby: The intention is good but by copying the WP code, you don't guarantee that a future evolution of WP won't break that.
So I suggest to call `WP_Post_List_Table::inline_edit()` instead. and count the DB queries done by the method. It may change for other reasons, an we would have false positives but we would catch interesting change.

Chouby (changes requested):

tests/phpunit/tests/test-admin-filters-post.php
accolette: ok, i have removed redondants parts but i have to let `_unregister_post_type()` in the tear down otherwise others CPT tests failed

tests/phpunit/tests/test-admin-filters-post.php
accolette: This method doesn't accept strings anymore, only arrays or `Traversable` objects.
I'll use `assertStringContainsString()` instead.

tests/phpunit/tests/test-admin-filters-post.php
Chouby: Indeed, I mixed things a bit. Post types and taxonomies are not unregistered at tear down (we sometimes register them is setup_before_class).

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -39,6 +39,13 @@ public function set_up() {
 		$GLOBALS['polylang']             = $this->pll_admin;
 	}
 
+	public function tear_down() {
+		parent::tear_down();
+		_unregister_post_type( 'doc' );
+		_unregister_post_type( 'article' );
+		_unregister_post_type( 'doc_not_translated' );
```
Chouby: I suggest to either:
- always use the same name to simplify code (each test requires only one cpt).
- or use names related to your test : hierarchical_translated, non_hierarchical_not_translated, etc...

At your convenience.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
 		$GLOBALS['hook_suffix'] = 'edit.php';
-		$_REQUEST['post_type']  = 'page';
+		$_REQUEST['post_type']  = $post_type;
 		set_current_screen();
 		$GLOBALS['wp_scripts'] = new WP_Scripts();
 		wp_default_scripts( $GLOBALS['wp_scripts'] );
-		do_action( 'admin_enqueue_scripts' );
+	}
 
+	private function get_admin_footer(): string {
```
Chouby:
```suggestion
	private function get_admin_footer_scripts(): string {
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
+		do_action( 'admin_enqueue_scripts' );
 		ob_start();
 		do_action( 'admin_print_footer_scripts' );
-		$footer = ob_get_clean();
+		return (string) ob_get_clean();
+	}
+
+	public function test_parent_pages_script_data_in_footer() {
```
Chouby:
```suggestion
	public function test_inline_script_for_pages() {
```
As a side note, I saw that the method `admin_enqueue_scripts()` still uses the old `wp_localize_script()`. We could use this PR to replace it with the more modern `wp_add_inline_script()`.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
 		$pages = array( 'en' => array( $en ), 'fr' => array( $fr ) );
 
-		$this->assertNotFalse( strpos( $footer, 'var pll_page_languages = ' . wp_json_encode( $pages ) ) );
+		$this->assertStringContainsString(
+			'var pll_page_languages = ' . wp_json_encode( $pages ),
+			$footer
+		);
+	}
+
+	public function test_hierarchical_cpt_script_data_in_footer() {
```
Chouby:
```suggestion
	public function test_inline_script_for_hierarchical_cpt() {
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
+
+		$this->set_current_edit_screen( 'doc' );
+		$footer = $this->get_admin_footer();
+
+		$result = array( 'en' => array( $en ) );
+
+		$this->assertStringContainsString( wp_json_encode( $result ), $footer );
+	}
+
+	public function test_non_hierarchical_cpt_script_data_in_footer() {
```
Chouby:
```suggestion
	public function test_inline_script_for_non_hierarchical_cpt() {
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
+		register_post_type( 'article', array( 'public' => true, 'hierarchical' => false ) );
```
Chouby: For the test to be complete, the post type must be translatable. Here there are 2 reasons not to add the inline script:
1. The post type is not hierarchical
2. The post type is not translated

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
+		$en = self::factory()->post->create( array( 'post_type' => 'article' ) );
+		self::$model->post->set_language( $en, 'en' );
+
+		$this->set_current_edit_screen( 'article' );
+		$footer = $this->get_admin_footer();
+
+		$this->assertStringNotContainsString( 'pll_page_languages', $footer );
+	}
+
+	public function test_hierarchical_untranslated_cpt_script_data_in_footer() {
```
Chouby:
```suggestion
	public function test_inline_script_for_hierarchical_untranslated_cpt() {
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
+		global $post;
+		$post = get_post( $post_id );
+		setup_postdata( $post );
+		// Now we trigger the dropdown build to check if it reuses the cached query.
+		ob_start();
+		$wp_list_table->inline_edit();
+		ob_end_clean();
+		$after_dropdown = $wpdb->num_queries;
+
+		$this->assertSame( $after_pll, $after_dropdown );
```
Chouby, line 591: I'm surprised there are no extra queries in `inline_edit()`. Nice...

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +471,116 @@ public function test_categories_script_data_in_footer() {
...
+		// so we call it before starting the count.
+		$wp_list_table = _get_list_table( 'WP_Posts_List_Table' );
+		$after_pll = $wpdb->num_queries;
+
+		// touch_time() (called internally by inline_edit() further) reads the global $post
+		// via get_post() with no argument — not set outside The Loop, so we set it up manually here.
+		$post_id = $post->ID;
+		global $post;
+		$post = get_post( $post_id );
+		setup_postdata( $post );
```
Chouby: You can simplify and remove 
```php
		$post_id = $post->ID;
		$post = get_post( $post_id );
```
as you already have the `$post` object. You can simply move the `global $post` a the top of the method.
See also: https://github.com/polylang/polylang/actions/runs/30439899079/job/90536434561?pr=1947#step:3:372
Avoid `setup_postdata()` which sets a lot of globals, probably unneeded for the test.
If `inline_edit()` requires more globals than only `$post`, then you can define them directly.

Chouby (changes requested):

Chouby (commented):
Just an additional idea `register_post_type()` + `add_filter( 'pll_get_post_types', ...` is used several times. You could create a new method `register_translated_post_type()`.

tests/phpunit/tests/test-admin-filters-post.php
accolette: No problem, but maybe as a separate issue instead? I'd rather do this after these two tickets are done, there's already a lot going on here, and it'll be clearer in my head to handle it on its own afterward.

src/admin/admin-filters-post.php
```diff
@@ -99,7 +99,12 @@ public function admin_enqueue_scripts() {
 
 		// Hierarchical post types
 		if ( 'edit' == $screen->base && is_post_type_hierarchical( $screen->post_type ) ) {
```
szepeviktor: Maybe add `&& post_type_supports( $screen->post_type, 'page-attributes' )` here

Chouby: Good catch :+1: 
@accolette See https://github.com/WordPress/WordPress/blob/7.0.2/wp-admin/includes/class-wp-posts-list-table.php#L1831

accolette: Sure, thank you !

szepeviktor (approved):

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +469,128 @@ public function test_categories_script_data_in_footer() {
...
 		$pages = array( 'en' => array( $en ), 'fr' => array( $fr ) );
 
-		$this->assertNotFalse( strpos( $footer, 'var pll_page_languages = ' . wp_json_encode( $pages ) ) );
+		$this->assertStringContainsString(
+			'var pll_page_languages = ' . wp_json_encode( $pages ),
+			$footer
+		);
+	}
+
+	private function register_post_type_for_dropdown_test( string $post_type, bool $hierarchical, bool $page_attributes = true, bool $translated = true ): void {
```
Chouby:
```suggestion
	private function register_post_type_for_dropdown_test( array $args ): void {
```
Given that you always use the same post type name, you can avoid this param and pass 'doc' directly to `register_post_type()`.

Then to improve the readability (boolean params are pretty unreadable especially when there are more than one), I suggest to accept an array of strings which could be `translated`, `hierarchical` and `page-attributes`.
In this method, you'll test the presence of each string in the array instead of the boolean params.
Thus your `@testWith` will be more readable with these strings.

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +469,128 @@ public function test_categories_script_data_in_footer() {
...
+		self::$model->post->set_language( $en, 'en' );
+
+		$this->set_current_edit_screen( 'doc' );
+		$footer = $this->get_admin_footer_scripts();
+
+		if ( $expects_output ) {
+			$this->assertStringContainsString( 'pll_page_languages', $footer );
+		} else {
+			$this->assertStringNotContainsString( 'pll_page_languages', $footer );
+		}
```
Chouby: I pushed you to use this assertion...
And now I'm suggesting something else to simplify:
```suggestion
		$this->assertSame( $expects_output, str_contains( $footer, 'pll_page_languages' ) );
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +469,128 @@ public function test_categories_script_data_in_footer() {
...
+	}
+
+	public function test_get_pages_query_is_cached_for_dropdown() {
+		global $post;
```
Chouby: It's preferable to have all globals at the beginning of the method.
```suggestion
		global $wpdb, $post;
```

tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -464,26 +469,128 @@ public function test_categories_script_data_in_footer() {
...
+
+		$this->register_post_type_for_dropdown_test( 'doc', true );
+
+		// touch_time() (called internally by inline_edit() further) reads the global $post
+		// via get_post() with no argument — not set outside The Loop, so we set it up manually here.
```
Chouby: We prefer this syntax for multiline comment (with only one star on the first line, when there 2 for doc blocks).
```suggestion
		/*
		 * touch_time() (called internally by inline_edit() further) reads the global $post
		 * via get_post() with no argument — not set outside The Loop, so we set it up manually here.
		 */
```

Chouby (changes requested):

PR #1949, opened 2026-08-03
Drop support of WP < 6.8

Chouby:
The PR raises the min WP version to 6.8 (released in April 2025) and thus removes some backward compatibility code.

### DIFF
--- a/js/src/post.js
+++ b/js/src/post.js
@@ -64,13 +64,8 @@ jQuery(
 											$.each(
 												terms,
 												function ( i ) {
-													// Backward compatibility with WordPress < 6.7.
-													// Support both old (WP < 6.7) and new (WP >= 6.7) ID formats.
-													// Old format: category-123 (WordPress < 6.7).
-													// New format: in-category-123-1 (WordPress >= 6.7).
 													const termId = pll_term_languages[ lg ][ tax ][ i ];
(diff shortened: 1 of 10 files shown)


szepeviktor:
All those changed versions remind me of my long time dream: https://github.com/polylang/wpml-to-polylang/pull/33

Hug0-Drelon (changes requested):
Should we remove these backward compatibilities as well ?
- https://github.com/polylang/polylang/blob/3.8.6/js/src/post.js#L67-L73
- https://github.com/polylang/polylang/blob/3.8.6/tests/phpunit/includes/testcase-preload-paths.php#L81-L84
- https://github.com/polylang/polylang/blob/3.8.6/tests/phpunit/tests/test-switcher-block-frontend.php#L70-L72

Hug0-Drelon:
Follow-up: removed remaining WP < 6.8 backward compatibility:

- `js/src/post.js`: keep only the WP ≥ 6.7 category checklist ID format
- `tests/phpunit/includes/testcase-preload-paths.php`: always set `WP_Block_Editor_Context::$name`
- `tests/phpunit/tests/test-switcher-block-frontend.php`: drop the WP < 6.8 test skip

Commit: e96b02ea

PR #1955, opened 2026-08-05
Add PHPUnit tests for Site Health (part 1)

accolette:
Fixes [#1125](https://github.com/polylang/polylang-pro/issues/1125)

## Why

Currently, there are no tests covering Site Health.

## What

Add PHPUnit tests to prevent regressions in the Site Health integration.

This PR covers the first four areas of the test plan. The tests for posts and terms are now performed through the public `info()` method rather than directly testing the internal helper methods.

The homepage Site Health test is also now only registered when the configured static front page actually exists. This avoids running the translation check when WordPress points to a page that has been deleted.

Following review feedback, the tests are split into namespaced classes under `tests/phpunit/tests/SiteHealth/`, sharing a common `TestCase` base class for the shared setup.

## Behaviour checked

The static front page behaviour was also checked manually by configuring a page as the homepage and then deleting it directly.

Without the new check, WordPress displays a 404 for the deleted homepage while Polylang's homepage Site Health test still reports a successful result.

The implementation now checks that the configured front page exists before registering the Polylang homepage Site Health test.

### `homepage_test` (`Front_Page_Test`)

* [x] All languages translated, missing translation, full structure
* [x] **Unhappy path** — the front page no longer exists

### `status_tests` (`Front_Page_Test`)

* [x] Static front page configured
* [x] Non-static front page
* [x] Empty `page_on_front`
* [x] **Unhappy path** — the homepage test is not added when the configured front page no longer exists
* [x] Existing Site Health tests are preserved when `pll_homepage` is added

### `info` — posts without language (`Posts_Test`)

* [x] Posts are returned grouped by post type
* [x] Posts with an existing language are excluded
* [x] **Unhappy path** — no warnings are added when no posts are missing a language

### `info` — terms without language (`Terms_Test`)

* [x] Terms are returned grouped by taxonomy
* [x] Terms with an existing language are excluded
* [x] **Unhappy path** — no warnings are added when no terms are missing a language

### `info_languages` (`Languages_Test`)

* [x] Existing `term_props` test, migrated as-is

## Planned in a follow-up PR
The remaining Site Health methods and cases will be covered in a follow-up PR:

* `info_languages` (remaining cases)
* `info` (remaining cases)
* `site_status_test_php_modules`
* `info_options`
* `get_flag`
* `exclude_options_keys`
* `exclude_language_keys`


tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
 		$this->assertSame( $en->get_tax_prop( 'term_language', 'term_taxonomy_id' ), $info['term_props']['value']['term_language/term_taxonomy_id'] );
 		$this->assertSame( $en->get_tax_prop( 'term_language', 'count' ), $info['term_props']['value']['term_language/count'] );
 	}
+
+	public function test_homepage_test_all_languages_translated() {
+		// Arrange
+		$home_pages = self::factory()->post->create_translated(
+			array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ),
+			array( 'post_title' => 'accueil', 'post_type' => 'page', 'lang' => 'fr' )
+		);
```
Hug0-Drelon:
```suggestion
		$home_pages = self::factory()->page->create_translated(
			array( 'post_title' => 'home', 'lang' => 'en' ),
			array( 'post_title' => 'accueil', 'lang' => 'fr' )
		);
```
You can use the page factory directly.

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		$this->assertSame( '<p>It is mandatory to translate the static front page in all languages.</p>', $test_result['description'], 'homepage_test() should have the expected description when all languages are translated.' );
+		$this->assertSame( '', $test_result['actions'], 'homepage_test() should have empty actions when all languages are translated.' );
+		$this->assertSame( 'pll_homepage', $test_result['test'], 'homepage_test() should have the expected test identifier.' );
+	}
+
+	public function test_homepage_test_missing_translation() {
+		// Arrange
+		$admin_id = self::factory()->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $admin_id );
```
Hug0-Drelon: Can you add a comment to specify that translation URLs require a connected user with correct rights?

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+
+		$home_en = self::factory()->post->create( array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ) );
```
Hug0-Drelon:
```suggestion
		$home_en = self::factory()->page->create( array( 'post_title' => 'home', 'lang' => 'en' ) );
```
Same suggestion as above.

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+			$test_result['description'],
+			'Description should contain a link targeting the French translation.'
+		);
+		$this->assertSame( '', $test_result['actions'], 'homepage_test() should have empty actions when a language is not translated.' );
+		$this->assertSame( 'pll_homepage', $test_result['test'], 'homepage_test() should have the expected test identifier.' );
+	}
+
+	public function test_homepage_test_when_front_page_does_not_exist() {
+		// Arrange
+		$deleted_page_id = self::factory()->post->create( array( 'post_type' => 'page', 'lang' => 'en' ) );
```
Hug0-Drelon:
```suggestion
		$deleted_page_id = self::factory()->page->create( array( 'lang' => 'en' ) );
```

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		wp_delete_post( $deleted_page_id, true );
+		update_option( 'show_on_front', 'page' );
+		update_option( 'page_on_front', $deleted_page_id );
+
+		// Act
+		$test_result = $this->site_health->homepage_test();
+
+		// Assert
+		$this->assertSame( 'good', $test_result['status'], 'homepage_test() should return "good" when the front page no longer exists.' );
```
Hug0-Drelon: I think we have a bug here. You test as behavior that should not occur IMHO. What do you think @Chouby?

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+	}
+
+	public function test_status_tests_adds_pll_homepage_test_when_static_front_page_is_set() {
+		// Arrange
+		$home_en = self::factory()->post->create( array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ) );
```
Hug0-Drelon:
```suggestion
		$home_en = self::factory()->page->create( array( 'post_title' => 'home', 'lang' => 'en' ) );
```

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		$this->assertSame(
+			array( $this->site_health, 'homepage_test' ),
+			$result['direct']['pll_homepage']['test'],
+			'The "pll_homepage" test should reference the homepage_test() callback.'
+		);
+	}
+
+	public function test_status_tests_preserves_existing_tests_when_static_front_page_is_set() {
+		// Arrange
+		$home_en = self::factory()->post->create( array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ) );
```
Hug0-Drelon:
```suggestion
		$home_en = self::factory()->page->create( array( 'post_title' => 'home', 'lang' => 'en' ) );
```

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		$post_no_lang_ids = $this->create_posts_without_lang( 2, 'post' );
+		$page_no_lang_ids = $this->create_posts_without_lang( 2, 'page' );
+		$post_en = self::factory()->post->create( array( 'post_type' => 'post', 'lang' => 'en' ) );
+
+		// Act
+		$result = $this->site_health->get_post_ids_without_lang();
+
+		// Assert :
+		$this->assertSame( array( 'post', 'page' ), array_keys( $result ), 'Result should be grouped by post type.' );
+		$this->assertEqualsCanonicalizing(
```
Hug0-Drelon: I suggest to use `assertSameSets` to compare arrays.

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		$result = $this->site_health->get_post_ids_without_lang( 2 );
+		$result_ids = explode( ',', $result['post'] );
+
+		// Assert
+		$this->assertCount( 2, $result_ids, 'Result should be limited to 2 post IDs when a custom limit is passed.' );
+		$this->assertEmpty(
+			array_diff( $result_ids, $post_no_lang_ids ),
+			'2 returned IDs should be among the created posts without language.'
+		);
+	}
```
Hug0-Drelon: Maybe we can use `@testWith`  for these three tests.

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		$result = $this->site_health->get_post_ids_without_lang( 0 );
+		$result_ids = explode( ',', $result['post'] );
+
+		// Assert
+		$this->assertCount( 7, $result_ids, 'A limit below 1 (other than the documented -1) should be treated as unlimited.' );
+	}
+
+	public function test_get_post_ids_without_lang_returns_empty_array_when_none_missing() {
+		// Arrange
+		self::factory()->post->create( array( 'post_type' => 'post', 'lang' => 'en' ) );
```
Hug0-Drelon:
```suggestion
		self::factory()->post->create( array( 'lang' => 'en' ) );
```
For the post factory, the default post type is `post`.

tests/phpunit/tests/test-admin-site-health.php
```diff
@@ -66,4 +69,379 @@ public function test_info_languages_term_props() {
...
+		// Arrange
+		$this->create_terms_without_lang( 7, 'category' );
+
+		// Act
+		$result = $this->site_health->get_term_ids_without_lang( 0 );
+		$result_ids = explode( ',', $result['category'] );
+
+		// Assert
+		$this->assertCount( 7, $result_ids, 'A limit below 1 (other than the documented -1) should be treated as unlimited.' );
+	}
```
Hug0-Drelon: Same suggestion as above, we could group these three tests using `@testWith`.

Hug0-Drelon (changes requested):
Here are general suggestions:
- remove `// Arrange // Act // Assert` comments
- split tests in separate classes and use a base test case to reuse methods, e.g. `Admin_Site_Health_Base`, `Admin_Site_Health_Languages` (existing test), `Admin_Site_Health_Front_Page`, `Admin_Site_Health_Posts` and `Admin_Site_Health_Terms`
- create a new folder for Site Health tests and move classes into it
- refactor `update_option( 'show_on_front', 'page' ); update_option( 'page_on_front', N );` in a new method such as `::set_page_on_front( `$page`, $show = 'page' )`.

tests/phpunit/tests/test-admin-site-health.php
accolette: I can't see $page proprerty in `PLL_UnitTest_Factory` or `WP_UnitTest_Factory`, if i apply it, it seems it doesn't work

Hug0-Drelon: oops, my bad. You're right.
I got mixed up with `PLL_UnitTest_Factory::$term|$category|$tag` properties...

tests/phpunit/tests/SiteHealth/TestCase.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace WP_Syntex\Polylang\Tests\SiteHealth;
```
Hug0-Drelon, line 3:
```suggestion
namespace WP_Syntex\Polylang\Tests\Site_Health;
```
We usually use snake case for our namespaces/classes.

tests/phpunit/tests/SiteHealth/test-Posts.php
```diff
@@ -0,0 +1,64 @@
...
+			explode( ',', $result['page'] ),
+			'Result should contain pages IDs without language.'
+		);
+		$this->assertStringNotContainsString( (string) $post_en, $result['post'], 'Result should not contain posts that already have a language.' );
+	}
+
+	/**
+	 * @testWith [-1, 7]
+	 *           [2, 2]
+	 *           [0, 7]
```
Hug0-Drelon, line 31: This case is weird, let's see if we should update production code so `$expected_count` is `7`. Ping @Chouby

tests/phpunit/tests/SiteHealth/TestCase.php
```diff
@@ -0,0 +1,88 @@
...
+
+	/**
+	 * @var PLL_Admin
+	 */
+	protected $pll_admin;
+
+	/**
+	 * @param WP_UnitTest_Factory $factory
+	 */
+	public static function wpSetUpBeforeClass( WP_UnitTest_Factory $factory ) {
```
Hug0-Drelon, line 25: Let's take the opportunity to use `pllSetUpBeforeClass` instead of `wpSetUpBeforeClass`.

You'll be able to use `$factory->language->create_many( 2 );`.

Hug0-Drelon (changes requested):

Hug0-Drelon (changes requested):
Can you rename `tests/phpunit/tests/SiteHealth/` folder into `tests/phpunit/tests/Site_Health`, so it matches `WP_Syntex\Polylang\Tests\Site_Health` namespace.

Hug0-Drelon (approved):
Moving to @Chouby to address these two remaining questions:
- https://github.com/polylang/polylang/pull/1955#discussion_r3750508096
- https://github.com/polylang/polylang/pull/1955#discussion_r3748554296

tests/phpunit/tests/Site_Health/TestCase.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param int    $number    Number of posts to create.
+	 * @param string $post_type Post type to use. Default 'post'.
+	 * @return int[] The created post IDs.
+	 */
+	protected function create_posts_without_lang( int $number, string $post_type = 'post' ): array {
+		$ids = array();
+
+		for ( $i = 0; $i < $number; $i++ ) {
+			$ids[] = self::factory()->post->create( array( 'post_type' => $post_type ) );
+		}
```
Chouby: There is a dedicated method `create_many()`. So maybe the whole method is not very useful. The same for terms.

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,107 @@
...
+
+		$test_result = $this->site_health->homepage_test();
+
+		$this->assertSame( array( 'label', 'status', 'badge', 'description', 'actions', 'test' ), array_keys( $test_result ), 'homepage_test() should return the expected array structure.' );
+		$this->assertSame( 'All languages have a translated homepage', $test_result['label'], 'homepage_test() should have the expected label when all languages are translated.' );
+		$this->assertSame( 'good', $test_result['status'], 'homepage_test() should have a "status" key set to "good".' );
+		$this->assertSame( array( 'label' => POLYLANG, 'color' => 'blue' ), $test_result['badge'], 'homepage_test() should have the expected badge.' );
+		$this->assertSame( '<p>It is mandatory to translate the static front page in all languages.</p>', $test_result['description'], 'homepage_test() should have the expected description when all languages are translated.' );
+		$this->assertSame( '', $test_result['actions'], 'homepage_test() should have empty actions when all languages are translated.' );
+		$this->assertSame( 'pll_homepage', $test_result['test'], 'homepage_test() should have the expected test identifier.' );
```
Chouby: In my opinion, it would more readable to have only one assertion on the whole array.
```php
$expected = array( ... );
$this->assertSameSetsWithIndex( $expected, $this->site_health->homepage_test() );
```

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,107 @@
...
+
+	public function test_homepage_test_missing_translation() {
+		// Translation URLs require a logged-in user with the correct capabilities.
+		$admin_id = self::factory()->user->create( array( 'role' => 'administrator' ) );
+		wp_set_current_user( $admin_id );
+
+		$home_en = self::factory()->post->create( array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ) );
+		$this->set_page_on_front( $home_en );
+
+		$this->pll_admin->links = new PLL_Admin_Links( $this->pll_admin );
```
Chouby, line 43: Why is it useful?

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,107 @@
...
+		$this->assertSame( 'critical', $test_result['status'], 'homepage_test() should have a "status" key set to "critical".' );
+		$this->assertSame( array( 'label' => POLYLANG, 'color' => 'blue' ), $test_result['badge'], 'homepage_test() should have the expected badge.' );
+		$this->assertStringContainsString( 'You must translate your static front page in', $test_result['description'], 'Description should mention the untranslated languages.' );
+		$this->assertStringContainsString( '>Français</a>', $test_result['description'], 'Description should contain a translation link for the missing language.' );
+		$this->assertMatchesRegularExpression( '/href="[^"]*new_lang=fr[^"]*"/', $test_result['description'], 'Description should contain a link targeting the French translation.' );
+		$this->assertSame( '', $test_result['actions'], 'homepage_test() should have empty actions when a language is not translated.' );
+		$this->assertSame( 'pll_homepage', $test_result['test'], 'homepage_test() should have the expected test identifier.' );
```
Chouby: Same as above.

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	public function test_homepage_test_when_front_page_does_not_exist() {
+		$deleted_page_id = self::factory()->post->create( array( 'post_type' => 'page', 'lang' => 'en' ) );
+		wp_delete_post( $deleted_page_id, true );
+		$this->set_page_on_front( $deleted_page_id );
+
+		$test_result = $this->site_health->homepage_test();
+
+		$this->assertSame( 'good', $test_result['status'], 'homepage_test() should return "good" when the front page no longer exists.' );
```
Chouby: Isn't this a bug?

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,107 @@
...
+				),
+			),
+		);
+
+		$result = $this->site_health->status_tests( $existing_tests );
+
+		$this->assertArrayHasKey( 'other_test', $result['direct'], 'Existing tests should not be overwritten.' );
+		$this->assertSame( $existing_tests['direct']['other_test'], $result['direct']['other_test'], 'Existing test data should remain unchanged.' );
+		$this->assertArrayHasKey( 'pll_homepage', $result['direct'], 'pll_homepage should still be added alongside existing tests.' );
+	}
```
Chouby, line 106: As an alternative, you could write a test with mote integration.
Call `WP_Site_Health::get_tests()` instead of our `status_test()` method.
This checks that the filter is applied and you'll see that many other tests are present in the returned value.

This is a way we often use to test actions and filters. Indeed, it's best practice to test only the public interface and not the internals of a class. In the case of actions and filters, callbacks must be public to allow WP to call our methods, but we could treat them as internal. Ex: changing the method name (also in `add_filter()`) would not beak the feature. A test calling directly `status_test()` would break. A test calling `WP_Site_Health::get_tests()` would not break.

If a breaking change is made to the WP calling method, the integration test will inform us.

tests/phpunit/tests/Site_Health/test-Posts.php
```diff
@@ -0,0 +1,64 @@
...
+namespace WP_Syntex\Polylang\Tests\Site_Health;
+
+class Posts_Test extends TestCase {
+
+	public function test_get_post_ids_without_lang_returns_posts_grouped_by_post_type() {
+		$post_no_lang_ids = $this->create_posts_without_lang( 2, 'post' );
+		$page_no_lang_ids = $this->create_posts_without_lang( 2, 'page' );
+		$post_en = self::factory()->post->create( array( 'post_type' => 'post', 'lang' => 'en' ) );
+
+		$result = $this->site_health->get_post_ids_without_lang();
```
Chouby: I would write the test without calling this method, which should be private. So I would rather call `info()`.

tests/phpunit/tests/Site_Health/test-Posts.php
```diff
@@ -0,0 +1,64 @@
...
+
+	/**
+	 * @testWith [-1, 7]
+	 *           [2, 2]
+	 *           [0, 7]
+	 *
+	 * @param int $limit          The limit to pass to get_post_ids_without_lang().
+	 * @param int $expected_count The expected number of returned post IDs.
+	 */
+	public function test_get_post_ids_without_lang_respects_limit( int $limit, int $expected_count ) {
```
Chouby: I would remove this test from here, as:
1. This should be a private method.
2. The feature is not useful as the limit is always 5. It could be hardcoded in the call to `get_terms_with_no_lang()`. 
 
In fact, it tests the internal `PLL_Model::get_objects_with_no_lang()` so this the method which should be tested (in a different place and probably in a different PR).

tests/phpunit/tests/Site_Health/test-Posts.php
```diff
@@ -0,0 +1,64 @@
...
+		$result_ids = explode( ',', $result['post'] );
+
+		$this->assertCount( $expected_count, $result_ids, "Result should contain $expected_count post IDs for limit $limit." );
+		$this->assertEmpty(
+			array_diff( $result_ids, $post_no_lang_ids ),
+			'All returned IDs should be among the created posts without language.'
+		);
+	}
+
+	public function test_get_post_ids_without_lang_uses_default_limit_when_no_argument_passed() {
```
Chouby: Same as above.

Chouby (changes requested):

tests/phpunit/tests/Site_Health/test-Front_Page.php
accolette: In this case, homepage_test() calls get_must_translate_message(), which builds a translation link via $this->links->get_new_post_translation_link() when a language is missing.
PLL_Admin::links is only set in init(), never called here, so it must be initialized manually to avoid a fatal error on this code path.

tests/phpunit/tests/Site_Health/test-Front_Page.php
accolette: Yes, this unhappy test reveals that behavior: if the front page is deleted, homepage_test() returns "good" instead of flagging the issue.

One thing worth noting: WordPress core already has a safeguard for this : _reset_front_page_settings_for_post(), hooked on before_delete_post since WP 3.7, resets show_on_front/page_on_front automatically when the linked page is deleted via wp_delete_post(). So this exact state shouldn't be reachable through a normal deletion; the test only reproduces it by setting page_on_front after the deletion, bypassing that safeguard.

Given that, do we still want to fix this behavior, or is documenting it with this test enough? And if we do want to fix it, should I do it directly in this PR, or track it in a separate issue?

accolette:
Unless I'm mistaken, all comments are fixed. Their is one remaining question here  : https://github.com/polylang/polylang/pull/1955#discussion_r3912016417

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,143 @@
...
+		$expected = array(
+			'label' => 'All languages have a translated homepage',
+			'status' => 'good',
+			'badge' => array(
+				'label' => POLYLANG,
+				'color' => 'blue',
+			),
+			'description' => '<p>It is mandatory to translate the static front page in all languages.</p>',
+			'actions' => '',
+			'test' => 'pll_homepage',
```
Chouby: Could you align double arrows? In Polylang, we exclude the rule check https://github.com/polylang/polylang/blob/3.8.7/phpcs.xml.dist#L33 because there is still quite a lot of legacy code not using it, but it's good practice to follow the coding standard anyway.

tests/phpunit/tests/Site_Health/test-Front_Page.php
Chouby: Ok I would move it to the setup together with the other global instances, even if useful only in this test.

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,143 @@
...
+		$this->assertStringContainsString(
+			'>Français</a>',
+			$result['description'],
+			'Description should contain a translation link for the missing language.'
+		);
+		$this->assertMatchesRegularExpression(
+			'/href="[^"]*new_lang=fr[^"]*"/',
+			$result['description'],
+			'Description should contain a link targeting the French translation.'
+		);
```
Chouby: I would match the entire description in one single assertion to simplify.

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,143 @@
...
+		$expected = array();
+		add_filter(
+			'site_status_tests',
+			function ( $tests ) use ( &$expected ) {
+				$expected = $tests;
+				return $tests;
+			},
+			1
+		);
+		$result = $site_health->get_tests();
```
Chouby: `get_tests()` is static. No need to instantiate `WP_Site_Health`.

Chouby (changes requested):

src/modules/site-health/admin-site-health.php
```diff
@@ -211,12 +211,18 @@ protected function get_flag( $language ) {
...
+			$page_on_front = (int) $page_on_front;
+
+			if ( get_post( $page_on_front ) ) {
+				$tests['direct']['pll_homepage'] = array(
+					'label' => esc_html__( 'Homepage translated', 'polylang' ),
+					'test'  => array( $this, 'homepage_test' ),
+				);
+			}
 		}
 		return $tests;
```
Chouby: Testing show_on_front first as it's useless to read the page option if this one is not what we expect.
Avoid pyramid steps.
```suggestion
		if ( 'page' !== get_option( 'show_on_front' ) ) {
			return $tests;
		}

		$page_on_front = get_option( 'page_on_front' );

		if ( ! is_numeric( $page_on_front ) || ! get_post( (int) $page_on_front ) ) {		
			return $tests;
		}

		$tests['direct']['pll_homepage'] = array(
			'label' => esc_html__( 'Homepage translated', 'polylang' ),
			'test'  => array( $this, 'homepage_test' ),
		);

		return $tests;
```

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,119 @@
...
+		$this->assertSame(
+			'<p>You must translate your static front page in <a href="">Français</a>.</p>',
+			$result['description'],
+			'homepage_test() should return the expected description.'
+		);
+	}
+
+	public function test_status_tests_does_not_add_homepage_test_when_front_page_does_not_exist() {
+		update_option( 'show_on_front', 'page' );
+		update_option( 'page_on_front', 999999 );
```
Chouby:
```suggestion
		update_option( 'page_on_front', 999999 ); // Intentionally big not to match an existing page. 
```

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,119 @@
...
+
+		$result = $this->site_health->homepage_test();
+
+		$this->assertSameSetsWithIndex(
+			$expected,
+			array_diff_key( $result, array( 'description' => true ) ),
+			'homepage_test() should return the expected array.'
+		);
+		$this->assertSame(
+			'<p>You must translate your static front page in <a href="">Français</a>.</p>',
```
Chouby: How does it work with the empty url in href? I would expect a complete url.

Chouby (changes requested):

tests/phpunit/tests/Site_Health/test-Front_Page.php
accolette: I moved the `PLL_Admin_Links` initialization to `set_up()` as suggested, but this causes an issue in `test_homepage_test_missing_translation()`.

`PLL_Admin_Links` stores the current user when it is instantiated. Since the administrator is created and set as the current user inside this specific test, it happens after `set_up()`. As a result, `PLL_Admin_Links` keeps the default user and the translation link is empty (`href=""`).

To get the expected translation URL, the administrator needs to be created and set as the current user before `PLL_Admin_Links` is instantiated. Since this is only required by this specific test, I kept the initialization there rather than duplicating it in the shared `set_up()`.

tests/phpunit/tests/Site_Health/test-Front_Page.php
accolette: I moved the PLL_Admin_Links initialization to set_up() as suggested, but this causes an issue in test_homepage_test_missing_translation().

PLL_Admin_Links stores the current user when it is instantiated. Since the administrator is created and set as the current user inside this specific test, it happens after set_up(). As a result, PLL_Admin_Links keeps the default user and the translation link is empty (href="").

To get the expected translation URL, the administrator needs to be created and set as the current user before PLL_Admin_Links is instantiated. Since this is only required by this specific test, I kept the initialization there rather than duplicating it in the shared set_up().

tests/phpunit/tests/Site_Health/test-Front_Page.php
Chouby: Sorry. This is not something that I expected to happen.
So you have to set the current user in `set_up()` too.
It's better to have a realistic test with the url.
NB: You can use `wp_set_current_user( 1 );` as 1 is the id of the admin created by default.

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,127 @@
...
+
+		$home_en = self::factory()->post->create( array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ) );
+		$this->set_page_on_front( $home_en );
+
+		/**
+		 * PLL_Admin_Links stores the current user when instantiated.
+		 * This test requires an administrator to generate the translation link,
+		 * so it must be initialized after setting the administrator as the current user.
+		 */
+		$this->pll_admin->links = new PLL_Admin_Links( $this->pll_admin );
```
Chouby: 1 is the default admin user ID.
I re-arrange to have related things together.
`/**` only for doc blocks. 
```suggestion
		$home_en = self::factory()->post->create( array( 'post_title' => 'home', 'post_type' => 'page', 'lang' => 'en' ) );
		$this->set_page_on_front( $home_en );

		/*
		 * We need PLL_Admin_Links to generate the link.
		 * PLL_Admin_Links requires a user.
		 */
		wp_set_current_user( 1 ); 
		$this->pll_admin->links = new PLL_Admin_Links( $this->pll_admin );
```

tests/phpunit/tests/Site_Health/test-Front_Page.php
```diff
@@ -0,0 +1,127 @@
...
+
+		$result = $this->site_health->homepage_test();
+
+		$this->assertSameSetsWithIndex(
+			$expected,
+			array_diff_key( $result, array( 'description' => true ) ),
+			'homepage_test() should return the expected array.'
+		);
+		$this->assertMatchesRegularExpression(
+			'/^<p>You must translate your static front page in <a href="[^"]*post-new\.php\?post_type=page&#038;from_post=' . $home_en . '&#038;new_lang=fr&#038;_wpnonce=[^"]+">Français<\/a>\.<\/p>$/',
```
Chouby: Let's hardcode everything we can. Only the nonce requires the regex:
```suggestion
		$this->assertMatchesRegularExpression(
			'%^<p>You must translate your static front page in <a href="http://example.org/wp-admin/post-new\.php\?post_type=page&#038;from_post=' . $home_en . '&#038;new_lang=fr&#038;_wpnonce=[^"]+">Français<\/a>\.<\/p>$%',
```

Chouby (changes requested):

PR #1962, opened 2026-08-07
Add assertions to check isolated tests behave as intended

Chouby:
We "isolated" 2 tests by running them in separate test suites before the other tests, in one case to make sure that the Polylang API is not loaded, in the other case to make sure that a constant is not set.  This PR asserts that the tests are executed in the correct conditions.

### DIFF
--- a/tests/phpunit/tests/plugins/test-model-maybe-create-language-terms.php
+++ b/tests/phpunit/tests/plugins/test-model-maybe-create-language-terms.php
@@ -12,6 +12,8 @@
 class Model_Maybe_Create_Language_Terms_Test extends PLL_UnitTestCase {
 
 	public function test_maybe_create_language_terms() {
+		$this->assertFalse( function_exists( 'PLL' ) );
+
 		// Translatable custom table.
 		require_once PLL_TEST_DATA_DIR . 'translatable.php';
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/plugins/test-model-maybe-create-language-terms.php
```diff
@@ -12,6 +12,8 @@
 class Model_Maybe_Create_Language_Terms_Test extends PLL_UnitTestCase {
 
 	public function test_maybe_create_language_terms() {
+		$this->assertFalse( function_exists( 'PLL' ) );
```
Hug0-Drelon:
```suggestion
		$this->assertFalse( function_exists( 'PLL' ), 'The test should run in a separate process' );
```
I suggest to add an error message to understand the intent under the assertion.

tests/phpunit/tests/test-install.php
```diff
@@ -133,6 +133,8 @@ public function test_should_hide_set_language_from_content_on_activation() {
 	 * This test must be executed before all uninstall tests.
 	 */
 	public function test_uninstall_without_removing_data() {
+		$this->assertFalse( defined( 'PLL_REMOVE_ALL_DATA' ) );
```
Hug0-Drelon:
```suggestion
		$this->assertFalse( defined( 'PLL_REMOVE_ALL_DATA' ), 'The test should run in a separate process' );
```
Same suggestion as above.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

PR #1963, opened 2026-08-07
fix(api): restore pll_the_languages() backward compatibility

Hug0-Drelon:
Fixes https://github.com/polylang/polylang-pro/issues/3084

## What?
- Restore `pll_the_languages()` list output to `<li>` elements only (no `<nav>`/`<ul>` wrapper).
- Default `show_wrapper` to `false` when omitted, with a `_doing_it_wrong()` notice telling developers to pass `show_wrapper` => false explicitly before the default switches to `true` in a near future.
- Restore the pre-3.9 default layout (`horizontal`) for the template tag.
- Keep the standard block wrapper rendering (`show_wrapper` forced to `true`; legacy `dropdown` attributes would otherwise disable it).
- Remove redundant `show_wrapper = true` assignment in the widget (already the `Settings` default).
- Add a WP 6.8 fixture for the navigation language switcher bad-CSS test.

## Why?
The switcher rework changed `pll_the_languages()` behavior:
- `show_wrapper` defaulted to `true`, breaking third-party code using `<ul><?php pll_the_languages(); ?></ul>`.
- `layout` defaulted to `vertical`, while Polylang 3.8.x used a horizontal list (see polylang/polylang-pro#3084).

## How?
- In `pll_the_languages()`, set `show_wrapper` to `false` when omitted (HTML output only) and trigger `_doing_it_wrong()` with migration guidance.
- Respect `show_wrapper` when it is passed explicitly.
- Set `layout` to `horizontal` when omitted and `dropdown` is not used.
- Force `show_wrapper` to `true` in the standard language switcher block render callback.

## Test plan
- [x] `npm run test:php -- --filter PLL_The_Languages_Test`
- [x] `npm run test:php -- --filter Switcher_Block_Test`
- [x] `npm run test:php -- --filter Switcher_Block_Frontend_Test`
- [x] List output has no wrapper and works when wrapped in a custom `<ul>`
- [x] `show_wrapper` => true displays the wrapper; omitted argument triggers `_doing_it_wrong()`
- [x] Default layout is `horizontal` (polylang/polylang-pro#3084)
- [x] Standard block list output is wrapped in a `<nav>` tag
- [x] Navigation block bad-CSS snapshot on WP 6.8


tests/phpunit/tests/test-pll-the-languages.php
```diff
@@ -291,8 +346,9 @@ public function test_should_return_select() {
 		$this->go_to( get_permalink( $posts['en'] ) );
 
 		$args = array(
-			'dropdown' => 1,
-			'echo'     => 0,
+			'dropdown'     => 1,
+			'show_wrapper' => true,
```
Marianne380: Why adding this?
Shouldn't we keep testing the legacy call unchanged to make sure its previous behavior is preserved?

Marianne380 (changes requested):

tests/phpunit/tests/test-pll-the-languages.php
Hug0-Drelon: You're right, we should. It even showed a bug... The `dropdown` legacy setting wasn't handled well (should convert to `select` layout with a wrapper).

Fixed in 556a7129.

Marianne380: That's what I thought.

src/api.php
```diff
@@ -22,11 +22,12 @@
  *     Optional switcher settings.
  *
  *     @type string   $layout                 Layout of the switcher. Possible values are `horizontal`, `vertical`,
- *                                            `dropdown`, and `select`. Default is `vertical`.
+ *                                            `dropdown`, and `select`. Default is `horizontal`.
  *     @type string   $alignment              Alignment of the items. Possible values are `left`, `center`, `right`,
  *                                            `stretched`. Default is `left` or `right`, depending on `is_rtl()`.
- *     @type bool     $show_wrapper           Display the wrapper or not. Default is `true` when the layout is `select`,
- *                                            and `false` otherwise.
+ *     @type bool     $show_wrapper           Display the wrapper or not. Default is `false`. Omitting this argument
```
Marianne380: Need update.

Marianne380 (changes requested):

src/api.php
Hug0-Drelon: Fixed in 794b8ec5

Marianne380 (approved):

src/api.php
```diff
@@ -65,6 +66,19 @@ function pll_the_languages( $args = array() ) {
...
 	}
 
+	if ( empty( $args['raw'] ) && ! isset( $args['show_wrapper'] ) ) {
+		_doing_it_wrong(
+			'pll_the_languages()',
+			'pll_the_languages() does not output a wrapper by default. Pass `show_wrapper` => false explicitly to keep this behavior. In a near future, pll_the_languages() will output a `<ul>` wrapper as well (i.e. `show_wrapper` will default to `true`).',
+			'3.9'
+		);
+		$args['show_wrapper'] = ! empty( $args['dropdown'] );
+	}
```
Chouby: I would suggest to also test `! isset( $args['layout'] ) && empty( $args['dropdown'] )` That way we would the message less often to new users.

Also don't you think we should handle https://github.com/polylang/polylang-pro/issues/3076 in this PR too? As the topic is very close and here we could avoid to deal with the wrapper for the legacy dropdown (as I expect the select tag to be always output, while we could optionnally output the div and label).

src/api.php
```diff
@@ -65,6 +66,19 @@ function pll_the_languages( $args = array() ) {
...
+
+	if ( ! isset( $args['layout'] ) && empty( $args['dropdown'] ) ) {
+		$args['layout'] = 'horizontal';
```
Chouby: Isn't it vertical by default today? Do we avoid CSS conflicts?

Chouby (changes requested):

src/api.php
Hug0-Drelon: It is for widgets and blocks, no `pll_the_languages` as per https://github.com/polylang/polylang-pro/issues/3084.

For CSS, bare list calls (`show_wrapper => false`) emit no layout classes, so no CSS conflict there IMHO.

src/api.php
Hug0-Drelon: > I would suggest to also test ! isset( $args['layout'] ) && empty( $args['dropdown'] ) That way we would the message less often to new users.

Good idea, done in dceea7cc4302b6b5e82256e7589dad03d5b0ccfd and abc770920c1edd04f334d802c30801cf1a36a32c.

Hug0-Drelon: > Also don't you think we should handle https://github.com/polylang/polylang-pro/issues/3076 in this PR too? As the topic is very close and here we could avoid to deal with the wrapper for the legacy dropdown (as I expect the select tag to be always output, while we could optionnally output the div and label).

It's a @Screenfeed choice if I'm correct. Wrapping with a `<div>` is usually harmless for CSS rules thanks to `pll-switcher-select` class attached to `<select>`.

TBH I don't understand why that's here in the first place.

src/api.php
Chouby: In my own tests, it's vertical by default. I don't know how @Chrystll obtained this result. But I guess in can be theme dependent.

More than that, I would expect that with `show_wrapper => false` none of our css rules apply. This means that vertical or horizontal should give the same result, and thus we should not need to specify.

src/api.php
Chouby: In my opinion it's a wrong idea to allow to output the select without the select tag. Only the div should be optional. Let's not repeat the error I made with the list.

To avoid CSS conflicts, we must not target the `pll-switcher-select` class as it may be already used by existing language switchers. We should target a class on the div instead.

Hug0-Drelon: See a proposal in [84e077f](https://github.com/polylang/polylang/pull/1963/commits/84e077f3622db4666bac3e82900cecb4eb107720).

src/api.php
Hug0-Drelon: See a proposal in [84e077f](https://github.com/polylang/polylang/pull/1963/commits/84e077f3622db4666bac3e82900cecb4eb107720).

src/api.php
```diff
@@ -65,6 +68,24 @@ function pll_the_languages( $args = array() ) {
...
+					'polylang'
+				),
+				'3.9'
+			);
+			$args['show_wrapper'] = false;
+		} elseif ( ! empty( $args['dropdown'] ) ) {
+			// 3.8 BC: bare `<select>`, no outer `<div>`.
+			$args['show_wrapper'] = false;
+		}
+		// `layout` => `select` without `show_wrapper` keeps Settings' default (`true`).
```
Chouby:
```suggestion
		if ( ! isset( $args['layout'] ) ) {
			// Backward compatibility with Polylang < 3.9.
			$args['show_wrapper'] = false;

			if ( empty( $args['dropdown'] ) ) {
				_doing_it_wrong(
					'pll_the_languages()',
					esc_html__(
						'pll_the_languages() does not output a wrapper by default. Pass `show_wrapper` => false explicitly to keep this behavior. In a near future, pll_the_languages() will output a `<ul>` wrapper as well (i.e. `show_wrapper` will default to `true`).',
						'polylang'
					),
					'3.9'
				);
			}
		}
```

Chouby (changes requested):

PR #1965, opened 2026-08-11
Fix WP Offload cache metadata synchronization

superdav42:
## Summary

- exclude WP Offload Media's volatile `amazonS3_cache` metadata from Polylang post-meta synchronization
- continue synchronizing the durable `amazonS3_info` and `as3cf_filesize_total` metadata required by the existing integration
- add regression coverage for the metadata allow/exclude behavior

## Problem

WP Offload Media updates `amazonS3_cache` while filtering front-end content. Polylang's synchronized-post handling can then copy that internal cache update to every translation.

On an uncached anonymous request from a multilingual WooCommerce site, this produced 624 `PLL_Sync_Metas` queries out of 960 total queries. Excluding the volatile cache reduced the same request to 314 queries and eliminated all 624 cache-sync queries.

`amazonS3_cache` is a derived URL-rewrite cache, unlike the offload records that this integration intentionally synchronizes, so translations can rebuild it independently.

## Files

- `src/integrations/wp-offload-media/as3cf.php`: remove `amazonS3_cache` from the metadata synchronization list
- `tests/phpunit/tests/plugins/test-wp-offload-media.php`: verify custom and durable offload metadata remain while the internal cache is excluded

## Verification

- `vendor/bin/phpcs src/integrations/wp-offload-media/as3cf.php tests/phpunit/tests/plugins/test-wp-offload-media.php`
- `vendor/bin/phpstan analyze --memory-limit=2000M`
- `vendor/bin/rector process src/integrations/wp-offload-media/as3cf.php tests/phpunit/tests/plugins/test-wp-offload-media.php --dry-run`
- direct method-level regression check under PHP

The repository's PHPUnit bootstrap was installed locally, but the integration test itself is left to CI because the local environment does not expose a disposable WordPress test database.

### DIFF
--- a/src/integrations/wp-offload-media/as3cf.php
+++ b/src/integrations/wp-offload-media/as3cf.php
@@ -37,6 +37,7 @@ class PLL_AS3CF {
 	 * @return array
 	 */
 	public function copy_post_metas( $metas ) {
+		$metas = array_diff( $metas, array( 'amazonS3_cache' ) );
 		$metas[] = 'amazonS3_info';
 		$metas[] = 'as3cf_filesize_total';
 		return $metas;
--- a/tests/phpunit/tests/plugins/test-wp-offload-media.php
+++ b/tests/phpunit/tests/plugins/test-wp-offload-media.php
@@ -1,0 +1,10 @@
+<?php
+
+class WP_Offload_Media_Test extends PLL_UnitTestCase {
+	public function test_copy_post_metas_excludes_internal_cache() {
+		$integration = new PLL_AS3CF();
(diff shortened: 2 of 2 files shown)


superdav42:
CI note: the only failing check is `WP 6.8.x + PHP 7.4`, in the unrelated `Switcher_Block_Frontend_Test::test_render_polylang_navigation_switcher_block` snapshot. The same matrix job already fails on this PR's base commit (`c5a889de`): https://github.com/polylang/polylang/actions/runs/31392306860/job/93466705262

The new WP Offload regression test passes across the matrix; all other PHPUnit, PHPCS/PHPStan, Playwright, ESLint, and typo checks passed.

PR #1967, opened 2026-08-11
fix(admin): hide language filter on Site Editor screen

Hug0-Drelon:
## What?
Hide the Polylang language filter in the admin bar on the Site Editor screen (`site-editor.php`).

## Why?
From WordPress 7.1, the admin bar is persistent in the Site Editor ([consistent navigation in WP 7.1](https://make.wordpress.org/core/2026/07/13/consistent-navigation-in-wordpress-7-1-with-persistent-toolbar/)). The language filter does not filter any content there and is misleading on Style and Templates screens.

Fixes polylang/polylang-pro#3080

## How?
Extend `PLL_Admin_Base::should_hide_admin_bar_menu()` to return `true` on `site-editor.php`, alongside existing post and term edit screens. Parameterize the PHPUnit test to cover post, term, and Site Editor pages.

## Test plan
- [ ] On a multilingual site, open the Site Editor and confirm the admin bar language filter is hidden
- [ ] Confirm the language filter still appears on list screens (e.g. Posts)
- [ ] Confirm the language filter is still hidden on post and term edit screens
- [x] Run `Admin_Test::test_admin_bar_menu_should_hide`

### DIFF
--- a/src/admin/admin-base.php
+++ b/src/admin/admin-base.php
@@ -612,21 +612,9 @@ abstract class PLL_Admin_Base extends PLL_Base {
 	 * @return bool
 	 */
 	public function should_hide_admin_bar_menu(): bool {
-		global $pagenow, $typenow, $taxnow;
+		global $pagenow;
 
-		if ( in_array( $pagenow, array( 'post.php', 'post-new.php' ), true ) ) {
-			return ! empty( $typenow );
-		}
-
-		if ( 'site-editor.php' === $pagenow ) {
-			return true;
-		}
-
-		if ( 'term.php' === $pagenow ) {
-			return ! empty( $taxnow );
-		}
-
-		return false;
+		return in_array( $pagenow, array( 'post.php', 'post-new.php', 'site-editor.php', 'term.php' ), true );
 	}
 
 	/**
--- a/src/admin/admin-base.php
+++ b/src/admin/admin-base.php
@@ -618,6 +618,10 @@ abstract class PLL_Admin_Base extends PLL_Base {
 			return ! empty( $typenow );
 		}
 
+		if ( 'site-editor.php' === $pagenow ) {
+			return true;
+		}
+
 		if ( 'term.php' === $pagenow ) {
 			return ! empty( $taxnow );
 		}
--- a/src/admin/admin-base.php
+++ b/src/admin/admin-base.php
@@ -605,7 +605,7 @@ abstract class PLL_Admin_Base extends PLL_Base {
 
 	/**
 	 * Tells if the Polylang's admin bar menu should be hidden for the current page.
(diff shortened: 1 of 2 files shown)


src/admin/admin-base.php
```diff
@@ -618,6 +618,10 @@ public function should_hide_admin_bar_menu(): bool {
 			return ! empty( $typenow );
 		}
 
+		if ( 'site-editor.php' === $pagenow ) {
```
Marianne380: Should we update the doc ?

Marianne380 (changes requested):

src/admin/admin-base.php
Hug0-Drelon: Good idea, done in ccffb906.

Marianne380 (approved):

Chouby (changes requested):
I would simplify the whole method with;
```php
return in_array( $pagenow, array( 'post.php', 'post-new.php', 'site-editor.php', 'term.php' ), true );
```
Indeed, I don't understand why we kept checking `$typenow` and `$taxnow`. Initially it was to remove the filter only for translated post types and taxonomies, but now looks like overkill checks.

Hug0-Drelon:
> I would simplify the whole method with;
> 
> ```
> return in_array( $pagenow, array( 'post.php', 'post-new.php', 'site-editor.php', 'term.php' ), true );
> ```
> 
> Indeed, I don't understand why we kept checking `$typenow` and `$taxnow`. Initially it was to remove the filter only for translated post types and taxonomies, but now looks like overkill checks.

Applied in [d927175](https://github.com/polylang/polylang/pull/1967/commits/d927175fc43cd5799179904a99239fad53e7ac1b).

tests/phpunit/tests/test-admin.php
```diff
@@ -75,13 +75,29 @@ public function test_admin_bar_menu_with_filtered_language() {
...
+	 *           [ "post.php", "post.php", "post" ]
+	 *           [ "site-editor.php", "site-editor.php", "" ]
+	 *           [ "term.php", "term.php", "category" ]
+	 *
+	 * @param string $pagenow The page now.
+	 * @param string $url     The URL of the page.
+	 * @param string $typenow The type now.
+	 * @return void
+	 */
+	public function test_admin_bar_menu_should_hide( $pagenow, $url, $typenow ) {
```
Chouby: We can now simplify the test by removing the 3rd param.

Chouby (changes requested):

tests/phpunit/tests/test-admin.php
Hug0-Drelon: Removed in be06e29d

PR #1969, opened 2026-08-12
Improve upsell when WooCommerce is detected

Chouby:
Fixes https://github.com/polylang/polylang-pro/issues/2862 about misreading
and replaces Polylang for WooCommerce by the Polylang Business Pack.

### DIFF
--- a/src/admin/admin-notices.php
+++ b/src/admin/admin-notices.php
@@ -240,8 +240,8 @@ class PLL_Admin_Notices {
 				<?php
 				printf(
 					/* translators: %1$s is link start tag, %2$s is link end tag. */
-					esc_html__( 'We have noticed that you are using Polylang with WooCommerce. To ensure compatibility, we recommend you use %1$sPolylang for WooCommerce%2$s.', 'polylang' ),
-					'<a href="https://polylang.pro/pricing/polylang-for-woocommerce/">',
+					esc_html__( 'We have noticed that you are using WooCommerce. To ensure compatibility, we recommend that you use the %1$sPolylang Business Pack%2$s.', 'polylang' ),
+					'<a href="https://polylang.pro/pricing/polylang-business-pack/">',
 					'</a>'
 				);
 				?>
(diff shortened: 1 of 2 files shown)


src/admin/admin-notices.php
```diff
@@ -240,8 +240,8 @@ private function pllwc_notice() {
 				<?php
 				printf(
 					/* translators: %1$s is link start tag, %2$s is link end tag. */
-					esc_html__( 'We have noticed that you are using Polylang with WooCommerce. To ensure compatibility, we recommend you use %1$sPolylang for WooCommerce%2$s.', 'polylang' ),
-					'<a href="https://polylang.pro/pricing/polylang-for-woocommerce/">',
+					esc_html__( 'We have noticed that you are using WooCommerce. To ensure compatibility, we recommend that you use the %1$sPolylang Business Pack%2$s.', 'polylang' ),
+					'<a href="https://polylang.pro/pricing/polylang-business-pack/">',
 					'</a>'
```
Screenfeed, line 245: Maybe we should use the same string in both places.

```php
/* translators: %s is the name of Polylang Business Pack product */
esc_html__( 'We have noticed that you are using WooCommerce. To ensure compatibility, we recommend that you use the %s.', 'polylang' ),
'<a href="https://polylang.pro/pricing/polylang-business-pack/">Polylang Business Pack</a>'
```

Another observation:
Admin notices don't tell who is printing them, so, sometimes they cannot be understood by users because they've been ripped off of their context. In this one:
- `We have noticed that...`: who is "We"?!
- `we recommend that you use the Polylang Business Pack`: oh it must be Polylang, but I had the information at the end of the sentence.

So, this is maybe something we could do:

```php
esc_html__( '%1$s: we have noticed... we recommend that you use the %2$s.', 'polylang' ),
esc_html( POLYLANG ),
'<a>...</a>'
```

What do you think?

Screenfeed (changes requested):

PR #1970, opened 2026-08-13
fix(switcher): render flag styles inline instead of enqueueing a stylesheet

Hug0-Drelon:
## What?
The language switcher now carries its flag styles inline, rendered where the flag markup is built (`Switcher\Element\Nav::get_label()` for the wrapping `<span>`, `Switcher\Element\Abstract_Element::set_flag()` for the ``).

`css/src/lib/switcher-flags.css` is no longer imported by the frontend stylesheets; only `navigation-language-switcher-editor.css` still imports it, since the block is not rendered by PHP in the editor.

Replaces https://github.com/polylang/polylang/pull/1944.

## Why?
Flags were not displayed correctly when the switcher was used only in a classic menu: no switcher stylesheet is enqueued in that case, so the flag rules never applied.

[#1944](https://github.com/polylang/polylang/pull/1944) fixed it by detecting that case and enqueueing the frontend styles from the menu. Rendering the styles alongside the markup removes that detection entirely and makes the flags correct wherever the switcher is rendered by PHP (classic menu, widget, shortcode, blocks), regardless of which stylesheets are enqueued.

## How?
- `Nav::get_label()` outputs the `pll-switcher-flag` span (width, aspect ratio, overflow, border radius) and the `pll-switcher-label` span (spacing) with inline styles.
- `Abstract_Element::set_flag()` already ran a `WP_HTML_Tag_Processor` on the flag to strip the size set by `PLL_Language::get_flag_html()`; it now replaces that `style` value with the image rules (`object-fit`, full width/height) instead of removing it.
- The block keeps its own flag customisations: `Abstract_Block::apply_flag_styles_to_markup()` appends its CSS variables to the existing `style` attribute instead of overwriting it.
- Both style blocks are commented to stay in sync with `css/src/lib/switcher-flags.css`.

## Test plan
- [ ] Classic menu with flags, flat list: flags sized and cropped correctly, correct spacing before names
- [ ] Classic menu with flags, dropdown: parent item and submenu items styled the same
- [ ] Classic menu with flags only (no names) and with codes
- [ ] Language switcher block (standard and navigation): flag width, border radius and label spacing controls still apply
- [ ] Language switcher widget and `pll_the_languages()` shortcode with flags
- [ ] `1:1` aspect ratio crops the flag instead of letterboxing it
- [ ] Block editor: flags still styled by `navigation-language-switcher-editor.css`


css/src/lib/switcher-flags.css
```diff
@@ -1,5 +1,8 @@
 /**
- * Shared flag styling for language switcher blocks.
+ * Flag styling for the language switcher block editor.
+ *
+ * Keep these styles in sync with the inline CSS in
+ * `src/Switcher/Element/Nav.php` (`Nav::get_label()`).
```
Screenfeed, line 5:
```suggestion
 * `src/Switcher/Element/Abstract_Element.php` (`Abstract_Element::set_flag()`)
 * and `src/Switcher/Element/Nav.php` (`Nav::get_label()`).
```

Screenfeed (changes requested):

Screenfeed:
A test fails because of `word-spacing: 30px;` being added. I think we've seen this already 🤔

Hug0-Drelon:
> A test fails because of `word-spacing: 30px;` being added. I think we've seen this already 🤔

It has been fixed in https://github.com/polylang/polylang/pull/1971.

Screenfeed (approved):

PR #1972, opened 2026-08-14
test(switcher): cover frontend assets enqueue for switcher and widget

Hug0-Drelon:
## What?
Adds PHPUnit coverage for frontend language switcher assets (`frontend-switcher.js` / `frontend-switcher.css`) via `pll_the_languages()` and the classic Languages widget.

Fixes https://github.com/polylang/polylang-pro/issues/3076

## Why?
Issue #3076 noticed that the legacy inline `<script>` on the select switcher is gone in 3.9. There is no production bug: the change handler now lives in the enqueued `frontend-switcher.js` file (Select/Dropdown layouts), and the classic widget still enqueues the shared CSS on `wp_enqueue_scripts`.

## How?
- Assert `pll_the_languages()` enqueues JS for `select` / `dropdown`, and does not for Nav layouts (`horizontal` / `vertical` / legacy `dropdown => 0`).
- Assert the classic widget enqueues CSS for all layouts, plus JS for `select` / `dropdown`.
- Guard against asset bleeding between tests (nothing enqueued before rendering).

## Test plan
- [x] Manually: place a Languages widget (select or dropdown) on the front, confirm `pll-language-switcher` CSS/JS load and language change works without an inline script in the markup
- [x] Manually: call `pll_the_languages( array( 'layout' => 'select' ) )` in a theme and confirm the footer script is present


Screenfeed (approved):
Just an idea: maybe `reset_asset_globals()` and `assert_frontend_assets_are_not_enqueued()` could be merged.

Chouby (changes requested):
I agree with @screenfeed. `assert_frontend_assets_are_not_enqueued()` should be merged in `reset_asset_globals()`.

Also now `Frontend_Assets_Test` and `Admin_Assets_Test` (together with the test case) are asymmetric.
`_test_scripts()` in `PLL_Assets_UnitTestCase` vs `assert_frontend_assets_are_enqueued_correctly()`
`get_polylang_assets()` once in the test case, once in the test file...

I propose to:
- move the test case and the test file in a seperate directory with namespaces.
- keep `assert_scripts_are_enqueued_correctly() (which I would rename to `assert_enqueued_assets()` to reduce the name length) in the test case together with an abstract `get_polylang_assets()`.
- Override `get_polylang_assets()` in the two test classes.
- Move `_test_scripts()` to `Admin_Assets_Test` and rename it `assert_admin_assets()` and for symmetry, rename `assert_frontend_assets_are_enqueued_correctly` to `assert_frontend_assets()`.

Hug0-Drelon:
> I agree with @Screenfeed. `assert_frontend_assets_are_not_enqueued()` should be merged in `reset_asset_globals()`.
> 
> Also now `Frontend_Assets_Test` and `Admin_Assets_Test` (together with the test case) are asymmetric. `_test_scripts()` in `PLL_Assets_UnitTestCase` vs `assert_frontend_assets_are_enqueued_correctly()` `get_polylang_assets()` once in the test case, once in the test file...
> 
> I propose to:
> 
> * move the test case and the test file in a seperate directory with namespaces.
> * keep `assert_scripts_are_enqueued_correctly() (which I would rename to `assert_enqueued_assets()`to reduce the name length) in the test case together with an abstract`get_polylang_assets()`.
> * Override `get_polylang_assets()` in the two test classes.
> * Move `_test_scripts()` to `Admin_Assets_Test` and rename it `assert_admin_assets()` and for symmetry, rename `assert_frontend_assets_are_enqueued_correctly` to `assert_frontend_assets()`.

Applied in 8a73121b with some other improvements.

PR #1976, opened 2026-08-19
[Language switcher rework] Harmonize sections in UI

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3061.

This harmonizes the UI with similar sections in classic widget, classic menu, content block, and navigation block.

1. Blocks: the "Flags Settings" section is moved one level down, at the same level than the "Display" and "Behavior" sections.
2. Classic: similar sections are added (new `tr`s for the widget, and new `fieldset` for the menu).

Note: the UI still shows/hides like before in classic widget: the system in place is generic enough to handle it without changing anything.

## Classic widget

Capture 2026-08-19 à 15 30 00
Capture 2026-08-19 à 15 30 26

## Classic menu

Capture 2026-08-19 à 15 31 33
Capture 2026-08-19 à 15 32 02

## Block

Capture 2026-08-19 à 15 35 39
Capture 2026-08-19 à 15 36 09


Hug0-Drelon (changes requested):
I stumbled upon a small bug when creating a class menu, the flag section is already expanded although "Display flags" is not active. See bellow:

Image

Screenfeed:
> I stumbled upon a small bug when creating a class menu, the flag section is already expanded although "Display flags" is not active.

Good catch. Fixed in [69685a0](https://github.com/polylang/polylang/pull/1976/commits/69685a0abf8b7ed19b869b52a734b476d0f62a01).

js/src/nav-menu.js
```diff
@@ -175,6 +173,43 @@ const pllNavMenu = {
 			);
 		},
 
+		/**
+		 * Creates and returns a wrapper row.
+		 *
+		 * @param {Object}  optionData           A field definition.
+		 * @param {boolean} {optionData.hide_if} Optional. "Hide row" conditions.
+		 * @param {Array}   menuValues           The input values.
+		 * @return {HTMLElement} The input element.
```
Hug0-Drelon:
```suggestion
		 * @return {HTMLElement|undefined} The input element.
```

js/src/nav-menu.js
```diff
@@ -175,6 +173,43 @@ const pllNavMenu = {
 			);
 		},
 
+		/**
+		 * Creates and returns a wrapper row.
+		 *
+		 * @param {Object}  optionData           A field definition.
+		 * @param {boolean} {optionData.hide_if} Optional. "Hide row" conditions.
+		 * @param {Array}   menuValues           The input values.
```
Hug0-Drelon:
```suggestion
		 * @param {Object}   menuValues           The input values.
```
Should we mention that the input is wrapped in a `<p>`?

js/src/nav-menu.js
```diff
@@ -175,6 +173,43 @@ const pllNavMenu = {
 			);
 		},
 
+		/**
+		 * Creates and returns a wrapper row.
+		 *
+		 * @param {Object}  optionData           A field definition.
+		 * @param {boolean} {optionData.hide_if} Optional. "Hide row" conditions.
```
Hug0-Drelon:
```suggestion
		 * @param {Object} {optionData.hide_if} Optional. "Hide row" conditions.
```

js/src/nav-menu.js
```diff
@@ -98,44 +97,43 @@ const pllNavMenu = {
...
-								wrapperAtts.class += ` pll-hidden-by-${ conditionName }`;
-							}
-						}
-					);
-				}
 
-				const inputWrapper = t.createElement( 'p', wrapperAtts );
+				// Create a section with title.
+				if ( optionData.section ) {
+					sectionWrapper = document.createElement( 'fieldset' );
```
Hug0-Drelon:
```suggestion
					sectionWrapper = t.createElement( 'fieldset' );
```
So we use the same API as the rest of the loop.

js/src/nav-menu.js
```diff
@@ -98,44 +97,43 @@ const pllNavMenu = {
...
+					settingsWrapper.append( sectionWrapper );
 
-				metabox.prepend( inputWrapper );
+					const sectionTitle = document.createElement( 'legend' );
```
Hug0-Drelon:
```suggestion
					const sectionTitle = t.createElement( 'legend' );
```

Hug0-Drelon (changes requested):

js/src/nav-menu.js
Screenfeed: In which case(s) does it return `undefined`?

Is it because of this?
```JS
for ( const [ conditionName, conditionValue ] of Object.entries(
	optionData.hide_if
) ) {
	if ( typeof pll_data.data[ conditionName ] === 'undefined' ) {
		// Should not happen.
		return;
	}
	...
```

I made a mistake here: I wanted to "continue", not to stop the function. So I'll replace this `return` by `continue`.

Screenfeed: Fixed in [0dbb76a](https://github.com/polylang/polylang/pull/1976/commits/0dbb76a61ef9775f8965d629e179310ac36bc65c).

js/src/nav-menu.js
Screenfeed: Yes, good idea. Done and fixed in [a76a2e8](https://github.com/polylang/polylang/pull/1976/commits/a76a2e8d738c021a45852a885fd2a8493dae24bc).

js/src/nav-menu.js
Screenfeed: Good catch. Done in [a76a2e8](https://github.com/polylang/polylang/pull/1976/commits/a76a2e8d738c021a45852a885fd2a8493dae24bc).

js/src/nav-menu.js
Screenfeed: Done in [697be19](https://github.com/polylang/polylang/pull/1976/commits/697be19adba5332db69d2bcee828b36d595484fb).

js/src/nav-menu.js
```diff
@@ -98,44 +97,43 @@ const pllNavMenu = {
...
-								wrapperAtts.class += ` pll-hidden-by-${ conditionName }`;
-							}
-						}
-					);
-				}
 
-				const inputWrapper = t.createElement( 'p', wrapperAtts );
+				// Create a section with title.
+				if ( optionData.section ) {
+					sectionWrapper = t.createElement( 'fieldset' );
```
Hug0-Drelon:
```suggestion
					sectionWrapper = t.createElement( 'fieldset', {} );
```
Otherwise we get the following JS error in the browser and the section won't render:
```
[Error] TypeError: Object.entries requires that input parameter not be null or undefined
	entries
	createElement (nav-menu.js:294)
	handleEvent (nav-menu.js:120)
```
Alternatively, we can use a default value for `pllNavMenu.createElement` `attribute  argument.

js/src/nav-menu.js
```diff
@@ -98,44 +97,43 @@ const pllNavMenu = {
...
+					settingsWrapper.append( sectionWrapper );
 
-				metabox.prepend( inputWrapper );
+					const sectionTitle = t.createElement( 'legend' );
```
Hug0-Drelon:
```suggestion
					const sectionTitle = t.createElement( 'legend', {} );
```
Follows previous fix.

Hug0-Drelon (changes requested):

js/src/nav-menu.js
Screenfeed: Crap, I didn't pay attention. I went for the optional parameter in [8cbffe8](https://github.com/polylang/polylang/pull/1976/commits/8cbffe810b84db535d4b6f295dcd19712ad8730b).

Hug0-Drelon (changes requested):
Can you update https://github.com/polylang/polylang/blob/c78912d3253dc38002df429fbd7400e03b53ed6c/.github/workflows/static-analysis.yml#L63 and add `js/src/nav-menu.js` (now the file is properly linted) ?

.github/workflows/static-analysis.yml
Hug0-Drelon:
```suggestion
    - name: Run ESLint
```

Hug0-Drelon (changes requested):

Hug0-Drelon (changes requested):
Unfortunately, there is a (new?) bug when setting flags to square. I get the following error in the browser:
```
SyntaxError: '[class*="pll-hidden-if-flag_aspect_ratio-"]:not(.pll-hidden-if-flag_aspect_ratio-1:1)' is not a valid selector.
```

With the following settings:

Image

Screenfeed:
Will check. It probably comes from `1:1`

Hug0-Drelon (approved):
LGTM 👍 

Could you rebase the PR onto `master`? so e2e CI is up-to-date.

Chouby:
Clean :+1:

PR #1977, opened 2026-08-20
[Language switcher rework] Fix old classic menu settings conversion

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3081.

## What

With a language switcher in classic menu, the menu item looses its settings after upgrading Polylang from 3.8.x to 3.9.

## Why

The conversion of the old language switcher settings to the new structure is not correctly done.

## How

User the right `Settings` class.

> [!NOTE]
> The case `dropdown = 1` + `hide_current = 1` is not possible in the classic menu with PLL 3.8.x, so `hide_current` will be `false` with PLL 3.9.
> To try this out with PLL 3.8.x:
> 1. Disable `dropdown`, the checkbox `hide_current` appears.
> 2. Enable `hide_current`.
> 3. Enable `dropdown`.
> 4. Save the menu.
> 5. Disable `dropdown`: the checkbox `hide_current` appears and is diabled.

### DIFF
--- a/src/admin/admin-nav-menu.php
+++ b/src/admin/admin-nav-menu.php
@@ -3,8 +3,8 @@
  * @package Polylang
  */
 
-use WP_Syntex\Polylang\Switcher\Settings\Settings;
 use WP_Syntex\Polylang\Switcher\Fields\Menu as Fields;
+use WP_Syntex\Polylang\Switcher\Settings\Menu as Settings;
 
 /**
  * Manages custom menus translations as well as the language switcher menu item on admin side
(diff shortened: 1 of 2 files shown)


Hug0-Drelon (approved):
I must say that it feels weird to test JS code in PHP 😅 

Otherwise LGTM 👍

Screenfeed:
It was much simpler to me 😅

Hug0-Drelon:
~~I've missed that but there is a regression in e2e tests.~~ Not linked to this PR IMHO.

PR #1979, opened 2026-08-21
[Language switcher rework] Support the custom HTML classes in menus

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3019.

## What

Block menu: in the "Advanced" section there is a field to add a custom HTML class. This class wasn't added to the markup.

This is also the case for the classic menu: there is a "CSS Classes" input that is hidden by default.

## Why

The feature is simpmly not implemented.

## How

Inline menu: add the classes to each `<li>`.
Dropdown menu: add the classes to the parent `<li>`. With the block menu, the classes are also added to the `<ul>`: this is because of the block `core/navigation-submenu`, which adds the same classes to these tags.

This PR also fixes the related tests.


src/modules/Blocks/Language_Switcher/Navigation/Block.php
```diff
@@ -121,7 +121,12 @@ public function render( $attributes, $content, $block ) {
 				$inner_nav_link_blocks[] = new WP_Block( $nav_link_block_args, $block->context );
 			}
 
-			$submenu_attributes               = $this->get_core_block_attributes( $attributes, $top_level_element );
+			$submenu_attributes = $this->get_core_block_attributes( $attributes, $top_level_element );
+
+			if ( ! empty( $attributes['className'] ) ) {
+				$submenu_attributes['className'] .= ' ' . $attributes['className'];
+			}
```
Hug0-Drelon: I'd move this to `get_core_block_attributes` to avoid duplication. Then unset `$attributes['className']` before calling `get_core_block_attributes` for submenu `<li>`.

We can even go a bit further and move `$submenu_attributes['className'] .= ' ' . wp_apply_generated_classname_support( $block->block_type )['class'];` into `get_core_block_attributes` if we accept to pass the whole `$block` `WP_Block` object instead of `$attributes` (the latter being accessible from the former, see https://github.com/WordPress/wordpress-develop/blob/9b24d2d08d47c4fdba9e8131deedb025809c0be8/src/wp-includes/class-wp-block.php#L223-L236).

Hug0-Drelon (changes requested):

src/modules/Blocks/Language_Switcher/Navigation/Block.php
Screenfeed: > I'd move this to `get_core_block_attributes` to avoid duplication. Then unset `$attributes['className']` before calling `get_core_block_attributes` for submenu `<li>`.

Good idea. Done in [d1e2728](https://github.com/polylang/polylang/pull/1979/commits/d1e272875676eccf8d1bcafb777899a3eafb7d67).

> We can even go a bit further and move `$submenu_attributes['className'] .= ' ' . wp_apply_generated_classname_support( $block->block_type )['class'];` into `get_core_block_attributes` if we accept to pass the whole `$block` `WP_Block` object instead of `$attributes` (the latter being accessible from the former, see [WordPress/wordpress-develop@`9b24d2d`/src/wp-includes/class-wp-block.php#L223-L236](https://github.com/WordPress/wordpress-develop/blob/9b24d2d08d47c4fdba9e8131deedb025809c0be8/src/wp-includes/class-wp-block.php#L223-L236)).

I didn't apply this because I think it would complicate things: with the dropdown layout we would need to differentiate the top level item from the sub-items. However, I moved the call to `wp_apply_generated_classname_support()` to the top of the method, so it is called only once (it was called for each element with the "inline" layout).

Hug0-Drelon: >  I moved the call to `wp_apply_generated_classname_support()` to the top of the method

Good idea 👍

src/modules/Blocks/Language_Switcher/Navigation/Block.php
```diff
@@ -103,26 +103,30 @@ public function render( $attributes, $content, $block ) {
...
 		if ( 'dropdown' === $settings->layout ) {
 			$top_level_element = $this->get_top_level_element( $settings );
 
 			if ( empty( $top_level_element ) ) {
 				return '';
 			}
 
-			$inner_nav_link_blocks = array();
+			$inner_nav_link_blocks   = array();
+			$attributes_wo_classname = array_diff_key( $attributes, array( 'className' => '' ) );
```
Hug0-Drelon, line 116: Can't we do `$this->get_core_block_attributes( unset( $attributes['className'] ), $element )` instead?

src/modules/Blocks/Language_Switcher/Navigation/Block.php
```diff
@@ -103,26 +103,30 @@ public function render( $attributes, $content, $block ) {
 			return '';
 		}
 
+		$block_class_name = wp_apply_generated_classname_support( $block->block_type )['class'];
```
Hug0-Drelon:
```suggestion
		$attributes['className'] .= '  ' . wp_apply_generated_classname_support( $block->block_type )['class'];
```
This would allow to get rid of `$xxx['className'] .= ' ' . $block_class_name;` done twice.

Hug0-Drelon (changes requested):

src/modules/Blocks/Language_Switcher/Navigation/Block.php
Screenfeed:
```
function unset(mixed $var, mixed ...$vars): void
```

`unset()` [returns `void`](https://www.php.net/manual/en/function.unset.php).

src/modules/Blocks/Language_Switcher/Navigation/Block.php
Screenfeed: Nice, I didn't see this one. Done in [bdcf9d3](https://github.com/polylang/polylang/pull/1979/commits/bdcf9d36a2852f262645483e3507728559281ad5).

Note: `$attributes['className']` seems to always be set, so no need to test its existence.

src/modules/Blocks/Language_Switcher/Navigation/Block.php
Hug0-Drelon: Or should I say; RTFM 😅

Hug0-Drelon (approved):

.editorconfig
```diff
@@ -10,6 +10,9 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = tab
 
+[*.html]
+insert_final_newline = false
```
Hug0-Drelon, line 14: 🥇

Hug0-Drelon (approved):

PR #1980, opened 2026-08-21
Hide the legacy widget in the block widgets screen

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3071.

## What

When displaying the block widgets screen, searching widgets with "lang" will display 2 widgets side-by-side: the classic one and the block one.
Polylang Pro hides the classic widget, so only the block widget remains.

## Why

Because reasons. And it's Friday.

## How

We decided to backport what is done in PLL Pro: hide the classic widget. We can't see value in keeping them both. Note: If a classic widget is already in use, it will still work, we just won't be able to add a new one.

### DIFF
--- a/tests/phpunit/tests/test-widget-languages.php
+++ b/tests/phpunit/tests/test-widget-languages.php
@@ -182,6 +182,14 @@ class Widget_Languages_Test extends PLL_UnitTestCase {
 		$this->assertMatchesRegularExpression( "/id=[\"']{$prefix}-hide_if_no_translation[\"'][^>]* checked=[\"']checked[\"']/", $form );
 	}
 
+	public function test_legacy_widget_is_hidden() {
+		$this->init_frontend();
+		do_action( 'widgets_init' );
+
+		$settings = get_legacy_widget_block_editor_settings();
+		$this->assertContains( 'polylang', $settings['widgetTypesToHideFromLegacyWidgetBlock'] );
+	}
+
 	/**
 	 * Stores the widget's options into the database.
 	 *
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-widget-languages.php
```diff
@@ -182,6 +182,14 @@ public function test_display_form_with_non_default_settings(): void {
 		$this->assertMatchesRegularExpression( "/id=[\"']{$prefix}-hide_if_no_translation[\"'][^>]* checked=[\"']checked[\"']/", $form );
 	}
 
+	public function test_legacy_widget_is_hidden() {
+		$this->init_frontend();
+		do_action( 'widgets_init' );
```
Hug0-Drelon, line 187: Can we simplify and do `new \WP_Syntex\Polylang\Widgets\Languages()` instead?

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-widget-languages.php
Screenfeed: This works:

```php
$this->init_frontend();
new Languages( $GLOBALS['polylang'] );
```

But I think we should keep `do_action( 'widgets_init' )`:
1. Your suggestion is not simpler in the end because we still need to create the `PLL_Frontend` instance (and the PLL object must be passed by reference so it cannot be inlined).
2. I think it's better to use `do_action()` than calling the hook callback directly in general, unless we cannot do otherwise.

Hug0-Drelon (approved):

tests/phpunit/tests/test-widget-languages.php
Chouby, line 187: Is it possible to replace the `init_frontend()` and `init_admin()` from this class by [context classes instances](https://github.com/polylang/polylang/tree/3.8.7/tests/phpunit/includes/context)?

Chouby (changes requested):

tests/phpunit/tests/test-widget-languages.php
Screenfeed: I couldn't find a way. Two of the four tests are constantly failing :(

PR #1982, opened 2026-08-24
chore(deps): drop @wordpress/scripts and add Dependabot

Hug0-Drelon:
## What?
- Replace `tests/e2e/package.json` with an ESM stub and keep Playwright deps on the plugin root.
- Lint the root `package.json` with `npmPkgJsonLint` instead of `wp-scripts`.
- Remove `@wordpress/scripts`.
- Add Dependabot for npm (production + named build tools in this repo) and Composer (`require` only).
- Request a review on Dependabot PRs (`dependabot-reviewer.yml`) instead of assigning. Comment when the bump is a production dependency.

## Why?
`lint:js` already uses ESLint. The leftover `@wordpress/scripts` install is unused, noisy for Dependabot, and package-json lint should not recurse into the e2e stub. Dependabot no longer supports a `reviewers` key, so review requests are a workflow.

## How?
Standalone `npm-package-json-lint` + `@wordpress/npm-package-json-lint-config`, ESLint `packageDir` for e2e imports, weekly Dependabot (Monday 07:00 Europe/Paris). `pull_request_target` requests `@Hug0-Drelon` on `dependabot[bot]` PRs and uses `dependabot/fetch-metadata` (`direct:production`) for the production comment.

## Test plan
- [x] `npm ci` succeeds at the repo root
- [x] `npm run lint:pkg-json` lints only `./package.json`
- [x] `npm run lint:js -- tests/e2e` does not flag root Playwright/e2e-utils imports
- [x] `npm run build` succeeds
- [x] No `wp-scripts` / `@wordpress/scripts` in package.json or workflows
- [ ] After merge, a Dependabot PR requests review; a production bump also gets the comment

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,7 @@
-# Same schedule and allow lists across Polylang plugins; Dependabot does not sync versions between repos.
+# Same schedule across Polylang plugins; Dependabot does not sync versions between repos.
+# allow — if a package does not match, Dependabot does not open a version PR or a security PR for it.
+# groups.build-tools — among packages that already passed allow, matching ones are collapsed into one PR instead of many.
+# open-pull-requests-limit: 3 — caps how many version-update PRs stay open (assignee and Monday 07:00 Europe/Paris are just process).
 version: 2
 updates:
   - package-ecosystem: npm
@@ -37,7 +40,9 @@ updates:
         patterns:
           - webpack
           - webpack-cli
-          - "*-loader"
+          - babel-loader
+          - css-loader
+          - sass-loader
           - sass
           - postcss
           - mini-css-extract-plugin
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,66 @@
+# Same schedule and allow lists across Polylang plugins; Dependabot does not sync versions between repos.
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    exclude-paths:
+      - tests/e2e/**
+    assignees:
+      - Hug0-Drelon
+    open-pull-requests-limit: 3
+    schedule:
+      interval: weekly
+      day: monday
+      time: "07:00"
+      timezone: Europe/Paris
+    allow:
+      - dependency-type: production
+      - dependency-name: webpack
+      - dependency-name: webpack-cli
+      - dependency-name: babel-loader
+      - dependency-name: "@wordpress/babel-preset-default"
+      - dependency-name: css-loader
+      - dependency-name: sass
+      - dependency-name: sass-loader
+      - dependency-name: postcss
+      - dependency-name: mini-css-extract-plugin
+      - dependency-name: css-minimizer-webpack-plugin
+      - dependency-name: terser-webpack-plugin
+      - dependency-name: clean-webpack-plugin
+      - dependency-name: copy-webpack-plugin
+      - dependency-name: "@wordpress/base-styles"
+      - dependency-name: "@wpsyntex/polylang-build-scripts"
+      - dependency-name: "@wpsyntex/distribute"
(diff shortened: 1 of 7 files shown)


Hug0-Drelon:
Dependabot process:

- **`allow`** — if a package does not match, Dependabot does not open a version PR or a security PR for it.
- **`groups.build-tools`** — among packages that already passed `allow`, matching ones are collapsed into one PR instead of many.

Weekly schedule (Monday 07:00 Europe/Paris) is process only.

Hug0-Drelon:
Rough picture of dropping `@wordpress/scripts` (`origin/master` vs this PR, from `package-lock.json` package entries). Current `node_modules` is after the shrink; `master` was not reinstalled to weigh the old tree.

| | Lockfile | Packages in the tree |
|---|---|---|
| This PR | 677 → 430 KiB (−37%) | 1919 → 1356 (**−563**, −29%) |

On disk now: about **1.0G**. If that tracked the package count, on the order of a few hundred MB less.

Ballpark: **~30% fewer packages**, **~⅓ smaller lockfile**. Webpack/ESLint/`@wordpress/eslint-plugin` still pull a large tree. What went away is the extra Jest/Puppeteer/markdown/stylelint stack that lived only under `@wordpress/scripts`.

.github/workflows/dependabot-reviewer.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Request review
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh pr edit "${{ github.event.pull_request.number }}" --add-reviewer Hug0-Drelon
+
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
```
Marianne380: Why not `v3` ?

.github/dependabot.yml
```diff
@@ -0,0 +1,62 @@
...
+  - package-ecosystem: npm
+    directory: /
+    exclude-paths:
+      - tests/e2e/**
+    schedule:
+      interval: weekly
+      day: monday
+      time: "07:00"
+      timezone: Europe/Paris
+    allow:
```
Marianne380, line 15: Just curious about the allow list: how did you decide which dev/build dependencies should be included here?

Marianne380 (changes requested):

.github/dependabot.yml
Hug0-Drelon: Rule of thumb: Dependabot should only bump what either ships with the plugin or changes the compiled JS/CSS we ship.

- dependency-type: `production` covers runtime deps (`@wpsyntex/polylang-react-library`).
- The named packages are the webpack stack we call from `npm run build` (webpack, loaders/plugins, Sass/PostCSS, `@wordpress/babel-preset-default`, `@wordpress/base-styles`, `@wpsyntex/polylang-build-scripts`, plus `glob` / `rimraf` used by that pipeline). Those are `devDependencies` in npm, so they would be skipped without an explicit allow.
Everything else in `devDependencies` is lint, tests, wp-env, or the release helper (`@wpsyntex/distribute`). We still update those when we need to, but not on a weekly Dependabot cadence. Same idea as Composer: require only.

`groups.build-tools` does not add packages; it only collapses the ones already allowed into a single PR.

Marianne380: Makes sense, thanks for the explanation!

.github/workflows/dependabot-reviewer.yml
Hug0-Drelon: 🙈 

Bumped in c41be8a984e8206c7750eab64c143eb2ae828092.

Marianne380 (approved):

PR #1985, opened 2026-08-27
Assert `map_meta_cap` globally

Screenfeed:
This adds an assertion to the hook `map_meta_cap`, to make sure that if a 2nd argument is passed to `current_user_can()`, it's an integer (not an object, like a post). Some exceptions are handled in the assertion.

This seems to be the best way to monitor OUR use of `user_can()`: monitor every call to `user_can()` and exclude some 3rd parties.

### DIFF
--- a/tests/phpunit/includes/map-meta-cap-trait.php
+++ b/tests/phpunit/includes/map-meta-cap-trait.php
@@ -1,0 +1,36 @@
+<?php
+
+/**
+ * Trait that makes sure that `user_can()` is used with a post ID as argument, and not a post object.
+ */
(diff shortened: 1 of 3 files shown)


Hug0-Drelon (approved):
LGTM 👍 

Do you confirm it fails on Polylang Pro `master` (initial issue)?

Screenfeed:
It fails on PLL if I remove `->ID` [here](https://github.com/polylang/polylang/blob/58dfafdf301eb58dc3ff922befc9f245ac67a99d/src/admin/admin-classic-editor.php#L193).
I'm about to commit some changes to PLL Pro, to cover more things with this, but I confirm it fails with the initial issue.

PR #1989, opened 2026-09-01
Fix issues reported by PHPStan

Screenfeed:
Fixes https://github.com/polylang/polylang/actions/runs/33483746075/job/99778904177.

## Details about some choices I made

```php
wp_register_script( 'foobar', plugins_url( ... ), ... );
```

`wp_register_script()` now expects its 2nd parameter (`$src`) to be a non empty string. The issue is that `plugins_url()` returns a string that can be empty: while it is unlikely to return an empty string, it's technically possible because of `apply_filters()`. Solution: if this happen, the script is not registered.

```php
wp_die( wp_json_encode( $return ) );
```

`wp_json_encode()` returns [`string|false`](https://developer.wordpress.org/reference/functions/wp_json_encode/#return). `wp_die()` [expects `string|WP_Error|int`](https://developer.wordpress.org/reference/functions/wp_die/#parameters), meaning that `false` is the issue.
First attempt: `wp_die( (string) wp_json_encode( $return ) )`. In that case, PHPCS asks for the string received by `wp_die()` to be escaped.
Chosen solution: `wp_die( wp_json_encode( $return ) ?: '[]' )`.

```php
count( $query->query )
```

`$query->query`'s type is `array|null`, so the solution is simple: `count( (array) $query->query )`.

```php
$cat = $cat->error_data['term_exists'] ?? $cat['term_id'];
```

`$cat`'s type is `array|WP_Error` (Schrödinger would say it can be both at the same time), so I chose to make it explicit instead of toying with mixed types on a one-liner.


src/default-term.php
```diff
@@ -139,16 +139,34 @@ public function create_default_term( $lang, $taxonomy ) {
...
 		// Check that the term was not previously created (in case the language was deleted and recreated).
-		$cat = $cat->error_data['term_exists'] ?? $cat['term_id'];
+		if ( is_wp_error( $cat ) ) {
+			if ( empty( $cat->error_data['term_exists'] ) ) {
+				return;
+			}
+			$cat = $cat->error_data['term_exists'];
+		} else {
+			$cat = $cat['term_id'];
+		}
```
Hug0-Drelon:
```suggestion
		if ( is_wp_error( $cat ) ) {
			return;
		}
		$cat = $cat['term_id'];
```
I don't think `wp_inster_term` returns a `WP_Error` containing term data when the term is inserted according to https://developer.wordpress.org/reference/functions/wp_insert_term/#return.

Hug0-Drelon (changes requested):

src/default-term.php
Screenfeed: You may have missed it:
```php
return new WP_Error( 'term_exists', __( 'A term with the name provided already exists with this parent.' ), $existing_term->term_id );
```

https://github.com/WordPress/wordpress-develop/blob/0d401064ee33074244ccfa3d7bd5a1201d5db237/src/wp-includes/taxonomy.php#L2601-L2604

Screenfeed: Note: we could also use `$cat->get_error_data( 'term_exists' )`, which is a bit cleaner.

What do you think about this?

```php
if ( is_array( $cat ) ) {
	$cat = $cat['term_id'];
} elseif ( ! empty( $cat->get_error_data( 'term_exists' ) ) ) {
	$cat = $cat->get_error_data( 'term_exists' );
} else {
	return;
}
```

Screenfeed: Done in [72e5b97](https://github.com/polylang/polylang/pull/1989/commits/72e5b97019417770fae0c9237545bd6794389254) but don't hesitate to tell me if you have a better idea.

Hug0-Drelon: `get_error_data` is better indeed.

Hug0-Drelon (approved):

src/Widgets/Languages.php
```diff
@@ -169,11 +169,11 @@ public function update( $new_instance, $old_instance ) { // phpcs:ignore Variabl
 	 * @since 3.9
 	 *
 	 * @param array $instance Current settings.
-	 * @return void
+	 * @return null
```
Chouby, line 172: I don't understand this change as it goes against parent class doc https://github.com/WordPress/WordPress/blob/7.1/wp-includes/class-wp-widget.php#L141 (PHPStan is wrong here).

src/Switcher/Assets.php
```diff
@@ -61,6 +61,7 @@ public static function register_styles(): void {
 	 */
 	public static function register_scripts(): void {
 		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
-		wp_register_script( self::FRONTEND_ASSET_HANDLE, plugins_url( "/js/build/frontend-switcher{$suffix}.js", POLYLANG_ROOT_FILE ), array(), POLYLANG_VERSION, true );
+		$src    = plugins_url( "/js/build/frontend-switcher{$suffix}.js", POLYLANG_ROOT_FILE );
+		wp_register_script( self::FRONTEND_ASSET_HANDLE, $src ?: false, array(), POLYLANG_VERSION, true );
```
Chouby: `pll_register_script` was created to avoid us this kind of pain. We could use it now everywhere you modified the call to `wp_register_script()`

src/functions.php
```diff
@@ -225,13 +227,16 @@ function pll_enqueue_script( string $name, array $dependencies = array(), array
...
  * @param string   $prefix       Optional. Prefix to use for the script handle. Default 'pll-'.
  * @return void
+ *
+ * @phpstan-param non-empty-string $name
  */
 function pll_register_script( string $name, array $dependencies = array(), array $args = array(), string $prefix = 'pll-' ): void {
 	$suffix  = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
 	$file    = '/js/build/' . $name . $suffix . '.js';
 	$src     = plugins_url( $file, POLYLANG_ROOT_FILE );
+	$src     = $src ?: false;
```
Chouby: If we don't have a src, there is nothing to register. Shouldn't we stop the execution instead of calling `wp_register_script()`?

Chouby (changes requested):

src/Switcher/Assets.php
Screenfeed: Yes but that means we must rename some files or their handler. For example here, the handler is `pll-language-switcher` while the file is named `frontend-switcher`.
Isn't this something we should do in another PR?

src/Widgets/Languages.php
Screenfeed: From my understanding, the line you're pointing proves the opposite.
The parent doc says:

```
@return string|null Default return is 'noform'. A subclass may opt to return null.
```

So PHPStan is right, our method should return `string` or `null`, not `void`. IMHO, the doc is wrong, the parent method should return `string|void`.

src/functions.php
Screenfeed: Done in [0006541](https://github.com/polylang/polylang/pull/1989/commits/0006541c675410d0eda52e94876fa46923ecfef7).

src/Switcher/Assets.php
Screenfeed: Done in [0fd8379](https://github.com/polylang/polylang/pull/1989/commits/0fd83790c081d316c0756058a3010d4010da67b0).
Note: there are still 20-ish calls to `wp_register_script()` and its friends.

src/Widgets/Languages.php
Chouby: :eyeglasses:

PR #1992, opened 2026-09-02
Fix duplicated posts page

Screenfeed:
Fixes https://github.com/polylang/polylang-pro/issues/3109.

In short: when the front page is supposed to be `example.com/fr/front-page/` (`redirect_lang` is `false`), redirect `example.com/fr/` to `example.com/fr/front-page/`.

### DIFF
--- a/src/frontend/frontend-static-pages.php
+++ b/src/frontend/frontend-static-pages.php
@@ -131,15 +131,36 @@ class PLL_Frontend_Static_Pages extends PLL_Static_Pages {
 	}
 
 	/**
-	 * Prevents the canonical redirect if we are on a static front page.
+	 * Manages redirections for the static front pages.
 	 *
 	 * @since 1.8
 	 *
-	 * @param string $redirect_url The redirect url.
+	 * @param string|false $redirect_url The redirect URL.
 	 * @return string|false
 	 */
 	public function pll_check_canonical_url( $redirect_url ) {
-		return $this->options['redirect_lang'] && ! $this->options['force_lang'] && ! empty( $this->curlang->page_on_front ) && is_page( $this->curlang->page_on_front ) ? false : $redirect_url;
+		global $wp_query;
+
+		if ( empty( $this->curlang->page_on_front ) ) {
+			return $redirect_url;
+		}
+
+		if ( $this->options['redirect_lang'] ) {
+			// Prevent the canonical redirect if we are on a static front page.
+			if ( empty( $redirect_url ) ) {
+				// Already prevented.
+				return $redirect_url;
+			}
+			return ! $this->options['force_lang'] && $wp_query->is_page( $this->curlang->page_on_front ) ? false : $redirect_url;
+		}
+
+		if ( 1 === $this->options['force_lang'] && ! $this->options['hide_default'] && $wp_query->is_home && ! $wp_query->is_posts_page ) {
+			// Redirect example.com/fr/ to example.com/fr/front-page/ if we're not hiding the page name, and not hiding
+			// the language slug for the default language.
+			return $this->curlang->get_home_url();
+		}
+
+		return $redirect_url;
 	}
 
 	/**
--- a/tests/phpunit/tests/test-canonical.php
+++ b/tests/phpunit/tests/test-canonical.php
@@ -13,6 +13,8 @@ class Canonical_Test extends PLL_Canonical_UnitTestCase {
 	private static $tag_en;
 	private static $page_for_posts_en;
 	private static $page_for_posts_fr;
+	private static $page_on_front_en;
+	private static $page_on_front_fr;
 
 	/**
 	 * @param WP_UnitTest_Factory $factory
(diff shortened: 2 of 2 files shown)


src/frontend/frontend-static-pages.php
```diff
@@ -131,15 +131,36 @@ public function pll_pre_translation_url( $url, $language, $queried_object_id ) {
...
+		if ( empty( $this->curlang->page_on_front ) ) {
+			return $redirect_url;
+		}
+
+		if ( $this->options['redirect_lang'] ) {
+			// Prevent the canonical redirect if we are on a static front page.
+			if ( empty( $redirect_url ) ) {
+				// Already prevented.
+				return $redirect_url;
+			}
```
Hug0-Drelon, line 153: Since we'd return `false` anyway with the next `return` statement, will it reduce complexity to remove this `if`?

src/frontend/frontend-static-pages.php
```diff
@@ -131,15 +131,36 @@ public function pll_pre_translation_url( $url, $language, $queried_object_id ) {
...
+			return ! $this->options['force_lang'] && $wp_query->is_page( $this->curlang->page_on_front ) ? false : $redirect_url;
+		}
+
+		if ( 1 === $this->options['force_lang'] && ! $this->options['hide_default'] && $wp_query->is_home && ! $wp_query->is_posts_page ) {
```
Hug0-Drelon, line 157: Seeing `! $wp_query->is_posts_page` added, I wonder if we should add a test for post page as well.

src/frontend/frontend-static-pages.php
```diff
@@ -131,15 +131,36 @@ public function pll_pre_translation_url( $url, $language, $queried_object_id ) {
...
+			// Redirect example.com/fr/ to example.com/fr/front-page/ if we're not hiding the page name, and not hiding
```
Hug0-Drelon, line 158:
```suggestion
			// Redirect example.com/en/ to example.com/en/front-page/ if we're not hiding the page name, and not hiding
```
I know it sounds silly, but we usually use English to refer to the default language 🤓

tests/phpunit/tests/test-canonical.php
```diff
@@ -79,6 +81,14 @@ public static function generate_shared_fixtures( WP_UnitTest_Factory $factory )
...
 
 		self::$model->post->save_translations( self::$page_for_posts_en, array( 'en' => self::$page_for_posts_en, 'fr' => self::$page_for_posts_fr ) );
+
+		self::$page_on_front_en = $factory->post->create( array( 'post_title' => 'front', 'post_type' => 'page' ) );
+		self::$model->post->set_language( self::$page_on_front_en, 'en' );
+
+		self::$page_on_front_fr = $factory->post->create( array( 'post_title' => 'accueil', 'post_type' => 'page' ) );
+		self::$model->post->set_language( self::$page_on_front_fr, 'fr' );
+
+		self::$model->post->save_translations( self::$page_on_front_en, array( 'en' => self::$page_on_front_en, 'fr' => self::$page_on_front_fr ) );
```
Hug0-Drelon, line 91: Should we delete these posts in `wpTearDownAfterClass`?

Hug0-Drelon (changes requested):

src/frontend/frontend-static-pages.php
Screenfeed: I did that specifically to prevent calling [`is_page()`](https://developer.wordpress.org/reference/classes/wp_query/is_page/) for nothing.

IT'S THE FUCKING RAMPOCALYPSE OK, WE MUST PREVENT USELESS PROCESSES!

(🤣)

src/frontend/frontend-static-pages.php
Screenfeed: Good thing we don't refer to the default language here then 😛 (the issue is present for any language).

tests/phpunit/tests/test-canonical.php
Screenfeed: Indeed I forgot. Weirdly, the other posts and terms are not deleted either 🤔

src/frontend/frontend-static-pages.php
Screenfeed: The page for posts [is already covered](https://github.com/polylang/polylang/blob/5af114f0cad9f0112674aca54a27d61e2bfae121/tests/phpunit/tests/test-canonical.php#L537-L593), but you're right, something was missing.

In [f45cb47](https://github.com/polylang/polylang/pull/1992/commits/f45cb473a4deb96c20ac7d249acf0420c3486c06) I set a value to the `page_on_front` option, so we make sure that the tests don't stop at `if ( empty( $this->curlang->page_on_front ) ) {` in `pll_check_canonical_url()`, and that `! $wp_query->is_posts_page` is used to not trigger the new redirection.

tests/phpunit/tests/test-canonical.php
Screenfeed: Done in [6fcacfc](https://github.com/polylang/polylang/pull/1992/commits/6fcacfc2e61dbe0bb7df8b5f647173dd7e2f6cee).

src/frontend/frontend-static-pages.php
Hug0-Drelon: Fair enough!

Hug0-Drelon (approved):

PR #2003, opened 2026-09-10
fix(auto-translate): restore tax_query fallback when term has no translation

Hug0-Drelon:
## What?
Restore the fallback to the original term in `translate_tax_query_recursive()` when no translation exists in the current language.

## Why?
A regression was introduced in #1891 when `get_translated_term_by()` was replaced by `PLL_Translated_Term::get_by()` combined with `array_filter()`. When a term has no translation in the current language, the `tax_query` terms array becomes empty and secondary queries return no results.

Polylang for WooCommerce surfaced this regression in CI: [PLLWC PHPUnit failure](https://github.com/polylang/polylang-wc/actions/runs/34387339432/job/102586827434#step:5:806) (`Filter_Blocks_Test::test_products_by_attribute_block`).

The regression was introduced in https://github.com/polylang/polylang/pull/1891/.

This restores the behavior that existed before #1891 and matches other auto-translate paths in the same class (`get_terms_args`, `category__in`, etc.) which already use `$translated ?: $original`.

## How?
- Keep the original term when `get_by()` returns an empty value.
- Stop filtering out falsy translated values from `tax_query` terms.
- Add a PHPUnit test covering a language without term translations.

## Test plan
- [x] `npm run env:composer -- cs -- src/frontend/frontend-auto-translate.php tests/phpunit/tests/test-auto-translate.php`
- [x] `npm run test:php -- --filter=Auto_Translate_Test`
- [x] CI green

### DIFF
--- a/src/frontend/frontend-auto-translate.php
+++ b/src/frontend/frontend-auto-translate.php
@@ -266,11 +266,12 @@ class PLL_Frontend_Auto_Translate {
 			if ( isset( $q['taxonomy'], $q['terms'] ) && $this->model->is_translated_taxonomy( $q['taxonomy'] ) ) {
 				$arr = array();
 				$field = isset( $q['field'] ) && in_array( $q['field'], array( 'slug', 'name' ) ) ? $q['field'] : 'term_id';
-				foreach ( (array) $q['terms'] as $t ) {
-					$arr[] = $this->model->term->get_by( $field, $t, $this->curlang, $q['taxonomy'] );
+				foreach ( (array) $q['terms'] as $term ) {
+					$translated = $this->model->term->get_by( $field, $term, $this->curlang, $q['taxonomy'] );
+					$arr[]      = $translated ?: $term;
 				}
 
-				$tax_queries[ $key ]['terms'] = array_filter( $arr );
+				$tax_queries[ $key ]['terms'] = $arr;
 			} else {
 				// Nested queries.
 				$tax_queries[ $key ] = $this->translate_tax_query_recursive( $q );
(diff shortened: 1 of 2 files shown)


src/frontend/frontend-auto-translate.php
```diff
@@ -266,11 +266,12 @@ protected function translate_tax_query_recursive( $tax_queries ) {
...
 				$field = isset( $q['field'] ) && in_array( $q['field'], array( 'slug', 'name' ) ) ? $q['field'] : 'term_id';
-				foreach ( (array) $q['terms'] as $t ) {
-					$arr[] = $this->model->term->get_by( $field, $t, $this->curlang, $q['taxonomy'] );
+				foreach ( (array) $q['terms'] as $term ) {
+					$translated = $this->model->term->get_by( $field, $term, $this->curlang, $q['taxonomy'] );
+					$arr[]      = $translated ?: $term;
 				}
 
-				$tax_queries[ $key ]['terms'] = array_filter( $arr );
+				$tax_queries[ $key ]['terms'] = $arr;
```
Screenfeed, line 274: With `$arr[] = $translated ?: $term;`, that means empty values come from `$term` itself instead of `$translated`. So is the removal of `array_filter()` still necessary?

Hug0-Drelon: I proposed to remove it to restore pre-#1891 behavior and align with other auto-translate paths in this class (`category__in`, `get_terms_args`, etc.) that use `$tr ?: $original` without filtering.

Screenfeed (approved):

PR #2004, opened 2026-09-11
Add PHPUnit tests for Site Health (part 2)

accolette:
Fixes [#1125](https://github.com/polylang/polylang-pro/issues/1125)

## Why

Following up on the first PR, the remaining Site Health methods (`load.php`, warnings, PHP modules check) were still untested.

## What

Add PHPUnit tests for `load.php`, `info()` (WPML/SimpleXML/multisite warnings), and `site_status_test_php_modules()`.

With this PR, PLL_Admin_Site_Health coverage is approximately 84.62% for methods (11/13) and 91.15% for lines (103/113).

### `wpml-config.xml` fixture and cached file list

The tests temporarily remove and restore the `wpml-config.xml` fixture to cover both presence and absence cases. `PLL_WPML_Config::get_files()` caches the discovered files, so the cache must be reset after changing the fixture on disk.

The reset is handled by a shared `TestCase` helper using `ReflectionProperty`. `ReflectionProperty::setAccessible()` is only used on PHP versions below 8.1, where it is required to access the protected property; it is avoided on newer PHP versions where it is deprecated.

### Testing `load.php`

`load.php` is a plain script (not a class or function) that reads a `$polylang` variable from its own scope. Tests `require` it directly, after preparing a local `$polylang` variable, so the real file runs unmodified rather than re-implementing its condition.

For the "no language configured" case, `PLL_Model::$languages` is replaced with a PHPUnit mock returning `false` for `has()`. This avoids deleting the shared languages from the database (used by `pllSetUpBeforeClass()` and other test classes), which caused cross-test pollution when tried directly.

### Testing `site_status_test_php_modules()`

Per the test plan, this filter is tested through WordPress' real integration point (`WP_Site_Health::get_test_php_extensions()`) rather than by calling the method directly. Since `get_test_php_extensions()` returns a computed `status`/`label`/`description` rather than the raw `$modules` array, tests use a spy filter callback (registered at a different priority) to capture the `$modules` array before and/or after Polylang's filter runs, and compare them directly.

### Multisite tests

`Multisite_Warnings_Test` is wrapped in `if ( is_multisite() ) :` and only runs via `test:php:multisite`.

It covers the `network_activated` field as well as the warnings generated when posts or terms are missing a language. Since `pll_warnings` can already exist in multisite because of `network_activated`, the tests check for the specific `post-no-lang` and `term-no-lang` fields within the existing `pll_warnings` section.

### `load.php` (`Load_Test`)

* [x] Site Health is loaded when `PLL_Admin` has languages configured
* [x] **Unhappy path** — Site Health is not loaded when no language is configured

### `info` — multisite (`Multisite_Warnings_Test`, runs only under `test:php:multisite`)

* [x] `network_activated` is `"Yes"` when Polylang is network activated
* [x] `network_activated` is `"No"` when Polylang is not network activated
* [x] `post-no-lang` is added to `pll_warnings['fields']` when a post is missing a language
* [x] `term-no-lang` is added to `pll_warnings['fields']` when a term is missing a language

### `info` — `pll_warnings` (`Warnings_Test`)

* [x] No `wpml`/`simplexml` warning field when nothing to report (no posts/terms without language, no WPML file)
* [x] Existing `$debug_info` entries are preserved when `pll_warnings` is added

### `info` — WPML / SimpleXML (`Warnings_Test`)

* [x] Existing WPML/SimpleXML tests migrated and adapted to reset the cached file list when changing the fixture
* [x] `network_activated` is absent when not multisite

### `site_status_test_php_modules` (`Warnings_Test`)

* [x] `simplexml` is added to the modules list, marked as required, when a `wpml-config.xml` file exists
* [x] The modules list is left untouched when no `wpml-config.xml` file exists
* [x] `simplexml` is added without altering the other pre-existing modules entries


tests/phpunit/tests/Site_Health/test-Load.php
Hug0-Drelon, line 1: I'm not sure this test is really necessary. Don't remove it right away, we'll see what @Chouby thinks of it first.

tests/phpunit/tests/Site_Health/test-Multisite_Warnings.php
```diff
@@ -0,0 +1,81 @@
...
+				'The network_activated entry should have the expected label.'
+			);
+			$this->assertSame(
+				'Yes',
+				$debug_info['pll_warnings']['fields']['network_activated']['value'],
+				'The network_activated entry should contain the expected value.'
+			);
+		}
+
+		public function test_multisite_does_not_add_warnings_when_no_posts_are_missing_a_language() {
```
Hug0-Drelon: I'd test the opposite, e.g. `test_multisite_add_warnings_when_posts_are_missing_a_language`.

tests/phpunit/tests/Site_Health/test-Multisite_Warnings.php
```diff
@@ -0,0 +1,81 @@
...
+
+			$this->assertArrayHasKey( 'pll_warnings', $debug_info, 'Debug information should contain an entry for pll_warnings.' );
+			$this->assertArrayNotHasKey(
+				'post-no-lang',
+				$debug_info['pll_warnings']['fields'],
+				'Should not report posts without language when all posts have a language.'
+			);
+		}
+
+		public function test_multisite_does_not_add_warnings_when_no_terms_are_missing_a_language() {
```
Hug0-Drelon: Same suggestion as above, we should first and foremost test the happy path. Suggesting `test_multisite_add_warnings_when_terms_are_missing_a_language`.

patchwork.json
```diff
@@ -0,0 +1,5 @@
+{
```
Hug0-Drelon: This can have side effects. I wonder is this causes some errors in https://github.com/polylang/polylang/actions/runs/34611403477/job/103302527084.

Could you try to remove it (as well as the tests redefining `extension_loaded` to see if it fixes the other tests?

Hug0-Drelon (changes requested):
Instead of `try/finally` pattern (which is correct code wise), we tend to use `tear_down` instead to restore initial states. I suggest to move `finally` code to `tear_down`.

patchwork.json
accolette: You're right, I removed `patchwork.json` and the test redefining `extension_loaded()`.

I also ran the full PHPUnit suite locally: The previous 7 errors no longer occur without Patchwork.
I pushed this version, and the CI confirmed the same behavior.

I removed the test in question.

tests/phpunit/tests/Site_Health/test-Languages.php
```diff
@@ -2,8 +2,22 @@
...
 namespace WP_Syntex\Polylang\Tests\Site_Health;
 
+use WP_Error;
+
 class Languages_Test extends TestCase {
 
+	private $languages_deleted = false;
+
+	public function tear_down() {
+		if ( $this->languages_deleted ) {
```
Hug0-Drelon: Instead of having a new property, you can check `self::factory()->pll_model->has_languages()`. It'll be simpler IMHO. And also prevent errors if some day we add new tests and forgot to set `$this->languages_deleted`.

Hug0-Drelon (changes requested):

tests/phpunit/tests/Site_Health/test-Languages.php
```diff
@@ -2,7 +2,16 @@
 
 namespace WP_Syntex\Polylang\Tests\Site_Health;
 
+use WP_Error;
+
 class Languages_Test extends TestCase {
+	public function tear_down() {
+		if ( ! self::factory()->pll_model->has_languages() ) {
```
Hug0-Drelon, line 9: Let's add a small comment to understand why we dot it.
```suggestion
		// Some tests are deleting all languages on purpose, @see test_info_languages_returns_empty_array_when_no_language_is_set.
		if ( ! self::factory()->pll_model->has_languages() ) {
```

tests/phpunit/tests/Site_Health/TestCase.php
```diff
@@ -26,11 +32,23 @@ public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
...
+
+		@mkdir( WP_CONTENT_DIR . '/polylang' );
+		copy(
+			PLL_TEST_DATA_DIR . 'wpml-config.xml',
+			WP_CONTENT_DIR . '/polylang/wpml-config.xml'
+		);
+
+		// Force a fresh scan: another test class may already have cached
+		// an empty file list before this one copied wpml-config.xml.
+		self::reset_wpml_files_cache();
```
Hug0-Drelon, line 44: We can refactor this and move `remove_wpml_config` as well as `restore_wpml_config` in `TestCase`. You'll have to make them `static` though.

Hug0-Drelon (changes requested):

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -7,7 +7,7 @@
 class Warnings_Test extends TestCase {
 	public function tear_down() {
 		if ( ! file_exists( WP_CONTENT_DIR . '/polylang/wpml-config.xml' ) ) {
```
Hug0-Drelon: I suggest to move `if ( ! file_exists( WP_CONTENT_DIR . '/polylang/wpml-config.xml' ) )` into `restore_wpml_config` and bail out if `false`.

Hug0-Drelon (changes requested):

Hug0-Drelon (approved):

tests/phpunit/tests/Site_Health/TestCase.php
```diff
@@ -50,4 +74,66 @@ protected function set_page_on_front( int $page, string $show = 'page' ): void {
...
+	protected static function reset_wpml_files_cache() {
+		$files_reflection = new ReflectionProperty( PLL_WPML_Config::class, 'files' );
+
+		// `setAccessible()` is required before PHP 8.1 to access non-public properties,
+		// but is deprecated in recent PHP versions where properties are accessible by default.
+		if ( PHP_VERSION_ID < 80100 ) {
+			$files_reflection->setAccessible( true );
+		}
+
+		$files_reflection->setValue( PLL_WPML_Config::instance(), null );
```
Chouby, line 103: I will use this pattern to remove PHP version check when it will become useless:
```suggestion
		$files_reflection = new ReflectionProperty( PLL_WPML_Config::class, 'files' );
		version_compare( PHP_VERSION, '8.1', '<' ) && $files_reflection->setAccessible( true );
		$files_reflection->setValue( PLL_WPML_Config::instance(), null );
```

tests/phpunit/tests/Site_Health/TestCase.php
```diff
@@ -26,11 +32,16 @@ public static function pllSetUpBeforeClass( PLL_UnitTest_Factory $factory ) {
 		parent::pllSetUpBeforeClass( $factory );
 
 		$factory->language->create_many( 2 );
+
+		// Ensure the fixture exists and force PLL_WPML_Config to rescan the file.
+		self::restore_wpml_config();
```
Chouby, line 37: This seems to be related to a specific test class so I suggest to move what's needed for wpml-config tests to this class. The test case should keep only what's common.

tests/phpunit/tests/Site_Health/TestCase.php
```diff
@@ -3,9 +3,15 @@
...
-use PLL_Admin_Site_Health;
+use WP_Debug_Data;
+use PLL_WPML_Config;
 use PLL_UnitTestCase;
+use ReflectionProperty;
 use PLL_UnitTest_Factory;
+use PLL_Admin_Site_Health;
+
+use function Brain\Monkey\setUp;
+use function Brain\Monkey\tearDown;
```
Chouby, line 14: I could not find where Brain Monkey is used.
```suggestion
use function Brain\Monkey;
```

tests/phpunit/tests/Site_Health/test-Languages.php
```diff
@@ -2,7 +2,17 @@
...
+
 class Languages_Test extends TestCase {
+	public function tear_down() {
+		// Some tests are deleting all languages on purpose, @see test_info_languages_returns_empty_array_when_no_language_is_set.
+		if ( ! self::factory()->pll_model->has_languages() ) {
+			self::factory()->language->create_many( 2 );
+		}
+
+		parent::tear_down();
+	}
```
Chouby, line 15: This is rather hacky. It would be better to move the tests to a separate class where languages are not created globally.

tests/phpunit/tests/Site_Health/TestCase.php
```diff
@@ -50,4 +74,66 @@ protected function set_page_on_front( int $page, string $show = 'page' ): void {
...
+
+	/**
+	 * Retrieves WordPress debug information.
+	 *
+	 * @return array The debug information.
+	 */
+	protected function get_debug_info() {
+		require_once ABSPATH . 'wp-admin/includes/class-wp-debug-data.php';
+
+		return WP_Debug_Data::debug_data();
```
Chouby, line 86: Used only in one method. you can inline it.

tests/phpunit/tests/Site_Health/test-Languages.php
```diff
@@ -35,4 +45,77 @@ public function test_info_languages_term_props() {
...
 	}
+
+	public function test_info_languages_is_applied() {
+		// Prevent the external WordPress.org request performed by WP_Debug_Data,
+		// which is unrelated to the behavior tested here.
+		$filter = function () {
+			return new WP_Error( 'test_http_request', 'HTTP request disabled for this test.' );
+		};
+
+		add_filter( 'pre_http_request', $filter );
```
Chouby, line 56: WP has already a method to do this.
```suggestion
		// Prevent the external WordPress.org request performed by WP_Debug_Data.
		add_filter( 'pre_http_request', array( $this, 'block_external_http_request' ) );
```

tests/phpunit/tests/Site_Health/test-Languages.php
```diff
@@ -35,4 +45,77 @@ public function test_info_languages_term_props() {
...
+		}
+	}
+
+	public function test_info_languages_returns_one_entry_per_language() {
+		$debug_info = $this->site_health->info_languages( array() );
+
+		$this->assertCount( 2, $debug_info, 'Result should contain one entry per configured language.' );
+		$this->assertArrayHasKey( 'pll_language_en', $debug_info, 'Result should contain an entry for English.' );
+		$this->assertArrayHasKey( 'pll_language_fr', $debug_info, 'Result should contain an entry for French.' );
+	}
```
Chouby, line 87: Almost duplicate of `test_info_languages_is_applied()`. You can merge both. You can use use `preg_grep()` to count the number of keys matching the expected format.

tests/phpunit/tests/Site_Health/test-Languages.php
```diff
@@ -35,4 +45,77 @@ public function test_info_languages_term_props() {
...
+
+	public function test_info_languages_returns_empty_array_when_no_language_is_set() {
+		self::delete_all_languages();
+
+		$debug_info = $this->site_health->info_languages( array() );
+
+		$this->assertEmpty( $debug_info, 'Result should be empty when no language is set.' );
+	}
```
Chouby, line 95: The test doesn't seem useful as the module is not loaded if there's no language.

tests/phpunit/tests/Site_Health/test-Load.php
```diff
@@ -0,0 +1,38 @@
...
+	public function test_when_no_languages_are_configured_site_health_is_not_loaded() {
+		$options     = new Options();
+		$model       = new PLL_Model( $options );
+		$model->languages = $this->createMock( Languages::class );
+		$model->languages->method( 'has' )->willReturn( false );
+
+		$links_model = $model->get_links_model();
+		$polylang    = new PLL_Admin( $links_model );
+
+		require POLYLANG_DIR . '/src/modules/site-health/load.php';
```
Chouby, line 31: You can share duplicate code.

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
...
+
+	public function test_should_not_add_network_activated_field_when_not_multisite() {
+		if ( is_multisite() ) {
+			$this->markTestSkipped( 'network_activated is always present on multisite.' );
+		}
+
+		$debug_info = $this->site_health->info( array() );
+
+		$this->assertArrayNotHasKey( 'network_activated', $debug_info['pll_warnings']['fields'] ?? array(), 'Debug information should not contain an entry for multisite.' );
+	}
```
Chouby, line 113: Move this test to a multisite test class. Reason: I like to avoid to mark tests as skipped in a situation where everything should work smoothly.

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
+<?php
+
+namespace WP_Syntex\Polylang\Tests\Site_Health;
+
+use WP_Site_Health;
+
+class Warnings_Test extends TestCase {
```
Chouby, line 7: Maybe rename this class to WPML_Config

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
...
+	public function tear_down() {
+		self::restore_wpml_config();
```
Chouby, line 9: It would be more logical to add the wpm-config in tests when needed and remove it in tear_down instead of the opposite.

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
...
+				'lang'      => 'en',
+			)
+		);
+
+		self::factory()->term->create(
+			array(
+				'taxonomy' => 'category',
+				'lang'     => 'en',
+			)
+		);
```
Chouby, line 89: Is this useful for this test?

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
...
+		$site_health = new WP_Site_Health();
+		$site_health->get_test_php_extensions();
+
+		$this->assertIsArray( $captured_modules, 'The filter callback should have captured the modules array.' );
+		$this->assertArrayHasKey( 'simplexml', $captured_modules, 'The simplexml module should have been added to the list.' );
+		$this->assertSame(
+			array( 'extension' => 'simplexml', 'required' => true ),
+			$captured_modules['simplexml'],
+			'The simplexml module should be marked as required.'
+		);
```
Chouby, line 137: This test is much too close from what is done in the production code (array manipulation). Just test the output of `get_test_php_extensions()`.

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	public function test_should_not_add_simplexml_to_modules_when_wpml_config_does_not_exist() {
```
Chouby, line 140: Same as above.

tests/phpunit/tests/Site_Health/test-Warnings.php
```diff
@@ -0,0 +1,213 @@
...
+		$site_health->get_test_php_extensions();
+
+		$this->assertSame(
+			$modules_before,
+			$modules_after,
+			'The modules list should be left untouched when no wpml-config.xml file is found.'
+		);
+	}
+
+	public function test_should_add_simplexml_without_altering_other_modules() {
```
Chouby, line 175: This one is also too close from the way it is done in production, but trickier to test. May be you can compare the output with and without Polylang.

Chouby (changes requested):

PR #2008, opened 2026-09-14
Sync IDs in blocks via `wpml-config.xml`

Screenfeed:
Closes https://github.com/polylang/polylang-pro/issues/1841.
Related to https://github.com/polylang/polylang-pro/issues/2952.

[Doc WPML](https://wpml.org/documentation/support/language-configuration-files/make-custom-gutenberg-blocks-translatable/#converting-ids).

Some examples:
- https://github.com/OnTheGoSystems/wpml-config/blob/3ce703dd2592580d88a6f494debef830c50cb250/Divi/wpml-config.xml#L1649
- https://github.com/OnTheGoSystems/wpml-config/blob/3ce703dd2592580d88a6f494debef830c50cb250/edd-multilingual/wpml-config.xml#L56
- https://github.com/OnTheGoSystems/wpml-config/blob/3ce703dd2592580d88a6f494debef830c50cb250/sitepress-multilingual-cms/wpml-config.xml#L98

## What

This allows to synchronize IDs in blocks (attributes + content) via a `wpml-config.xml` file.

## How

Ex:

```xml
<gutenberg-block type="foo/bar" translate="0">
	<key name="the-posts">
		<key name="*" type="post-ids" sub-type="page" />
	</key>
	<key name="the-terms">
		<key name="*" type="taxonomy-ids" sub-type="post_tag" />
	</key>
	<xpath type="post-id" sub-type="post">//*[@class="foobar-post-id"]</xpath>
	<xpath type="taxonomy-ids" sub-type="category">//*[@class="foobar-cat-id"]</xpath>
</gutenberg-block>
```

1. `type="post-ids"` and `type="taxonomy-ids"` tell if the values are post IDs or term IDs.
2. `sub-type` contains the post type name or the taxonomy name. However it is optional.
3. The `translate` attribute has no impact on this mechanism, but it should be `1` if the attributes contain other non-IDs data to translate.

Point 2 requires us to introduce a type `post_mixed` in the filters: the post type will be determined right before the translation, when we get access to the IDs themselves. See https://github.com/polylang/polylang-pro/pull/3140.


src/modules/wpml/wpml-config.php
```diff
@@ -712,6 +831,51 @@ private function is_supported_field( SimpleXMLElement $field ): bool {
 		return true;
 	}
 
+	/**
+	 * Returns the traduction type of the given node, for IDs translation.
```
Hug0-Drelon, line 848:
```suggestion
	 * Returns the traduction type of the given node, for IDs translation.
	 *  All post types are inferred to `post` except for `attachment` and `wp_block`.
```
Suggesting a more descriptive documentation. This leads to different behaviors internally.

src/modules/wpml/wpml-config.php
```diff
@@ -712,6 +831,51 @@ private function is_supported_field( SimpleXMLElement $field ): bool {
...
+	private function get_ids_traduction_type( SimpleXMLElement $field ): string {
+		$type = $this->get_field_attribute( $field, 'type' );
+
+		if ( 'taxonomy-ids' === $type ) {
+			return 'term';
+		}
+
+		if ( 'post-ids' !== $type ) {
+			return '';
+		}
```
Hug0-Drelon: This is checked upstream, should we trust and simplify since it's  all private methods?

src/modules/wpml/wpml-config.php
```diff
@@ -652,13 +714,15 @@ private function get_field_attribute( SimpleXMLElement $field, $attribute_name )
 	 * Gets attributes values recursively.
 	 *
 	 * @since 3.6
+	 * @since 3.9 Parameter `$translate` added.
 	 *
-	 * @param  SimpleXMLElement $field A XML node.
+	 * @param SimpleXMLElement $field     A XML node.
```
Hug0-Drelon, line 717: Could we use `SimpleXMLElement::addAttribute` ([see](https://www.php.net/manual/en/simplexmlelement.addattribute.php)) to specify the translate status and avoid drilling the `$translate` argument to several methods?

src/modules/wpml/wpml-config.php
```diff
@@ -540,12 +593,21 @@ protected function extract_blocks_parsing_rules() {
...
-								break;
+							if ( 'json' === $encoding ) {
+								// For WPML, `json` means `json,urlencode` (and is the only format supported in this context).
+								$parsing_rules['encoding'][ $block_name ][ key( $rule ) ] = 'json,urlencode';
 							}
 
-							// For WPML, `json` means `json,urlencode` (and is the only format supported in this context).
-							$parsing_rules['encoding'][ $block_name ][ key( $rule ) ] = 'json,urlencode';
+							foreach ( array( 'post', 'attachment', 'wp_block', 'term' ) as $trad_type ) {
+								$ids_attributes = $this->get_ids_attributes( $trad_type, $child, $translate );
```
Hug0-Drelon, line 600: Calling `get_ids_attributes` in a loop makes us parse the element three times (and it doesn't count the children).

Could we get all IDs attributes at once and order the output as per `[ 'post', 'attachment', 'wp_block', 'term' ]` or something?

Hug0-Drelon (changes requested):
Should we honor "sub-type (optional): can be the specific entity of the type if it’s already known ... If not defined, the specific entity will be guessed." as per WPML documentation?

src/modules/wpml/wpml-config.php
Screenfeed: That means we have to add the attribute to each child node, I'm not sure it's a good idea :/

src/modules/wpml/wpml-config.php
Screenfeed: I tried but I couldn't find any better way.
Imagine this:

```xml
<gutenberg-block type="foo/bar" translate="0">
	<key name="first">
		<key name="a" type="post-ids" sub-type="page" />
		<key name="b">
			<key name="*" type="taxonomy-ids" sub-type="post_tag" />
		</key>
	</key>
	<key name="second">
		<key name="c" type="taxonomy-ids" sub-type="category" />
		<key name="d">
			<key name="*" type="post-ids" sub-type="post" />
		</key>
	</key>
</gutenberg-block>
```

This should result in this:

```php
array(
	'post' => array(
		'first' => array(
			'a' => true,
		),
		'second' => array(
			'd' => array(
				'*' => true,
			),
		),
	),
	'term' => array(
		'first' => array(
			'b' => array(
				'*' => true,
			),
		),
		'second' => array(
			'c' => true,
		),
	),
)
```

But basically the main difficulty IMHO is that to determine the `'term'`, `'post'`, etc. key we must go to the deepest child because it's the one that contains the `sub-type` attribute.

Screenfeed:
> Should we honor "sub-type (optional): can be the specific entity of the type if it’s already known ... If not defined, the specific entity will be guessed." as per WPML documentation?

I totally missed that part. I guess (lol) we should but that means the post type must be determined once we receive the post IDs (no problem with terms). This implies changes in PLL Pro (in `PLL_Sync_Ids::translate()` I suppose). It may be doable by creating a new key `post_mixed` along `post`, `attachment`, and `wp_block`.
Note: currently `post` is returned in case the sub-type is not specified.

What do you think?

EDIT: in progress.
EDIT 2: done.

src/modules/wpml/wpml-config.php
Screenfeed: Good idea. Done in [dbaf727](https://github.com/polylang/polylang/pull/2008/commits/dbaf727bc1d1a90b6720b051d33baa2513b73cdc).

src/modules/wpml/wpml-config.php
Screenfeed: I went the other around in [76dfd87](https://github.com/polylang/polylang/pull/2008/commits/76dfd87d8cd839496ea72955dcb58b580eedd36d) because it was checked only in one place (`xpath` case), not in the `key` case.

src/modules/wpml/wpml-config.php
Hug0-Drelon: I don't see why, do you have an example in mind?

Screenfeed:
```xml
<gutenberg-block type="foo/bar" translate="0">
	<key name="the-posts">
		<key name="toto" type="post-ids" sub-type="page" />
		<key name="first-level">
			<key name="second-level" />
		</key>
	</key>
</gutenberg-block>
```

Before that, this block was skipped entirely as soon as we had `translate="0"`.
Now we must traverse all the tree to determine for each node if it must be translated or not:
- A node with `type="post-ids"` will always be translate, we don't need the value of the `translate` attribute.
- A node without a `type` attribute: we must know if it needs to be translated.

But maybe there's a better way to do this.

PR #2014, opened 2026-09-15
The "All" status filter can be the current one

Screenfeed:
Fixes #3137.

## What

The "All" status filter on the posts page wasn't showing as the current one after clicking on it.

## Why

WP [detects](https://github.com/WordPress/WordPress/blob/8d60e6b30b9633f5082ae6ece1e9233d51e24c79/wp-admin/includes/class-wp-posts-list-table.php#L357) the current filter with a simplistic way for the "All" status. See [`WP_Posts_List_Table::is_base_request()`](https://github.com/WordPress/WordPress/blob/8d60e6b30b9633f5082ae6ece1e9233d51e24c79/wp-admin/includes/class-wp-posts-list-table.php#L241). If anything is added to the URL, like `foo=1`, it messes it up.

## How

Add `all_posts=1` to the URL to force it as the current status: simple, less error prone, WP already uses it for posts.
Another way would be to compare the query args in the "All"'s URL with the content of `$_GET`.


tests/phpunit/tests/test-admin-filters-post.php
```diff
@@ -592,4 +592,26 @@ public function test_get_pages_query_is_cached_for_dropdown() {
...
+		_get_list_table( 'WP_Posts_List_Table' )->views();
+		$html = ob_get_clean();
+
+		$doc = new DOMDocument( '1.0', 'UTF-8' );
+		$internal_errors = libxml_use_internal_errors( true );
+		$doc->loadHTML( $html );
+		libxml_clear_errors();
+		libxml_use_internal_errors( $internal_errors );
+
+		$list = ( new DOMXpath( $doc ) )->query( '//ul/li[@class="all"]/a[@class="current"]' );
```
Hug0-Drelon, line 612: I find it brittle to use a XPath for this (if WordPress changes something, we're F). You see me coming but no, I won't suggest `WP_HTML_Tag_Processor` 😅

Should we instead test this with Playwright?

Hug0-Drelon (changes requested):

tests/phpunit/tests/test-admin-filters-post.php
Screenfeed: I'm not sure it's worth it: I've check on WP 4.9 and it's the same markup.
Both XPath and e2e could break some day (by changing the markup for the 1st, by changing the CSS for the 2nd), so I chose the solution that was the simplest for me. But if you think we really should use e2e instead I can try it.

Hug0-Drelon: > Both XPath and e2e could break

Not so true, e2e could of course break, but not that easily (remember we use locators in Playwright, which resolve differently than CSS selectors or XPath).

Hug0-Drelon (approved):
LGTM,

One thing remains though, what should we do for https://github.com/polylang/polylang/pull/2014#discussion_r4034450169 @Chouby?

tests/phpunit/tests/test-admin-filters-post.php
Screenfeed: How would you locate the link? Something like this?

```JS
page.locator( '.subsubsub [aria-current]' )
```

Then check that it contains the text "All"?

Reminder of the markup:

```html
<ul class="subsubsub">
	<li class="all"><a href="..." class="current" aria-current="page">All <span class="count">(47)</span></a> |</li>
	<li class="active"><a href="...">Active <span class="count">(5)</span></a> |</li>
	...
</ul>
```

src/admin/admin-filters-post.php
```diff
@@ -357,7 +357,12 @@ public function add_per_language_edit_views( $views ) {
 		}
 
 		if ( isset( $views['all'] ) ) { // In case a filter removes this key.
-			$views['all'] = str_replace( 'edit.php?', 'edit.php?lang=all&', $views['all'] );
+			$replacement = 'lang=all&';
+
+			if ( ! str_contains( $views['all'], 'all_posts=1' ) ) {
+				$replacement .= 'all_posts=1&';
+			}
+			$views['all'] = str_replace( 'edit.php?', 'edit.php?' . $replacement, $views['all'] );
```
Chouby: I wrote it like this the first time but `add_query_arg()` should do the job better:
 ```suggestion
			$views['all'] = add_query_arg(
				array( 
					'lang'      => 'all',
					'all_posts' => 1,
				),
				$views['all']
			);
```

**Review by Chouby — CHANGES_REQUESTED** [review]

File: src/admin/admin-filters-post.php (hunk 2 above)
Comment (Screenfeed) [reply src/admin/admin-filters-post.php]: Homer bushes

Comment (Screenfeed) [reply src/admin/admin-filters-post.php]: We got powned like rookies 🤣
`$views['all']` doesn't contain the URL, but the entire link HTML 🤦‍♂️

Comment (Chouby) [reply src/admin/admin-filters-post.php]: Ah! That was the reason why I proceeded like this. I couldn't remember.

Comment (Screenfeed) [reply src/admin/admin-filters-post.php]: If we want to use `add_query_arg()`, we'll need `preg_match()` to grab the URL. I'm not sure it's worth it.

Comment (Screenfeed) [reply src/admin/admin-filters-post.php]: We forgot that the URL is escaped, so we need to decode it, add the args, then escape again. That seams a bit complicated and fragile.

```php
if ( ! empty( $views['all'] ) && preg_match( '/\shref="(?<HREF>edit\.php[^"]*)"/', $views['all'], $matches ) ) { // In case a filter removes this key.
	$href = add_query_arg(
		array(
			'lang'      => 'all',
			'all_posts' => 1,
		),
		htmlspecialchars_decode( $matches['HREF'], ENT_QUOTES )
	);
	$views['all'] = str_replace( $matches['HREF'], esc_attr( $href ), $views['all'] );
}
```

[Test](https://regex101.com/r/4Uoa60/1).

PR #2015, opened 2026-09-16
Fix orphan translation group after the language of a term is changed

Chouby:
The issue has been reported in https://wordpress.org/support/topic/polylang-3-8-9-detached-term_translations-after-term-language-change/ however it can be reproduced in a simpler way just by assigning a language to a term then by changing the language.

- When we set the language of a term, we call `PLL_Translated_Object::set_language()` in which we attempt to update its translation group: https://github.com/polylang/polylang/blob/3.9-beta1/src/translated-object.php#L102
- In which we attempt to delete the translation in the old language: https://github.com/polylang/polylang/blob/3.9-beta1/src/translated-object.php#L171
- Here we look for the term ID in the translation group description : https://github.com/polylang/polylang/blob/3.9-beta1/src/translated-object.php#L171
- To remove the corresponding key.

The problem is that `get_translations()` temporarily returns an array with 2 keys for the same ID, the old and the new language. `array_search()` returns only one of them and then Murphy's law applies.

The fix is to remove all language keys corresponding for the searched term ID.

A test is added with double check for the translation group count and the translation group relationships.


tests/phpunit/tests/test-translated-term.php
```diff
@@ -245,4 +245,18 @@ public function test_should_not_update_translations_group_when_removing_all_tran
 		$this->assertSame( self::$model->post->get_translation( $terms['en'], 'fr' ), 0 );
 		$this->assertSame( self::$model->post->get_translation( $terms['fr'], 'en' ), 0 );
 	}
+
+	public function test_set_language_should_not_create_orphan_group() {
+		$id = self::factory()->term->create();
+		self::$model->term->set_language( $id, 'en' );
+		self::$model->term->set_language( $id, 'de' );
+
+		$terms = get_terms( array( 'taxonomy' => 'term_translations', 'hide_empty' => false ) );
```
Hug0-Drelon, line 254:
```suggestion
		$terms = get_terms( array( 'taxonomy' => 'term_translations', 'hide_empty' => false ) );
		$this->assertNotEmpty( $terms );
```

I think we should check the array before asserting while looping through it.

src/translated-object.php
```diff
@@ -235,11 +235,15 @@ public function delete_translation( $id ) {
...
-				unset( $descr[ $slug ] );
-			}
+			/*
+			 * Search the ID to remove only among our language keys
+			 * in case some plugin stores the same value with different key.
+			 * Remove all keys with this ID as `get_translations()` may return
+			 * temporarily 2 languages for the same ID (old and new language)
+			 * when `set_language()` is called.
+			 */
+			$slugs = array_keys( $this->get_translations( $id ), $id );
```
Hug0-Drelon, line 245:
```suggestion
			$slugs = array_keys( $this->get_translations( $id ), $id, true );
```
Should we use strict comparison here?

Hug0-Drelon (changes requested):

src/translated-object.php
Chouby: we can. Everything is sanitized.

Hug0-Drelon (approved):

